PR #24, opened 2020-09-22
HOLD React Card Example

urbanisierung:
During our Hacksession we also had the idea to add some eyecandy and hightlight some guides. With MDX and React this seems to be quite easy:

image

/cc @jwulf


menski:
@akeller can we delete this, or is that anything you think we want to use in the future

akeller:
We can delete @menski, I think this would be easy to do in the future if we see the need, but I don't see a reason to keep the PR.

PR #170, opened 2021-02-01
Pricing page updates

akeller:
Content in these pages are already out of date (ex. we have 3 plans, including Enterprise but only 2 are documented).

Do not merge until I confirm there are no other changes necessary.


akeller:
Closing this. Looks like I did something wrong(?). Will fix an reopen.

Issue #141, opened 2021-01-08
Remove pricing plan info from docs


akeller:
Documentation is not a good place to display this information as the names of the pricing plans/tiers may change and we probably don't want a history of that information stored in git. 

https://docs.camunda.io/docs/product-manuals/cloud-console/manage-plan/available-plans

I'm happy to link to a product page with this information, but this is not appropriate to have in documentation.

akeller:
Via email this is good to go.

PR #171, opened 2021-02-01
Pricing & plan updates

akeller:
Covers #141


akeller:
Felix is aware of these changes via email.

urbanisierung (approved):
/lgtm

@akeller Just one question: Do we plan to put the information about how billing works somewhere else?

akeller:
> @akeller Just one question: Do we plan to put the information about how billing works somewhere else?

This information will be available on the product page. I want to work with product marketing to make something like a calculator available too.

Issue #165, opened 2021-01-28
Operate does not support TLS communication


akeller:
User feedback - https://zeebe-io.slack.com/archives/C6WGNHV2A/p1611651908045300

![image](https://user-images.githubusercontent.com/1051363/106175779-2163b480-615c-11eb-80dc-35266c674323.png)

After a conversation with @jwulf we should add a note to the Helm chart docs, something like -

"Currently Operate does not support TLS communication"

We can also include Josh's comments on Slack as a workaround.

akeller:
@pihme @ralfpuchert does the above make sense to you? Can we add a note "currently Operate does not support TLS communication" in the Helm chart docs?

pihme:
I have no insight in the matter and will defer this to @ralfpuchert.

ralfpuchert:
Hello, I have nothing to do with Helm chart docs, but I propose to add this to the deployment guides of Operate and Tasklist.
Do you want me to create issues for that?

pihme:
@ralfpuchert So it is true that Operate doesn't support TLS?

ralfpuchert:
@pihme As Svetlana wrote (see the screenshot) in communication with zeebe.

pihme:
Ok, then we should indeed add some remarks to the Helm charts docs. If you have a phrase in Operate doc, I would copy that and describe the same in the helm charts section.

ralfpuchert:
@pihme Operate docs have now a note about TLS connection to zeebe and is merged into `release_0_27`

akeller:
Should have closed this in April/May, but closing it now 😃

PR #175, opened 2021-02-08
docs(operate/tasklist): add tls not supported note

ralfpuchert:
* Add note that TLS communication is not supported in Operate and Tasklist

Related with #165


ralfpuchert:
Done in #209

PR #177, opened 2021-02-09
Clarify job delivery guarantees/idempotency requirements

### LINKED ISSUE #176: Document that Zeebe is an at least once system
We should clearly indicate that Zeebe is an at least once system. This means records may be processed more than once, jobs may be activated more than once, etc. 

We already document in the exporter reference that exporting is at least once, so we should just do it for jobs/workers.

If you find anything else, please add it as well, but I can't think of it now.

MiguelPires:
closes #176


npepinpe (approved):

npepinpe:
I guess we just merge it? Bot doesn't seem to work properly :man_shrugging:

PR #179, opened 2021-02-12
Update workflow-lifecycles.md

ghRocs:
fix 2 items for the "Element id"

### DIFF
--- a/docs/product-manuals/zeebe/technical-concepts/workflow-lifecycles.md
+++ b/docs/product-manuals/zeebe/technical-concepts/workflow-lifecycles.md
@@ -124,12 +124,12 @@ Given the above workflow a successful execution will yield the following records
     </tr>
     <tr>
         <td>ELEMENT_COMPLETING</td>
-        <td>order-placed</td>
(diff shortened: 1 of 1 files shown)


pihme (approved):
Nice catch!

pihme:
@urbanisierung / @menski  any idea how I can get the build-docs task to complete?

urbanisierung:
@pihme it seems that the `build-docs` github action did not trigger. Can you push an empty line somewhere (maybe to the readme) to check if it triggers then?

pihme:
@urbanisierung Maybe it is because it comes from a fork? Strangely enough I was able to commit to the forked repo.

urbanisierung:
Found some github docs regarding that issue, and it seems we have to enable actions for forked repos:
image

But then all github secrets can also be used within forked repos :/

pihme:
@ghRocs I recreated the change in another PR. We had some trouble merging this one, because our tooling doesn't yet play nice with cloned branches. 

Thanks for the contribution!

PR #186, opened 2021-02-25
Console: Create cluster with channels

ultraschuppi:
- merged create cluster page from product handbook and getting started
- added section about channels

will manually merge when feature is on prod


urbanisierung (approved):
/lgtm

PR #189, opened 2021-03-02
162 add sitemap

### LINKED ISSUE #162: Add sitemap
We need a sitemap for good SEO. 

While there is a [plugin](https://v2.docusaurus.io/docs/api/plugins/@docusaurus/plugin-sitemap/) it looks like we can configure this through `@docusaurus/preset-classic`.

urbanisierung:
/closes #162


akeller (approved):
As far as I can tell, this looks good. I assume it will generate a sitemap on build and Google's crawler just knows to look at it? Stretching the limits of my web knowledge 😅

urbanisierung:
> As far as I can tell, this looks good. I assume it will generate a sitemap on build and Google's crawler just knows to look at it? Stretching the limits of my web knowledge

Exactly, that's how it should work :)

PR #191, opened 2021-03-04
Community updates

akeller:
Aligning to the rebranding with a generic reference to the Slack and forum.

### DIFF
--- a/docs/product-manuals/zeebe/open-source/get-help-get-involved.md
+++ b/docs/product-manuals/zeebe/open-source/get-help-get-involved.md
@@ -7,17 +7,17 @@ We provide a few different public-facing Zeebe support and feedback channels so 
 
 ### Zeebe user forum
 
-The best place to ask questions about Zeebe and to troubleshoot issues is the Zeebe user forum.
+The best place to ask questions about Zeebe and to troubleshoot issues is the forum.
 
 The Zeebe team monitors the forum closely, and we do our best to respond to all questions in a timely manner.
 
-[Go to the Zeebe user forum](https://forum.zeebe.io)
+[Go to the forum](https://forum.zeebe.io)
 
 ### Public Slack group
 
-There is a public Zeebe Slack group where you can ask one-off questions, share community contributions, and connect with other Zeebe users.
+There is a public Slack group where you can ask one-off questions, share community contributions, and connect with other Zeebe users.
 
-[Join the Zeebe Slack group](https://zeebe-slack-invite.herokuapp.com/)
+[Join the Slack group](https://zeebe-slack-invite.herokuapp.com/)
 
 ### Create an issue in GitHub
 
--- a/docs/product-manuals/zeebe/open-source/get-help-get-involved.md
+++ b/docs/product-manuals/zeebe/open-source/get-help-get-involved.md
@@ -5,7 +5,7 @@ title: "Get help & get involved"
 
 We provide a few different public-facing Zeebe support and feedback channels so that users can ask questions, report problems, and make contributions.
 
-### Zeebe user forum
+### Camunda Cloud user forum
 
 The best place to ask questions about Zeebe and to troubleshoot issues is the forum.
 


akeller:
@mary-grace @menski looking closer at this page I'm wondering if it would be best to move the Slack and forum info up and out of the Zeebe specific area of the docs. The contributions for Zeebe still make sense at this level. What are your thoughts here? 

Should I also change references to the Zeebe team monitoring?

mary-grace (changes requested):
Line 8 should be changed as well. We could say `Camunda Cloud user forum` or more generically, `Discourse forum.`

More generally, I think we could likely combine the forum & Slack sections. Ideally, I'd like to see people posting about their contributions and connecting with other users in both Discourse & Slack, and I think we could communicate that here. If that complicates things too much for right now, however, I'm fine to push this forward with only the change to line 8.

akeller:
I changed line 8 after @mary-grace's comment.

I'm going to leave the rest as-is for now.

Because this is in a Zeebe specific section of the docs, I think we need to consider moving it file structure wise.

/docs/product-manuals/
|-- Concepts
|-- Clients
|-- Cloud Console
|-- Modeler
|-- Zeebe
  |-- Introduction
  |-- Technical concepts
  |-- Deployment guide
  |-- Open Source community
    |-- Community contributions
    |-- **Get help & get involved**
    |-- Release Cycle
    |-- Deprecated features
|-- Operate
|-- Tasklist

I would recommend we move the whole *Open source community* directory into the /docs/reference section or at a minimum move the *Get help & get involved* page. @menski @pihme any thoughts here? Since the forum and Slack would be available for all Camunda Cloud devs, it doesn't make sense to bury it within the Zeebe docs. This isn't fitting right for me in its current place.

docs/product-manuals/zeebe/open-source/get-help-get-involved.md
```diff
@@ -5,19 +5,19 @@ title: "Get help & get involved"
...
-### Zeebe user forum
+### Camunda Cloud user forum
 
-The best place to ask questions about Zeebe and to troubleshoot issues is the Zeebe user forum.
+The best place to ask questions about Zeebe and to troubleshoot issues is the forum.
 
 The Zeebe team monitors the forum closely, and we do our best to respond to all questions in a timely manner.
 
-[Go to the Zeebe user forum](https://forum.zeebe.io)
+[Go to the forum](https://forum.zeebe.io)
```
menski, line 14: Should we already use the new URL? https://forum.camunda.io

docs/product-manuals/zeebe/open-source/get-help-get-involved.md
```diff
@@ -5,19 +5,19 @@ title: "Get help & get involved"
...
 
 ### Public Slack group
 
-There is a public Zeebe Slack group where you can ask one-off questions, share community contributions, and connect with other Zeebe users.
+There is a public Slack group where you can ask one-off questions, share community contributions, and connect with other Zeebe users.
 
-[Join the Zeebe Slack group](https://zeebe-slack-invite.herokuapp.com/)
+[Join the Slack group](https://zeebe-slack-invite.herokuapp.com/)
```
menski, line 20: Unrelated to this change just noticed that we maybe need a new slack invite app

menski (approved):
Maybe one change for the Forum URL and we should think about the slack invite app. @mary-grace I think the infra Team is owning that asset, we can have a chat about it in our next sync.

pihme:
@akeller  My two cents on the structure questions: I am in favor of moving the `Get Help & get involved` page upwards to a more prominent position.

I am not so keen on moving the `Open Source Community` section into references or outside of Zeebe. As far as I know, Zeebe is the only Open Source component yet. 

But in any case, I think the overall Open Source strategy of Camunda and the Camunda Cloud product line should drive where this sections ends up. I am not sure who has an authoritative say on this at the moment (e.g. whether it is product or DevRel).

akeller:
I'll move these comments and feedback to a new issue about the structure so we can continue the discussion.

PR #200, opened 2021-04-06
Correct the XML representation of sequential subprocesses

saig0:
* correct the XML representation for sequential subprocesses


Zelldon (approved):
Awesome 💪

PR #205, opened 2021-04-08
Added terms 'connector' and 'bridge'

berndruecker:


akeller:
@berndruecker are these applicable to Camunda Cloud and Camunda Platform? I'd love to have one, single glossary in the future right now we just have the Camunda Cloud one we pulled over from Zeebe.

akeller:
As an FYI, there is a commit in the `0.27` PR to handle updates to the glossary for workflow -> process name change - https://github.com/camunda-cloud/camunda-cloud-documentation/pull/148/commits/2c7fff5d90febe99e110e1f4c0b826e40df95160

berndruecker:
Re: Camunda Platform: Interesting question! In Camunda Platform I think we tended to use connectors for "non-active" reusable components that connect to other systems (e.g. HTTP Connector). This is an option, that we don't have in Camunda Cloud, as "there are no Java Delegates", but only "External Task". In Platform we used bridge for "active" components that connect to other systems (e.g. RPA Bridge).
But as far as I am aware we never defined these terms precisely - but my feeling says there is this difference (which is not possible in Camunda Cloud).

PR #206, opened 2021-04-08
Updated link to readme 

berndruecker:


akeller (commented):
This looks good to me, but I'm wondering if we should add other Community Hub link changes to this PR as well.

@celanthe can you review the `products-manuals/clients` to see if there are additional links we can update to reference the Community Hub?

berndruecker:
are we small in PR budget? ;-)

akeller:
> Is the PR budget running out? ;-)

😝  No but a merged PR to master doesn't update the docs. We have to do a tagged release to update the live docs so I'd love to combine things so its easier on my brain!

PR #208, opened 2021-04-09
Add Public API page

menski:
Add a page to document the public API of Camunda Cloud. Public API is defined as an API which we guarantee backwards compatibility. At the moment we agreed that for 1.0 we only see the Java Client API as public API.

I choose to put it in the reference section as I didn't see fit in other sections, and wasn't sure how a new section could look like. Happy to take suggestions.

### DIFF
--- a/docs/reference/overview.md
+++ b/docs/reference/overview.md
@@ -8,6 +8,7 @@ slug: /reference/
 This section contains general reference material for Camunda Cloud, including API references.
 
 ## APIs
+- [Public API](public-api.md)
 - [Console API clients](cloud-console-api-clients.md)
 - [Console API (REST)](cloud-console-api-reference.md)
 - [Zeebe API (gRPC)](grpc.md)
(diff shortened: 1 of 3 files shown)


npepinpe (approved):
Looks fine. My only concern is that it's unclear if these are the only backwards compatibility guarantees we offer or not. imo it's a bit backwards: we should talk about backwards compat, then introduce the public api and its guarantees. But that's an improvement we can do later.

akeller (commented):
Please add some links 
* Link to the API reference page(s) -> I'm confused if the Camunda Cloud Public API includes Console API (REST), Zeebe (gRPC), and Tasklist API
* **Backwards Compatibility for Public API** section link to the release cycle/policy page if readers are interested in more information

@npepinpe raises a good point about ordering, but I don't have strong feelings either way.

akeller:
@falko yesterday you mentioned missing some public API information. Does this cover what you are looking for?

menski:
Thanks I always forget these links. :see_no_evil:

PR #210, opened 2021-04-15
Proposal for Structural Changes

akeller:
Created the following pages in reference:
* Licenses
* Notices
* Supported environments

Moved and modified: 
* Release cycle (from open-source/zeebe to reference) _-> Renamed **Release policy** to align with Camunda Platform_
* Deprecated features _-> Renamed **Announcements** to align with Camunda Platform_

Modified:
* Contact

Added to footer:
* Licenses
* Notices
* Release policy (formerly Release cycle)
* Clickable logo to Camunda.com

**Open questions**

- [x] Is there any other license information that we need to pull in? 
- [x] I assumed there were no security notices for Camunda Cloud yet, is that correct?
- [x] @menski can you update the release cycle with information relevant for Camunda Cloud? ~~I modified it slightly, but the section for Zeebe 1.0 should be rewritten since we know there are breaking changes.~~

PR #217, opened 2021-04-22
Remove release-cycle pre-1.0 note

korthout:
Considering Zeebe release `1.0.0` is nearing, I assume we should remove the note about breaking changes before Zeebe 1.0

### DIFF
--- a/docs/reference/release-cycle.md
+++ b/docs/reference/release-cycle.md
@@ -20,18 +20,6 @@ patch releases. Patch releases are offered on a best effort basis for the
 currently supported versions.
 
 
-### Breaking changes before Zeebe 1.0
-
-Given how early we are in the Zeebe journey, we're not ready to make it a 1.0
-release, and we want it to be transparent that this was a deliberate decision.
-If we do end up having to make breaking API changes, we'd rather do so before
-we get to 1.0 and not by moving to 2.0 just a few months after a 1.0 release.
-
-Of course, even in Zeebe's pre-1.0 state, we'll always make our best effort to
-avoid breaking changes, to communicate early and often about planned changes,
-and if possible, to provide a migration path if we do need to make such a change.
-
-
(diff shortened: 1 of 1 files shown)


akeller:
@korthout can you take a look at this PR? I'm advocating we move this release-cycle content more up-level to handle all Camunda Cloud components. I am having trouble figuring out how we want to describe things pre-1.0. Maybe you have inputs there?

korthout:
@akeller Ah good point. I was unaware of that change. I just saw this page and thought let's improve it 😄 

To be honest, the more I think about it. Perhaps it makes sense to leave the comment up while we still support pre-1.0 minor versions. Consider a strange bug that requires us to make breaking changes in order to fix it, then we might have a 0.26.x or a 0.27 release to fix a bug with actual breaking changes. I don't imagine that this will actually happen though. 

@npepinpe Do you have any opinions on this?

npepinpe:
We will keep supporting 0.26.x until 1.1 - we can remove the note at that point. But I don't think it's a big deal to be honest if we remove it now. I'm happy to defer to our product team.

pihme:
@akeller I removed myself as reviewer and added you, instead, so you can coordinate this change with #210.  

In my personal opinion, I would remove this note in the 1.0.0 content. We do have versioned documentation now and in the older versions this can stay at the current position and with the current content. 
But I don't see a need to talk about pre 1.0.0 version in the 1.0.0 documentation.

And for 1.0.0 aside from removing the note it might make sense to talk about public APIs and stability that customers can expect. But this is probably more a question to product management and legal department, what we want to communicate here.

korthout:
I agree with this one:
> I don't see a need to talk about pre 1.0.0 version in the 1.0.0 documentation.

@akeller I've rebased the changes onto `structural-GA-changes` and changed the target for this PR.

PR #220, opened 2021-04-26
Add startup probe

deepthidevaki:
Add documentation for the new endpoint for startup probe added by PR https://github.com/camunda-cloud/zeebe/pull/6852


pihme (changes requested):
The change looks good. The one aspect I am concerned about is the target branch of the PR.

I think it should be `release/0.27` because it targets the next version to be released.

I might be wrong, though, especially if you have backported the code changes to 0.26. If so, please let me know.

deepthidevaki:
It should be on the next release. The new endpoint is not backported to 0.26.

pihme (approved):
/lgtm

PR #224, opened 2021-05-04
Release 0 27 redirect example

urbanisierung:
@akeller @jwulf @menski Here's an example for client side redirects.

The redirect plugin does not work with the live preview (see https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-client-redirects)

But you can try it out by building the project and manually starting an http server (like https://www.npmjs.com/package/http-server)

```bash
npm run build
cd build
http-server -o
```

This example redirects `/docs/guides/setting-up-development-project-old` to `/docs/guides/`

Hope that helps!


akeller:
Thanks for pulling this together! Do we have an opportunity to do anything on the server-side for redirects? I don't like this for a long-term redirect strategy, but for some of the structural changes I'm making for V1.0 I want to have redirects in place when I move content from a product manual section to the reference section, for example.

akeller:
@jwulf will you be leveraging this redirect method?

akeller:
I'm going to close this since it doesn't look like we ended up using this.

PR #229, opened 2021-05-06
Add Osano and Google Analytics

urbanisierung:
To get the scripts in the head on every site we need to implement a minor workaround by using the footer element and adding a `<head>` tag (see https://dev.to/rallipi/adding-scripts-to-every-page-of-a-docusaurus-project-3fee)

The Osano Banner will be displayed:

https://dev.to/rallipi/adding-scripts-to-every-page-of-a-docusaurus-project-3fee
image

And it looks like the google analytics plugin also works but this needs to be verified via the Google Analytics Console.

I recognized some errors in the console from the osano script when running the page locally.


akeller:
I added a commit with the Privacy Statement link.

Can you please take a look at adding the Cookie Preferences link? `<a href="#" onclick="Osano.cm.showDrawer('osano-cm-dom-info-dialog-open')">Cookie Preferences</a>`

urbanisierung:
replaced by https://github.com/camunda-cloud/camunda-cloud-documentation/pull/231

PR #241, opened 2021-05-10
Replace Zeebe Modeler with Camunda Modeler

menski:
- ignore versioned docs
- ignore getting started guide


akeller (approved):
This looks good to me. I caught a few typos, but they are not a part of this PR so I'll handle them in `release_0_27` directly. Thank you!

PR #248, opened 2021-05-10
Getting Started Updates for GA

akeller:


menski:
With the slug we now have `/docs/getting-started/` for the sign-up & login page and for the other pages `/docs/guides/getting-started/` is that expected? Or should the slug be also `/guides/getting-started/`?

akeller:
@menski good catch! Slug is now updated to '/guides/getting-started/`

Issue #257, opened 2021-05-18
Include Best Practices Blog Posts links


akeller:
berndruecker is in the middle of creating at least 3 Camunda Cloud Best Practices in the form of blog posts. 

* https://blog.bernd-ruecker.com/drafting-your-camunda-cloud-architecture-connecting-the-workflow-engine-with-your-world-3d94e8d404d6
* To be published - https://blog.bernd-ruecker.com/service-integration-patterns-with-bpmn-and-camunda-cloud-53b0f458e49
* To be written

We need to find a good place to add these blogs for reference.

A future topic of discussion would be splitting the Best Practices into a separate space or including them in the Camunda Platform Best Practices. For now, we will link to them in the docs as blog posts.

berndruecker:
@falko FYI

akeller:
@berndruecker when your 3rd blog is ready to go, be sure to send me the URL or add it here!

falko:
https://blog.bernd-ruecker.com/writing-good-workers-for-camunda-cloud-61d322cad862

berndruecker:
Thanks, Falko! I haven't found back this issue - but created a PR instead: https://github.com/camunda-cloud/camunda-cloud-documentation/pull/316

From my perspective, we could also close this issue now - with the best practices page already existing I am happy to simply provide PR's whenever I think a future blog post from me deserves a mention there. OK?

akeller:
Sounds good! Thank you both :)

PR #267, opened 2021-06-02
docs(product-manuals): Include Best Practices Blog Posts links (#257)

akeller:
Closes 257 for 2/3 blogs (last blog has not been written).

### DIFF
--- a/docs/product-manuals/best-practices.md
+++ b/docs/product-manuals/best-practices.md
@@ -1,0 +1,11 @@
+---
+id: best-practices
+title: Best Practices
+---
+
+Similar to our [Camunda Best Practices](https://camunda.com/best-practices/_/), we offer Best Practices with a mix of conceptual and practical implementation information for working with Camunda Cloud. Currently, these are available through blog posts. 
+
+* [Drafting Your Camunda Cloud Architecture: Connecting The Workflow Engine With Your World](https://blog.bernd-ruecker.com/drafting-your-camunda-cloud-architecture-connecting-the-workflow-engine-with-your-world-3d94e8d404d6)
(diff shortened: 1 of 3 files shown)


akeller:
@berndruecker @falko I put the links in Product Manuals > Zeebe > Technical Concepts. Does this work for you both?

berndruecker:
Hey Amara! I think this does work - but I also wondered if it might make sense to introduce an own top-level link to either something along the lines of "best practices" or "further resources" to make it more visible? 
I don't think it is part of the technical concepts of Zeebe - but if we don't want the hassle of introducing something new here we can still do that for the moment.

akeller:
We could create a section under Product Manuals > Concepts. Visibility-wise, would that work better for you? 

Thinking further into the future, do you think we should include Best Practices within the documentation? Or separate it out like our Camunda Platform docs and Best Practices today? If we want them to be separate in the future, I'm hesitant to have a dedicated page or dedicated section in the docs.

berndruecker:
Spontaneously I like that place ("section under Product Manuals") a bit better!

I don't have a strong opinion on having best practices as part of the docs or separate - the are arguments for both. My two cents: In the docs I think they are easier to find so people are hopefully more aware of the content - which is a good thing. But could lead to another discussion about gating the content for lead-gen :-)

akeller:
I will adjust this PR and create a new section! Thanks for your feedback.

akeller:
@christinaausley can you take a look at these changes for style & grammar?

christinaausley (commented):
Clean copy and concise -- no major grammatical/punctual adjustments needed!

PR #275, opened 2021-06-09
Document current partition layout and guarantees 

menski:


docs/product-manuals/zeebe/technical-concepts/partitions.md
```diff
@@ -37,6 +37,59 @@ A partition is a persistent append-only event stream. Initially, a partition is
...
+If no other configuration is specified the partitions are always distributed with the following guaranteed layout described in this pseudo code:
+
+```
+function nodes_for_partition(partition_id, number_of_nodes, replication_factor) {
+    for replica in range(0, replication_factor) {
+        node_id = ((partition_id + replica - 1) % number_of_nodes) + 1
+        print(node_id)
+    }
+}
+```
```
npepinpe, line 53: Not sure this is particularly helpful or clearing up what was just said. I think the table the view is actually much better to clarify things, i.e. the examples below.

docs/product-manuals/zeebe/technical-concepts/partitions.md
```diff
@@ -37,6 +37,59 @@ A partition is a persistent append-only event stream. Initially, a partition is
 
 For fault tolerance, data in a partition is replicated from the _leader_ of the partition to its _followers_. Followers are other Zeebe broker nodes that maintain a copy of the partition without performing event processing.
 
+## Partition Distribution
+
+Partitions are distributed in a round-robin fashion across all brokers in the cluster considering the number of nodes, number of partitions and the replication factor. For example the first partition will always be hosted by the first node, plus the next nodes based on the replication factor. The second partition will be hosted on the second node and the following to fulfil the replication factor.
```
npepinpe, line 42:
```suggestion
Partitions are distributed in a round-robin fashion across all brokers in the cluster considering the number of nodes, number of partitions and the replication factor. For example the first partition will always be hosted by the first node, plus the next nodes based on the replication factor. The second partition will be hosted on the second node and the following to fulfill the replication factor.
```

docs/product-manuals/zeebe/technical-concepts/partitions.md
menski: Thanks for the feedback, I was not sure about it in the first place. @akeller may I ask you for your opinion before deleting this part from the PR?

akeller: Just now seeing this and it looks like it was not part of the PR. Do you still need my thoughts here? I'm not sure this level of detail is super helpful. If we have advanced users interested in this kind of information we can add it in the future.

PR #300, opened 2021-06-29
Bpmn coverage

saig0:
* rework the BPMN coverage page
* replace the list of supported elements with a complete overview of BPMN elements
* mark all supported BPMN elements
* link the supported BPMN elements to their documentation pages
* apply the changes for all versions: latest, 1.0, 0.26, 0.25

was inspired by #46

closes #135

PR #301, opened 2021-06-29
Update BPMN coverage page

saig0:
Based on #300

* mark send tasks, business rule tasks, and script tasks as implemented


korthout (approved):
LGTM 👍

PR #307, opened 2021-07-07
Enable contextual-search for Algolia

### LINKED ISSUE #280: The search shows duplicated results for different versions
When I use the search in the documentation then it shows the same result multiple times. Each entry links to a different version.

image

As a user, this looks confusing and it doesn't show the version of the entry.

We could solve this problem by enabling the following feature: https://docusaurus.io/docs/search#contextual-search
If enabled then the search shows only the results of the selected version.

saig0:
* contextual-search is useful for versioned docs to avoid multiple results from different versions

See here: https://docusaurus.io/docs/search#contextual-search

closes #280

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -137,6 +137,7 @@ module.exports = {
     algolia: {
       apiKey: "f1b2a46296ae374b7f5a5c627341c354",
       indexName: "camunda",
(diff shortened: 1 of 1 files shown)


akeller (approved):
Thanks for adding the link in the comments!

Issue #304, opened 2021-07-01
Links on the BPMN coverage page are broken


saig0:
The BPMN coverage page contains all BPMN elements. The supported elements are highlighted. By clicking on a highlighted element, it should open the related doc page.

This works if the page is opened using the navigation. In this case, it doesn't contain a trailing slack `https://docs.camunda.io/docs/reference/bpmn-processes/bpmn-coverage`.

But if the page is opened directly with the URL in the browser then it doesn't work. Here, it adds a training slack to the URL `https://docs.camunda.io/docs/reference/bpmn-processes/bpmn-coverage/`

The trailing slack breaks the relative links on the page.

Maybe, related to https://github.com/facebook/docusaurus/issues/3372#issuecomment-857931379 which is released with version `2.0.0-beta.1`.

saig0:
@akeller / @menski please triage this issue (priority, when & who should work on it).

akeller:
I took a look at this and bumped the necessary dependencies and set `trailingSlash: false,` which appears to stop the trailing slash from being added if going to the URL directly, without a trailing slash in the URL, but it still doesn't seem to prevent someone from opening the link directly and including a trailing slash.

Let me create a branch for testing. I want to make sure this doesn't break anything else. I clicked around in the FEEL docs.

akeller:
This is merged in and live. I'm not able to reproduce the slash after adding the config. I'll also make sure we cleanup any links in that may include a trailing slash and get those removed. 

Please reopen if this is still broken!

slorber:
> I took a look at this and bumped the necessary dependencies and set trailingSlash: false, which appears to stop the trailing slash from being added if going to the URL directly, without a trailing slash in the URL, but it still doesn't seem to prevent someone from opening the link directly and including a trailing slash.

Using `trailingSlash: false` only affect how Docusaurus output HTML files:
- `/path.html` instead of `/path/index.html`
- links and canonical URLs without trailing slash

The rest is not under Docusaurus control. Hosts generally serve `/path.html` without a trailing slash, but may not redirect urls with a trailing slash with the non-trailing slash version. 

Use this resource to figure out the behavior of popular jamstack hosts: https://github.com/slorber/trailing-slash-guide

Note Docusaurus recommends against using relative links:
https://docusaurus.io/docs/next/docs-markdown-features#referencing-other-documents

akeller:
[...]

Issue #269, opened 2021-06-04
Update Guide has wrong url for 026 to 100


felix-mueller:
When opening https://docs.camunda.io/docs/1.0/guides/update-guide/introduction/ somehow sometimes the link to 0.26 -> 1.0 is dead, it sometimes points to https://docs.camunda.io/docs/1.0/guides/update-guide/introduction/026-to-100 instead of https://docs.camunda.io/docs/1.0/guides/update-guide/026-to-100
The side bar navigation however seems to work most of the times.

Sounds like URL slugs are causing problems

akeller:
I haven't been able to get this to fail since you reported it. Is this still a problem you can reproduce?

akeller:
Nevermind, this is the trailing slash problem like #304

akeller:
This is merged in and live. I'm not able to reproduce the slash after adding the config. I'll also make sure we cleanup any links in that may include a trailing slash and get those removed.

PR #311, opened 2021-07-12
Trailing slash config

akeller:
This PR addresses trailing slash concerns in #304 and #269.

* Add `trailingSlash` config, set to false
* Remove `trailingSlash` from sitemap plugin (deprecation warning)
* Update previous version docs to add .md to fix broken links


akeller:
Appears to stop the trailing slash from being added if going to the URL directly, without a trailing slash in the URL, but it still doesn't seem to prevent someone from opening the link directly and including a trailing slash.

menski (approved):
:shipit:

PR #317, opened 2021-07-15
Switch orange to green on BPMN coverage

falko:
The orange highlighting currently used in the BPMN coverage map
feels like a warning color that one would expect to be used for the unsupported elements.
In addition the color is not a Camunda brand color.
The replacement color is from the official Icemunda color scale
and fits nicely to the hover color.


falko:
@felix-mueller FYI

akeller:
Good catch! And a good reminder that I need to figure out a better way to completely overhaul the Camunda colors on Docusuaurus.

akeller:
I'll merge this in, but it likely will not go live until next week when we do a publish. Thanks!

PR #343, opened 2021-08-24
Zeebe editorial fix on instructions

christinaausley:
Updated Modeler section, providing updated next steps direction. Ran a small editorial pass across entire document for grammar and punctuation.


akeller (commented):
This makes much more sense to me. Did you get a chance to look at the other deployment guides in the Zeebe section? Will you also make a similar change?

PR #345, opened 2021-08-25
Testing tab structure for OS DX

akeller:
Testing some structural improvements for cross-OS compatibility without crowding the docs!


christinaausley:
I really like how these tabs look on the page -- super clean without listing every option out on the same page.

akeller:
We'll want to standardize on when to use tabs - should we say if there is more than one option? Two options? I see this being useful for programming language examples too.

christinaausley:
@akeller Discussed this in our 1:1 yesterday but commenting here for documentation purposes -- I like tabs for any more than one option given the cleanliness of the display.

PR #352, opened 2021-09-02
Add action to publish to datacenter

menski:
This change adds two new GitHub workflows to publish the docs to the Apache in the datacenter in Frankfurt. The old workflow to publish the GitHub pages still exists and will continue to work. Until we redirected the DNS entry nothing will change.


akeller (approved):
Thank you for no-indexing stage! This looks good to me :)

PR #357, opened 2021-09-08
Documentation/components implementation

christinaausley:
Replaced all references of product manuals with components according to docs vision. Passed local build but subject to further testing before merge.


akeller:
@menski for redirects, is it possible to have something like 

`https://docs.camunda.io/docs/product-manuals/*` -> `https://docs.camunda.io/docs/components/*`

Or do we have to list the URLs out individually for each page?

menski:
@akeller yes this is possible with something like
```
RewriteRule docs/product-manuals/(.*) docs/components/$1
```

You can try it out here https://htaccess.madewithlove.be/?share=e2453cfc-7ad3-44ed-a2cb-65755314c66d

akeller:
```
RewriteRule docs/product-manuals/(.*) docs/components/$1
RewriteRule docs/product-manuals docs/components
```

akeller (approved):
These look good! I'm only really concerned about the redirects, but I think these should be covered as well. Merge this to master today, then next week we can have @menski push live!

PR #359, opened 2021-09-13
Added intermediate none event to documentation

remcowesterhoud:
Add intermediate throw event support to the documentation


docs/reference/bpmn-processes/none-events/none-events.md
```diff
@@ -19,6 +19,12 @@ A process or subprocess can have multiple none end events. When a none end event
 
 If an activity has no outgoing sequence flow, it behaves the same as it would be connected to a none end event. When the activity is completed, the current execution path ends.
 
+## Intermediate None Event (throwing)
```
saig0, line 22: :wrench: We should align the name with the other sections.

```suggestion
## Intermediate none events (throwing)
```

docs/reference/bpmn-processes/none-events/none-events.md
```diff
@@ -19,6 +19,12 @@ A process or subprocess can have multiple none end events. When a none end event
...
+
+None events can be used to indicate some state achieved in the process.
```
saig0, line 24: :wrench: We could provide more context from a user perspective.

```suggestion
Intermediate none events can be used to indicate some state achieved in the process. They are especially useful for monitoring to understand how the process is doing, for example, as milestones or KPIs.
```

saig0 (changes requested):
@remcowesterhoud thanks :rocket: 

I've some minor suggestions. Please have a look.

Additionally,
* :x: we should mention that an intermediate none event can have variable output mappings (in a section below like [here](http://localhost:3000/docs/reference/bpmn-processes/message-events/message-events#variable-mappings). In the section, we should also mention that start events can have output mappings that is often used to initialize variables :sweat_smile: 
* :wrench: we could update the process diagram to include the intermediate none event

saig0 (approved):
@remcowesterhoud looks good :+1:

PR #363, opened 2021-09-16
Origin/editorialreview/zeebe

christinaausley:
Editorial review of Zeebe docs for grammar, punctuation, organization, and terminology.

### DIFF
--- a/docs/product-manuals/zeebe/deployment-guide/getting-started/deploy-a-process.md
+++ b/docs/product-manuals/zeebe/deployment-guide/getting-started/deploy-a-process.md
@@ -3,9 +3,11 @@ id: deploy-a-process
 title: "Deploy a process"
 ---
 
-_In this section, we're going to start up the Zeebe broker as well as Camunda Operate, a tool that gives you visibility into deployed processes and running process instances and contains tooling for fixing problems in those process instances._
+In this section, we'll start up the Zeebe broker and Camunda Operate—a tool that gives you visibility into deployed processes and running process instances, and contains tooling for fixing problems in those process instances.
 
-_We offer Operate free of charge for unrestricted **non-production** use because we think it's a great tool for getting familiar with Zeebe and building initial proofs-of-concept. And at this time, Operate is available for non-production use only. In the future, we'll offer an Operate enterprise license that allows for production use, too._
+:::note
+We offer Operate free of charge for unrestricted **non-production** use only because we think it's a great tool for getting familiar with Zeebe and building initial proofs-of-concept. In the future, we'll offer an Operate enterprise license that allows for production use, too.
+:::
 
 Before we run the Zeebe broker, we need to configure an Elasticsearch exporter in the Zeebe configuration file. Which leads to the question: what's an exporter, and why is Elasticsearch a part of this tutorial?
 
@@ -15,15 +17,19 @@ Elasticsearch is also what Camunda Operate uses to store data, so to run Operate
 
 ## Docker compose
 
-If you are using Docker and [zeebe-docker-compose](https://github.com/zeebe-io/zeebe-docker-compose) then follow the instructions in the README file in the `operate` directory of that repository to start Zeebe and Operate. Once you have done that, skip the following section, and continue from [check the status of the broker](#check-the-status-of-the-broker).
+If you are using Docker and [zeebe-docker-compose](https://github.com/zeebe-io/zeebe-docker-compose), follow the instructions in the README file in the `operate` directory of that repository to start Zeebe and Operate. Then, you can skip the following section, and continue from [check the status of the broker](#check-the-status-of-the-broker).
 
-If you are using individual components, then you will need to manually configure and start components.
+If you are using individual components, you'll need to manually configure and start components.
 
 ## Manually configure and start Zeebe and Operate
 
-_These instructions are for using separate components, and are not necessary when using Docker._
+:::note
+These instructions are for using separate components, and are not necessary when using Docker.
+:::
 
-First, copy the following lines into a new file `getting-started.yaml` file (in the `config` directory of the Zeebe broker).
+Take the following steps:
+
+1. Copy the following lines into a new file `getting-started.yaml` file (in the `config` directory of the Zeebe broker).
 
 ```
 zeebe:
@@ -35,11 +41,11 @@ zeebe:
 
 These settings enable the Zeebe Elasticsearch exporter.
 
-> **Note:** Some command examples might not work on Windows if you use cmd or
-> Powershell. For Windows users we recommend to use a bash-like shell, i.e. Git
-> Bash, Cygwin or MinGW for this guide.
+:::note
+Some command examples might not work on Windows if you use cmd or Powershell. For Windows users, we recommend using a bash-like shell such as Git Bash, Cygwin, or MinGW for this guide.
+:::
 
-Next, open Terminal or another command line tool and start up Elasticsearch.
+2. Open Terminal or another command line tool and start up Elasticsearch.
 
 ```
 cd elasticsearch-6.7.0
@@ -57,26 +63,26 @@ bin/elasticsearch
 bin\elasticsearch.bat
 ```
 
-You'll know that startup was successful when you see something like:
+You'll know startup was successful when you see something like the following:
 
 ```
 [2019-04-05T10:26:22,288][INFO ][o.e.n.Node ] [oy0juRR] started
 ```
 
-Then start the Zeebe broker in another Terminal window.
+3. Start the Zeebe broker in another Terminal window.
 
 ```
 ./bin/broker --spring.config.location=file:./config/getting-started.yaml
 ```
 
-And finally, start Operate in yet another Terminal window. Note that you'll need port 8080 in order to run Operate and access the UI, so be sure to check that it's available.
+4. Start Operate in another Terminal window. Note that you'll need port 8080 to run Operate and access the UI, so be sure to check it's available.
 
 ```
 cd camunda-operate-distro-1.0.0-RC2
 bin/operate
 ```
 
-To confirm that Operate was started, go to [http://localhost:8080](http://localhost:8080). You should see the following:
+To confirm Operate was started, go to [http://localhost:8080](http://localhost:8080). You should see the following:
 
 ![Zeebe Configuration File](assets/Operate-Login-Page.png)
 
@@ -88,7 +94,7 @@ You can use the Zeebe CLI to check the status of your broker. Open a new Termina
 
 If you are using Docker, change into the `zeebe-docker-compose` directory.
 
-If you are using separate components, then change into the Zeebe broker directory.
+If you are using separate components, change into the Zeebe broker directory.
 
 Run the following:
 
@@ -121,9 +127,9 @@ Brokers:
     Partition 0 : Leader
 ```
 
-For all Zeebe-related operations moving forward, we'll be using Zeebe's command-line interface (CLI). In a real-world deployment, you likely wouldn't rely on the CLI to send messages or create job workers. Rather, you would embed Zeebe clients in worker microservices that connect to the Zeebe engine.
+For all Zeebe-related operations moving forward, we'll use Zeebe's command-line interface (CLI). In a real-world deployment, you likely wouldn't rely on the CLI to send messages or create job workers. Rather, you would embed Zeebe clients in worker microservices that connect to the Zeebe engine.
 
-But for the sake of keeping this guide simple (and language agnostic), we're going to use the CLI.
+For the sake of keeping this guide simple (and language agnostic), we're going to use the CLI.
 
 ## Deploy the worklfow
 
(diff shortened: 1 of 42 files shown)


docs/product-manuals/zeebe/deployment-guide/getting-started/deploy-a-process.md
```diff
@@ -3,9 +3,11 @@ id: deploy-a-process
 title: "Deploy a process"
 ---
 
-_In this section, we're going to start up the Zeebe broker as well as Camunda Operate, a tool that gives you visibility into deployed processes and running process instances and contains tooling for fixing problems in those process instances._
+In this section, we'll start up the Zeebe broker and Camunda Operate—a tool that gives you visibility into deployed processes and running process instances, and contains tooling for fixing problems in those process instances.
```
pihme: :thought_balloon: With the new phrasing connection "and" it is less clear to me that Operate is the "tool that gives you visibility". Arguably, it maybe wasn't all that clear before.

So maybe something along the lines of: "In this section we'll start up the Zeebe broker - the underlying worklfow execution engine - and Operate - a tool that gives you..."

But then there might be too many hyphens. 

Also not sure why it's Zeebe broker but **Caumunda**  Operate since both are Camunda products. But that is more a marketing thing.

docs/product-manuals/zeebe/deployment-guide/getting-started/tutorial-setup.md
```diff
@@ -18,21 +18,21 @@ If this is your first time working with Zeebe, we expect this entire guide to ta
...
 
-1.  [Camunda Modeler](https://docs.camunda.io/docs/product-manuals/modeler/camunda-modeler/install-the-modeler): A desktop modeling tool that we'll use to create and configure our process before we deploy it to Zeebe.
-1.  [Zeebe Distribution](https://github.com/camunda-cloud/zeebe/releases/tag/0.20.0): The Zeebe distribution contains the workflow engine where we'll deploy our process model; the engine is also responsible for managing the state of active process instances. Included in the distro is the Zeebe CLI, which we'll use throughout the tutorial. Please use Zeebe 0.20.0.
+1.  [Camunda Modeler](https://docs.camunda.io/docs/product-manuals/modeler/camunda-modeler/install-the-modeler): A desktop modeling tool we'll use to create and configure our process before we deploy it to Zeebe.
+1.  [Zeebe Distribution](https://github.com/camunda-cloud/zeebe/releases/tag/0.20.0): The Zeebe distribution contains the workflow engine where we'll deploy our process model; the engine is also responsible for managing the state of active process instances. Included in the distro is the Zeebe CLI, which we'll use throughout the tutorial. Use Zeebe 0.20.0.
```
pihme: **Mandatory:** Wow, we definitely need to change this. They should use the latest version always. Best would be a variable so it always updated with each documentation version. Applies both to the reference in the text, and to the link)

docs/product-manuals/zeebe/deployment-guide/getting-started/tutorial-setup.md
```diff
@@ -18,21 +18,21 @@ If this is your first time working with Zeebe, we expect this entire guide to ta
...
 1.  [Camunda Operate](https://github.com/camunda-cloud/zeebe/releases/tag/0.20.0): An operations tool for monitoring and troubleshooting live process instances in Zeebe. Operate is currently available for free and unrestricted _non-production use_.
 1.  [Elasticsearch 6.8.0](https://www.elastic.co/downloads/past-releases/elasticsearch-6-8-0): An open-source distributed datastore that can connect to Zeebe to store process data for auditing, visualization, analysis, etc. Camunda Operate uses Elasticsearch as its underlying datastore, which is why you need to download Elasticsearch to complete this tutorial. Operate and Zeebe are compatible with Elasticsearch 6.8.0.
```
pihme: **Mandatory** This also needs to be changed to 7.x, but here I am not sure

@npepinpe Do we have a minimum version for elastic? How can we make sure that this part of the documentation stays up to date?

docs/product-manuals/zeebe/deployment-guide/operations/resource-planning.md
```diff
@@ -75,72 +73,73 @@ Environment Variables
...
-- `state` - the active state. Deployed processes, active process instances, etc. Completed process instances or jobs are removed.
-- `snapshot` - a state at a certain point in time
+- `segments` - The data of the log split into segments. The log is only appended, and its data can be deleted when it becomes part of a new snapshot.
+- `state` - The active state. Deployed processes, active process instances, etc. Completed process instances or jobs are removed.
+- `snapshot` - A state at a certain point in time.
 
 > **Pitfalls**
 >
-> If you want to avoid exploding your disk space usage, here are a few pitfalls to avoid:
+> To avoid exploding your disk space usage, here are a few pitfalls to avoid:
```
pihme: I still find the word "exploding" strange. Does that make sense to you, as a native speaker?

pihme (changes requested):
Review part 1

Some comments. Those marked *Mandatory* are mandatory changes, all others are optional

docs/product-manuals/zeebe/open-source/exporters.md
```diff
@@ -47,87 +44,69 @@ all exporters interact in the same way with the broker, which is defined by the
...
-> **Note:** although Zeebe tries to reduce the amount of duplicate records an
-> exporter has to handle, it is likely that it will have to; therefore, it is
-> necessary that export operations be idempotent.
->
-> This can be implemented either in the exporter itself, but if it exports to an
-> external system, it is recommended that you perform deduplication there to reduce
-> the load on Zeebe itself. Refer to the exporter specific documentation for how
-> this is meant to be achieved.
+:::note
+Although Zeebe tries to reduce the amount of duplicate records an exporter has to handle, it is likely that it will have to. Therefore, it is necessary that export operations be idempotent. This can be implemented either in the exporter itself, but if it exports to an external system, it is recommended you perform deduplication there to reduce the load on Zeebe itself. Refer to the exporter-specific documentation for how this is meant to be achieved.
```
pihme: I find the second half of this sentence hard to understand: "Although Zeebe tries to reduce the amount of duplicate records an exporter has to handle, it is likely that it will have to."

pihme (commented):
review part II (and finished)

More comments

docs/product-manuals/zeebe/deployment-guide/getting-started/deploy-a-process.md
christinaausley: Agreed on this -- your proposed text is much more clear. Added into latest commit.

docs/product-manuals/zeebe/deployment-guide/getting-started/tutorial-setup.md
christinaausley: Would you prefer to update and link to this: https://camunda.com/blog/2021/02/announcing-zeebe-operate-and-tasklist-1-0-0-alpha1/ ?

docs/product-manuals/zeebe/deployment-guide/operations/resource-planning.md
christinaausley: This is a little awkward. How's this:

"To avoid exceeding your disk space, here are a few pitfalls to avoid"

docs/product-manuals/zeebe/open-source/exporters.md
christinaausley: How's the following:

"Zeebe tries to reduce the amount of duplicate records an exporter handles, but it is helpful to ensure export operations are idempotent."

christinaausley:
@pihme Thank you so much for your review and very helpful comments! I've proposed changes to resolve these issues in the comments.

akeller:
@menski & @felix-mueller want to review all the deployment guides in the future, but I see there were some comments that came up in review for the Zeebe deployment guide.

My thoughts - if there are quick fixes to do today, let's do them. If there are more complicated changes (like how to reference new versions, what to link where, etc.) we can wait on this. I'd like to have minimal places to update with each new version and make use of pages like our Update Guides, Announcements, and other pages that *should* be updated with each version.

docs/product-manuals/zeebe/deployment-guide/getting-started/tutorial-setup.md
pihme: I would probably rephrase it to "Use the latest release" and then link to this page: https://github.com/camunda-cloud/zeebe/releases
The blog article is already outdated. Right now it would be: https://camunda.com/blog/2021/07/camunda-cloud-110-released/

But in a couple of weeks it will be 1.1.2. 

The link to the github page is a stable one, and then user can pick out the latest release for their needs

docs/product-manuals/zeebe/open-source/exporters.md
pihme: I feel both are not ideal. Exporters don't **have** to be idempotent. It is just a convenient shortcut to describe how they **should** operate.

Maybe we would be better served to describe the contract that an exporter has to follow.

Something like (tech jargon, not a suggestion how to phrase it): 
- exporters are called with at least once semantics for the records to be exported
- this means they might be called more than once for any given record
- deduplication can be performed based on the record key, which is unique

Also not sure what we mean by "exporter-specific documentation". I searched the documentation for a helpful section, but couldn't find it. If it is there, we should link to it. If not, I think we should remove this reference.

saig0:
@pihme nothing to add from my side :rocket:

docs/product-manuals/zeebe/open-source/exporters.md
christinaausley: Here's what I've updated to, does it make sense to you or are there pieces you would add/remove?

"It's best to reduce the amount of duplicate records an exporter handles alongside Zeebe. Exporters are called at least once for the records to export, which means they may be called more than once for any given record. We recommend performing deduplication based on the unique record key to reduce the load on Zeebe."

pihme: @saig0/@npepinpe : What do you think of the proposal? 

Personally, I am stumbling a bit over "reduce the load on Zeebe". I don't think that's the goal here, and also if deduplication is done badly it could hurt Zeebe performance. Plus, I think deduplication should rather be done on the backend.

At the same time, I feel getting it exactly right might be too complicated.

docs/product-manuals/zeebe/deployment-guide/getting-started/tutorial-setup.md
npepinpe: 7.x is the minimum mandatory version with Zeebe 1.x. I'm not sure how we would keep it up to date in an automated way =/ To be fair, the next time this needs to be done is with major versions, so should not happen very often. 

Generally, the tutorial would need to be updated, no? It still points to Zeebe 0.20.0, which is very outdated at this point, but I imagine this falls outside the scope of this PR?

docs/product-manuals/zeebe/open-source/exporters.md
npepinpe: You're right that reducing the load on Zeebe is not the goal of de-duplication. Further, this de-duplication should occur on your target system, not in Zeebe (in this sense, here it _is_ to reduce the load on Zeebe :smile:). 

So the important bits:

- Zeebe exports with at least once semantics, so you _will_ have to deal with duplicates
- De-duplication can be performed based on the position and partition ID - not all records have keys, but every record has a partition ID, and a unique position relative to that partition.

We can drop the reduce load part, and focus on communicating that it's crucial for the backend system to deal with duplicates. We could keep something to the effect of trying to do as little as possible in the exporter to reduce the load on the broker though. A note also on not blocking would be good, since blocking means blocking on the actor threads entirely, which can have a big impact on the system.

The exporter specific documentation here means the documentation for a specific exporter, e.g. Elasticsearch Exporter documentation, or the Kafka exporter documentation, etc. It's also possible some parts of our docs don't make sense in the new structure as well.

npepinpe (approved):
Nothing to add on my side - I think there's some outdated things (i.e. the tutorial on 0.20.0 for example) that were raised already, and there may be more, but (and correct me if I'm wrong) I think those were outside the scope of this PR.

docs/components/zeebe/open-source/exporters.md
```diff
@@ -94,7 +94,7 @@ set by the exporter itself once it can guarantee a record has been successfully
 updated.
 
 :::note
-Zeebe exports with at least once semantics, so you will have to deal with duplicates. It's best to reduce the amount of duplicate records an exporter handles alongside Zeebe, doing as little as possible in the exporter to reduce the load on the broker. We recommend performing deduplication based on the unique record key. We also recommend refraining from blocking, given blocking on the actor threads entirely, which may have a large impact on the system.
+Zeebe exports with at least once semantics, so you will have to deal with duplicates. It's best to reduce the amount of duplicate records an exporter handles alongside Zeebe, doing as little as possible in the exporter to reduce the load on the broker. We recommend performing deduplication in your target system. Deduplication can be performed based on the position and partition ID. Not all records have keys, but every record has a partition ID, and a unique position relative to that partition. We also recommend refraining from blocking, given blocking on the actor threads entirely, which may have a large impact on the system.
```
pihme, line 97: Suggestion: 

Zeebe exports with at least once semantics, so you will have to deal with duplicates. It's best to reduce the amount of duplicate records an exporter handles alongside Zeebe, doing as little as possible in the exporter to reduce the load on the broker. We recommend performing deduplication in your target system. Deduplication can be performed based on the partition ID and record index. Not all records have keys, but every record has a partition ID, and a unique position relative to that partition. We also recommend refraining from blocking operations in the exporter, as it will have a negative impact on performance of the system.

npepinpe: Sounds good, but I would keep using the word record position and not record index to keep the language consistent across the whole documentation (including the Javadoc).

christinaausley: Adjusted and resolved, thanks!

PR #364, opened 2021-09-17
Fix redirects and add trailing slash

menski:
- Update `.htaccess` with better regular expression matching and return `301` return code
- Configure docusaurus to always attach trailing slashes to better deal with Apaches default behavior of adding trailing slashes
- Update all broken links to adjust to trailing slash change

### DIFF
--- a/docs/components/clients/java-client/index.md
+++ b/docs/components/clients/java-client/index.md
@@ -27,7 +27,7 @@ Use the latest released version from [Maven Central](https://search.maven.org/ar
 
 ## Spring Integration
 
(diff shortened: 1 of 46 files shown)


akeller (approved):
This looks good to me. I checked a few links on stage and everything appears to be working as expected. I also removed the trailing slash from some URLs and was still able to get to the page as expected.

Now we just cross our fingers and hope the trailing slash issue is gone forever 🤞 

Thanks again for your assistance here @menski!

akeller:
@christinaausley FYI this will likely cause some merge conflicts in your open PRs with URL paths.

PR #367, opened 2021-09-20
add executionPlatform to quickstart.bpmn

### LINKED ISSUE #366: Example diagram is opened as Camunda Platform in Camunda Modeler
__Describe the Bug__

Model which can be downloaded from this Camunda Cloud docs page: https://docs.camunda.io/docs/guides/getting-started/model-your-first-process is opened as a Camunda Platform diagram in Camunda Modeler.

Camunda Platform is the default engine currently. To change that behaviour, the diagram has to:

- have Zeebe namespace, or
- have `xmlns:modeler="http://camunda.org/schema/modeler/1.0" modeler:executionPlatform="Camunda Cloud"` in the definitions

__Expected Behavior__

Model is opened as a Camunda Cloud diagram right away.

ultraschuppi:
closes #366


akeller (commented):
I don't think the `id="Definitions_0rbumvm"` changes. Can you double check? Also I found a couple other places we need to change. This file is just in the version-1.0 docs, not latest.

akeller (approved):

PR #369, opened 2021-09-24
fix: rename the IAM landing page

Ben-Sheppard:
I would like to rename the IAM landing page from something generic to something a little more inviting. Therefore I decided to go with `What is IAM?` which is a clear sign post to a user that within the page they will find out what IAM actually is and the role it plays.

### DIFF
--- a/docs/components/iam/overview.md
+++ b/docs/components/iam/overview.md
@@ -1,7 +1,0 @@
----
-id: overview
-title: "IAM"
-sidebar_label: "Overview"
----
-
-IAM provides authentication and authorization functionality along with user management.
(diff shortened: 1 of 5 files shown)


menski (changes requested):
As far as I can see the overview page is already live https://docs.camunda.io/docs/product-manuals/iam/overview, which means we should take care of not breaking existing links. Therefore I would recommend to add a redirect to this file https://github.com/camunda-cloud/camunda-cloud-documentation/blob/master/static/.htaccess

Can you take care of that? Otherwise let me know I can do it

menski (approved):
Looks good, thanks for adding the redirects

PR #383, opened 2021-10-11
Document zeebe gateway interceptors

### LINKED ISSUE #379: Document custom gateway interceptors
Camunda Cloud 1.2 will bring a new CCSM feature. Loading arbitrary interceptors in the gateway. See: https://github.com/camunda-cloud/zeebe/pull/7839. This functionality needs to be documented.

Presently, we feel this documentation belongs under `Product Manuals > Zeebe > Deployment Guide`. In the future this might be moved to `CCSM Features`.

The following should be documented:
- High level information about interceptors (this should link to grpc doc)
- How to write and package an interceptor with a simple example
- How to deploy/load an interceptor into the gateway
- Troubleshooting

korthout:
This adds documentation about the gRPC interceptors that can be loaded into the gateway.

closes #379
relates to camunda-cloud/zeebe#7839


docs/components/zeebe/deployment-guide/interceptors.md
```diff
@@ -0,0 +1,172 @@
...
+
+> This functionality is currently only available in Camunda Cloud Self-Managed.
+
+All communication from a client to a broker must first pass through a gateway.
+There they can be intercepted before being dispatched. Zeebe provides a way to
+load arbitrary interceptors into the gateway. Some typical examples of what you
+can accomplish with this include:
+
+- enforcing custom authorization rules on incoming calls
+- monitoring and logging of incoming calls
```
npepinpe, line 15:
```suggestion
- monitoring and logging of incoming calls (e.g. https://github.com/grpc-ecosystem/java-grpc-prometheus)
- distributed tracing (e.g. https://github.com/open-telemetry/opentelemetry-java-instrumentation)
```

docs/components/zeebe/deployment-guide/interceptors.md
```diff
@@ -0,0 +1,172 @@
...
+
+An implementation must adhere to the following requirements:
+- it implements
+  [ServerInterceptor](https://grpc.github.io/grpc-java/javadoc/io/grpc/ServerInterceptor.html)
+- it has public visibility
+- it has a public default constructor (i.e. no-arg constructor)
+
+Let's consider an interceptor that provides logging of incoming calls as an
+example. Other ServerInterceptor examples can be found in the official grpc-java
+[examples](https://github.com/grpc/grpc-java/tree/v1.41.0/examples).
```
npepinpe, line 31: :question: Should we point to master instead of a specific tagged version? :thinking:

docs/components/zeebe/deployment-guide/interceptors.md
```diff
@@ -0,0 +1,172 @@
...
+      @Override
+      public void onMessage(final ReqT message) {
+        LOGGER.trace("intercepted a call");
+        super.onMessage(message);
+      }
+    };
+  }
+}
+```
+
```
npepinpe, line 67: :wrench: I would add an example of the outcome of using this interceptor as well. e.g. "Loading this interceptor, we would then see `intercepted a call` for every request that passes through." - but better worded :smile:

docs/components/zeebe/deployment-guide/interceptors.md
```diff
@@ -0,0 +1,172 @@
...
+contain all classes (i.e. including all classes your own classes depend upon at
+runtime).
+
+There are many ways to do this, but for simplicity we'll use `javac` and `jar`
+directly. Note, that means we have to define a java manifest file by hand, in
+order to place the libraries' classes on the classpath.
+
+```sh
+# to compile LoggingInterceptor.java, we'll need to provide the grpc and slf4j api libraries
+javac -classpath .:lib/grpc-api.jar:lib/slf4j-api.jar ./LoggingInterceptor.java
```
npepinpe, line 79: :wrench: You don't need to add grpc-api and slf4j-api. These are provided by the gateway itself. The only time you need to provide them is if you're using a different version - however since we'll load the server interceptor from Zeebe, you probably want to fix grpc-api at least to always be the same version as Zeebe's. I'm not sure how to word all this in an easy to understand way, though - the idea is you only need to add libraries which are not already provided by the gateway, _but_ that's risky and makes you rely on them. I guess the ideal would be for us to have a list of dependencies which we guarantee are always provided - SLF4J, grpc, etc. ? Let me know what you think. I think it's also not too bad to otherwise tell users to always package everything, and add a note for advanced users who know what they're doing and explain how the JAR class loading work - you first look up a class in the JAR itself, and if it's not found, fallback to the Zeebe application's classloader. So some dependencies may be omitted at your own risk to reduce JAR file size.

docs/components/zeebe/deployment-guide/interceptors.md
```diff
@@ -0,0 +1,172 @@
...
+```yaml
+zeebe:
+  gateway:
+    ...
+
+    # allows specifying multiple interceptors
+    interceptors:
+
+      # determines the interception order
+      0:
```
npepinpe, line 122: :wrench: This isn't necessary though - you can also just do:

```yaml
interceptors:
  - id: logging-interceptor
    className: LoggingInterceptor
    jarPath: /tmp/LoggingInterceptor.jar
```

The `0` is only necessary if you want to configure it via environment variables in the environment variable name.

docs/components/zeebe/deployment-guide/interceptors.md
```diff
@@ -0,0 +1,172 @@
...
+`zeebe.gateway.interceptors` expects a list of interceptor configurations). The
+interception order index determines the order in which a call is intercepted by
+the different interceptors. Interceptors with a lower index always wrap an
+interceptor with a higher index. The interceptor at index 0 is thus the
+outermost interceptor. In other words, calls are intercepted first by the
+interceptor at index 0, followed by the interceptor at index 1, etc.
+
+## Troubleshooting
+
+Here we describe a few common errors. Hopefully, this will help you recognize
```
npepinpe, line 148: :+1:

docs/components/zeebe/deployment-guide/interceptors.md
```diff
@@ -0,0 +1,172 @@
...
+# lib/slf4j.jar
+```
+
+## Loading an interceptor into a gateway
+An interceptor can be loaded into your gateway as a fat JAR. For each
+interceptor, you need to provide your gateway with:
+- an interception order index
+- an identifier to identify this specific interceptor
+- where to find the JAR with the interceptor class
+- the fully qualified name of the interceptor class
```
npepinpe, line 103: :wrench: I would explain a bit what that means - not everyone is familiar with this terminology. Maybe at least give an example.

docs/components/zeebe/deployment-guide/interceptors.md
```diff
@@ -0,0 +1,172 @@
...
+Your ServerInterceptor implementation could not be found. Make sure you've
+configured the `className` correctly in the [gateway
+configuration](#loading-an-interceptor-into-a-gateway) and that your [JAR
+contains your class](#packaging-an-interceptor).
+
+**io.camunda.zeebe.gateway.interceptors.impl.InterceptorLoadException**
+Something went wrong trying to load your interceptor. Make sure your [JAR is
+packaged](#packaging-an-interceptor) correctly, i.e. it contains all runtime
+dependencies and specifies them in the manifest file's classpath. The exception
+should provide a clear description, but generally we distinquish the following
```
npepinpe, line 166:
```suggestion
should provide a clear description, but generally we distinguish the following
```

npepinpe (changes requested):
:+1: Very nice!

Regarding your questions:

- should we mention that the gateway is a Java application and that therefore the ServerInterceptor must be a Java class (or compatible)?
  - Yes, and we should mention that since it runs within Zeebe, the language level must be the same as Zeebe's or lower.
- should we mention that the jarPath is actually optional?
  - I would add a sub section about class loading in general. If the jarPath is omitted, then the interceptor is loaded directly from the Zeebe classpath. If it's present, then it's loaded from the JAR itself. When executing the interceptor, any classes are first looked up in the JAR as well, and if not found, will fall back to the Zeebe classpath. This means anything in Zeebe is available, but if you have dependencies not in Zeebe they must be in the JAR, and if you have dependencies in Zeebe but of different version they will be available in your interceptor (but won't affect Zeebe itself).
- should we mention anything about blocking in the onCall method vs blocking in the listener?
  - We could add a small best practices section which detail this? But I'm not sure if it's super important, meaning I don't know if many users will care about this.
- did I miss anything in my example?
  - I would just add a small example of the outcome maybe.

docs/components/zeebe/deployment-guide/interceptors.md
korthout: Oh interesting! I'll make to change the description accordingly

docs/components/zeebe/deployment-guide/interceptors.md
korthout: In order to compile this class, you need to add these api's. Otherwise you'll end up with compile errors.

Considering runtime dependencies, I though that we use a disconnected classloader for the interceptors. So they cannot mess with the zeebe classes. Perhaps I misunderstood that. As I understand it now from your description, our classloader is parent of the interceptors classloader, if a class in their loader wants to look-up a class defined in our classloader then it can actually reach it. Is that correct?

Coming back to the original example, I feel like an exhaustive list of dependencies we guarantee are available is a bit crazy. I'd rather we tell them to provide all runtime dependencies themselves and provide an actual fat JAR. That should work in all cases and is easiest to explain

PR #391, opened 2021-10-12
Create APIs & Clients section

akeller:
This PR creates a new APIs & Clients section. We'll migrate existing APIs & Clients docs, making it more visible for developers to find and work with them. We'll also setup redirects.


menski (approved):
Hi Christina, thank you so much for this, and sorry for my delayed review.

I really like the new structure,and think it's good that you combined the Console API topics into a single page.

One question I had when looking at the structure is if we intentionally choose the order inside of the API and Clients sections? At the moment they look a bit random for me, and are neither alphabetical sorted nor by "importance" nor a order in which I would expect to read them. For example the API section is currently ordered like this: Tasklist, Public API, Console, Zeebe

My proposal would be the following orders:
- API section: Public API, Zeebe, Tasklist, Console
- Clients section: Java, Go, CLI (this is already the case :+1:)
- Community Clients: Spring, JavaScript/NodeJs, C#, Python, Ruby, Rust

Regarding the redirects I check them, but we will better see if we missed something when it is on stage. I only found that the URL https://docs.camunda.io/docs/reference/cloud-console-api-clients/ is not covered, I guess we need 
```
RewriteRule ^docs/reference/cloud-console-api-clients/$ /docs/apis-clients/cloud-console-api-reference [R=301,L]
```

Feel free to decide on your own if you want to change the order, and you don't have to wait for a second review from me. I think its good to merge :shipit:

christinaausley:
@menski Thank you so much for your review -- great catches on the suggested fixes here. Went ahead and added a commit with your changes. Will merge once @akeller reviews either today/tomorrow.

akeller:
Good call on the ordering! I also agree on better seeing how things work on stage. LGTM!

Issue #266, opened 2021-06-02
Update Zeebe deployment guide to version 1.0


korthout:
# Product Manuals > Zeebe > Deployment guide > Getting started tutorial > Tutorial setup

[This tutorial](https://docs.camunda.io/docs/product-manuals/zeebe/deployment-guide/getting-started/tutorial-setup/) is based around an old version of zeebe (0.20). However, for a user it is hard to migrate from 0.20 to 1.0. They would be setting up things using configurations that no longer exists, using an old elastic search version, etc.

Let's update this guide to 1.0.

Zelldon:
Not sure what was the state before but it is still referencing elastic 6.8

@akeller do we have any plans on this?

Zelldon:
Always fine with deleting stuff it makes things easier

akeller:
I'll review (merge into existing guide if needed) and delete it with my upcoming docs changes.

PR #392, opened 2021-10-13
Create CCSM section

### LINKED ISSUE #270: Make the difference between Self-Managed vs. SaaS in terms of updates more clear
The page https://docs.camunda.io/docs/1.0/guides/update-guide/026-to-100 is very detailed which is very good for Self-Managed users.
For SaaS users it is a bit confusing since there are many steps which they don't have to do. 
We should somehow make clear what SaaS users have to do vs. Self-Managed users.
This would make it way easier to consume for many of our users.

akeller:
This pull request move all component deployment guides to a new dedicated CCSM section. The only content reviewed in this migration is Zeebe's Deployment tutorial/getting started guide. All other deployment guides will be reviewed in the future. Cross referencing links between CCSaaS content and CCSM content for each component in case users arrive at one and need to be routed to the other.

Redirects must be in place before this is merged! 

Closes #270 and #266.


akeller:
Pushing this to avoid more merge conflicts. Please see stage environment for review and open a new PR with any changes!

PR #405, opened 2021-10-21
improved error event docs using best practices

berndruecker:
As discussed via Slack - this PR should align usage of terms "tech/business error" with book and adjusts the example with these best practices


christinaausley:
Reviewed for grammar and style. I was only confused by line 38 -- could you clarify @berndruecker?

PR #411, opened 2021-10-27
Updated link to maintained/default docker-compose file

berndruecker:
Looking at the installation docs we are currently discussion, I recognized that the docs point to the community extension around docker-compose, which is quite outdated at the moment (old Zeebe versions, no tasklist, ...).

At the same time, the Zeebe team maintains one docker-compose file always using the latest version: https://github.com/camunda-cloud/camunda-cloud-get-started/blob/master/docker-compose.yaml - so we should simply point to this one instead.

### DIFF
--- a/docs/components/zeebe/deployment-guide/docker/install.md
+++ b/docs/components/zeebe/deployment-guide/docker/install.md
@@ -18,7 +18,7 @@ The easiest way to develop with Zeebe is using Docker. Docker provides a consist
 
 ### Docker configurations for docker-compose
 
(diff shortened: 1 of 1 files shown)


akeller:
@berndruecker the build failed on this, looks like an issue with the link(s). I can take a look at this, but if you get to it sooner run `npm run start` AND `npm run build` to see how those links are resolving.

berndruecker:
Ah damn it - I see the problem and will fix it. Sorry

berndruecker:
Replaced by https://github.com/camunda-cloud/camunda-cloud-documentation/pull/411

PR #414, opened 2021-10-27
address improvements on docs based on analytics research

christinaausley:
Minor adjustments to most popular pages based on analytics research to improve performance, add next steps, and optimize SEO.


christinaausley:
@akeller Do you feel adding keywords to these high performing pages would be worthwhile?

PR #425, opened 2021-11-02
feat: add to list of environmental variables

Ben-Sheppard:
Solves https://github.com/camunda-cloud/camunda-cloud-documentation/issues/390

I noticed in the forums that there were a couple of configuration variables that aren't as clear as they should be, notably `ENFORCE_HTTPS` and `ENFORCE_ACCESS_CONTROL`

### DIFF
--- a/docs/self-managed/iam/deployment/configuration-variables.md
+++ b/docs/self-managed/iam/deployment/configuration-variables.md
@@ -16,8 +16,8 @@ FEATURE_LDAP | Toggle LDAP support within IAM | false
 
 Environment variable | Description | Default value
 -------|----------|-------
-ENFORCE_ACCESS_CONTROL | Controls the enforcing of permissions for the IAM component | false
-ENFORCE_HTTPS | Controls if the URLs supplied for client configurations have to be http/https. | true
+ENFORCE_ACCESS_CONTROL | Controls the enforcing of permissions for the IAM component, if set to false, all users can access user, role and permission management. | false
+ENFORCE_HTTPS | Controls if the URLs specified for client configuration must be `https://` | true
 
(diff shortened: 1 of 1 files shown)


docs/components/iam/deployment/configuration-variables.md
```diff
@@ -12,6 +12,13 @@ Environment variable | Description | Default value
 -----|-------------|--------------
 FEATURE_LDAP | Toggle LDAP support within IAM | false
 
+### Functionality
+
+Environment variable | Description | Default value
+-------|----------|-------
+ENFORCE_ACCESS_CONTROL | Controls the enforcing of permissions for the IAM component | false
```
dlavrenuek, line 19: I would also add more details like "If set to `false`, all users can access user, roles and permission management."

docs/components/iam/deployment/configuration-variables.md
```diff
@@ -12,6 +12,13 @@ Environment variable | Description | Default value
...
+ENFORCE_HTTPS | Controls if the URLs supplied for client configurations have to be http/https. | true
```
dlavrenuek, line 20: When set to false, the urls can be http or https, so it would be more correct to state that the urls must be `https` when set to true. Not sure on the wording

dlavrenuek (changes requested):

christinaausley (approved):
From an editorial perspective, this is good to go.

dlavrenuek (dismissed):
Looks good 👍

christinaausley (approved):
From an editorial perspective, this is good to go.

dlavrenuek (approved):

PR #434, opened 2021-11-08
Console November Changes

ultraschuppi:
- new properties panel
- organization settings is now known as organisation management
already live on prod


christinaausley (approved):
From a technical writing perspective, this is good to go. Thanks for updating all of the references!

PR #436, opened 2021-11-09
Align sidebar page ordering in CCSM section (#435)

### LINKED ISSUE #435: Sidebar in CCSM section should list Zeebe first
Currently, the order of the menu items in the CCSM section is not aligned with how a user would usually install these components.
In most cases, a user will first install Zeebe and then the other components. Also, the ordering is not consistent with the bullet point list on the Overview page.

ACs:
- In the CCSM section, order of sections in the sidebar is aligned with bullet point list on the overview page

-----

Screenshots for illustration

image

image

meyerdan:
make page ordering in sidebar of CCSM section consistent with bullet
point list in on overview page and aligned with the order in which we
believe users will typically install the components (e.g. Zeebe first)

closes #435

### DIFF
--- a/sidebars.js
+++ b/sidebars.js
@@ -286,43 +286,6 @@ module.exports = {
   "Self-Managed": [
       "self-managed/overview",
       {
-      "Tasklist": [
-        "self-managed/tasklist-deployment/install-and-start",
-        "self-managed/tasklist-deployment/configuration",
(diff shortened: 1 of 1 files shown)


akeller (approved):
This looks great! Thanks Daniel! Keep in mind pushing to `master` does not automatically update the live version.

PR #437, opened 2021-11-10
Remove k8 operator

Zelldon:
K8 operator is no longer maintained and we should not encourage to use it.

\cc @menski 

closes https://github.com/camunda-cloud/camunda-cloud-documentation/issues/431


akeller (approved):
RIP 🪦 

Thanks for cleaning this up!

PR #438, opened 2021-11-11
Announce deprecation of `zeebe-test`

oleschoenburg:
We are deprecating `zeebe-test`: https://github.com/camunda-cloud/zeebe/issues/8143. We have already started to work on [a replacement](https://github.com/camunda-cloud/camunda-cloud-testing) but I've not mentioned that in the deprecation announcement because it's not finished yet.

@akeller This is my first documentation contribution so I'm happy to receive any suggestions. We can also discuss if you'd prefer we add more information to the announcement, for example a link to the new `camunda-cloud-testing` project.


akeller (approved):
This looks great! Simple and clear 😄

PR #443, opened 2021-11-17
Added hint where to download things

berndruecker:
I find it more convenient to repeat the download at the moment when I really need it. But I could also understand if we don't want to repeat that information all over the place. Take it as a suggestion and feel free to close.


akeller:
I like this 👍

PR #448, opened 2021-11-18
clarify java client cloud bootstrap

ingorichtsmeier:
Hi, I had difficultites bootstrapping my Java Client against my Camunda Cloud cluster. You can find the details of the problem and two solutions in this forum post: https://forum.camunda.io/t/bootstrap-java-worker-failed-with-unrecognized-field-refreshtoken/2610

After I got it running, I started to correct the docs. I would like to reference them directly in exercises in Camunda Academy.

My questions: 
* Is my change a valid improvement? 
* What is the best way to bootstrap a java client?
* Where can I find the current Java-Docs?

As I would like to link the Java Docs, I found that the link used in this page: https://javadoc.io/doc/io.zeebe/zeebe-client-java/latest/io/zeebe/client/ZeebeClientBuilder.html only has a version 1.0.0-alpha7 as latest version.

I would like to replace the bold class names with a link to the classes used to configure the client.

### DIFF
--- a/docs/apis-clients/java-client/index.md
+++ b/docs/apis-clients/java-client/index.md
@@ -44,6 +44,7 @@ In Java code, instantiate the client via the **ZeebeClientCloudBuilder** as foll
             .withClusterId(clusterId)
             .withClientId(clientId)
             .withClientSecret(clientSecret)
+            .withRegion("bru-2")
             .build();
 
     cloudClient.newTopologyRequest().send().join();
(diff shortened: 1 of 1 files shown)


akeller:
In general I like these changes, but I would like to have @menski do a technical review first and then I'll take care of a style guide review. 

I had a similar question about the Javadocs, I found the same like you did.

PR #455, opened 2021-11-26
Enable zoomable images

menski:


akeller (commented):
I like the idea here, but for larger screenshots like in the Modeler section, it makes the images even more difficult to see and work with. We can exclude those but there might be others we need to exclude as well.

ingorichtsmeier:
I prefer to use a fixed windows size to take the screenshots and limit it to 1200*800 pixels. I've installed a sizer tool to resize the browser or modeler to reasonable size. 

Pictures become harder to read as the screen size grows to 4K and beyond.

akeller:
@pepopowitz I wanted to get this topic on your radar. We'll likely close this PR and create a new one.

This incorporates an image zoom plugin for Docusaurus which works on some of our images but doesn't work with others. Can you investigate a little closer as you learn more about how Docusaurus works? Do we need to have a standard image size? Is there an alternative image zoom plugin? 

Something to keep in mind is we do have some screenshot scripting/automation running for components like Modeler. Our goal is to have something similar for all components, but this means we can't immediately introduce standard image sizes.

akeller:
@pepopowitz I'm closing this PR and creating a new issue for Zoomable images in the backlog. Given all the other changes Docusaurus core introduced recently, my guess is this will be a different implementation when we get to it.

PR #461, opened 2021-11-29
Update 110-to-120.md

felix-mueller:


felix-mueller:
@menski got feedback about this section in docs. Probably makes sense to adjust it - just wanted. to make a proposal, wdyt?

menski (approved):
thanks Felix

PR #469, opened 2021-11-30
Update Temporal expressions at expressions.md

tales-paiva:
Minor typo at the Temporal expressions section.


menski (approved):
Thanks again for fixing this

PR #474, opened 2021-12-01
Create troubleshooting section CCSM with log levels

### LINKED ISSUE #465: Add log level info to docs
Pull info from Zeebe wiki on log level info and modify for end user audience - https://github.com/camunda-cloud/zeebe/wiki/Logging#levels

----

TRACE: information which is helpful only if you want to trace the execution of a particular component. In general TRACE logging should be using granular loggers, i.e. loggers tied to very specific components, such that we can enable them for those to avoid creating too much noise.

DEBUG: information which can provide helpful context when debugging. To help you decide whether it is TRACE or DEBUG, or INFO, ask yourself: is this useful for a user, or just for a developer trying to diagnose what happened? It's also perfectly acceptable to add DEBUG messages right after an INFO one to provide more context.

INFO: information about the system which is useful for the user (note: in the case of the broker, user here is the SRE deploying it), e.g. leader changes, a new node added to or removed from the membership, etc.

WARN: expected errors (e.g. connection time outs, the remote node is unavailable, etc.) which may indicate that parts of the system are not working, and would require attention if they persist, but may resolve by themselves

ERROR: errors which require a person to look into them, e.g. log corruption, inconsistent log, anything which could shut down a partition, etc.

akeller:
Closes #465.

Any further thoughts here?

### DIFF
--- a/docs/self-managed/troubleshooting/log-levels.md
+++ b/docs/self-managed/troubleshooting/log-levels.md
@@ -1,0 +1,26 @@
+---
+id: log-levels
+title: "Log levels"
+description: "Learn about what information you can expect in various log levels and how to handle them"
+---
+
+When working with Camunda Cloud, you may see various messages in your logs. Not all messages require action.
+
+## Understanding log levels
+
+Camunda Cloud uses the following log levels:
+
+* TRACE: information which is helpful only if you want to trace the execution of a particular component. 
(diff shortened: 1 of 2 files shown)


akeller:
@christinaausley once I get more feedback, I'll add you as a reviewer for a style guide and editorial review.

menski (approved):
looks good to me

christinaausley (approved):
From a technical writing standpoint, LGTM. Made minor editorial adjustments in latest commit.

PR #477, opened 2021-12-02
change tasks to operations

ingorichtsmeier:
I got confused with the original wording. There is no option to complete tasks in Operate. In the GUI they a named operations.


akeller (approved):
Thank you @ingorichtsmeier!

akeller:
@christinaausley FYI, there might be other areas of the docs that don't match the UI for Operate.

PR #484, opened 2021-12-07
Fix wrong anchor links and add link checker to docs build workflow

menski:
This PR does two things to fix our internal link issues.

1. Add link check to build docs workflow to discover broken internal links before merge
    - adds steps to setup local apache webserver to test links
    - remove https redirect from htaccess to test locally
    - add mime type for bpmn, this is already configured in our prod apache  but is needed for the local test
2. Fix broken anchor links on master
    - remove non-existing anchor links
    - update anchors links to new locations
    - update links in all versions of docs

After we merge this PR we should be able to detect internal link issues before merging PRs. There is still the risk that external links are broken. I would opt not to check internal links on every PR, as that might be temporary issues and should block a PR, but I'm happy to be challenged on that. I will investigate to add external link check to our nightly check, to discover potential broken links in a timely manner.

### DIFF
--- a/.github/workflows/build-docs.yaml
+++ b/.github/workflows/build-docs.yaml
@@ -20,3 +20,18 @@ jobs:
         run: npm ci
       - name: Build
         run: npm run build
+      - name: Remove https redirect
+        run: |
+          sed -i '/SERVER_PORT/d;/SERVER_NAME/d' build/.htaccess
+      - name: Run web server
+        run: |
+          docker run -d --name webserver -v "$PWD/build":/app -p 8888:8080 bitnami/apache:latest
+          echo "container_ip=$(docker inspect -f '{{range.NetworkSettings.Networks}}{{.IPAddress}}{{end}}' webserver)" >> $GITHUB_ENV
+      - name: Waiting for URL available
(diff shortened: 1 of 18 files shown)


christinaausley (dismissed):
LGTM from a technical writing perspective. Made one minor adjustment in latest commit.

akeller:
> After we merge this PR we should be able to detect internal link issues before merging PRs. There is still the risk that external links are broken. I would opt not to check internal links on every PR, as that might be temporary issues and should block a PR, but I'm happy to be challenged on that. I will investigate to add external link check to our nightly check, to discover potential broken links in a timely manner.

Is there a way to check on release tag creation? I agree blocking a PR might be annoying, but I'd want to correct this ahead of pushing live.

akeller (dismissed):
This is awesome, thank you! I left a question, but regardless I approve this.

menski:
> Is there a way to check on release tag creation? I agree blocking a PR might be annoying, but I'd want to correct this ahead of pushing live.

We can for sure put this into the release pipeline as a preliminary check before deploying to production, I created #491 to track this improvement.

PR #486, opened 2021-12-08
Add possibility to render BPMN models

berndruecker:
With this change we can add BPMN models to a page, that are rendered by BPMN.io. You just need to add the following div:

`<div bpmn="bpmn-files/someModel.bpmn" callouts="taskA,taskB" />`

I will use this in the best practices


berndruecker:
I close it for now and simply bring that in with the best practices content - I think that is easier to handle

PR #499, opened 2021-12-13
chore(notices): add log4j2 notices

menski:


docs/reference/notices.md
```diff
@@ -8,7 +8,63 @@ description: "Let's take a closer look at security notices, reporting vulnerabil
...
+#### Product affected:
+
+IAM
+
+#### Impact:
+
+IAM bundles log4j libraries for which the following CVE has been published: https://nvd.nist.gov/vuln/detail/CVE-2021-44228.
+Specifically, IAM bundles log4j-api and log4j-to-slf4j. However, IAM does not bundle the log4j-core library which contains the vulnerability referred to by the CVE. As a result, Camunda does not consider IAM to be affected by the vulnerability.
+
+Still, Camunda recommends applying fixes as mentioned in Solutions below.
```
akeller, line 27:
```suggestion
Still, Camunda recommends applying fixes as mentioned in the Solution section below.
```

akeller (changes requested):
Small suggestions.

docs/reference/notices.md
```diff
@@ -8,7 +8,63 @@ description: "Let's take a closer look at security notices, reporting vulnerabil
...
+You are using IAM version <= 1.2.5
+
+#### Solution
+
+Camunda has provided the following releases which contain a fix
+
+- [IAM 1.2.6](https://github.com/camunda-cloud/zeebe/releases/tag/1.2.6)
+
+
+### Notice 1
```
meyerdan, line 40: I think I would make this "Notice 2" and IAM "Notice 1" so that Zeebe, Operate, Tasklist is on top. I know it does not make a difference long-term but right now it might make a difference

menski: I reorder the notices to have Zeebe be the most recent one

akeller (approved):
Looks good. Thanks for making the small changes!

PR #507, opened 2021-12-14
1.4.0 - Zeebe process test

remcowesterhoud:
Added documentation about the `zeebe-process-test` library. 

I have moved the deprecated `zeebe-test` page to a separate page and replaced the existing `testing` page with the documentation for `zeebe-process-test`.

This creates a bit of an awkward situation. In other parts of the documentation we are referring to `zeebe-process-test`, even though this library does not have a stable release yet. A stable version will come in Q1 2022.
I did decide to replace the references, as before they were referencing the deprecated project `zeebe-test`. In my mind it's better to reference to an upcoming project, than to a deprecated project.


pihme:
> This creates a bit of an awkward situation. In other parts of the documentation we are referring to `zeebe-process-test`, even though this library does not have a stable release yet. A stable version will come in Q1 2022.
> I did decide to replace the references, as before they were referencing the deprecated project `zeebe-test`. In my mind it's better to reference to an upcoming project, than to a deprecated project.

I see the awkwardness. I don't agree with your conclusion though. I think we should reference whatever is working right now. 
Could you summarize what the state of the deprecated code is? Is it just deprecated or is it removed with the upcoming minor release?

remcowesterhoud:
@pihme It is deprecated for removal in release `1.4.0`. With this same release `zeebe-process-test` will get a stable version.

pihme:
> @pihme It is deprecated for removal in release `1.4.0`. With this same release `zeebe-process-test` will get a stable version.

Then we should not merge it to master just yet, in my opinion. I think we once had a branch for "content for next release", but that would still be 1.3.0 and not 1.4.0.

@akeller How are we handling content for upcoming releases these days?

@remcowesterhoud Please let me know if you disagree.

christinaausley (dismissed):
From a technical writing perspective, this looks good. Well done on the clarity and conciseness of the Zeebe test file.

Made a few editorial/Markdown adjustments in the latest commit, and ensured all headings were in sentence-case format.

akeller:
> @akeller How are we handling content for upcoming releases these days?

I'll admit we've been working on so many things, we haven't been good about creating a release branch for upcoming releases. If this needs to go out with 1.3, I can create a release branch and we can change the target from `master` to `1.3`. Would that work? Otherwise it can hang out as-is, but I would recommend you convert it to a draft PR.

docs/apis-clients/java-client/zeebe-test.md
```diff
@@ -0,0 +1,74 @@
+---
+id: zeebe-test
+title: "Zeebe test"
+---
+
+You can use the `zeebe-test` module to write JUnit tests for your job worker and BPMN process. This provides a JUnit rule to bootstrap the broker and some basic assertions.
```
pihme, line 6: :x: If we target 1.4 as release date for this change, then I think this document is outdated. Please decide whether this PR is for now (and foreshadowing a new way to test things, or for 1.4 and replacing the old way to test things). Then change text accordingly.

pihme (changes requested):
Like the changes, but not fully consistent yet.

Please also look at Amaras comment regarding scheduling for future releases.

docs/apis-clients/java-client/zeebe-test.md
remcowesterhoud: This PR will be for 1.4. I've discussed with Ole and his plan was to remove this page in 1.4 anyways. So instead of fixing this, I will remove it.

People who are using a lower version and are still using this should still be able to find it using an older version of the docs.

christinaausley:
I've made minor adjustments in my latest commit, including the following:

- Minor typos.
- Adding a comma in "If, then" statements.
- Removing spacing when adding a slash between words, for example: testcontainer/embedded rather than testcontainer / embedded.
- Sentence-case headers
- Active voice

Nice work here. Looks good on my end.

christinaausley (approved):

pihme (dismissed):
Looks good to me. 

My only reservation is that the tone / level of technical detail doesn't really fit to the rest of the documentation. But if the tech writers are ok with it, so am I.

There is one broken link (see the failing check)

remcowesterhoud:
@pihme Could you give me another stamp of approval? I fixed the url and now the approvals are gone.

pihme (approved):
Stamp!

akeller:
I see @menski labeled this with `release/1.4.0`. Can you please adjust the target base from `master` to `release-1.4.0`?

akeller:
Merging in `release-1.4.0` should fix the build issues 😄  standby and cross your fingers 🤞 . If successful, I will merge.

PR #510, opened 2021-12-15
Mixpanel Integration

urbanisierung:
goal: add mixpanel events to docs
technical implementation:
- auth0 to silently get user token (so we have a user context with userId, orgId, ...)
- mixpanel to send always the same event `docs`. the mixpanel lib enriches the events with additional information like `current_url`, `browser`, `os`, ...
- osano analytics check: events will only be sent if we got consent from the user


urbanisierung:
@akeller I've converted the draft PR to a real one. From a technical point of view this is ready to ship.

akeller (approved):
Let's go!! 🚀

PR #511, opened 2021-12-15
Add icon and color to Self-Managed nav bar

akeller:


akeller:
Rethinking this, will open a new PR.

PR #514, opened 2021-12-16
chore(docs): rename authentication to transport layer security section

### LINKED ISSUE #493: Authentication section title
[This section of the docs](https://docs.camunda.io/docs/self-managed/zeebe-deployment/security/authentication/
) is called “Authentication”.

Strictly speaking, that is kinda accurate, because it is describing authentication of the server through TLS.

However, most people would think of this as encryption plus maybe server authentication (it’s more like identity validation).

When the audience is thinking Authentication, they are more likely to be thinking of client authentication, ie: “_how do I allow authenticated clients only to perform operations_” .

So I think we should change this section title to “Transport Layer Security”.

menski:
closes #493

### DIFF
--- a/docs/self-managed/zeebe-deployment/security/security.md
+++ b/docs/self-managed/zeebe-deployment/security/security.md
@@ -7,6 +7,6 @@ description: "This document analyzes Zeebe's security features."
 
 Zeebe supports two security features you should be aware of:
 
-- **[Authentication](../authentication)** - allows you to secure communication between clients and gateways.
+- **[Transport Layer Security](../transport-layer-security)** - allows you to secure communication between clients and gateways.
 - **[Authorization](../authorization)** - allows you to supply access credentials to the client so these can be validated by a reverse proxy placed before the gateway.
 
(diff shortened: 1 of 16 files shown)


akeller:
Just confirming, you intentionally want to update this everywhere @menski? I don't necessarily disagree, I just wasn't expecting 16 files to be changed to address this 😅

menski:
I thought it would make more sense to keep it consistent between versions. Don't really see the why we want to have a page with the same content call differently between versions.

akeller (approved):
LGTM thanks for adding the redirects too!

menski:
This has to be reworked as we merged https://github.com/camunda-cloud/camunda-cloud-documentation/pull/441

akeller:
@menski do you think you'll get around to this for `release-1.4.0`?

akeller:
@akeller rework or open new PR with these rename changes

akeller:
@menski Quick question, since I may get to this later this week - is there a reason you wrote out "transport layer security" instead of using the well known abbreviation TLS? And do we need to specify the version? Is it TLS 1.3? Or do you just want to distinguish between SSL and TLS?

menski:
> is there a reason you wrote out "transport layer security" instead of using the well known abbreviation TLS? And do we need to specify the version

As I'm never sure what are well known abbreviation, that's why I spelled it out. I wouldn't specify a version. it depends on the java implementation

akeller:
Will open a new PR with smaller changes to add TLS abbreviation.

PR #516, opened 2021-12-17
Mixpanel Events: fix duplicate events

urbanisierung:
Problem: Mixpanel events are sent twice, I think it's somewhere part of the react/docusaurus lifecycpe api
Workaround: send max one event per second


ultraschuppi (approved):
parcour!

PR #537, opened 2022-01-05
Update Operate-Tasklist screenshots for variables panel

huygur:


akeller:
@menski @santanaluiz was this missed for 1.3? Can you review and merge?

menski:
@santanaluiz will take care of the review, we will push it live after we merged it

santanaluiz (dismissed):
The screenshots look mostly fine. The only point is that we have some inconsistencies.
Like part of them show the "Non production license" and part not.

Seems like not a big deal from my point of view. 
@menski wdyt?

menski:
Let's move forward and fix it with the next round of screenshot updates, I guess in general we should not produce screenshots with the non-prod warning. @santanaluiz maybe something you can communicate into the team

menski:
@huygur if you have time, go ahead

menski:
Thanks you @huygur :+1:

menski (approved):

PR #547, opened 2022-01-07
feat(optimize): add ccsm configuration setup documentation

RomanJRW:
relates to OPT-5827

Configuration page for Optimize CCSM. Please do not merge, we still need to test this as part of a complete docker-compose file with all other components, and also confirm that no config is missing here.

It's also not clear where documentation should live for shared components such as IAM/Zeebe/Elasticsearch. Open to suggestions for how to make this clearer and/or more complete

### DIFF
--- a/docs/self-managed/optimize-deployment/setup.md
+++ b/docs/self-managed/optimize-deployment/setup.md
@@ -1,0 +1,62 @@
+---
+id: setup
+title: "Self-Managed setup"
+description: "Install and configure Optimize Self-Managed."
+---
+
+## Install using Docker
+
+The `camunda/optimize:latest` docker image can be used to run Optimize in Self-Managed as a container. Certain environment
+variables need to be set for this to work correctly. See below for an example of how this could be done as
+part of a docker-compose file:
+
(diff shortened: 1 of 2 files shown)


docs/self-managed/optimize-deployment/setup.md
```diff
@@ -0,0 +1,62 @@
+---
+id: setup
+title: "Self-Managed setup"
+description: "Install and configure Optimize Self-Managed."
+---
+
+## Install using Docker
+
+The `camunda/optimize:latest` docker image can be used to run Optimize in Self-Managed as a container. Certain environment
+variables need to set for this to work correctly. See below for an example of how this could be done as
```
megglos, line 10: is the be missing here?
```suggestion
variables need to be set for this to work correctly. See below for an example of how this could be done as
```

megglos (approved):
LGTM! Small language suggestion.

christinaausley (approved):
One comment and minor adjustments made in latest commit. LGTM, but let me know if you have any questions.

PR #562, opened 2022-01-13
Add warning for known issue in Zeebe 1.3.0

oleschoenburg:
Warn users about a known issue in the Zeebe 1.3.0 docker image

### DIFF
--- a/docs/guides/update-guide/120-to-130.md
+++ b/docs/guides/update-guide/120-to-130.md
@@ -9,6 +9,16 @@ The following sections explain which adjustments must be made to migrate from Ca
 
 ## Server
 
+### Zeebe
+
+:::caution
+The docker image for Zeebe 1.3.0 does not set a system locale.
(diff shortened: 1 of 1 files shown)


docs/guides/update-guide/120-to-130.md
```diff
@@ -9,6 +9,16 @@ The following sections explain which adjustments must be made to migrate from Ca
 
 ## Server
 
+### Zeebe
+
+:::caution
+The docker image for Zeebe 1.3.0 does not set a system locale.
+This can lead to user data corruption, in particular with non-ascii characters (such as umlauts) in process variables.
```
menski, line 16:
```suggestion
This can lead to problems with user data, like process variables, which use non-ASCII characters (such as umlauts).
```

My proposal is to not call this corruption, as this is makes users quite uncomfortable, wdyt?

menski: we agreed it is data corruption

menski (approved):

PR #600, opened 2022-02-11
Removing deprecated references

Hafflgav:
Removing references to zeebe-docker-compose because it is depricated


berndruecker (approved):

berndruecker:
Yeah great!

Actually we should even think about removing simple monitor here - but that's for another day :-)

PR #485, opened 2021-12-08
Fixes modeler properties panel

Hafflgav:
Changed: 
* Minor Changes in the service, send/receive, business-rule and script task documentation 
* Additional fixes for Call Activities 
* And for Message Events

A rather major change has been made to the paragraph of variable mappings.  
Please review these changes and rephrase if necessary. :) 

It would be awesome if you could leave this branch open since I am going to do some more stuff in here.

PR #601, opened 2022-02-11
Added variable propagation in xml

Hafflgav:
Originates from PR #485 (Fixes modeler properties panel)


korthout (approved):
LGTM 🏅 Thanks for this!

EDIT: Please consider also changing this for the older product versions.

PR #635, opened 2022-02-22
Supporting DMN

akeller:
**This content was initially pulled over from Camunda Platform docs and should be reviewed entirely!**

* All links have been commented with `[//]: # ` and inline links have been left alone (unless they caused runtime issues, in which they link to `#`. 
* Only some internal to dmn-engine links have been adjusted

Local builds and commits will fail for link issues and redirects, which is why this PR will be initially in draft mode.

FYI @saig0 @korthout @barmac @christinaausley


akeller:
Quick note about location and structure - this is subject to change and I'll likely revisit it later this week. This may also include renaming "Modeler" as a section to "Modeling" or "Modeler & Modeling" to include the Modeler components, BPMN, & DMN.

akeller:
@saig0 need any help over here?

saig0:
@akeller thank you for the offer. I checked the PR. Unfortunately, these are the wrong pages :sweat_smile: 
I'm working on a new PR that copies the DMN reference docs from C7.

saig0:
Closing this PR in favor of #775.

PR #642, opened 2022-02-25
Typo

nevries:


akeller (approved):
Great catch! Thank you! 

As an FYI, we have not extensively reviewed all Best Practices yet, so if you find more before we do, please PR them! 🤗

PR #651, opened 2022-02-28
chore(desktopModeler): update screenshots

MaxTru:


marstamm (approved):
LGTM, altough the "alpha.1" might be a bit weird. I guess we will update them again when the Major version was released?

MaxTru:
> LGTM, altough the "alpha.1" might be a bit weird. I guess we will update them again when the Major version was released?

Yes, also we should think about allowing users to provide a custom configuration to the Modeler which specifies a "custom value" for the version indicator. Will create an issue

MaxTru:
Cf https://github.com/camunda/camunda-modeler/issues/2790 for follow-up

PR #654, opened 2022-02-28
resolving error link in forms guide

christinaausley:
It seems the same link, but variation between either `latest` or `7.16` triggers entirely separate docs. I've adjusted the link to `7.16`, but do you have thoughts here @akeller on why `latest` defaults back to the guide? I couldn't find any relevant redirects but perhaps I've missed something here.


akeller:
Is this content not in docs.camunda.io? We are redirecting `latest` and `7.16` to docs.camunda.io content, so this is expected behavior.

akeller:
Ok, so we are trying to get info from https://github.com/camunda/camunda-docs-manual/blob/master/content/reference/forms/camunda-forms/_index.md? 

Let's go ahead and get this moved into Modeler section since it looks like this was a miss. If it's applicable to both Desktop and Web Modeler, create a folder "Forms" under "DMN" and add this as a new page. This increases the scope, but don't feel like you need to drop everything to make this happen. Let me know if you need help.

christinaausley:
@akeller Done. Should I contact IT for the redirect?

akeller:
@christinaausley can you sync with @MaxTru to make sure that this info is applicable to Camunda Forms on Camunda Cloud + Web Modeler and Camunda Cloud + Desktop Modeler? Async via this PR or more synchronous via Slack or meeting, whichever you two prefer.

christinaausley:
@akeller Synced with Max and can confirm it is, Platform + Cloud as well as Web and Desktop. Do you want this placed within DMN or within its own folder?

akeller:
I know we just talked, but following up for completeness - please place this in its own "Forms" folder for a bit of future proofing.

christinaausley:
@akeller Excellent! This is already placed within its own folder within Modeler -- LGTM?

Let me know how to handle the redirect as well.

akeller:
@christinaausley you'll need to create a helpdesk ticket for IT. Give them the old URL and the new URL. Since this content should have been part of the Modeler migration, have them include `latest`, `develop`, and `7.17`.

If you are feeling extra generous, you can provide them with the rewrite rules yourself. They are similar to how we do them for internal Cloud redirects. You can see some examples here on a gist I made previously - https://gist.github.com/akeller/693c651cc7ad77473ea25261d586cb3e (Keep in mind the URLs with # don't actually work, but I don't think you'll run into that here)

christinaausley:
Update: Rewrites are in place so merging to master.

PR #657, opened 2022-02-28
resolving lingering comments in documentation

christinaausley:
Removing all lingering inline comments.


akeller (approved):
Wow, good catch on finding the others! Thank you!

PR #662, opened 2022-03-03
feat(desktopModeler): migrate GitHub docs to cloud-docs

MaxTru:
* This migrates the existing camunda modeler docs (https://github.com/camunda/camunda-modeler/tree/develop/docs) 1:1 to this documentation repo


akeller (commented):
Looks like the images are not coming through quite right, I'll take a look and see if we can get this cleaned up! Otherwise, it looks good to me.

akeller:
> Looks like the images are not coming through quite right, I'll take a look and see if we can get this cleaned up! Otherwise, it looks good to me.

Whoops sorry, the issue was on my side! This looks great!

akeller (approved):

PR #664, opened 2022-03-04
feat(.htaccess): provide static route for desktop-modeler

MaxTru:
* This will allow us to use the .../desktop-modeler/ URL as entry point
  when linking to desktop modeler
* This allows us to dynamically adjust in the future which entry point
  will be used, without changes in the linking docs


akeller (approved):
This approach makes sense to me, thank you!

PR #669, opened 2022-03-04
updating migration tooling link

christinaausley:
Resolving 404 error with migration tooling link.


akeller (approved):
Thanks! I think this was redirecting just fine on GitHub's side, but always good to get these cleaned up and changed.

PR #680, opened 2022-03-11
Docs guides humantask edits

christinaausley:
Resolving @berndruecker comments and suggestions for Human Task Orchestration guide.


christinaausley:
@berndruecker
1. I've added a clear note on creating forms, but per our Slack conversation, would we prefer to remove the form all together? For length of doc purposes, I didn't step through forms (as we have a separate guide dedicated to this,) but rather tried to tease the concept so users would be encouraged to learn more. I did include the forms guide as a link in the next steps as well. Would you rather remove the form entirely from this guide, or add a sentence directing the user to the forms guide?
2. I've added more description on the exclusive gateway and how the token moves forward.
3. I adjusted the screenshot with the sequence flow error, but note the blue "comment" field is actually representation of the included form JSON. Would we prefer to apply the BPMN model inclusion you mentioned? I'm not sure how this works, but happy to learn if you'd like to make an adjustment in this PR.

docs/guides/getting-started-orchestrate-human-tasks.md
```diff
@@ -41,7 +41,7 @@ To create an automated process with user tasks, take the following steps:
 9. Append a gateway to your user task by dragging it onto the dashboard from the palette on the left side of the screen, or by clicking on the user task `Decide what's for dinner` and clicking on the element you'd like to create next. In this case, we've selected the diamond icon to create a gateway.
 10. Create two sequence flows (represented by the arrows) from the gateway and two new user tasks based on what the user decides to eat. In this case, we've named ours `Prepare chicken` and `Prepare salad`.
```
berndruecker, line 42: I think exactly in step 10 we need a hint that expressions will be needed, otherwise this process cannot be deployed successfully.
And to create the expressions, you will need a variable name, that comes from the form eaelier on. This is IMHO not yet clear

docs/guides/getting-started-orchestrate-human-tasks.md
```diff
@@ -41,7 +41,7 @@ To create an automated process with user tasks, take the following steps:
...
 11. Attach an end event to the two user tasks.
-![gateway example](./img/gateway-example-1.png)
+![gateway example](./img/gateway-example-dinner.png)
```
berndruecker, line 44: Here you could use bpmn.js instead and add:

`<div bpmn="getting-started-guides/prepare-dinner.bpmn" />`

You could also add callouts to tasks by adding: callouts="DecideDinner,PrepareChicken" 

And then explain this by refering to the callout here in the text:

`<span className="callout">1</span>`

But not urgent at the moment I think - just wanted to flag for awareness.

berndruecker (changes requested):

docs/guides/getting-started-orchestrate-human-tasks.md
christinaausley: @akeller Thoughts here? Do you think pulling in expressions, callouts, etc, will make this too complex for a new user? Do we think pulling out the form will make things too thin? If the former sounds reasonable, I can plug in a few sentences referencing expressions.

docs/guides/getting-started-orchestrate-human-tasks.md
christinaausley: I'm not familiar with implementing BPMN attachments in Markdown. Have I inserted this correctly in the latest commit? On build it doesn't look quite right.

christinaausley: @berndruecker I can also just cherry pick out this commit (where I added the BPMN) and you can implement the BPMN in a subsequent PR. Just let me know what's best.

docs/guides/getting-started-orchestrate-human-tasks.md
christinaausley: @berndruecker I've added some details on the creation of the form and affiliated expressions within my latest commit. Thoughts?

docs/guides/getting-started-orchestrate-human-tasks.md
```diff
@@ -40,8 +40,21 @@ To create an automated process with user tasks, take the following steps:
 ![user task example](./img/user-task-example.png)
 9. Append a gateway to your user task by dragging it onto the dashboard from the palette on the left side of the screen, or by clicking on the user task `Decide what's for dinner` and clicking on the element you'd like to create next. In this case, we've selected the diamond icon to create a gateway.
 10. Create two sequence flows (represented by the arrows) from the gateway and two new user tasks based on what the user decides to eat. In this case, we've named ours `Prepare chicken` and `Prepare salad`.
+   - Note that the sequence flows require [expressions](../components/concepts/expressions.md) to access variables from the form we'll create below to determine what to eat for dinner. To add an expression, click on the sequence flow to view the properties panel, and open the **Condition** tab to insert a conditional expression.
 11. Attach an end event to the two user tasks.
-![gateway example](./img/gateway-example-1.png)
+
+<div bpmn="./img/prepare-dinner.bpmn" />
```
berndruecker: I fixed how this should look like in this commit: https://github.com/camunda-cloud/camunda-cloud-documentation/commit/c60ec0b90cab06bd7b50337cb862d9699ca7225f. But to make it work you have to cherry-pick the other commit from master

akeller:
Anything you need from me here?

docs/guides/getting-started-orchestrate-human-tasks.md
christinaausley: Which commit needs to be removed from master?

christinaausley:
@akeller Nope -- should have this wrapped up this week.

PR #684, opened 2022-03-14
chore(formsReference): remove outdated info

MaxTru:
* We don't explain how to create forms in the about page. However, that
  is okay, since it is obvious. Just removing this to not have invalid
content in the docs.


akeller (approved):
Thank you! LGTM!

PR #694, opened 2022-03-17
docs(supported-environments): simplify updates

### LINKED ISSUE #561: Simplify Supported Environments page
https://docs.camunda.io/docs/reference/supported-environments/

Currently supported environments page goes into every component separately. 
We should consider simplifying this. If there are differences, we should really think about the WHY and try to align as much as possible. This might also lead to changes in supported environments in the end.

akeller:
Based on internal docu and discussion.

This is a draft since we are waiting for details from Optimize.

Closes #561.

### DIFF
--- a/docs/reference/supported-environments.md
+++ b/docs/reference/supported-environments.md
@@ -43,6 +43,12 @@ Requirements for the components can be seen below:
 | Tasklist | OpenJDK 11+ | Elasticsearch 7.16.x |
 | Optimize | OpenJDK 11+ | Elasticsearch 7.8.x - 7.16.x |
 
+:::note Elasticsearch support
+[Elastic's Elasticsearch](https://www.elastic.co/elasticsearch/) is the only supported verison of Elastic compatible with Camunda Platform 8. 
+
+AWS Opensearch is not supported at this time.
+:::
+
 ### Version Matrix
 
 This overview shows which Zeebe version works with which Modeler, Operate, Tasklist and Optimize:
--- a/docs/reference/supported-environments.md
+++ b/docs/reference/supported-environments.md
@@ -14,7 +14,7 @@ description: "Find out where to run Camunda Platform 8 components for SaaS and S
 - **Zeebe Go Client**: Go 1.13+
 - **zbctl**: Windows, MacOS, and Linux (latest)
 
-_Hint: There are more [community-maintained Camunda Platform 8 clients](./apis-clients/community-clients/)._
+_Hint: There are more [community-maintained Camunda Platform 8 clients](./apis-clients/community-clients/index.md)._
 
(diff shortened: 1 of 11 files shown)


docs/reference/supported-environments.md
```diff
@@ -1,85 +1,68 @@
...
-- Oracle JDK/JRE 11
-- Open JDK/JRE 11 including builds of the following products:
- - Adopt OpenJDK
+### Desktop Modeler
+- Windows 7 / 10
+- Mac OS X 10.11
+- Ubuntu LTS (latest)
 
-### Docker
+## Camunda Cloud Self-Managed
```
felix-mueller, line 30: probably we can call it Camunda Platform 8 Self-Managed

docs/reference/supported-environments.md
```diff
@@ -1,85 +1,68 @@
 ---
 id: supported-environments
 title: "Supported environments"
-description: "Let's take a closer look at supported environments alongside Zeebe, Operate, and Tasklist."
+description: "Find out where to run Camunda Cloud components for SaaS and Self-Managed, including Optimize for both Camunda Cloud and Camunda Platform 7."
 ---
 
-## Zeebe
+## Camunda Cloud SaaS & Self-Managed
```
felix-mueller, line 7: Camunda Platform 8 SaaS & Self-Managed

docs/reference/supported-environments.md
```diff
@@ -1,85 +1,68 @@
 ---
 id: supported-environments
 title: "Supported environments"
-description: "Let's take a closer look at supported environments alongside Zeebe, Operate, and Tasklist."
+description: "Find out where to run Camunda Cloud components for SaaS and Self-Managed, including Optimize for both Camunda Cloud and Camunda Platform 7."
```
felix-mueller, line 4: Camunda Platform 8 and Camunda Platform 7

akeller:
@felix-mueller good catch, my brain is still not in full rename mode! 

Anything else?

akeller:
I thought we were having merge conflict-type issues, but it turns out the build is failing for real reasons!

Some of the older migration update versions for Optimize include references to connecting to Elasticsearch 6.2.0+. Since we are supporting Camunda Platform 7 users, we likely need to have this information as well. Or do we only need to support through 3.4.x, and if so, can I remove the migration update guides for 2.1 through 3.3?

felix-mueller:
I added Optimize 3.3 because officially still support it till end of june 22. 
We should keep the update guides, but I don't think we need to cover all old versions here for now in supported envs. 
do you think we are done with this?

akeller:
Looks like the build action was stuck, let me rerun this, and then we can get it merged!

PR #696, opened 2022-03-18
chore(desktopModeler): add API stability claimer

MaxTru:
* The element template and plugin API might have breaking changes in the
  future
* Since we in the past (at least for C7) guranteed stability, we should
  clearly highlight, that breaking changes might occur.
* closes bpmn-io/internal-docs#491


nikku:
Works for me, thanks @MaxTru

akeller (approved):
I adjusted the formatting to use a [Docusaurus note admonition](https://docusaurus.io/docs/markdown-features/admonitions).

PR #704, opened 2022-03-22
chore(element-templates): bump schema version

marstamm:


docs/components/modeler/desktop-modeler/element-templates/additional-resources.md
```diff
@@ -5,6 +5,6 @@ title: Additional resources
 
 Use the [Camunda Cloud Connect plugin](https://docs.camunda.org/cawemo/latest/technical-guide/integrations/modeler/) to integrate the Camunda Modeler with [Cawemo](https://cawemo.com/) and retrieve templates from a Cawemo catalog project.
 
-Try playing around with custom elements and some [example templates](https://github.com/camunda/camunda-modeler/blob/master/resources/element-templates/samples.json).
+Try playing around with custom elements and some [example templates](https://github.com/camunda/camunda-modeler/tree/master/resources/element-templates).
```
MaxTru, line 8: I think we should directly point to cloud (C8) and platfrom (c7) samples.

Something like:
```
Try playing around with custom elements and our example templates for [Camunda Platform 7](https://github.com/camunda/camunda-modeler/tree/master/resources/element-templates/samples.json) or [Camunda Platform 8](tbd).
```

Reason: make it simpler, avoid confusion.

WDYT?

marstamm: I could not find a link to the RPA example in the docs, do we link to them from somewhere else? I would argue that those examples are valuable to some users as well and would not be discoverable otherwise

marstamm: For the cloud link location, I choose `cloud-samples.json`, as this would be available if we released develop today. If we decide to rename/move the example file, we would have to update this link

marstamm:
Ready for re-review

MaxTru (approved):

PR #716, opened 2022-03-28
Document new DeployResource RPC and its usage

### LINKED ISSUE #707: Document new DeployResource rpc
The upcoming release will introduce a new DeployResource RPC (see camunda/zeebe#8924) to the [Zeebe API (gRPC)](https://docs.camunda.io/docs/apis-clients/grpc/).

We'll need to make sure that our API doc is up to date, so we'll have to document this new RPC.
The current design can be found [here](https://github.com/camunda/zeebe/blob/korthout-8065-grpc-deploy-request/gateway-protocol/src/main/proto/gateway.proto#L475-L487), but is subject to change until camunda/zeebe#8924 is accepted and merged.

### LINKED ISSUE #708: Mark DeployProcess RPC as deprecated
The upcoming release will introduce a new DeployResource RPC (see https://github.com/camunda/zeebe/pull/8924) to the [Zeebe API (gRPC)](https://docs.camunda.io/docs/apis-clients/grpc/), which will replace the existing [DeployProcess RPC](https://docs.camunda.io/docs/apis-clients/grpc/#deployprocess-rpc).

We'll need to make sure that our API doc is up to date, so we'll have to document that the DeployProcess RPC will be deprecated with the upcoming release.

korthout:
This PR:
- documents the new DeployResource RPC
- deprecates the existing DeployProcess RPC
- updates all zbctl, Go-client and Java-client usage in the docs to their respective new uses of this RPC.

closes #707 
closes #708

### DIFF
--- a/docs/apis-clients/go-client/get-started.md
+++ b/docs/apis-clients/go-client/get-started.md
@@ -135,7 +135,7 @@ The broker stores the process under its BPMN process id and assigns a version.
 ```go
 	// After the client is created
 	ctx := context.Background()
-	response, err := client.NewDeployProcessCommand().AddResourceFile("order-process.bpmn").Send(ctx)
+	response, err := client.NewDeployResourceCommand().AddResourceFile("order-process.bpmn").Send(ctx)
 	if err != nil {
 		panic(err)
 	}
@@ -253,7 +253,7 @@ func main() {
 
 	// deploy process
 	ctx := context.Background()
-	response, err := zbClient.NewDeployProcessCommand().AddResourceFile("order-process-4.bpmn").Send(ctx)
+	response, err := zbClient.NewDeployResourceCommand().AddResourceFile("order-process-4.bpmn").Send(ctx)
 	if err != nil {
 		panic(err)
 	}
--- a/docs/apis-clients/go-client/get-started.md
+++ b/docs/apis-clients/go-client/get-started.md
@@ -35,7 +35,7 @@ module github.com/zb-user/zb-example
 
 go 1.17
 
-require github.com/camunda-cloud/zeebe/clients/go v1.2.9
+require github.com/camunda/zeebe/clients/go/v8 v8.0.0
 ```
 
 3. Set the connection settings and client credentials as environment variables:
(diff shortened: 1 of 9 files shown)


docs/apis-clients/grpc.md
```diff
@@ -647,6 +677,84 @@ Returned if:
 
 - Retries is not greater than 0.
 
+## Deprecated RPCs
+
+The following RPCs are exposed by the gateway service, but have been deprecated.
```
saig0, line 682: :wrench: We could move this section under `Technical error handling`. The error handling is more relevant than duplicated RPCs. 

At some point, it may be good to extract the RPCs into different pages.

docs/apis-clients/go-client/get-started.md
```diff
@@ -135,7 +135,7 @@ The broker stores the process under its BPMN process id and assigns a version.
 ```go
 	// After the client is created
 	ctx := context.Background()
-	response, err := client.NewDeployProcessCommand().AddResourceFile("order-process.bpmn").Send(ctx)
+	response, err := client.NewDeployResourceCommand().AddResourceFile("order-process.bpmn").Send(ctx)
```
saig0, line 138: :thought_balloon: Do we need to adjust the output too?

The output of the command is written in the docs under the code examples. If the output is changed then we should adjust it on this page too.

saig0 (approved):
@korthout looks good :+1: Thank you for updating all docs :rocket: 

I have only two minor comments. Please have a look.

docs/apis-clients/go-client/get-started.md
korthout: 👍 Good point. 

I did not check this. I went with a search tool through all matches and made changes where I thought they were necessary. I missed this one, but I totally agree that output should be updated as well.

korthout: This took a bit more time because we discovered a problem in the v8 client (see https://github.com/camunda/zeebe/pull/9062).

korthout:
@saig0 Since I've had to adjust a bit more in the getting started guide, I think it makes sense that you take another look. Once approved, feel free to merge it.

saig0 (approved):
Good job @korthout. Thank you for the adjustments :+1:

PR #718, opened 2022-03-29
chore: ease transform transitions

menski:
Adds a transition to the transformations to reduce the jumping and flickering


akeller (approved):
THANK YOU!

PR #720, opened 2022-03-29
Add banner

akeller:
Banner with link to meta page.


akeller:
WDYT of this banner text @christinaausley? Simple but I think it gets the message across. Link will not work since it's not relative.

christinaausley:
I like it a lot! I would change the wording slightly to "Learn more about navigating the updated documentation here." with the period after "here." Additionally, maybe it's just my browser, but it's cut off in my local build?
Screen Shot 2022-03-30 at 11 43 29 AM

akeller:
@christinaausley I cannot reproduce this, so let's remember to look at the staging environment when we push this to see if it's still happening to you.

PR #728, opened 2022-03-30
add token simulation page to web modeler doc

nithinssabu:


christinaausley (approved):
This is so cool. I made a few adjustments in my latest commit, including:

- Sentence case header
- Descriptive description
- Bolding button/UI references

Approving from a technical perspective, but I think this should be targeted to merge with `release-1.4.0` as the `master` branch is technically "frozen" currently.

As far as location goes, @akeller may disagree, but I think this is long enough to warrant its own page currently. If it was shorter, I'd say go ahead and add it, but we can always just link to it from the `Model your first diagram` page so folks are aware of the feature and can take a deeper dive into its capabilities if they'd like to after getting started. Just my thoughts.

akeller:
Hi folks - before I take a closer look at this, is this part of the upcoming release or something we need to change immediately? Please adjust the merge target to `release-1.4.0`.

akeller:
> As far as location goes, @akeller may disagree, but I think this is long enough to warrant its own page currently. If it was shorter, I'd say go ahead and add it, but we can always just link to it from the `Model your first diagram` page so folks are aware of the feature and can take a deeper dive into its capabilities if they'd like to after getting started. Just my thoughts.

I agree with @christinaausley here. And she even beat me to asking about the merge target 🚀

nithinssabu:
> Hi folks - before I take a closer look at this, is this part of the upcoming release or something we need to change immediately? Please adjust the merge target to `release-1.4.0`.

This is something to release immediately. We can release the feature only after this page is available in the documentation as there is a link to this page from the web modeler. Everything else is ready on our side.

CatalinaMoisuc:
> > Hi folks - before I take a closer look at this, is this part of the upcoming release or something we need to change immediately? Please adjust the merge target to `release-1.4.0`.
> 
> This is something to release immediately. We can release the feature only after this page is available in the documentation as there is a link to this page from the web modeler. Everything else is ready on our side.

Thanks @akeller  and @christinaausley for the quick review! 💪🏻

Then let's just merge it and release Token Simulation.
@nithinssabu please organize with the Cloud team a release of the documentation to production.

CatalinaMoisuc (approved):

PR #741, opened 2022-04-01
Add security notice for Zeebe & Spring4Shell

oleschoenburg:
Adds a security notice for Zeebe regarding Spring4Shell. 

Some notes for the review

* Patched versions are not yet released, so I've left the "solution" part very generic.
* Older versions of Zeebe are also bundling unpatched versions of Spring but I've not mentioned them here since we don't support them anymore.
* We _could_ use more reassuring language when explaining that Zeebe is very likely not at all affected.
* We should probably extend this notice with information from other components like Operate and Tasklist

### DIFF
--- a/docs/reference/notices.md
+++ b/docs/reference/notices.md
@@ -12,24 +12,29 @@ Camunda publishes security notices after fixes are available.
 
 #### Publication Date:
 
-April 1st, 2022
+April 11th, 2022
 
-### Product affected:
+#### Product affected:
 
-Zeebe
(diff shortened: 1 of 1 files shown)


christinaausley:
@oleschoenburg Note that I've adjusted a minor typo in my latest commit but otherwise will leave this to @menski 👍

menski (dismissed):
I updated the notice to include all projects

menski (approved):

PR #751, opened 2022-04-04
chore(desktopModeler): change title of deploy diagram page

MaxTru:
* This makes the desktop modeler guide easier to read. 
* `Connect to Camunda Platform 8` is rather unspecific (what does connect mean?). Actually this page is about deploying after all and nothing else


akeller (approved):
Makes sense to me!

PR #764, opened 2022-04-05
Renaming for migration guide

berndruecker:


berndruecker:
my bad - wrong merge target

Issue #761, opened 2022-04-05
Update Migration Guide


berndruecker:
- Renaming
- Double-check content

berndruecker:
https://github.com/camunda-cloud/camunda-cloud-documentation/pull/765

akeller:
Merged 😄

PR #765, opened 2022-04-05
Final touches on Migration Guide (#761)

berndruecker:


berndruecker:
@akeller I wondered if we can easily change the URL from "migrating-from-Camunda-Platform.md" to "migrating-from-Camunda7.md" - do I simply need to provide a rewrite rule for it in https://github.com/camunda-cloud/camunda-cloud-documentation/blob/master/static/.htaccess? Then I think I would also add this to this PR

berndruecker:
I wanted to make the URL shorter and easier to remember. Product Marketing allows to speak about Camunda 7 and Camunda 8 - and especially around migration, this gets pretty verbose otherwise. But I am happy to keep the platform in the official URL as a good compromise :-) Thx! I might add this to the PR (once I am over my Github Trauma of not mastering branches correctly ;-))

docs/guides/migrating-from-Camunda-Platform.md
```diff
@@ -533,47 +487,32 @@ The ```bpmn message name``` is used in both products and doesn't need migration.
 
 ## Adjusting your DMN models
 
-For Camunda Cloud, [a former community extension](https://github.com/camunda-community-hub/dmn-scala), built by core Camunda developers, is productized. This engine has a higher coverage of DMN elements. This engine can execute DMN models designed for Camunda Platform 7, however, there are some small differences which can be assessed looking at [this issue](https://github.com/camunda/camunda-modeler/issues/2525).
+For Camunda 8, [a former community extension](https://github.com/camunda-community-hub/dmn-scala), built by core Camunda developers, is productized. This engine has a higher coverage of DMN elements. This engine can execute DMN models designed for Camunda 7, however, there are some small differences listed below.
```
felix-mueller, line 490: I think from Modeling perspective there are two different types again, right? One DMN that works on C7 only and one that works on C8 only? Should we mention how it can be manually migrated?

@MaxTru can maybe give more context

MaxTru: Camunda 7 DMN files also work on Camunda 8 DMN Engine, **if** the Camunda specifics were not set: (history time to live, version tag, expression languages different than FEEL (basically what is already mentioned here). 

So IMO the text proposed here is already accurate.


Context for Modeling: from Modeling perspective we differentiate DMN for C7 and DMN for C8 diagrams by custom attributes which we set as part of the definitions:
```
<definitions ...  xmlns:modeler="http://camunda.org/schema/modeler/1.0" modeler:executionPlatform="Camunda Cloud" modeler:executionPlatformVersion="8.0.0">
```
=> the respective DMN Modeling editor does then offer / not offer to model the C7 custom attributes (e.g., history time to live)

berndruecker: Looks like we can keep the text for now (Which might not be perfect anyway)?

berndruecker:
Ready for review

docs/guides/migrating-from-Camunda-Platform.md
felix-mueller: Hmm, as a User I would still ask:
"How would I open a C7 DMN as a C8 DMN to make it "compatible" if I have one of these attributes?" Like how do I go to the C8 mode in modeler?

MaxTru: True, but (just as for BPMN) we do not have an "easy" answer in terms of "click this migration button". So the answer could only be to manually clean up the diagram (in the XML), which I believe would not be satisfying.

So if this is a need, then we should maybe prioritize to offer some semi-automated / guided migration feature soon I would say. WDYT?

cc @christian-konrad

akeller:
@berndruecker I'm in the process of fixing your build issue and specifying Camunda Platform 7 and Camunda Platform 8. With URLs I'm fine to shorten them (we can always iterate and do this later), but in the written docs content let's spell them all the way out while users are still getting used to distinguishing them.

akeller (approved):
LGTM! Special thanks to @christinaausley for getting the clean build!

docs/guides/migrating-from-Camunda-Platform.md
berndruecker: I can look at if the migration tooling (https://github.com/camunda-community-hub/camunda-platform-to-cloud-migration) is able to do the required changes - but I am sure this will not be there before launch (as the biggest complexity is to get the environment running with DMN files and probably use a different set of dms-js or whatever). @MaxTru Unless you have a good idea for a quick fix on https://github.com/camunda-community-hub/camunda-platform-to-cloud-migration/tree/main/camunda-modeler-plugin-platform-to-cloud-converter to be applicable to DMN files and make necessary changes?

berndruecker: Moving discussion to https://github.com/camunda-cloud/camunda-cloud-documentation/issues/780 so we can close it here

PR #766, opened 2022-04-05
Optimize opt 6100

grlimacan:


docs/self-managed/optimize-deployment/setup.md
```diff
@@ -56,6 +57,7 @@ CAMUNDA_OPTIMIZE_ZEEBE_PARTITION_COUNT | The number of partitions configured in
 CAMUNDA_OPTIMIZE_SHARING_ENABLED | Enable/disable the possibility to share reports and dashboards. | true
 CAMUNDA_OPTIMIZE_UI_LOGOUT_HIDDEN | Disables the logout button (logout is handled by Identity). | true
 SPRING_SECURITY_OAUTH2_RESOURCESERVER_JWT_JWK_SET_URI | Authentication for the Public REST API using a resource server to validate the JWT token. Complete URI to get public keys for JWT validation | null
+CAMUNDA_OPTIMIZE_API_AUDIENCE | Optimize tries to match this with the <code>aud</code> field contained in the JWT token. Only used when the JWK_SET_URI above is set. This is not necessarily the same as <code>CAMUNDA_OPTIMIZE_IDENTITY_AUDIENCE</code>, in case a service other than Identity is being used to generate the API tokens| optimize
```
RomanJRW, line 60:
```suggestion
CAMUNDA_OPTIMIZE_API_AUDIENCE | Optimize tries to match this with the <code>aud</code> field contained in the JWT token. Only used when the JWK_SET_URI above is set. This is not necessarily the same as <code>CAMUNDA_OPTIMIZE_IDENTITY_AUDIENCE</code>, in case a service other than Identity is being used to generate the API tokens| optimize-api
```
Should this be consistent with the example above? Both should probably match the default

RomanJRW (approved):
LGTM, one comment 👍

PR #767, opened 2022-04-05
Documented max payload size

berndruecker:
https://github.com/camunda-cloud/camunda-cloud-documentation/issues/396

### DIFF
--- a/docs/components/best-practices/development/handling-data-in-processes.md
+++ b/docs/components/best-practices/development/handling-data-in-processes.md
@@ -72,7 +72,7 @@ Camunda does not treat BPMN **data objects** (<img src="/img/bpmn-elements/data-
 
 Do not excessively use process variables. As a rule of thumb, store *as few variables as possible* within Camunda.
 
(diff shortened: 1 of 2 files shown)


berndruecker:
@npepinpe approved it content-wise: "one small thing, it could be smaller if the user has configured the maxMessageSize to be smaller, but it cannot really be greater as we don't provide a way to configure the gRPC payload size limit."

@berndruecker : "Make sense to me. I don't think we have to discuss the "maxMessageSize" that prominently, as I would consider that the exception that it is configured - while everybody can be hit by this limitation you should be aware of. Thanks a lot!"

berndruecker:
@akeller Good to go from my end

akeller:
FYI Looks like a change in this PR is causing the build issue.

berndruecker:
Wired - but I found a typo that might be it @akeller

christinaausley (approved):
Adjusted typos and added a note admonition in my latest commit. Looks good on my end.

akeller:
Teamwork makes the dream work 💪

PR #770, opened 2022-04-05
Console docs to product links

christinaausley:
This PR includes renaming and redirect efforts for Console product to docs links based on the existing [Handbook page](https://confluence.camunda.com/display/HAN/Direct+Links+From+Products+to+Docs).


christinaausley:
@akeller -- Would you also like me to rename Cloud Console API Reference and implement the appropriate redirect for this doc?

akeller:
I just pushed the C8 Console PR, which gave you conflicts. Sorry about that! 

If I'm understanding correctly, you are looking at changing the URL for the page currently here - https://c8.docs.camunda.io/docs/apis-clients/cloud-console-api-reference/

Yes, go ahead with the change. 

A quick note here (and maybe you can verify by logging into the product) but Cloud API Management and Console API will likely retain their name (unless the team changed it). This is technically cloud in the lowercase "c" sense, shortened from cloud computing but for the URL I would recommend `/console-api-reference/`.

akeller:
Changes to `product-links.txt` may cause weird build issues since that's part of the link checking that happens when you commit to a remote branch, so when this is ready to merge we should be cautious of what other PRs are open at that time.

akeller:
I think this is still missing some of the recent changes, but now I'm realizing they are in the sections I haven't renamed yet 🙈 don't mind me! I'll review and see if we can get this merged.

akeller (approved):
LGTM!

PR #775, opened 2022-04-06
Add DMN reference docs

saig0:
## Description

* copy the DMN reference docs, including the images, from [Camunda Platform 7](https://docs.camunda.org/manual/7.16/reference/dmn/)
* adjust page headers
* fix references and images
* remove content for unsupported features
* add hints to use names without symbols


saig0:
@remcowesterhoud please have a **brief** look at the changes. You don't need to review the whole content because it is basically copied from C7 and we don't want to do major improvements now.

PR #776, opened 2022-04-06
docs(zeebe-deployment): add TLS abbreviation

akeller:
@menski I only found one place where we didn't reference TLS by abbreviation, so I went ahead and updated it. Most of the current documentation (I assume since your original PR was opened) uses the abbreviation 😄


menski:
works for me, way less changes as my original PR :shipit:

PR #778, opened 2022-04-07
Element templates updates

pinussilvestrus:
Related to https://github.com/bpmn-io/internal-docs/issues/499

* updated the supported schema version
* added documentation for `documentationRef`
* added `icon` documentation

I keep the icons docs rather short, intentionally for now. We considered adding more best practice guidelines for those, but since @currandwyer is out this week, I think we can later add them if we see the need.


MaxTru (approved):
:rocket:

PR #782, opened 2022-04-07
change the request string

davidgs:
The `--request` syntax doesn't actually work on macOS. the `-X POST` syntax is cross-platform.


akeller:
@davidgs I'm happy to change this, but I'm admittedly a bit confused. In the [cURL man pages](https://curl.se/docs/manpage.html) these are equivalent options. Can I ask what terminal you are using?

PR #785, opened 2022-04-07
cross linking between guides and concepts

christinaausley:
Offering a series of cross links between our Concepts and Guides documentation, as much is related for a beginner's reference.


akeller:
I'm liking these choices! Let me review with fresh eyes tomorrow 😄

akeller (approved):
I'm good with these and I'm confident we will iterate in the future. LGTM!

PR #786, opened 2022-04-07
Exporters to self managed

akeller:
No content changes, only moving pages.


akeller:
@menski @felix-mueller any issues with this move? Anything you want to add?

felix-mueller:
fine for me

PR #793, opened 2022-04-11
Add hint for testing rewrite rules to the guidelines

remcowesterhoud:
Running the documentation does not allow testing redirects. The rules in .htaccess can be used with online tooling. To help people with this an example tool has been added to the guidelines.


remcowesterhoud:
@saig0 As discussed last week, I've added the hint to the docs :)

PR #795, opened 2022-04-11
Docs optimize environment fix

christinaausley:
Adjusting environment requirement as Optimize requires at least 7.10.


akeller (approved):
Approving solely because this is in fact the number 10, like you wanted it to be 😆

PR #800, opened 2022-04-11
adjust and implement C7 links

christinaausley:
I've checked through all of the existing C7 links within C8 docs and adjusted just a few that can remain within C8 docs. I've also taken a look at where we might need links to C7 (DMN, for example,) and feel we've done an appropriate job of linking to C7 where needed without causing confusion.

I've flagged a sentence in best practices I'll need to update with a link once connectors docs are merged (and I'll need to revisit the connectors docs for links once merged anyways.)


akeller:
@christinaausley connectors docs merged just now to `release-1.4.0`

akeller (approved):
Works for me, thank you!

PR #805, opened 2022-04-12
docs(getting-started): updates for C8

akeller:
@ultraschuppi FYI 🙂


akeller:
You'll notice I didn't update the links from Camunda Cloud to Camunda Platform yet.

ultraschuppi:
@akeller i will do another round of screenshots updates once we hit production. you took screenshots from preprod - they are not entirely the same ;)

PR #808, opened 2022-04-12
Move to Java 17

ingorichtsmeier:
as it is supported now with Camunda 7.17.

### DIFF
--- a/docs/components/best-practices/architecture/deciding-about-your-stack-c7.md
+++ b/docs/components/best-practices/architecture/deciding-about-your-stack-c7.md
@@ -29,7 +29,7 @@ Use the following stack:
 
 3. Use your favorite IDE, for example Visual Studio Code, IntelliJ or Eclipse.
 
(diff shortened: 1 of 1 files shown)


berndruecker:
Awesome - thanks @ingorichtsmeier - as promised 👍  I wondered why you added Oracle JDK, for C8 we currently recommend OpenJDK, see https://docs.camunda.io/docs/components/best-practices/architecture/deciding-about-your-stack/. I think we should harmonize both recommendations. Any good reasons for Oracle?

ingorichtsmeier:
Well, I just clicked on the links to check if they provide a download... And they did. 

I will look for an openjdk link in the afternoon. Please delay the merge of this pull request.

ingorichtsmeier:
Ok, done.

Don't wonder about "OpenJDK JDK 17", which reads a bit awkward, but it's the official headline of this page: https://jdk.java.net/17/

berndruecker (approved):

PR #811, opened 2022-04-13
Add Micronaut to list of community clients

tobiasschaefer:
I'm the lead developer of the Micronaut Zeebe Client. Please add "Micronaut" to the list of community clients, see PR.

Actually, I'm also missing "Micronaut" in the sidebar (see screenshot). I'm not sure if this PR will fix that as well - probably not. Would be nice if you could add that as well.

image

### DIFF
--- a/docs/apis-clients/overview.md
+++ b/docs/apis-clients/overview.md
@@ -45,6 +45,7 @@ Community clients supplement the official clients. These clients have not been t
 
 - [C#](community-clients/c-sharp.md)
 - [JavaScript/NodeJS](community-clients/javascript.md)
+- [Micronaut](community-clients/micronaut.md)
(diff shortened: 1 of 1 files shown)


akeller:
@tobiasschaefer let me expand the scope here to add Micronaut to previous versions **and** check that this is in the sidebar. I thought this was in the previous commit, but looks like it was missed (so much has changed recently!).

akeller:
@tobiasschaefer I took a look at this and your proposed fix in this PR is already on `master`, but I updated `sidebars.js` to add Micronaut and updated the list in the Sidebar so it's also alphabetical order like the overview page. You'll see these changes in #821 

I'm going to close this PR but tag you in the other for awareness, but I think we have this sorted 😄

Issue #299, opened 2021-06-28
set up a formatter for the markdown


korthout:
## Progress

- [x] #815 
- [x] #833 
- [x] #874 
- [x] #876
- [x] #883

## Background

In reviewing code it makes sense to have a automated formatter to end discussions about styling. The same goes for Markdown files.

I know that prettier has [markdown support](https://prettier.io/blog/2017/11/07/1.8.0.html#markdown-support), which also does line wrapping for text through [prose-wrap](https://prettier.io/docs/en/options.html#prose-wrap) and is a node package so should work with the development environment of this repo. 

Personally, I'm no fan of the line wrapping at fixed line length for Markdown, but at least it would add consistency to Markdown styling.

_Originally posted by @korthout in https://github.com/camunda-cloud/camunda-cloud-documentation/pull/298#discussion_r659823177_

korthout:
@akeller How do you feel about this?

akeller:
I do not write with line wrapping in mind, code or sentences, so it makes sense that we should look to have an automated formatter. That said, I don't have any strong opinions on this, but I have used Prettier before and have no objections to it. Plus I want to be part of a [hype train](https://prettier.io/docs/en/why-prettier.html#ride-the-hype-train) and I see they have an extension on VS Code. 👍 
 
I do know it is a very controversial topic, [for example](https://prettier.io/docs/en/why-prettier.html#ride-the-hype-train). In the [Google Technical Writing Style guide](https://developers.google.com/style/html-formatting#line-length) that we are looking to use as our backbone today, they suggest 80 characters, with notable exceptions like links. 100 vs. 80 doesn't matter to me, but I agree let's standardize.

@christinaausley do you have any thoughts here? I don't think we considered this for the style guide draft (yet).

korthout:
IMHO, I think that char length wrapping of readable text is not **as** useful as with "normal" code. Editors and visualizers are generally capable to do soft wrapping of the lines. And most editors (like VSCode) even offer rendered Markdown previews.

Instead, in my ideal world, Markdown would be wrapped after each entire sentence. This would help version control in better showing changes.

Consider the following in which I'd replace the word three for four.
```
In this great example, we start with a short sentence. Then we follow it with another three
sentences. But they are all hard wrapped at 100 chars. So, when I change the word three to four
then it would not be clear in the diff what I'm actually talking about. Three x became four x, but
x is not known, because it exists on a line that did not have to change.
```
```
[...]

PR #815, opened 2022-04-14
Enable prettier to format files

korthout:
Here's a first setup to format files using prettier. I recommend reading the commits individually, as I've described the intensions in more detail there.

Generally:
- You can run it using `npm run format`. 
- By removing entries from the `.prettierignore` file, we can transition to formatting the files in steps.
- A few first files have been formatted to show case how to continue.
- CI-integration to verify that everything is formatted, still needs to be added. For this we could use `prettier --check .`

See #299


korthout:
I've rebased on top op `master` to resolve the conflicts.

akeller:
@pepopowitz FYI :)

pepopowitz:
This is amazing, @korthout! Thanks for doing this. 

1. To confirm that I understand your expectations with this PR: 

- In-scope of this PR, preventing us from merging

  - set up a prettier github action to fail when files are introduced without being formatted

- Out-of-scope of this PR, and coming in follow-up PRs

  - uncomment the remaining folders in .prettierignore, and format the associated files. This can be done one folder at a time.

2. If I've got that accurate, are you intending on working on setting up the github action soon, or is this something you'd prefer I pick up? (I love to pair so it's also something we could do together 🙂 )

3. What do you think about the idea of adding a [git pre-commit hook](https://prettier.io/docs/en/precommit.html) that runs the formatter, to minimize the chances of someone pushing up an unformatted commit, seeing the github action fail, and pushing up an ensuing "format" commit?

korthout:
Hi @pepopowitz, sorry for the late reply. I'm actually out sick, but still want to quickly answer your questions.

> * In-scope of this PR, preventing us from merging
>   
>   * set up a prettier github action to fail when files are introduced without being formatted

IMO, this is something we need to have rather sooner than later. Otherwise, there is no guarantee that formatting is applied. This would eventually likely lead to someone having to run the formatter once in a while, which in turn might then lead to changes in many files which is prone to merge conflicts. In these cases, it's generally best just to verify it in the CI. 

Another option is to auto-apply it in the CI and allow a bot to make and push commits. Personally, I'm not a fan of this because unreviewed changes could lead to unforeseen mistakes. I'd argue for verification in CI over application in CI.

I don't necessarily think it must be part of this PR. It's fine to tackle this shortly after this PR.

> * Out-of-scope of this PR, and coming in follow-up PRs
>   
>   * uncomment the remaining folders in .prettierignore, and format the associated files. This can be done one folder at a time.

Absolutely. I'd argue that the docs folder doesn't have to be done all at once, but its entry in the .prettierignore can be replaced by the files/folders it contains which can then be done one folder at a time. Even though the formatting is applied automatically, I believe someone should review the changes and PRs shouldn't be too big.

> 2. If I've got that accurate, are you intending on working on setting up the github action soon, or is this something you'd prefer I pick up? (I love to pair so it's also something we could do together 🙂 )

Originally, I wanted to just get it all working 😆 But, I just didn't have the time to finish it completely.

If you want, I'd be happy to hand over the topic to you. Perhaps you can review this PR and then continue with the CI stuff. Of course, I can also stay involved if you want me to. Would you be willing to take it over?

> 3. What do you think about the idea of adding a [git pre-commit hook](https://prettier.io/docs/en/precommit.html) that runs the formatter, to minimize the chances of someone pushing up an unformatted commit, seeing the github action fail, and pushing up an ensuing "format" commit?

I'm not so familiar with using pre-commit hooks, TBH. In the Zeebe team, we generally build the project locally with maven which also auto-applies the formatting. For documentation, that might be different because writing docs and building the project might be unrelated for some devs. So, it might make more sense to run a pre-commit hook. I think we can just try this. If devs complain because it breaks their git operations or it takes too much time, it can always be reverted. WDYT?

pepopowitz:
Thanks for responding, Nico, hope you feel better soon. It looks like we're on the same page.

> Another option is to auto-apply it in the CI and allow a bot to make and push commits. Personally, I'm not a fan of this because unreviewed changes could lead to unforeseen mistakes. I'd argue for verification in CI over application in CI.

+1, I have the same opinion.

> ...the docs folder doesn't have to be done all at once, but its entry in the .prettierignore can be replaced by the files/folders it contains which can then be done one folder at a time. Even though the formatting is applied automatically, I believe someone should review the changes and PRs shouldn't be too big.

+1 to small PRs 

> Would you be willing to take it over?

I think so! I'll be meeting later today with @akeller and I'll talk to her about where this could fit in with my priorities, and I'll let you know where we settle afterward. 

> [re: git hooks] I think we can just try this. If devs complain because it breaks their git operations or it takes too much time, it can always be reverted. WDYT?

Yes, this sounds great. 

---

Given all that, here's a list that I think captures our intentions -- let me know if anything's missing or I misunderstood.

- [ ] This PR: ready for review
- [ ] Followup PR: Add a GitHub Action to run prettier
- [ ] Followup PRs: Include remaining folders in prettier checks
- [ ] Followup PR: Add a pre-commit hook to auto-format files when they're updated

.github/workflows/publish-stage.yaml
```diff
@@ -20,7 +20,7 @@ jobs:
       - name: Disable Indexing
         run: "sed -i 's/noIndex: false/noIndex: true/g' docusaurus.config.js"
       - name: Update URL
-        run: "sed -i 's!url: \"https://docs.camunda.io\"!url: \"https://stage.docs.camunda.io\"!g' docusaurus.config.js"
+        run: 'sed -i ''s!url: "https://docs.camunda.io"!url: "https://stage.docs.camunda.io"!g'' docusaurus.config.js'
```
pepopowitz, line 23: (no action required) 

This looks right to me, but just leaving a placeholder for myself to remember that this part of the staging deploy changed, in case it ends up being an issue.

sidebars.js
```diff
@@ -98,18 +97,18 @@ module.exports = {
...
                     "components/modeler/web-modeler/connectors/available-connectors/rest",
                     "components/modeler/web-modeler/connectors/available-connectors/sendgrid",
-                  ]
-                }
-              ]
-            }
+                  ],
+                },
+              ],
+            },
```
pepopowitz, line 111: ❤️ yesssss, trailing commas

sidebars.js
```diff
@@ -405,13 +402,10 @@ module.exports = {
...
     {
-      "Installation": [
+      Installation: [
         "self-managed/platform-deployment/index",
         {
-          "Kubernetes": [
-            "self-managed/platform-deployment/kubernetes",
-            "self-managed/platform-deployment/kubernetes-helm",
-          ],
+          Kubernetes: ["self-managed/platform-deployment/kubernetes"],
```
pepopowitz, line 408: Potentially blocking: 

There's one removed item here (`kubernetes-helm`) -- is that intentional?

sidebars.js
```diff
@@ -426,19 +420,16 @@ module.exports = {
...
     },
     {
-      "Zeebe": [
+      Zeebe: [
         "self-managed/zeebe-deployment/index",
         {
           Configuration: [
-            "self-managed/zeebe-deployment/configuration/configuration",
-            "self-managed/zeebe-deployment/configuration/logging",
-            "self-managed/zeebe-deployment/configuration/gateway-health-probes",
```
pepopowitz, line 441: Potentially blocking:

Also several removed here. Are these old, and that's why they were removed?

pepopowitz (commented):
This looks great! 

There were a few lines removed that I wasn't sure about -- I want to confirm they were intentional removals before approving/merging.

pepopowitz:
Following up re: taking over this work -- @akeller and I talked about it today and yes, I will own this moving forward!

.github/workflows/publish-stage.yaml
korthout: Good idea 👍 

TBH, this one was a surprise to me too, but apparently, YAML escapes quotes `'` with another quote `''`.

sidebars.js
korthout: 👍 Of course, this is just the default setting. I think it's good, but if we prefer something else we can change it in the `.prettierrc.json`

sidebars.js
korthout: Nope, good catch. Might have happened in the conflict resolution

sidebars.js
korthout: Also unintentional. Good catch

pepopowitz (approved):
Sorry for the delayed response @korthout! This looks great, and I'll plan to either merge tomorrow, or Monday. I will do my best to take care of any merge conflicts that show up by then, but might reach out for help if any tricky ones appear.

Issue #817, opened 2022-04-15
homepage feature images look low-resolution


pepopowitz:
The feature images on the homepage are scaled to a size larger than the originals, resulting in a blurry low-res effect: 

<img width="1415" alt="image" src="https://user-images.githubusercontent.com/1627089/163595590-fd72d42c-b706-4b81-a36e-b731d927877c.png">

<img width="382" alt="image" src="https://user-images.githubusercontent.com/1627089/163595761-8147bd75-9561-4642-be1a-8a30cfa11c9e.png">

pepopowitz:
Resolved in #818

PR #818, opened 2022-04-15
style(features): scale feature icons down a bit

pepopowitz:
Addresses #817 

The source images for the feature icons are 150x150. They were previously displaying as 200x200, resulting in some low-res stretchiness. This PR scales them down, displaying them as slightly smaller than the sources (130x130) to better align with the proposed UI from marketing (which can be found in the slack conversation linked below). 

There is still a slight bit of blurriness to the icons, the edges don't seem totally sharp. My interpretation is that they were built that way intentionally, but if people still find it unsettling I can push for marketing to make new images with sharper edges. 

Before:

image

After: 

image

See [this thread](https://camunda.slack.com/archives/C01UMP0V17Z/p1649780222187189) for history. 

I've added reviewers based on the original slack conversation, with intention to self-merge when I receive positive-or-neutral feedback.

### DIFF
--- a/src/pages/styles.module.css
+++ b/src/pages/styles.module.css
@@ -32,6 +32,6 @@
 }
 
 .featureImage {
-  height: 200px;
-  width: 200px;
+  height: 130px;
+  width: 130px;
 }


Sijoma (commented):
I like it. 👍

akeller (approved):
Works for me, mostly because it brings the bottom two "above the fold" on my smallest screen.

PR #825, opened 2022-04-20
Update variables.md

fml2:


akeller:
@fml2 thanks for your PR. Can you please sign the CLA? I think you've signed previously, but we had to make a small update recently which retriggered signature collection. Thanks!

PR #827, opened 2022-04-20
docs(best-practices): remove extra best practices landing page

akeller:
@berndruecker I assigned you as a reviewer because you know this content better than anyone else 😉 . Is this an extra page? I don't see if when I navigate in the docs and I think it was just a miss when you migrated over the other Best Practices content.


berndruecker (approved):
This is the old page - deletion is exactly the right thing to do. Good finding!

akeller:
Rerunning the build process because it didn't really fail, it looked like it was canceled.

PR #832, opened 2022-04-22
chore(deps): pin docusaurus dependencies

pepopowitz:
Pins the docusaurus dependencies to version 2.0.0-beta.15, until we can upgrade them properly. 

## Why?

There are at least two breaking changes introduced in newer versions of the 2.0 beta that have been reported in the Camunda slack. They appear as these errors during `npm start`: 

```
[ERROR] ValidationError: "algolia.appId" is required. If you haven't migrated to the new DocSearch infra, please refer to the blog post for instructions: https://docusaurus.io/blog/2021/11/21/algolia-docsearch-migration
```

and

```
[ERROR] ValidationError: colorMode.switchConfig is deprecated. If you want to customize the icons for light and dark mode, swizzle IconLightMode, IconDarkMode, or ColorModeToggle instead.
```

Pinning the dependencies to version 2.0.0-beta.15 will force that version to be installed, preventing us from seeing those breaking change errors.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,11 +5,12 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "camunda-cloud-documentation",
       "version": "0.0.0",
       "dependencies": {
         "@auth0/auth0-react": "^1.9.0",
-        "@docusaurus/core": "^2.0.0-beta.15",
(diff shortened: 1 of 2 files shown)


akeller (approved):
Ship iiiit 🚀

PR #833, opened 2022-04-22
chore(format): exclude build artifacts from prettier

pepopowitz:
1. Adds two build artifact folders to the .prettierignore. They are generated by docusaurus, and excluded from git, and not worth us spending time formatting. 
2. Updates the formatting for a github action

This work is a follow-up to #815. @korthout noticed those folders being formatted in that PR, but we fixed it here in a follow-up for ease of PR review.

### DIFF
--- a/.github/workflows/publish-c8.yaml
+++ b/.github/workflows/publish-c8.yaml
@@ -20,7 +20,7 @@ jobs:
       - name: Disable Indexing
         run: "sed -i 's/noIndex: false/noIndex: true/g' docusaurus.config.js"
       - name: Update URL
-        run: "sed -i 's!url: \"https://docs.camunda.io\"!url: \"https://c8.docs.camunda.io\"!g' docusaurus.config.js"
+        run: 'sed -i ''s!url: "https://docs.camunda.io"!url: "https://c8.docs.camunda.io"!g'' docusaurus.config.js'
       - name: Build
         run: npm run build
       - name: Publish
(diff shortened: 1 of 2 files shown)


akeller (approved):
I'm approving based on 

> They are generated by docusaurus, and excluded from git, and not worth us spending time formatting.

One additional thing to consider that you could argue is or is not in the scope of this PR is borrowing ideas from Docusaurus's `.prettierignore` and include a few other artifacts and folders while you are thinking about it. Linking their file [here](https://github.com/facebook/docusaurus/blob/main/.prettierignore).

akeller:
> One additional thing to consider that you could argue is or is not in the scope of this PR is borrowing ideas from Docusaurus's `.prettierignore` and include a few other artifacts and folders while you are thinking about it. Linking their file [here](https://github.com/facebook/docusaurus/blob/main/.prettierignore).

Realized I was nice and vague here, but specifically wanted to call out that they include `docusaurus.config.js` in their file.

pepopowitz:
I didn't think to check there, good idea!

I do think we should include `docusaurus.config.js` in the formatting, because it's code that _we_ write. They also have some other `.json` files excluded (like sidebars), but I _also_ think those should be formatted, since it's again json that we write. 

And I didn't see anything else worthy of exclusion in their config, but let me know if you do!

pepopowitz:
also @akeller I just noticed the `dx` label you applied -- first time I've noticed it, but I'll apply it moving forward.

PR #857, opened 2022-04-26
add java version column

christinaausley:
Adding JDK matrix


akeller (approved):
Looks good, but please confirm with @megglos before merging. I think Tobias Conz had other feedback, but I'm not sure it was in the scope of this PR.

megglos (approved):
LGTM!

PR #863, opened 2022-04-27
Feedback on backpressure pages

akeller:
Thanks @falko for your feedback! I've added you as a reviewer for this PR. If you have a chance to review that would be great. 

@christinaausley in our docs we use both backpressure and back-pressure. I didn't correct this and instead added keywords to the frontmatter on both pages to handle this spelling. Can you check if this is something that is standardized in the software engineering space to use (or not use!) the hyphen? We should standardize on something, but whatever we decide we should keep the keywords as-is to assist with search.

### DIFF
--- a/docs/apis-clients/grpc.md
+++ b/docs/apis-clients/grpc.md
@@ -2,6 +2,7 @@
 id: grpc
 title: "Zeebe API (gRPC)"
 description: "Zeebe clients use gRPC to communicate with the cluster."
+keywords: ["backpressure", "back-pressure", "back pressure"]
 ---
 
 [Zeebe](../components/zeebe/zeebe-overview.md) clients use [gRPC](https://grpc.io/) to communicate with the cluster.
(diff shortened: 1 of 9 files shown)


christinaausley:
Agreed here. I'll check on preferred standardization, add to tech writing glossary, and launch a PR to fix any necessary instances this week.

christinaausley:
@akeller The common terminology is `backpressure`. I've updated the glossary with this term, but would you like me to launch an additional PR to adjust any instances of `back pressure` or `back-pressure`?

akeller:
> @akeller The common terminology is `backpressure`. I've updated the glossary with this term, but would you like me to launch an additional PR to adjust any instances of `back pressure` or `back-pressure`? Can also adjust in this commit if preferred.

Please - make sure all 3 are listed as keywords in the frontmatter on the pages backpressure is mentioned. Adjusting in this PR makes sense to me.

akeller:
@falko I will go ahead and merge. If you have more feedback on how we are using backpressure or your search experience with topics related to backpressure, reach out to @christinaausley.

PR #874, opened 2022-04-29
chore(ci): verify prettier formatting in build-docs workflow

pepopowitz:
Adds a step to the build-docs workflow to check prettier formatting. This is a follow-up to #815.

## What it looks like when format is valid (as seen in [this workflow run](https://github.com/camunda/camunda-platform-docs/runs/6231159619))

image

## What it looks like when format is invalid (as seen in [this workflow run](https://github.com/camunda/camunda-platform-docs/runs/6231238715))

image

### Follow-up work, not included in this PR

- add a git pre-commit hook to auto-format changes (coming soon)
- format remaining folders in the docs (coming later, de-prioritized for now)

PR #876, opened 2022-04-29
improve(dx): auto-format files on pre-commit

pepopowitz:
Adds a pre-commit hook via husky & lint-staged to auto-format staged changed files. 

This is a followup to #815 and #874. Now that we're using prettier to format, and requiring formatted files in CI, we'll have fewer commits titled "prettier" if we auto-format files before they're committed. 

## How this PR came to be

1. I followed the directions from [lint-staged](https://github.com/okonet/lint-staged#installation-and-setup). I haven't used husky since they went to version 7, so the files that got created during setup didn't look familiar....but the differences align with [husky's migration guide](https://typicode.github.io/husky/#/?id=migrate-from-v4-to-v7) so it seems legit. 

2. I wanted to update the matched extensions for running `lint-staged`, since out of the box they were `{js,md,css}` and it seemed like we'd have more than that in this project. I initially tried matching `*.*` but this broke on `.gitignore` files, and I decided it was too aggressive. Instead, I grabbed a breakdown of file extensions in the project, using this command inspired by [a StackOverflow comment](https://stackoverflow.com/questions/1842254/how-can-i-find-all-of-the-distinct-file-extensions-in-a-folder-hierarchy#comment62085285_1842270): 

```sh
find . -type f -not -path "./node_modules/*" -not -path "./.docusaurus/*" -not -path "./build/*" | perl -ne 'print $1 if m/\.([^.\/]+)$/' | sort | uniq -c | sort -n
```

This gave me the file extensions & count for each, excluding the node_modules/.docusaurus/build folders: 

```sh
❯ find . -type f -not -path "./node_modules/*" -not -path "./.docusaurus/*" -not -path "./build/*" | perl -ne 'print $1 if m/\.([^.\/]+)$/' | sort | uniq -c | sort -n
   1 19
   1 2
   1 4
   1 5
   1 6
   1 diff
   1 graphqls
   1 htaccess
   1 ico
   1 nojekyll
   1 patch
   1 prettierignore
   2 MD
   2 gitignore
   2 sh
   2 xlsx
   2 xml
   3 1
   3 3
   3 css
   3 graffle
   3 txt
   4 idx
   4 pack
   5 0
   5 18
   5 mp4
   8 dmn
   9 jpg
  10 json
  13 sample
  17 yaml
  25 js
  27 pptx
  28 yml
 140 mdx
 164 gif
 184 bpmn
 694 svg
1646 md
2262 png
```

I looked through this list and grabbed the extensions that I thought would be useful to auto-format, and settled on `{css,js,json,md,mdx,yaml,yml}`. 

3. I tested a few cases locally: 

- [x] unformatted staged changes are formatted before commit
- [x] unformatted/unstaged changes in partially-staged file are unformatted after commit
- [x] pre-commit hook runs on a fresh clone
- [x] prettier-ignored files are not formatted on commit


korthout (approved):
Interesting stuff @pepopowitz 🎸 LGTM

I was wondering a bit about the double definition for files to format (once in package.json > lint-staged and once in .prettierignore), but I don't think it hurts.

menski (dismissed):
lgtm

:camping: from your list of file endings I noticed that it listed `MD`, and when I checked it seems our `CONTRIBUTING.MD` file has a upper case ending. I think either we should lowercase that ending or add MD to the file endings, otherwise this file will not be touched by the prettier.

:memo: After this is merged, contributors are required to run `npm install` locally to setup the pre-commit hooks. Maybe something we should inform people in the `#ask-documentation` channel

pepopowitz:
@menski Good call on both fronts -- (1) I'll update the extension pattern to include MD, and (2) I'll notify that channel when I've merged.

package.json
```diff
@@ -44,6 +45,11 @@
     ]
   },
   "devDependencies": {
+    "husky": "^7.0.4",
+    "lint-staged": "^12.4.1",
     "prettier": "2.6.2"
+  },
+  "lint-staged": {
+    "*.{css,js,json,md,mdx,yaml,yml}": "npm run format"
```
pepopowitz, line 53:
```suggestion
    "*.{css,js,json,MD,md,mdx,yaml,yml}": "npm run format"
```

Per code-review

CONTRIBUTING.MD
```diff
@@ -137,26 +137,26 @@ For further guidelines and best practices to keep in mind as a reviewer, please
...
+[contributor license agreement]: https://cla-assistant.io/camunda-cloud/camunda-cloud-documentation
 [code style guidelines]: https://github.com/camunda-cloud/zeebe/wiki/Code-Style
 [testing best practices]: https://docs.camunda.io/docs/apis-clients/java-client/zeebe-process-test/
-[Conventional Commits]: ./conventional-commits
+[conventional commits]: ./conventional-commits
 [template and settings files]: https://github.com/camunda/camunda-bpm-platform/tree/master/settings
-[Technical Writing Style Guide]: https://github.com/camunda-cloud/camunda-cloud-documentation/blob/master/howtos/technical-writing-styleguide.md
-[Pull Requests and Code Review]: https://github.com/camunda-cloud/zeebe/wiki/Pull-Requests-and-Code-Reviews
-[SemVer]: http://semver.org/
+[technical writing style guide]: https://github.com/camunda-cloud/camunda-cloud-documentation/blob/master/howtos/technical-writing-styleguide.md
```
menski, line 160: Just a comment, no need for action, but probably interesting to know. It seems the prettier touches these links by lower-casing the words, but not their usage inside the content, i.e. `Does your writing follow our [Technical Writing Style Guide]?` (Line 54)

menski (dismissed):
Thanks for incorporating the changes

CONTRIBUTING.MD
korthout: I think it doesn't matter because these are link reference labels. These are not visible in the document.

AFAIK, prettier uses the Commonmark specification, which says that link labels of [link references](https://spec.commonmark.org/0.30/#link-reference-definitions) are [matched case-insenstive](https://spec.commonmark.org/0.30/#example-205).

🤔 So I guess prettier just applies lower-case for consistency in the reference list 🤷‍♂️ I don't know why it wouldn't then in the reference usage, but again: it doesn't seem like it really matters 😆

pepopowitz: I confirmed that the links still go where they're intended to. I also think it's strange that it only updates in one place, but I am also not concerned about it.

akeller (approved):
I only looked a the `package.json` update since @pepopowitz mentioned that was what needed to be updated due to a marge conflict. I see no issues, which is why I'm approving.

PR #891, opened 2022-05-09
chore(deps): remove unused 'color' dependency

pepopowitz:
Removes the `color` dependency, which I don't think we are using. 

Causes no obvious regressions when I run the docs locally.

I'm tagging reviewers that might have more context about why it was pulled in, based on my spelunking described below.

## Why did I spend time on this?

I was going through dependabot PRs and saw #768, a major-version bump for the `color` package. I wanted to update it, but I like to know where things are used so I can smoke test them before I merge dependency updates. I didn't find anywhere that `color` was used.

## Why don't I think it is used anywhere?

I searched for `"color"`, `'color'`, `color(`, and `color.`, and saw no significant results. I attempted to search for the whole word `color` but there were too many results to meaningfully identify its usage.

These first two tell me that we never import or require this dependency (as in `import color from "color"`): 
image
image

These last two tell me that we never _use_ [the `color` API](https://github.com/Qix-/color/tree/e752c390d4ab3f3659b22ef6cf91333c544cf7b8): 
image
image

I am not sure if there's any other way for docusaurus to magically make use of this dependency, but it appears to me it's not used.

## When did it get pulled in? 

As part of #99 - be warned that that is a very large release, and the github UI slows considerably while trying to navigate it. 

More specifically, in [this commit](https://github.com/camunda/camunda-platform-docs/pull/99/commits/8d4146d48497b210fe0c19d6079f06b79f238317#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519), titled "add version 0.25". Be warned that this _also_ is a very large commit, and the github UI slows considerably while trying to navigate it. 

I can't glean any information from that commit about _why_ it was added. I'm tagging @urbanisierung and @akeller for a review because they might have been involved in that work, but I fully expect neither of them to remember a dependency added a year and a half ago.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,6 @@
         "@docusaurus/preset-classic": "2.0.0-beta.15",
         "@mdx-js/react": "^1.6.22",
         "clsx": "^1.1.1",
-        "color": "^3.2.1",
         "docusaurus-gtm-plugin": "0.0.2",
         "mixpanel-browser": "^2.43.0",
         "react": "^17.0.2",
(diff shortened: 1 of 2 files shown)


berndruecker (approved):
I follow your reasoning and think we should go ahead and remove it (and just keep a close eye to what could probably break). I also haven't quickly found any reference to it. If it comes with docusaurus it should be a transitive dependency I guess.

PR #892, opened 2022-05-09
initial footer restructure

christinaausley:
A more logical footer rework and restructure with updated user resources.

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -97,23 +97,19 @@ module.exports = {
       },
       links: [
         {
-          title: "Docs",
+          title: "About",
           items: [
             {
-              label: "Introduction",
-              to: "docs/guides",
+              label: "How to use our docs",
+              to: "meta",
             },
             {
-              label: "Licenses",
-              to: "docs/reference/licenses",
+              label: "Try free",
+              href: "https://camunda.com/get-started",
             },
             {
-              label: "Notices",
-              to: "docs/reference/notices",
-            },
-            {
-              label: "Release cycle",
-              to: "docs/reference/release-policy",
+              label: "Contact",
+              to: "contact",
             },
           ],
         },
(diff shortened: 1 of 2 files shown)


christinaausley:
@pepopowitz I'll keep out of this branch for the rest of the week if you'd like to hop in, take a look, and make adjustments. Looking forward to your thoughts on it now that we can see a local build live.

If we'd like to clean up the Reference section, this does mean we would need to move the "removed" tabs elsewhere. Thoughts on where these could live within docs?

pepopowitz:
> If we'd like to clean up the Reference section, this does mean we would need to move the "removed" tabs elsewhere. Thoughts on where these could live within docs?

@christinaausley I _think_ the only one in here that's not accounted for in either the footer or the proposed revamped reference section is "Announcements": 

image

From [the google doc](https://docs.google.com/document/d/1NqV-_P_71JPLqG4pYeB8W_U7dhAHEsW3FNNF8tnXOZ0/edit#heading=h.wfjb0kz6pfgr), I would have expected that to be added under the "Camunda" section of the footer, though:

image

..although it is highlighted in yellow, and I didn't do that, so maybe you intentionally left it out?

pepopowitz:
Having said that I would also be fine with leaving Announcements in the Reference section if you think that's a better place than the Camunda column of the footer.

pepopowitz (approved):
This looks great, and you should merge when you're happy with it!!!!

I made a comment in a previous message about the "Announcements" page (https://docs.camunda.io/docs/reference/announcements/) -- that's the only link we remove from References that will need a home, and at some point I think we talked about putting it under "Camunda" in the footer. I defer to you in regards to where you think it belongs.

christinaausley:
@pepopowitz I did highlight a few sections of that doc to keep track of what I had changed and what I had yet to change. I'll leave the Reference section as is currently and take another look at it next Monday with some fresh eyes.

Do you want to hop in this branch to confirm spacing/visuals of the footer look clean to you? As you mentioned, we can hold off on moving forward with the Algolia/updates and therefore the external link icon until @akeller returns.

Once we have that polished up, we can go ahead and merge 💯

pepopowitz:
> Do you want to hop in this branch to confirm spacing/visuals of the footer look clean to you? 

Yes -- I think they look centered! It feels a bit off because each column is fairly wide & left-justified, but it looks like all columns are equal width & centered within the page to me. It's a little easier to see with some elements highlighted via the browser dev tools:

image

pepopowitz:
> we can hold off on moving forward with the Algolia/updates and therefore the external link icon

And I think I might have been unclear on https://github.com/camunda/camunda-platform-docs/issues/492#issuecomment-1122555085 -- I am planning to hold on the algolia/updates, but I opened #896 against our _current_ version of docusaurus, and plan to revisit it when we do the dependency updates.

PR #896, opened 2022-05-10
feat(footer): display external link icon for external links

pepopowitz:
Adds an external link icon to links in the footer which are external, without having to specify anything in docusaurus.config.js. 

## How it looks

image

## What's all included

1. Displays an icon based on the link URL. If it's an external link, we pop it in a new tab, _and_ display the external link icon.

    Note that the implementation of this logic is based on [docusaurus version 2.0.0-beta.15's implementation of the footer](https://github.com/facebook/docusaurus/blob/32ec84ef3c0a238436e913b2026ab809e5750fa8/packages/docusaurus-theme-classic/src/theme/Footer/index.tsx#L45-L52). We override/customize the footer in [our own `src/theme/footer/index.js`](https://github.com/camunda/camunda-platform-docs/blob/be0014171b365903f606d7eac2c6ded5555c5a3e/src/theme/Footer/index.js#L20), and due to drift of this duplicated file, we missed the external link icon feature when docusaurus introduced it. So I basically took the logic from their footer and put it into ours.

2. The icon itself is included in this PR. Docusaurus has an ExternalLinkIcon that looks nearly identical to this one (maybe even identical), but I chose to use the one mentioned by @akeller in [this comment](https://github.com/camunda/camunda-platform-docs/issues/492#issuecomment-1098388606) since it came from our Brand Assets. 

3. The ExternalLinkIcon includes a `title` that reads "(opens in a new window)", to support users with assistive devices. [This WCAG doc](https://www.w3.org/TR/WCAG20-TECHS/G201.html) (see Example 1) describes the need to include text in links that are opening in a new window to avoid disorientation. [This (admittedly old) CSS-tricks article describes adding text titles to SVG for accessibility](https://css-tricks.com/accessible-svgs/#aa-2-inline-svg). Combining these two, we get:
    * for sighted users, a link with an "external" icon for sighted users, and a tooltip for the icons that reads "opens in a new window"
    * for users with screenreaders, the text "opens in a new window" appended to the link name. See the screen-reader version in action (including audio, which is turned off by default by GitHub):

https://user-images.githubusercontent.com/1627089/167711941-8a7b64ca-ce5b-4930-a5df-ab774d16c0ea.mov

5. I removed some comments & unused code. These bits were made obsolete in https://github.com/camunda/camunda-platform-docs/commit/3cf14c343082f820d76f1af191e8e76e5e1296f6.

## Notes for the future

* When we update the docusaurus dependency, we'll need to revisit our `src/theme/footer/index.js`. In recent releases, docusaurus has changed the implementation of the footer, so our custom override will probably break (or at least lack features) with an update.


src/theme/ExternalLinkIcon/index.js
```diff
@@ -0,0 +1,24 @@
+import React from "react";
+
+import styles from "./styles.module.css";
+
+function IconExternalLink({ width = 15, height = 15 }) {
```
pepopowitz, line 5: Oops. I named the component one thing (Icon...) and the file something else (External...). I'll push up a fix to synchronize this.

christinaausley (approved):

PR #899, opened 2022-05-11
Document new multi-instance properties for completion condition

### LINKED ISSUE #884: Document new multi-instance properties for completion condition
Support for multi-instance properties has been added to the completion condition expression evaluation with https://github.com/camunda/zeebe/pull/8460.

>numberOfInstances and numberOfActiveInstances is now available for use in the [completion condition](https://docs.camunda.io/docs/components/modeler/bpmn/multi-instance/#completion-condition) expression. Although they are available in that expression, they do not exist as process variables. These properties take precedence over process variables with the same name.

We should document this new feature.

lzgabel:
The BPMN 2.0 specification defines the following properties of a multi-instance body instance:

- `numberOfInstances`: The total number of instances created.
- `numberOfActiveInstances`: The number of instances currently active.
- `numberOfCompletedInstances`: The number of instances already completed.

These properties are available for use in the `completionCondition` expression. 

closes #884

### DIFF
--- a/docs/components/modeler/bpmn/multi-instance/multi-instance.md
+++ b/docs/components/modeler/bpmn/multi-instance/multi-instance.md
@@ -99,9 +99,9 @@ The BPMN 2.0 specification defines the following properties of a multi-instance 
 - `numberOfInstances`: The number of instances created.
 - `numberOfActiveInstances`: The number of instances currently active.
 - `numberOfCompletedInstances`: The number of instances already completed.
-- `numberOfTerminatedInstances`: The number of instances already terminated, this will always be `0`.
+- `numberOfTerminatedInstances`: The number of instances already terminated.
 
-These properties are available for use in the `completionCondition` expression (eg. when 50% of the instances have completed, the multi-instance body is completed). Although they are available in this expression, they do not exist as process variables. These properties take precedence over process variables with the same name.
+These properties are available for use in the `completionCondition` expression. For example, using these properties you can express "complete the multi-instance body when 50% or more of the instances already completed" as `= numberOfCompletedInstances / numberOfInstances >= 0.5`. Although these properties are available in this expression, they do not exist as process variables. These properties take precedence over process variables with the same name.
 
 Multiple boolean values or comparisons can be combined as disjunction (`and`) or conjunction (`or`).
 
@@ -118,7 +118,7 @@ For example:
 
 = numberOfCompletedInstances = 2
 
-= numberOfCompletedInstances/numberOfInstances >= 0.5
+= numberOfCompletedInstances / numberOfInstances >= 0.5
 ```
 
 
--- a/docs/components/modeler/bpmn/multi-instance/multi-instance.md
+++ b/docs/components/modeler/bpmn/multi-instance/multi-instance.md
@@ -94,6 +94,14 @@ target: output
 
 A `completionCondition` defines whether the multi-instance body can be completed immediately when the condition is satisfied. It is a [boolean expression](/components/concepts/expressions.md#boolean-expressions) that will be evaluated each time the instance of the multi-instance body completes. Any instances that are still active are terminated and the multi-instance body is completed when the expression evaluates to `true`.
 
+The BPMN 2.0 specification defined the following properties of a multi-instance body instance:
+
+- `numberOfInstances`: the total number of instances created.
+- `numberOfActiveInstances`: the number of currently active instances.
+- `numberOfCompletedInstances`: the number of already completed instances.
+
+These properties are available for use in the `completionCondition` expression. Although they are available in that expression, they do not exist as process variables. These properties take precedence over process variables with the same name.
+
 Multiple boolean values or comparisons can be combined as disjunction (`and`) or conjunction (`or`).
 
 For example:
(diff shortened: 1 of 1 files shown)


christinaausley (approved):
From a technical perspective, I've made a few very minor adjustments in my latest commit. Following a look from @korthout, LGTM!

docs/components/modeler/bpmn/multi-instance/multi-instance.md
```diff
@@ -94,6 +94,14 @@ target: output
 
 A `completionCondition` defines whether the multi-instance body can be completed immediately when the condition is satisfied. It is a [boolean expression](/components/concepts/expressions.md#boolean-expressions) that will be evaluated each time the instance of the multi-instance body completes. Any instances that are still active are terminated and the multi-instance body is completed when the expression evaluates to `true`.
 
+The BPMN 2.0 specification defines the following properties of a multi-instance body instance:
+
+- `numberOfInstances`: The number of instances created.
+- `numberOfActiveInstances`: The number of instances currently active.
+- `numberOfCompletedInstances`: The number of instances already completed.
```
korthout, line 101: ❌   The `numberOfTerminatedInstances` are also defined and available for use in the completion condition expression. Let's add it here as well. 

Generally, this value will not work that well in the completion condition, i.e. the completion condition is only evaluated for completing instances, but instances won't complete anymore as soon as the multi-instance is terminating. So, the value will always be 0 when used in the completion condition, simply because of the way the zeebe engine deals with termination. However, the value is available in the completion condition and will take precedence over process variables with the same name. So we should mention it here as well.

docs/components/modeler/bpmn/multi-instance/multi-instance.md
```diff
@@ -94,6 +94,14 @@ target: output
...
+
+These properties are available for use in the `completionCondition` expression. Although they are available in this expression, they do not exist as process variables. These properties take precedence over process variables with the same name.
```
korthout, line 103: 🔧 I'd like to see an additional sentence that describes what they can be used for. Perhaps we provide the example from below (if 50% of the instances have completed). WDYT?

docs/components/modeler/bpmn/multi-instance/multi-instance.md
```diff
@@ -108,6 +116,10 @@ For example:
 = orderCount > 15 or totalPrice > 50
 
 = list contains([6,7], today().weekday)
+
+= numberOfCompletedInstances = 2
+
+= numberOfCompletedInstances/numberOfInstances >= 0.5
```
korthout, line 122: 👍  I really like the new additions.

🔧  This list of examples is becoming pretty long which might confuse users. Let's reduce it a bit by removing the expressions that use `orderCount`.

docs/components/modeler/bpmn/multi-instance/multi-instance.md
```diff
@@ -94,6 +94,14 @@ target: output
 
 A `completionCondition` defines whether the multi-instance body can be completed immediately when the condition is satisfied. It is a [boolean expression](/components/concepts/expressions.md#boolean-expressions) that will be evaluated each time the instance of the multi-instance body completes. Any instances that are still active are terminated and the multi-instance body is completed when the expression evaluates to `true`.
 
+The BPMN 2.0 specification defines the following properties of a multi-instance body instance:
```
korthout, line 97: 🔧  `multi-instance body instance` is correct, but might be a bit confusing to newcomers. Let's just call it `multi-instance body` like in the paragraph above.

korthout (changes requested):
Thanks @Lzgabel Looking good already. 👍 

I have a few comments, please have a look. 

EDIT: I've applied [the emoji code](https://github.com/camunda/zeebe/blob/main/CONTRIBUTING.md#review-emoji-code) in my review comments, as you're probably used to from me 😄

Lzgabel:
Hi @korthout. Please take a moment to check this out. 🙇

docs/components/modeler/bpmn/multi-instance/multi-instance.md
```diff
@@ -94,13 +94,14 @@ target: output
...
-The BPMN 2.0 specification defines the following properties of a multi-instance body instance:
+The BPMN 2.0 specification defines the following properties of a multi-instance body:
 
 - `numberOfInstances`: The number of instances created.
 - `numberOfActiveInstances`: The number of instances currently active.
 - `numberOfCompletedInstances`: The number of instances already completed.
+- `numberOfTerminatedInstances`: The number of instances already terminated, this will always be `0`.
 
-These properties are available for use in the `completionCondition` expression. Although they are available in this expression, they do not exist as process variables. These properties take precedence over process variables with the same name.
+These properties are available for use in the `completionCondition` expression (eg. when 50% of the instances have completed, the multi-instance body is completed). Although they are available in this expression, they do not exist as process variables. These properties take precedence over process variables with the same name.
```
korthout, line 104: 🔧 I'd like to suggest the following changes, focussing on an example expression along with the use case example.

 ```suggestion
These properties are available for use in the `completionCondition` expression. For example, using these properties you can express "complete the multi-instance body when 50% or more of the instances already completed" as `= numberOfCompletedInstances / numberOfInstances >= 0.5`. Although these properties are available in this expression, they do not exist as process variables. These properties take precedence over process variables with the same name.
```

korthout (changes requested):
Thanks @lzgabel 👍 

A few last comments

lzgabel:
Hi @korthout. Please check this out. 🙇

korthout (approved):
Thanks @lzgabel LGTM 🏆 

@christinaausley There's some new changes. Please have another look.

christinaausley (approved):

PR #904, opened 2022-05-12
[DO NOT MERGE] very rough POC of multi-version docs

pepopowitz:
I'm not planning to merge this. It's a rough outline of what it would look like to have most of the docs using one version scheme, and a single section using a different version scheme.

I'm creating this PR as a reference point, and will close it immediately. 

Most docs show versions 1.3 and 1.2: 

image

One new section named "community" was added, with versions 5 and 5.1: 

image

Note that the URL for this new community section starts with "/community" instead of "/docs". This is what it looks like with multiple docs instances configured.

(The list of files changed is massive, as I cleared out a whole bunch of docs so I could focus on the problem at hand. If your browser is slow while viewing this PR, that's why.)


pepopowitz:
Closing as this is just for reference.

PR #906, opened 2022-05-13
[DO NOT MERGE] POC - unique versioning for some sections of the docs

pepopowitz:
This PR will not be merged. 

It's a proof of concept for how we could implement multi-versioning within the docs -- most of the docs use one versioning scheme, while some sections (Optimize, in this example) use a different one. 

[Demo of how this PR affects navigating the docs](https://www.loom.com/share/ca27f401e19f4e9b956bf6919028c118)


src/theme/NavbarItem/DefaultNavbarItem.js
```diff
@@ -0,0 +1,2 @@
+import OriginalDefaultNavbarItem from "@theme-original/NavbarItem/DefaultNavbarItem";
+export default OriginalDefaultNavbarItem;
```
pepopowitz, line 2: This was fun, I learned all about the concept of ["swizzling"](https://docusaurus.io/docs/swizzling), which is how we customize core docusaurus bits. 

In this file, and several below it, we're literally wrapping the original docusaurus implementation. This seems silly, but we _have_ to do it, because further down we want to override the implementation of another component _in this same folder_. If you replace one item inside of `NavbarItem`, you have to replace all of them. And since we don't actually want to change the implementation, we wrap them.

src/theme/NavbarItem/index.js
```diff
@@ -0,0 +1,29 @@
+import React from "react";
```
pepopowitz, line 1: But here we _do_ want to change the implementation, of the `NavbarItem` itself. The versions that appear in the navbar dropdown should vary depending on which section of the site you're visiting. 

This component checks the current path, and if it pattern matches to the `optimize` section, sends the new docs plugin ID through to the version dropdown. Every other page falls back to the main `docs` versions.

pepopowitz (commented):
Adding notes to help explain the changes.

src/theme/NavbarItem/DefaultNavbarItem.js
pepopowitz: (I also learned that [it's not just a made-up word](https://en.wikipedia.org/wiki/Swizzling_(computer_graphics)))

pepopowitz:
Closing this as it's for POC purposes.

PR #928, opened 2022-05-27
chore(format): enable formatting for docs/self-managed, and format docs/self-managed.

pepopowitz:
Partially addresses https://github.com/camunda/camunda-platform-docs/issues/883.

Enables prettier formatting for the ./docs/self-managed folder, and formats the files in that folder.

There was one syntactical change that I had to make, and I'll point it out inline.


docs/self-managed/optimize-deployment/configuration/service-config.yaml
```diff
@@ -440,30 +439,30 @@ historyCleanup:
...
-    backgroundColor: '#FFFFFF'
+    backgroundColor: "#FFFFFF"
 
 eventBasedProcess:
   # A list of userIds that are authorized to manage (Create, Update, Publish & Delete) event based processes.
   authorizedUserIds: []
-    # A list of groupIds that are authorized to manage (Create, Update, Publish & Delete) event based processes.
-    authorizedGroupIds: []
+  # A list of groupIds that are authorized to manage (Create, Update, Publish & Delete) event based processes.
+  authorizedGroupIds: []
```
pepopowitz, line 465: I had to outdent this block in order to commit -- yaml didn't like it being indented deeper than `authorizedUserIds` above it. 

I assumed this was the intention -- if you know these settings and know that I was incorrect, let me know!

docs/self-managed/optimize-deployment/configuration/system-configuration.md
```diff
@@ -130,74 +130,74 @@ at least one engine configured at all times. You can configure multiple engines
...
+| engines.${engineAlias}.authentication.enabled  | false                             | Toggles basic authentication on or off. When enabling basic authentication, please be aware that you also need to adjust the values of the user and password.                                                                                                                                                                                                                                                                                           |
+| engines.${engineAlias}.authentication.user     |                                   | When basic authentication is enabled, this user is used to authenticate against the engine.<br /><br /> Note: when enabled, it is required that the user has <ul><li>READ & READ*HISTORY permission on the Process and Decision Definition resources</li> <li>READ permission on \_all* ("\*")Authorization, Group, User, Tenant, Deployment & User Operation Log resources</li></ul> to enable users to log in and Optimize to import the engine data. |
```
pepopowitz, line 143: @christinaausley the build broke on this changed line, prettier had a hard time formatting the asterisks and TBH it didn't do a great job. 

I intend to fix it, but I want to make sure I'm fixing it correctly. Do you know what the `("*")` on this line is intended to indicate? It seems like a strange outlier that I think I might be able to remove. 

For reference, [here is the section on the live site](https://docs.camunda.io/docs/self-managed/optimize-deployment/configuration/system-configuration/#connection-to-camunda-platform-7). The line (and `(*)`) in question looks like this: 

image

docs/self-managed/optimize-deployment/configuration/service-config.yaml
christinaausley: Good question for @menski

docs/self-managed/optimize-deployment/configuration/system-configuration.md
menski: I think this is about Optimize and maybe @megglos or @RomanJRW can better answer this question

docs/self-managed/optimize-deployment/configuration/service-config.yaml
menski: I think this was unintended and the fix is correct if I look at the default [config file.](https://github.com/camunda/camunda-optimize/blob/67de788cb2263628e7247305bf7d3261c787dcbe/distro/src/config/environment-config.yaml#L291-L294)

docs/self-managed/optimize-deployment/configuration/system-configuration.md
pepopowitz: For the sake of moving forward, I updated this line so that it _almost_ matches what's currently in production, and doesn't confuse prettier. 

It turns out prettier was _really_ getting confused by the underscore in READ_HISTORY, resulting in weird things happening with the rest of the line. In https://github.com/camunda/camunda-platform-docs/pull/928/commits/4763b2f41396dd1c32cf849823c9b5cd2cbb38d5, I surrounded READ_HISTORY with backticks to turn it into a code block, so that prettier doesn't touch it. I also surrounded READ with backticks, for consistency. 

With the code blocks added, this item now looks like this: 

image

My hope is that we can merge this as-is, and if there are changes that come up from @megglos or @RomanJRW I will add them in a follow-up PR. 

@christinaausley would you mind reviewing again to verify these changes conform to our writing guidelines?

christinaausley (approved):

PR #929, opened 2022-05-30
Added content from blog post to migration guide

berndruecker:
Added content from https://camunda.com/blog/2022/05/what-to-do-when-you-cant-quickly-migrate-to-camunda-8/ to the migration guide.

Not sure if we need to copy over the images - at the moment I just referred to the one existing on the Camunda blog. Let me know what you think!.


christinaausley (approved):
Minor technical adjustments in latest commit for clarity. Pushed a second commit as Prettier didn't run in my first. LGTM!

berndruecker:
Thanks Christina!

PR #946, opened 2022-06-01
chore(self-managed): polish introduction

MaxTru:
* I found this page via google
* Reading the intro sentence <The alternative way to use Camunda Platform 8 ...> read super strange when this is the first page to start with. I immediately asked myself "alternative to what"?
* I propose this as polishing


christinaausley (approved):

akeller:
Thanks for the update @MaxTru and thanks for the review @christinaausley! LGTM!

PR #947, opened 2022-06-01
chore(self-managed): cross-link desktop modeler

MaxTru:
* This will ensure easy discoverability and allow users to understand more, if they would like to

### DIFF
--- a/docs/self-managed/overview.md
+++ b/docs/self-managed/overview.md
@@ -26,7 +26,7 @@ The following components are available for Camunda Platform 8 Self-Managed:
 * Optimize
 * Identity (not available in Camunda Platform 8 SaaS)
 
-Camunda Platform 8 Self-Managed users may also use Desktop Modeler to build BPMN diagrams and work with Camunda Forms.
+Camunda Platform 8 Self-Managed users may also use [Desktop Modeler](../../components/modeler/desktop-modeler/install-the-modeler) as an addition to these components. The Desktop Modeler can be used by Process Developers to build BPMN diagrams, DMN diagrams, or Camunda Forms for automation.
 
 While documentation for Optimize is available in the Self-Managed section, we continue to refine the experience for Camunda Platform 8 Self-Managed users.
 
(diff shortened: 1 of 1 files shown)


docs/self-managed/overview.md
```diff
@@ -26,7 +26,7 @@ The following components are available for Camunda Platform 8 Self-Managed:
 * Optimize
 * Identity (not available in Camunda Platform 8 SaaS)
 
-Camunda Platform 8 Self-Managed users may also use Desktop Modeler to build BPMN diagrams and work with Camunda Forms.
+Camunda Platform 8 Self-Managed users may also use [Desktop Modeler](../../components/modeler/desktop-modeler/install-the-modeler) as an addition to these components. The Desktop Modeler can be used by Process Developers to build BPMN diagrams, DMN diagrams, or Camunda Forms for automation.
```
akeller, line 29: @christinaausley thoughts on capital "Process Developers"? My instinct says to lowercase this. Do we use it anywhere else?

christinaausley: Ah yes. This is also used [here](https://docs.camunda.io/docs/components/best-practices/architecture/deciding-about-your-stack/). Across the rest of our documentation, however, we always reference *developer* rather than *Developer*, so I would lowercase this.

Regarding the best practices link I shared, I'll launch a PR to ensure this is lowercase throughout.

akeller: Thank you! I'll update it here.

akeller, line 29:
```suggestion
Camunda Platform 8 Self-Managed users may also use [Desktop Modeler](../../components/modeler/desktop-modeler/install-the-modeler) as an addition to these components. The Desktop Modeler can be used by process developers to build BPMN diagrams, DMN diagrams, or Camunda Forms for automation.
```

akeller (changes requested):

akeller (approved):

PR #948, opened 2022-06-01
chore(format): set long-term strategy for formatting old versions

pepopowitz:
Partially addresses https://github.com/camunda/camunda-platform-docs/issues/883.

1. Enables prettier formatting for all `versioned_sidebars`. 
2. Formats all `versioned_sidebars`.
3. Expands the `versioned_docs` prettier exclusion into the exact set of older versions. Moving forward, the current `docs` and any new versions we cut will be expected to meet prettier's formatting rules. Older versions will remain ignored by prettier -- it doesn't seem worth taking the time/effort to update them.


menski (approved):

akeller (approved):
@pepopowitz we briefly discussed this in our 1:1 and I'm totally onboard with your proposed formatting strategy -  "Older versions will remain ignored by prettier -- it doesn't seem worth taking the time/effort to update them."

Thank you for the work here!

PR #956, opened 2022-06-03
[DO NOT MERGE] messing with workflow changes

pepopowitz:
This is a PR from a forked repo!

(Don't merge it though.)


pepopowitz:
This forked branch worked exactly like I wanted it to -- no checks until the PR, then one check, then a check for each additional commit. 

But it's based on a workflow configuration that won't work (see #957 for why; #955 is the configuration change.)

PR #962, opened 2022-06-06
chore: remove unused mdx-js/react dependency

pepopowitz:
While investigating the build failure in #919, I got curious about whether this dependency is even needed. 

I don't think it is, so this PR removes it. 

MDX support [is built into docusaurus](https://6220d0362c028b000827f851--docusaurus-2.netlify.app/docs/2.0.0-beta.15/markdown-features/react), even in the version we're using. I suspect that maybe it initially wasn't, and that's why the dependency [was added early on](https://github.com/camunda/camunda-platform-docs/blame/c13600365b206bf4d5de6e779e249f56063730dc/package.json).

I've tested the site locally, and found that docs that are using MDX render fine without this dependency. For example, the tabs we have throughout the site: 

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,6 @@
         "@auth0/auth0-react": "^1.10.1",
         "@docusaurus/core": "2.0.0-beta.15",
         "@docusaurus/preset-classic": "2.0.0-beta.15",
-        "@mdx-js/react": "^1.6.22",
         "clsx": "^1.1.1",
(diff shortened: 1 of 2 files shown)


akeller:
Can you double-check the Best Practices section? I know Bernd had to do some work to get the BPMN diagrams to render correctly and I want to make sure there isn't something using it there. 

Otherwise, this makes sense to me to remove an unused dependency.

akeller (approved):
Remove it and see what happens 😈

PR #974, opened 2022-06-15
Improve wording to avoid misunderstandings

berndruecker:
See https://forum.camunda.io/t/current-user-assignee-in-camunda-8-tasklist-claimtask-mutation/38062


berndruecker:
I think this does not make too much sense to fix in isolation now - but it should be tackled as part of https://github.com/camunda/tasklist/issues/1213 instead. Commented there and close it here for now.

PR #976, opened 2022-06-17
fix: resolve broken builds by updating package-lock and specify node version via .nvmrc

pepopowitz:
This PR specifies a node version in a .nvmrc file, and updates the package-lock based on that version. This is intended to fix our current situation of all builds failing for the last ten days 😬 😅 

### TL;DR

A new version of node v16 was released, including a newer version of npm, and we needed to update the package-lock file accordingly. 

#### The journey

Looking through [the list of recent workflow runs](https://github.com/camunda/camunda-platform-docs/actions/workflows/build-docs.yaml), it looks like the builds stopped failing between [this successful check](https://github.com/camunda/camunda-platform-docs/runs/6776614878?check_suite_focus=true) and [this unsuccessful one](https://github.com/camunda/camunda-platform-docs/runs/6814730392?check_suite_focus=truehttps://github.com/camunda/camunda-platform-docs/runs/6776614878?check_suite_focus=true). 

The error message in the failing checks is about the package.json and package-lock.json not being in sync:

image

#### But I couldn't update the lockfile locally

I tried to re-run `npm install` locally, to update the package-lock, but there were no changes!!!

This made me suspicious of node/npm versions. 

#### The smoking gun

I noticed a difference between those two checks -- the successful one was using node v16.15.0 & npm v8.5.5, and the failing one was using node v16.15.1 & npm v8.11.0. So, yeah -- different.

I confirmed by updating my local environment to use v16.15.1, and _now_ when I ran `npm install`, I got an updated package-lock file. That's what's in this PR.

## Why the .nvmrc?

For people like me using a node version manager, the .nvmrc is automatically read by local tooling to use the specified version of node. By specifying this from now on, we'll increase the likelihood that we'll spot this kind of difference sooner rather than later. 

### What about our workflows? Do they read the nvmrc?

No, not yet. Right now, we're specifying [a node version in the workflows](https://github.com/camunda/camunda-platform-docs/blob/b74bbb2f95f142dd5991db9087c16a938b5a59e8/.github/workflows/build-docs.yaml#L24), and we're specifying it broadly -- `16`. This is kind of the reason for these build failures -- the new version of v16 was released at the beginning of this month, and our builds started using it, but likely most of us weren't using that version of node locally, so none of us generated an updated package-lock.

In the future, I think we should update our configs to use the nvmrc file as [described in this PR](https://github.com/actions/setup-node/pull/338), instead of specifying the version in each workflow. But right now I just want to get things building again, before I go on another week of vacation 😄 

I could also see us saying that the nvmrc specification should be at the major level, e.g. `16`, instead of the patch level like it is in this PR, e.g. `16.5.1`. This scenario, of an updated minor/patch node version causing build failures, is uncommon, and there is probably more value in the free updates we get by specifying only the major version. 🤷


akeller:
> In the future, I think we should update our configs to use the nvmrc file as https://github.com/actions/setup-node/pull/338, instead of specifying the version in each workflow. But right now I just want to get things building again, before I go on another week of vacation 😄

I'm good with this and happy to merge.

PR #984, opened 2022-06-23
add diagram errors page to web modeler doc

nithinssabu:
Closes https://github.com/camunda/web-modeler/issues/1555


nithinssabu:
> Hi @nithinssabu - have changed a few wordings / terminology in the doc - let me know if you agree

@MoAtCamunda I agree with all the other changes except for the use of the word **debug rules**. Lint rules / validation rules sounds better in my opinion.

For my programmer mind, debugging is a process that I follow to find errors in my code and I make use of lint errors, stack traces, console error messages, etc. to do this.

Just sharing my opinion here and finally it's your decision what to use here.

docs/components/modeler/web-modeler/diagram-errors.md
```diff
@@ -0,0 +1,33 @@
...
+
+### Interactivity
+
+The debug errors are interactive - Clicking on the element ID in the error line highlights the corresponding element in the canvas and points to the specific property in the properties panel where you can resolve the issue.
+
+![error panel](img/diagram-errors/interactivity.png)
+
+### Engine errors
+
+Not all errors are currently covered by the debug rules. Engine errors which are shown when you deploy a process and shown in the same debug panel. At present, we do not support interactivity for these errors.
```
CatalinaMoisuc, line 31: We should rephrase this to be positive. Please take this as a quick suggestion, not a proposal for rephrasing per se.

"If you fixed all the lint errors and further errors occur when deploying the diagram, they will be shown in the same panel. Keep in mind that some errors could also occur i.e. due to your cluster not being healthy or available etc. and while you can see these errors, you can't interact with them in the same way you can with the lint errors."

docs/components/modeler/web-modeler/diagram-errors.md
```diff
@@ -0,0 +1,33 @@
...
+
+Web Modeler continuously validates implementation properties for a process diagram while the user is modeling, based on a set of debug rules. The debug rules are added to the debug panel at the bottom of the web modeler. Simply expand the debug panel to be able to see the errors.
+
+![error panel](img/diagram-errors/error-panel.png)
+
+### Expand / collapse
+
+The debug panel can be expanded or collapsed by clicking anywhere on the header. The panel is collapsed by default and the latest state (expanded or collapsed) is remembered when you next open the web modeler.
+
+### Version selection
```
CatalinaMoisuc, line 17:
```suggestion
### Engine version selection
```

docs/components/modeler/web-modeler/diagram-errors.md
```diff
@@ -0,0 +1,33 @@
+---
+id: diagram-errors
+title: Diagram errors
+description: This feature will assist you in debug and fix errors in your processes.
```
CatalinaMoisuc, line 4: I think we should establish an absolute minimum set of terms we use for this feature.

As a User reading this, I might get very confused when reading: "lint errors", "engine errors", "debug panel" etc.

I would rather narrow the terms down to two types of errors:
* Lint errors
* Deploy errors

And since the panel will always display errors, then use the
* Diagram error panel
term for the component. Also, why call it "debug panel" if the title on the component is "Diagram errors"?
We should ensure that the terms in the UI are aligned with the documentation.

CatalinaMoisuc (changes requested):

akeller:
@christinaausley can you review this PR? Let me know if you have any questions or need clarification.

docs/components/modeler/web-modeler/diagram-errors.md
```diff
@@ -0,0 +1,33 @@
+---
+id: diagram-errors
+title: Diagram errors
+description: This feature assists you in debugging and fixing errors in your processes.
+---
+
+### Lint errors
+
+Based on a set of debug rules, Web Modeler continuously validates implementation properties for a process diagram while the user is modeling. The debug rules are added to the debug panel at the bottom of Web Modeler. Expand the debug panel to view the errors.
```
christinaausley, line 9: I would clarify here in the case someone cannot see the direct image. For example, "Expand the debug panel to view the errors by clicking the **Diagram errors** header."

Given the following section alludes to this, you could say "Expand the debug panel to view the errors by clicking the **Diagram errors** header. The panel is collapsed by default and the latest state (expanded or collapsed) is remembered for the next time you open Web Modeler." Then, you could just remove the `### Expand and collapse` section as I don't think we need a section dedicated to this.

docs/components/modeler/web-modeler/diagram-errors.md
```diff
@@ -0,0 +1,33 @@
...
+
+### Interactivity
+
+The debug errors are interactive. Clicking on the element ID in the error line highlights the corresponding element in the canvas and points to the specific property in the properties panel where you can resolve the issue.
+
+![error panel](img/diagram-errors/interactivity.png)
+
+### Engine errors
+
+Not all errors are currently covered by the debug rules. Engine errors which are shown when you deploy a process and shown in the same debug panel. Currently, we do not support interactivity for these errors.
```
christinaausley, line 31: Do you mean "Engine errors which are shown when you deploy a process **are** shown in the same debug panel."?

christinaausley (dismissed):
My latest commit includes some technical adjustments for clarity, including capitalization of Web Modeler.

Please note I left two comments, one a suggestion and one a necessary correction for clarification. Once these comments are resolved, LGTM from a technical perspective.

nithinssabu:
@MoAtCamunda As promised, can you please add the reasoning here for why we want to go with the terms "Debug errors" and "Debug panel"?

docs/components/modeler/web-modeler/diagram-errors.md
nithinssabu: Yes, that's what I meant. I changed it to `thrown` and also changed the text taking into account [this suggestion](https://github.com/camunda/camunda-platform-docs/pull/984#discussion_r905931637).

christian-konrad:
@nithinssabu @CatalinaMoisuc @MoAtCamunda 

For now, I suggest it should look like this:
  
**TLDR:**
The panel should be called **error panel**, though we are open to changing the name in the future. 

The page could be called **Fix problems in your diagram**.

The feature scope is **diagram validation**, but it may grow in the future into **debugging**.

**Full reasoning:**
The overall feature scope (the outcome; the job it carries out for the user) is **debugging**. We coin this term on our roadmap.

For what we offer _now_, we may not call it debugging (in order not to raise expectations too high), since the panel does not help in finding _bugs_ (misbehavior of the model implementation), but only in finding diagram errors that actually prevents the model from being deployed at all. IDEs call this kind of feature „Code Inspection“ (see the JetBrains documentation for reference [https://www.jetbrains.com/help/idea/code-inspection.html](https://www.jetbrains.com/help/idea/code-inspection.html)). So, let’s call the scope of what we are currently limited to with our tooling (the Web Modeler) **diagram validation** (or **model validation**) and call the documentation page **Fix problems in your diagram**.

One of the features we provide to assist users with **model validation** is the **error panel**.  

It _may_ later transform into a **problems panel, validation panel** or even a **debugging panel** once it not only displays errors but also, for example, warnings or suggestions, or provides additional tabs to support further debugging cases (like a tab where a user can test processes partially by providing mock data, or a tab to inspect data flow throughout a process). Until then, it should continue to be called **error panel** both in the UI (Desktop: Errors / Web: Diagram errors) and the docs so as not to confuse users.  

The technique we apply to detect/identify and show errors (and warnings + suggestions in the future) is **linting** (applying lint rules through a static checker). This is how we solve that problem (and this is only one possibility of many), but it is not the overall outcome/gain we sell. Developers know that term very well, but not business people. We should _not_ use this term now in public communications; yet it is still valid to use the term in the context of customizing the Desktop Modeler, as keen users can use the bpmn-lint plugin to write their own linting rules. We may offer custom linting rules in the Web Modeler in the future (e.g. in the context of governance rules), but until then, we don’t use this term.

For now, we should call these types of errors **design time errors** (or even **diagram errors**) and if we want to mention it, the process of identifying those is **linting**, which is a kind of **static model analysis.**

So, they are easily differentiable from **deploy time errors** (prior _engine errors_) or just **deployment errors** (do we also catch errors when starting instances?).

We can not just call the design time errors „_debug errors_“, as debugging in general is a broader scope and most people understand it as inspecting current application state (variables, call stack, UI/DOM…) or traces of states, e.g. by using breakpoints. (Also, "debug errors" to me would probably describe errors in the debugger itself).


Bildschirmfoto 2022-06-28 um 20 08 04

nithinssabu:
> For now, I suggest it should look like this:

@christian-konrad Thanks for your feedback. I love the reasoning here. 

I made the suggested changes to the best of my understanding. Please read through it once more and approve the PR if you think it's ready to be merged.

CatalinaMoisuc (approved):

PR #994, opened 2022-06-28
docs(identity): add troubleshooting for SSL issues

Ben-Sheppard:
Resolves https://github.com/camunda-cloud/identity/issues/752

We have noticed issues appearing when deploying the platform to environment where container to container communication is performed using IPs which fall outside of the local IP-range. 

By default, Keycloak requires SSL to be configured for all "external" requests, which renders the components unable to authenticate as Keycloak considers the requests as originating from external sources.

This PR introduces a troubleshooting section for this behaviour and offers one possible solution.


docs/self-managed/troubleshooting/platform-deployment.md
```diff
@@ -0,0 +1,17 @@
+---
+id: platform-deployment
+title: "Platform deployment"
+---
+
+## Connection issues between Identity and Keycloak
+
+By default, Keycloak requires all non-local IP ranges to use SSL. It has been noted that in certain
+cloud environments where communication between containers is performed using IPs outside the local range,
+components are not able to function correctly.
```
dlavrenuek, line 10: I would add an example of the logged exception (there should be some examples in the JIRA issues), this way you can get to this page just by googling the error message

docs/self-managed/troubleshooting/platform-deployment.md
```diff
@@ -0,0 +1,17 @@
...
+
+To resolve this, we would suggest disabling the requirement for SSL connections by following the steps
+in the [SSL modes](https://www.keycloak.org/docs/16.1/server_admin/#_ssl_modes) section of the Keycloak
+docs and setting the `Require SSL` field value to `none` for both the `master` and `camunda-platform` realms.
```
dlavrenuek, line 14: The default solution should be to configure the ip range for containers to be "local", so this section should contain the specific IP ranges. As a fallback (only if configuring local IP ranges is not possible) we can then suggest disabling SSL requirements in keycloak. It has to be noted that this is a two step process:
1. disable SSL on `master` realm, restart Identity
2. disable SSL on Camunda realm, restart Identity

dlavrenuek (changes requested):

dlavrenuek (dismissed):
The provided information and the structure looks good from my side 👍

docs/self-managed/identity/troubleshooting/common-problems.md
```diff
@@ -0,0 +1,55 @@
...
+```
+
+This can be caused by:
+
+1. The Keycloak service has not started/is not ready
+2. Identity making requests from an external IP address
+
+### Solution 1: The Keycloak service has not started/is not ready
+
+The Keycloak service can take time to start due to the supporting systems, please provide Keycloak with adequate time
```
akeller, line 23: This is a bit of a nit pick - but do we have any idea about how much time? Or is the guidance to monitor the logs like suggested below?

docs/self-managed/identity/troubleshooting/common-problems.md
```diff
@@ -0,0 +1,55 @@
...
+   to achieve this:
+   1. In the `master` realm set `Require SSL` to `none` by following the steps in [SSL modes](https://www.keycloak.org/docs/16.1/server_admin/#_ssl_modes)
+   2. Restart the Identity service
+   3. In the `camunda-platform` realm set `Require SSL` to `none` by following the steps in [SSL modes](https://www.keycloak.org/docs/16.1/server_admin/#_ssl_modes)
+   4. Restart the Identity service again
+   5. Identity should now start successfully
+
+:::warning
+We would only recommend that requirements for SSL are disabled in a development environment.
+:::
```
akeller, line 55: I really like how you worded this section @Ben-Sheppard! "We suggest you consider these options" is perfect for this scenario. We should never make a recommendation to disable SSL, but we should guide users through understanding their options depending on their environment and their tolerance. Great job here!

docs/self-managed/identity/troubleshooting/index.md
```diff
@@ -0,0 +1,45 @@
+---
+id: index
+title: "Troubleshooting Identity"
+sidebar_label: "Overview"
+---
+
+## Issue: Identity pod crashloops/continually restarts
+
+If the Identity pod crashloops it is likely that there is an issue with the connection to Keycloak.
```
akeller, line 9:
```suggestion
If the Identity pod crashloops, or the pod continually restarts, it is likely that there is an issue with the connection to Keycloak.
```

Is "crashloop" specific to Kubernetes? Again, nit pick, but I would add this to the text (even though its in the header). It's a pretty intuitive word for me, but I'm not sure if it would translate well for non-native English speakers or devs outside of the Kubernetes space.

docs/self-managed/identity/troubleshooting/common-problems.md
Ben-Sheppard: Thank you for the feedback!

docs/self-managed/identity/troubleshooting/index.md
Ben-Sheppard: Good question, in this instance its a specific error status within Kubernetes `CrashLoopBackOff` but I agree that the general term `crashloops` potentially does not translate well.

I'll add in the text as it covers all cases 👍🏻

Ben-Sheppard:
> For someone who is troubleshooting, do you think they know enough to understand how to restart the Identity service? Or is that documented somewhere else we can link to? I always grapple with the right amount of context for system administration type steps.

This is a great question, I too question how much context is too much. My perspective here is that for a user to get to the stage where this issue appears they will have had to progress through a few other stages that require the use of similar commands, or, stages that link to external documentation in some form that means further links are unlikely to provide great value.

docs/self-managed/identity/troubleshooting/common-problems.md
akeller: Instead of mentioning time at all, I would direct users to monitor the logs for the lines you've listed below.

docs/self-managed/identity/troubleshooting/index.md
akeller: Thanks! 

FYI @christinaausley "crashloops" might be a good term for you to research. I'm seeing it online as CrashLoops, crash loop, crashloops, etc. Singular, plural, various casing and spacing. Like Ben mentions, it's related to CrashLoopBackOff.

akeller (approved):
Looks great Ben! Thanks for the care and attention to detail you put into this PR.

PR #995, opened 2022-06-28
make console api rest more visible

christinaausley:
Make Console API REST more visible at top of page with added note.


docs/apis-clients/console-api-reference.md
```diff
@@ -4,6 +4,14 @@ title: Console API clients (REST)
 description: "To interact with Camunda Platform 8 programmatically without using the Camunda Platform 8 UI, create Cloud API clients."
 ---
 
+## Console API (REST)
+
+For all requests, include the access token for Cloud API in the Authorization header: `authorization:Bearer ${TOKEN}`.
+
+:::note
+A detailed API description can be found [here](https://console.cloud.camunda.io/customer-api/openapi/docs/#/).
```
akeller, line 12: I would call this interactive for users with a valid C8 account. Also might be good to check in with @ultraschuppi to see if this is available for users after their trial is over (and they haven't converted to a paid tier).

akeller: I'm also logged into C8, but I'm not able to see anything render at this URL. The Favicon loads, but the page doesn't. Is there anything else we need to mention to help someone reach the content on this page?

akeller: image

🤷‍♀️

akeller: This is rendering for me, thanks @ultraschuppi!

ultraschuppi:
just to make this crystal clear

https://console.cloud.camunda.io/customer-api/openapi/docs/#/ is public, is is not even looking different if you have a C8 account, if it is active or expired, enterprise or whatever. does not matter - it does not check for a running session

users may add a jwt token on the ui to actually call the corresponding methods

akeller:
> A detailed API description can be found [here](https://console.cloud.camunda.io/customer-api/openapi/docs/#/).

I would make a slight adjustment

> A detailed API description can be found [here](https://console.cloud.camunda.io/customer-api/openapi/docs/#/) via Swagger. With a valid access token, this offers an interactive API experience against your Camunda Platform 8 cluster.

or something similar. No need to link to how to create one since we are already on the right page for that.

ultraschuppi:
> So is there anything we need to concretely add to the text here? Should we mention a jwt token on the UI will assist in calling the corresponding methods, or leave as-is? Should we mention this is applicable for users with a valid C8 account? Just curious about what you feel is most relevant/applicable here.

- developers who know about openapi should be used to this haptics already
- it is described here - https://docs.camunda.io/docs/apis-clients/console-api-reference/

my feeling: no need to add additional stuff

akeller (approved):

PR #1011, opened 2022-07-01
chore(docs): update Optimize FE and BE dependencies

RomanJRW:


megglos (approved):
Lgtm!

docs/reference/dependencies.md
```diff
@@ -944,7 +944,7 @@ All of these libraries are required for core functionality.
 - [@types/unist@2.0.3](https://github.com/DefinitelyTyped/DefinitelyTyped) (MIT)
 - [bail@2.0.1](https://wooorm.com) (MIT)
 - [bpmn-js-disable-collapsed-subprocess@0.1.4](https://github.com/barmac) (MIT)
-- [bpmn-js@8.9.1](https://github.com/nikku) (MIT-)
+- [bpmn-js@8.9.1](https://github.com/nikku) (MIT\*)
```
akeller, line 947: Is this `\*` intentional? It's rendering as `MIT*`.

RomanJRW: I think it should be `MIT*`. I suspect that in the past someone did a find/replace for `-` to `*`, leaving us with a bunch of `MIT-` entries. The escape character was added by the prettier. So I think we should go ahead and merge 🙂

PR #1012, opened 2022-07-01
Document start instructions for CreateProcessInstanceRequest gRPC message

### LINKED ISSUE #943: Update GRPC API with start instructions for CreateProcessInstance RPC
Concerns the [CreateProcessInstance RPC](https://docs.camunda.io/docs/apis-clients/grpc/#createprocessinstance-rpc) and the [CreateProcessInstanceWithResult RPC](https://docs.camunda.io/docs/apis-clients/grpc/#createprocessinstancewithresult-rpc).

korthout:
The gRPC message `CreateProcessInstanceRequest` now supports start instructions which allow a user to switch from starting at the default none start event and instead, start 1 or multiple tokens at specific elements in the process. See https://github.com/camunda/zeebe/issues/9366 for more details on the feature.

This updates the Zeebe API (gRPC) documentation with the new start instructions definition. It also rectifies a previous change that wasn't pulled into the docs.

Note that no new errors are introduced for this feature. This may change in the future as command rejection reasons will be expanded.

This feature will be part of the upcoming 8.1 release, and so this PR targets the corresponding `release_8_1` branch.

closes #943

### DIFF
--- a/docs/apis-clients/grpc.md
+++ b/docs/apis-clients/grpc.md
@@ -174,7 +174,7 @@ Only processes with none start events can be started through this command.
 message CreateProcessInstanceRequest {
   // the unique key identifying the process definition (e.g. returned from a process
   // in the DeployProcessResponse message)
-  int64 processKey = 1;
+  int64 processDefinitionKey = 1;
   // the BPMN process ID of the process definition
   string bpmnProcessId = 2;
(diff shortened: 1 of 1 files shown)


saig0 (approved):
@korthout looks good :+1:

PR #1015, opened 2022-07-01
chore(deps): Update docusaurus from beta-15 to beta-16, including configuring new Algolia instance

pepopowitz:
The main goal of this PR is to update from docusaurus beta-15 to beta-16, but it includes a couple side-quests required by this update: 

* Configures a standalone Algolia DocSearch instance
* Removes empty sidebar items

### Before: Production site before publishing a new doc

Note that there is no link in the sidebar titled "Create decision tables using DMN" under Next Steps. This doc was merged but not yet published:
 
image

### Before: Local search results (using new algolia index) before publishing the new doc

Note that there are no search results for the new article, searched by exact title: 

image

### ...And then I published the new doc

Note that the DMN article now appears under Next Steps, so it's live:

image

### ...And then I triggered a re-crawl of the new index

Note that the search results locally, pointed at the new algolia instance, found the new doc by exact title: 

image

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -191,10 +191,10 @@ module.exports = {
       copyright: `Copyright © ${new Date().getFullYear()} Camunda`,
     },
     algolia: {
-      apiKey: "f1b2a46296ae374b7f5a5c627341c354",
+      // These keys are for our new standalone algolia instance!
+      apiKey: "d701d38126d1a43866047d3ab97680d1",
+      appId: "6KYF3VMCXZ",
       indexName: "camunda",
-      contextualSearch: true, // useful for versioned docs (https://docusaurus.io/docs/search#contextual-search)
-      searchParameters: {}, // Optional (if provided by Algolia)
     },
     // Disabling Dark Mode
     // https://github.com/camunda-cloud/camunda-cloud-documentation/issues/125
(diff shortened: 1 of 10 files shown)


docusaurus.config.js
```diff
@@ -191,10 +191,10 @@ module.exports = {
       copyright: `Copyright © ${new Date().getFullYear()} Camunda`,
     },
     algolia: {
-      apiKey: "f1b2a46296ae374b7f5a5c627341c354",
+      // These keys are for our new standalone algolia instance!
+      apiKey: "d701d38126d1a43866047d3ab97680d1",
```
pepopowitz, line 195: GitGuardian is not happy about this apikey being exposed, but it doesn't seem like there's a lot of evil that someone could do with it. Maybe the worst is to try to take down Algolia and pin it on us. 

We could take the extra time to inject it at build time, but even if we did, it'd still be exposed on the built public website.

akeller: As you mentioned over Slack, this is the way Facebook/Docusaurus is doing it so 🤷‍♀️ .

I have no issues with this. This is working as intended.

akeller (approved):
The appendix, much like the human organ, just hangs out until it's annoying and then gets removed.

PR #1016, opened 2022-07-01
adding appropriate links to Zeebe Chaos

christinaausley:
Incorporated the links we discussed via Slack. Would you prefer to see them elsewhere?


Zelldon (approved):
Thanks @christinaausley ! Looks good to me I first thought that we wanted to add the content from the blogs, but I like this way as well :+1:

christinaausley:
@Zelldon Would you prefer to pull in direct content from the blogs rather than link to the resource?

akeller:
Chiming in here and then I'll see myself out - linking out, in this case, makes more sense to me than making it "official" docs. One of these topics feels quite conceptual and the other is best kept as a blog for context.

PR #1019, opened 2022-07-04
Link to new Go getting started guide

npepinpe:
We added a new getting started guide for the Go client, but forgot to update the link in the docs.

One thing to note: the code in the getting started guide on GitHub, and the one here in the docs, differs. I thought about aligning them, but decided against, primarily because the [code on GitHub](https://github.com/camunda/camunda-platform-get-started/tree/main/go) is more of a complete sample application, and the one here in the docs is just showing how to do specific things.

### DIFF
--- a/docs/apis-clients/go-client/get-started.md
+++ b/docs/apis-clients/go-client/get-started.md
@@ -6,7 +6,7 @@ sidebar_label: "Getting started guide"
 
 In this tutorial, you will learn how to use the Go client in a Go application to interact with Camunda Platform 8.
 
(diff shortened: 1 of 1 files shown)


npepinpe:
Not sure why the build fails when all I did was change some Markdown... :thinking:

korthout:
> Not sure why the build fails when all I did was change some Markdown... 🤔

@npepinpe I don't see a failed build 😕 Merging is blocked because of a missing approval. Is that what you meant?

korthout (approved):
Thanks @npepinpe LGTM 💯

christinaausley (approved):

npepinpe:
The build was initially failing because I had forked from master and not main :sweat_smile:

PR #1023, opened 2022-07-05
Document create process instance at user-defined elements

### LINKED ISSUE #944: Document "Start Process Instance from Anywhere" feature
This concerns [Process instance creation](https://docs.camunda.io/docs/components/concepts/process-instance-creation/).

We might also want to add an example like [Create a process instance with result](https://docs.camunda.io/docs/apis-clients/java-client-examples/process-instance-create-with-result/)

korthout:
Describes:
- how a user can start a process instance at different elements than the default none start event.
- why a user might want to do this.
- when not to use this.

And provides a code example.

This does not:
- describe how to use it as a workaround for the missing Process Instance Migration feature (that should be described in the feature request issue).
- link to the start instructions gRPC definition, but perhaps it should, please let me what you think.
- add an example like https://docs.camunda.io/docs/apis-clients/java-client-examples/process-instance-create-with-result/, but perhaps it should, please let me know what you think.

closes #944


saig0 (approved):
@korthout looks good :+1:

Issue #1024, opened 2022-07-05
Bring docusaurus dependencies up to date


pepopowitz:
Docusaurus dependencies have been stuck at version beta-15 for some time, due to Algolia DocSearch migration. For security & maintenance reasons, we should bring them as up to date as possible. 

Work included in this issue (some may be combined if they prove to be small/easy updates): 

- [x] #1015 
- [x] #1036 
- [x] #1068 
- [x] #1085
- [x] #1100
- [x] #1025 
- [x] #1114 

cc @akeller for visibility

akeller:
Cheering you on from the sidelines! Let me know if you need extra eyes on anything.

pepopowitz:
Marking this done!!!! We're not fully up-to-date because they just released 2.0.0, but this got us close enough that future updates should be pretty routine.

PR #1025, opened 2022-07-05
chore: remove unused DocVersionSuggestions customization

pepopowitz:
Partially addresses #1024 

DocVersionSuggestions was introduced to globally suppress the warning about viewing an older version. It was introduced at a time that docusaurus would show the message even on our "current" version of the docs, which was confusing, so we turned it off entirely. 

Since it was introduced, two things have happened (actually at the same time): 
1. Docusaurus [added significant configurability of the warning banner](https://github.com/facebook/docusaurus/pull/5052). 
2. Docusaurus [changed the name of this customized file](https://github.com/facebook/docusaurus/pull/5052/files#diff-875dd9b45837be247d089aed4e70e19b7c8cb9efc731c2e84df325b34f5a570d), making this customization moot -- it wasn't actually doing anything anymore.

Since the base functionality is what we want, we can delete this unused customization.


pepopowitz:
CI is red until #1026 gets merged

akeller (approved):
Onward and upward!

akeller:
@pepopowitz this is good to merge IMO

PR #1064, opened 2022-07-18
Fix callout number related to tasklist-mockup.png

MitchelNijdam-Rockstars:
The numbers in the callouts were `1, 3, 3` instead of `1, 2, 3`.


akeller (approved):
Wow! Great catch here @MitchelNijdam-Rockstars thank you!

PR #1065, opened 2022-07-18
chore: update screenshots for modeler 5.1.0

marstamm:


akeller (approved):
@marstamm these screenshots look updated to me! Should we go ahead and merge and release so they are live (since Modeler 5.1 is already released)?

PR #1066, opened 2022-07-18
fix gsg structure overview

christinaausley:
According to https://github.com/camunda/camunda-platform-docs/issues/1037, adjusting the order of the components and adding details on the guides section regarding the getting started steps and the next steps.


akeller (commented):
One bonus period and a request - could you change the name of the markdown file? I'm actually not sure if it will require a redirect/rewrite since the URL is /guides/ per the slug.

akeller (approved):
Thank you! This looks so much better!

PR #1070, opened 2022-07-20
Omit public modifier for test classes

Hafflgav:
As being recommended by the JUnit 5 User Guide. 
See: https://junit.org/junit5/docs/snapshot/user-guide/#writing-tests-classes-and-methods 

(This was brought up during a meetup talk I was giving about this topic. I thought it would be good to improve our documentation in that sense as well)

### DIFF
--- a/docs/components/best-practices/development/testing-process-definitions.md
+++ b/docs/components/best-practices/development/testing-process-definitions.md
@@ -70,7 +70,7 @@ A test can now look like the following example. The complete source code is avai
 ```java
 @SpringBootTest
 @ZeebeSpringTest
-public class TestTwitterProcess {
+class TestTwitterProcess {
 
     @Autowired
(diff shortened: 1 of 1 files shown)


Hafflgav:
Thanks @akeller, I only investigated the testing best-practices as they should be accurate. 

So far I have not run a bigger search. Feel free to do so and pot me to the correct pages. :)

akeller:
Did a little more digging and I couldn't find any other test code snippets. It looks like most of the test code is on [this repo](https://github.com/camunda/zeebe-process-test). We can get this merged!

akeller (approved):

akeller:
@christinaausley, for context - we should probably get something together about coding best practices for snippets in docs. I'll follow up in our 1:1 this week and we can go from there.

PR #1077, opened 2022-07-21
docs rename selfmanaged

christinaausley:
According to https://github.com/orgs/camunda/projects/27/views/1, renaming files named `overview`, `index`, or other vague terms like `authentication` or `configuration` to clarify the file subject and ease analytics research.

### DIFF
--- a/docs/self-managed/platform-deployment/platform-8-deployment.md
+++ b/docs/self-managed/platform-deployment/platform-8-deployment.md
@@ -1,5 +1,5 @@
 ---
-id: index
+id: platform-8-deployment
 title: "Camunda Platform 8 deployment"
 sidebar_label: "Overview"
 ---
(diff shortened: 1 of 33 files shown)


docs/self-managed/platform-deployment/platform-8-deployment.md
```diff
@@ -0,0 +1,64 @@
+---
+id: platform-8-deployment
+title: "Camunda Platform 8 deployment"
+sidebar_label: "Overview"
```
akeller, line 4: Do you think the sidebar label still makes sense as "Overview"? My gut says yes since you'll have the proper context around the rest of the sidebar items.

akeller (approved):
I left one non-blocking comment. Thanks again @christinaausley for working though this!

docs/self-managed/platform-deployment/platform-8-deployment.md
christinaausley: I actually think clarifying this would be better as it's more of an "action" doc on deployment -- how do you feel about naming the sidebar label "Deployment"?

akeller: Oh I like that!

akeller (approved):

PR #1088, opened 2022-07-26
decision report

christinaausley:
Related to https://github.com/camunda/camunda-platform-docs/issues/692


akeller (approved):
Thank you @christinaausley! This was a little bit of a nit pick, but I think consistency and standardization with the UI is important for users.

PR #1090, opened 2022-07-27
chore(tutorials): rename repo from Camunda Cloud to Camunda Platform

HanselIdes:
Reference: new [repo name](https://github.com/camunda/camunda-platform-tutorials)


akeller (approved):
Thanks @HanselIdes! Looks like this got missed during the renaming. The GitHub URLs redirect (although I experienced some weirdness with the link to the worker-java 🤷‍♀️  but now I can't reproduce it after subsequent clicks) but the terminal instructions were obviously not going to work with the project directory renamed.

PR #1091, opened 2022-07-27
docs(contributing): Asking for us to format your changes is an option

pepopowitz:
Updates the CONTRIBUTING.md guide to make it clear that, as discussed in #1089, asking for someone else to format changes is an option.


akeller (approved):
Thanks @pepopowitz this makes sense to me! If I have time to unblock folks, I may extend it the other way as well but we don't need to be that detailed 😄

PR #1093, opened 2022-07-28
fix microservice orchestration counting error

Camundaan:
Fixed the numbering bug under "Design your process with BPMN"

### DIFF
--- a/docs/guides/getting-started-orchestrate-microservices.md
+++ b/docs/guides/getting-started-orchestrate-microservices.md
@@ -35,15 +35,15 @@ Start by designing your automated process using BPMN. This guide introduces you 
 
 ![Task with dropdown showing config, including service task](./img/microservice-orchestration-config-service-task.png)
 
-1. Add a descriptive name using the properties panel. For this guide, we'll use **Microservice Example**. Since you previously opened the **General** section of the properties panel, it is likely still open when working with your service task configuration.
-2. In the properties panel, expand the **Task definition** section and use the **Type** field to enter a string used in connecting this service task to the corresponding microservice code. For this guide, we'll use **orchestrate-something** as the type. You will use this while [creating a worker for the service task](#create-a-worker-for-the-service-task). If you do not have an option to add the **Type**, use the wrench icon and select **Service Task**.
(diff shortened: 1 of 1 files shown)


christinaausley (approved):

christinaausley:
> @christinaausley can you take a look at this and the other similar PR #1094? Was the counting error @Camundaan spotted because we moved other content around? I want to make sure this isn't a symptom that we need to update other things.

In previous versions, the numbering is correct in the docs. It is only in the latest version some of the numbering is off. Do we think this is a migration issue? Formatting? Looping in @pepopowitz in the case he'd also like to take a look.

The only other place I noticed an issue is in the [DMN guide](https://docs.camunda.io/docs/guides/create-decision-tables-using-dmn/). I'll launch a PR to correct.

pepopowitz:
Sharing some info re: the formatting of these numbered lists -- 

Prettier, our formatter, is smart enough to auto-increment numbered lists for us. So if we write `1. ` for three straight items, it will turn it into `1.`, `2.`, and `3.`. 

In the docs in these PRs, the numbered lists are interrupted by images or bullet points. For example, in this one, we have a list that goes something like: 

```
1. ...
2. ...
3. ...
4. ...

![....an image](...an image)

5. ...
6. ...
7. ...
```

Prettier interprets this as two separate lists, because of the interruption in the middle. And it just assumes that with the second part of the list, we're starting it with the correct number, and it will auto-increment for us after that. If we wanted to, we could write this list as:

```
1. ...
2. ...
3. ...
4. ...

![...an image](...an image)

1. ...
2. ...
3. ...
```

This is what happened in at least one of the "incorrect" numbered lists. Prettier assumes we know what we're talking about when we start that second list -- the interruption in the middle disconnects the two lists. 

All this to say...prettier let us write the incorrect numbers, because it can't tell whether it's all one list.

So my guess is that we "broke" the number order unintentionally, because there are multiple sub-numbered-lists that we need to synchronize manually...and prettier couldn't tell us it was wrong, because it couldn't tell.

akeller:
Hey @Camundaan we fixed this in #1099 and merged it into `main`. @christinaausley can follow up with you about how to break up changes and work best with us and GitHub.

PR #1094, opened 2022-07-28
fix humantask orchestration counting error

Camundaan:


christinaausley (approved):
Thanks for making these fixes @Camundaan -- currently investigating the cause.

akeller:
Hey @Camundaan we fixed this in #1099 and merged it into `main`. @christinaausley can follow up with you about how to break up changes and work best with us and GitHub.

PR #1095, opened 2022-07-28
Revert "Add timer cycle with start time"

korthout:
Reverts camunda/camunda-platform-docs#1084

That PR was not supposed to target `main`, but `release_8_1`.


remcowesterhoud (approved):
Thanks!

PR #1084, opened 2022-07-26
Add timer cycle with start time

remcowesterhoud:
Adds documentation about the timer cycle with start time. E.g. `R/2022-01-01T10:00:00+02:00[Europe/Berlin]/P1D`

closes #986

PR #1071, opened 2022-07-20
rename components overview pages

christinaausley:
According to https://github.com/orgs/camunda/projects/27/views/1, renaming vague files in Components section of documentation.

PR #1096, opened 2022-07-28
Add timer cycle with start time (#1084)

remcowesterhoud:
Add timer cycle with start time docs.

Let's try this again @korthout 😄

### DIFF
--- a/docs/components/modeler/bpmn/timer-events/timer-events.md
+++ b/docs/components/modeler/bpmn/timer-events/timer-events.md
@@ -78,9 +78,12 @@ If the duration is zero or negative, the timer fires immediately.
 ### Time cycle
 
 A cycle defined as ISO 8601 repeating intervals format; it contains the duration and the number of repetitions. If the repetitions are not defined, the timer repeats infinitely until it is canceled.
+It's possible to define a start time. By doing this the timer triggers for the first time on the given start time. Afterwards it will follow the interval as usual.
 
 - `R5/PT10S`: Every 10 seconds, up to five times
 - `R/P1D`: Every day, infinitely
+- `R/2022-01-01T10:00:00+02:00[Europe/Berlin]/P1D`: Every day infinitely, starting from the 1st of January 2022 at 10 am
+- `R3/2022-04-027T17:20:00+02:00Z/P1D`: Every day up to three times, starting from the 27th of April 2022 at 5:20 pm
 
 ## Additional resources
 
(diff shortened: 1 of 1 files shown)


korthout:
@remcowesterhoud Please also await a review from @christinaausley before merging

korthout (approved):

remcowesterhoud:
Will do, I have no clue why the build is failing as I didn't touch any links. So that is completely unrelated to this PR.

korthout:
Hmm 🤔 I also don't know. Maybe @pepopowitz can help us here.

pepopowitz:
I think I can explain it! 

We've been renaming vague doc filenames (#1069, #1071, #1077, ...). It looks like this branch & `release_8_1` are pretty far off `main`. You might think that the link-check shouldn't be affected by that, since the links would be based on the branch being checked....but it turns out that [the link-check pulls the sitemap from _production_](https://github.com/camunda/camunda-platform-docs/blob/main/.github/workflows/build-docs.yaml#L38-L39), not from this branch. So it's checking production URLs against your branch, which fails for all the renames that aren't in your branch yet. 

I think that if you were to rebase `release_8_1` (or merge `main`, if that's easier), then rebase this branch on `release_8_1` (or merge, whatever works best for you), you should get to green. 

Long term, I wonder if the link-check would be better off pulling the sitemap from the current branch, rather than production....I'll dig into that async.

docs/components/modeler/bpmn/timer-events/timer-events.md
```diff
@@ -78,9 +78,12 @@ If the duration is zero or negative, the timer fires immediately.
 ### Time cycle
 
 A cycle defined as ISO 8601 repeating intervals format; it contains the duration and the number of repetitions. If the repetitions are not defined, the timer repeats infinitely until it is canceled.
+It's possible to define a start time. By doing this, the timer triggers for the first time on the given start time. Afterwards, it will follow the interval as usual.
 
 - `R5/PT10S`: Every 10 seconds, up to five times
 - `R/P1D`: Every day, infinitely
+- `R/2022-01-01T10:00:00+02:00[Europe/Berlin]/P1D`: Every day infinitely, starting from January 1, 2022 at 10 a.m.
```
christinaausley, line 85: Are both of these times mentioned CEST? We might want to note that.

christinaausley (approved):
Added a few commas and updated dates/times according to [style guidelines](https://confluence.camunda.com/display/HAN/Camunda+Writing+Style+Guide). One comment for clarification on the timezone.

docs/components/modeler/bpmn/timer-events/timer-events.md
remcowesterhoud: Good shout, I've added the timezone to the description. Thank you!

PR #1097, opened 2022-07-28
chore(product-links): update prop panel links

akeller:
@pepopowitz your review is mostly for awareness and to merge when you'd like so it doesn't interrupt your link check testing. Or maybe you want to interrupt it...? 🤷‍♀️


pepopowitz (approved):
Yes, let's please interrupt, anything to stop these out of control builds 😅

PR #1098, opened 2022-07-28
debug(ci): enable debug flag for link checking

pepopowitz:
We've been noticing that link-checking is suddenly sporadically taking a very very long time ([example 1](https://github.com/camunda/camunda-platform-docs/actions/runs/2746917140), [example 2](https://github.com/camunda/camunda-platform-docs/actions/runs/2751525426), [example 3](https://github.com/camunda/camunda-platform-docs/actions/runs/2752589151)), but we don't really know why. 

This PR adds [the debug flag](https://github.com/filiph/linkcheck/blob/master/lib/linkcheck.dart#L204) to our link-checking steps, with the hopes of gaining insight into why the link-checking is taking so long. 

I anticipate this change to be temporary, and we'll revert after gaining (or not gaining) insight.


christinaausley (approved):
Editorially this is correct 😬

menski (approved):
:shipit:

PR #1106, opened 2022-08-01
Remove old include directive

npepinpe:
This PR removes left over from the migration of the old Zeebe docs, which included some old unsupported include directive.

### DIFF
--- a/docs/self-managed/zeebe-deployment/configuration/configuration.md
+++ b/docs/self-managed/zeebe-deployment/configuration/configuration.md
@@ -156,14 +156,6 @@ Zeebe uses Log4j2 framework for logging. In the distribution and the docker imag
 
 To enable Google Stackdriver compatible JSON logging, set the environment variable `ZEEBE_LOG_APPENDER=Stackdriver` before starting Zeebe.
 
-### Default logging configuration
-
-- `config/log4j2.xml` (applied by default)
-
-```
-{{#include ../../../dist/src/main/config/log4j2.xml}}
-```
-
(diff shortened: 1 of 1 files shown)


npepinpe:
I also noticed there's a separate `logging.md` in the same folder which describes in more detail the logging configuration, so this block was entirely useless imo.

pepopowitz:
Hi friends!! I want to make sure you're aware of #1115 -- a PR that I plan to merge in the next week. 

It makes some fundamental changes to where our docs are sourced: 
1. 8.0 docs will now be sourced from `/versioned_docs/version-8.0` instead of `/docs`
2. Anything in `/docs` will be presented as the next/unreleased version

### Action required

It looks to me like this PR is intended to update the current 8.0 docs. Assuming that's true, one of these things need to happen: 

1. We need to merge this before I merge #1115. This is my favorite option if the PR is ready, because it means no extra work for any of us 😄 
2. After I merge #1115, we'll need to update this PR to make changes in the `/versioned_docs/version-8.0` folder instead of `/docs`. 

If this PR is intended for future 8.1 docs, we don't need to do anything! After #1115 merges, this PR _will_ be pointing at the unreleased 8.1 docs, and you can merge it at your convenience. 

---

I'll hold off on merging #1115 until at least August 8th, but probably not much after. If someone is able to get this one merged before then, great! If not, I'll assume that @npepinpe, as the author of the PR, will update this PR as needed. 

Thanks! If you have questions or concerns, let me know.

akeller (approved):
Thank you!!

PR #1114, opened 2022-08-02
chore: rewrite Footer customization based on docusaurus 2.0.0-rc.1

pepopowitz:
Partially addresses #1024

In the process of updating docusaurus in #1024, our customized Footer component became out-of-date. This PR rewrites that component based on the correct source, with no functional changes.

It includes a few other small tangential details: 

1. A README.md in the `src/theme` folder describing what lives there
2. Notes on the ColorModeToggle component about when & why it was swizzled
3. Remove an unused file (which was replaced by a previous upgrade & the `docusaurus-gtm-plugin`)
4. Rewrite the ExternalLink icon based on the correct docusaurus source

## Proof that the footer hasn't changed

image

## Follow-up

- After this is merged & deployed, test that MixPanel still sends messages in production! ([According to this Slack thread](https://camunda.slack.com/archives/C026U8GBNSW/p1652278810298179)) that's the only place I can test it. 😰 )


src/theme/Footer/index.js
```diff
@@ -1,140 +1,24 @@
...
-import Link from "@docusaurus/Link";
-import { useThemeConfig } from "@docusaurus/theme-common";
-import useBaseUrl from "@docusaurus/useBaseUrl";
-import clsx from "clsx";
-import mixpanel from "mixpanel-browser";
-import React from "react";
-import IconExternalLink from "../IconExternalLink";
-import styles from "./styles.module.css";
 
-function FooterLink({ to, href, label, prependBaseUrlToHref, ...props }) {
```
pepopowitz, line 20: One nice thing about newer versions of docusaurus -- they made it easier to "wrap" theme components, which allowed us to delete a lot of this file.

src/theme/Footer/styles.module.css
```diff
@@ -1,41 +0,0 @@
...
- */
-
-.footerLogoLink {
-  opacity: 0.5;
-  transition: opacity 0.15s ease-in-out;
-}
-
-.footerLogoLink:hover {
-  opacity: 1;
-}
```
pepopowitz, line 15: These first two styles are from the classic theme, we can delete them because the theme takes care of them for us.

src/theme/Footer/styles.module.css
```diff
@@ -1,41 +0,0 @@
...
-.osano-cm-link {
-  color: #002f66 !important;
-}
-.osano-cm-close {
-  background: transparent !important;
-}
-.osano-cm-view--type_consent .osano-cm-list-item:nth-child(5) {
-  display: none;
-}
```
pepopowitz, line 41: These Osano styles are for the consent popup, but the ones that are actually used are [in src/custom.css](https://github.com/camunda/camunda-platform-docs/blob/d689488d5b7aef3252c0792b39505ece4e220057/src/css/custom.css#L81).

christinaausley (approved):
No changes on my end here!

src/theme/Footer/styles.module.css
pepopowitz: Visual proof that there is no _regression_ due to removing these styles: 

image

I say _regression_ because @akeller raised concerns that the styles in the code do not match the styles given to us by marketing, which I will follow up on (and will not block the merge of this PR).

akeller: This was my 🗑️, but I followed up with you on Slack about some discrepancies and to check in with the web marketing team. This doesn't block this PR.

akeller (approved):
I see no blocking issues and the CSS performs as expected. Thanks for cleaning this up! I love PRs that delete more than they add!!! 

Regarding styles - Button styling is different on docs vs. camunda.com to match the respective button experience. Please don't change this because consistency should match the experience of the site.

The link is also a different color than I would expect.

image

akeller:
That is by far the worst and laziest screenshot I've taken in my life, and I'd just like to apologize to the internet by letting everyone know I'm signing off for the day.

pepopowitz:
Merging, and I'll deal with the cookie consent styling difference(s) in a follow-up.

Issue #1116, opened 2022-08-02
Configure multi-instance versioning


pepopowitz:
Epic: https://github.com/camunda/product-hub/issues/231
Prerequisite: #1115 

Optimize versions are different than the versions of the other Camunda components. [Docusaurus recommends using multiple instances of the docs plugin](https://docusaurus.io/docs/next/docs-multi-instance#mobile-sdks-documentation) to enable users to see different versions for different docs.

### Preparation (ahead of time)
- #1167 (done in #1170)
- #1193 

### Move docs over
1. ✅ Create a "next" Optimize version
     1. ✅ #1166
          This moves a single doc into the new instance, including all configuration necessary to iteratively move more docs over afterward:
          - Add a second Optimize instance of the docs plugin
          - Add a second Optimize set of sidebars for the new instance.
          - Configure sidebars for consistent cross-plugin navigation
          - Move one doc (`what-is-optimize`) into new instance
          - Update links in documentation for the moved doc (and docs that reference it)
     2. ✅ Iteratively move remaining "next" Optimize docs over: #1204 and #1214

2. ✅ Create a "3.8.0" Optimize version
    1.  ✅ #1234
        This will be merged into `main` all at once, because this is the current version of documentation and we will confuse people if we release it in pieces.
        - Configure the 3.8.0 version of Optimize documentation
        - Customize NavbarItem so that the version dropdown shows the correct versions, based on which docs are being viewed (in #1235)
        - Extract everything from docs/8.0/components/optimize to optimize/3.8.0 (in #1236)
        - Add rewrite rules for optimize/3.8.0 docs (in #1238)
        - Revert #1193 (because we want these things in the sitemap now) (in #1239)
3. ✅ #1286
    For consistency with the new folders planned to move, we'll want to move all the existing Optimize docs into a `/components/` folder.
    1. Move `/optimize/*` into `/optimize/components/*` for vNext
    2. Move `/optimize_versioned_docs/version-3.8.0/*` into `/optimize_versioned_docs/version-3.8.0/*` for v3.8.0
4. ✅  Move Optimize docs from apis-clients & self-managed sections into Optimize instance
    1. ✅  vNext: #1293
       1. Move `/docs/apis-clients/optimize-api/` docs into vNext 
       2. Move `/docs/self-managed/optimize-deployment/` docs into vNext
   2. ✅  v3.8.0: #1300 
      1. Move `/docs/apis-clients/optimize-api/` docs into 3.8.0 
[...]

Issue #1115, opened 2022-08-02
Default to latest tagged version; provide unstable "next" version


pepopowitz:
Epic: https://github.com/camunda/product-hub/issues/231

Our current approach to presenting versions is to default to the “latest” version, and have all other versions be tagged with a number. Docusaurus’s approach, and [the approach they best support](https://docusaurus.io/docs/versioning#configuring-versioning-behavior), is to default to the last tagged version, have past versions available by their tagged number, and an unstable “next” version available at a /next URL.

We should switch our approach to that of Docusaurus, with an unstable “next” version. Short-term, we will gain the ability to iterate on our restructuring of content. Rather than being forced to do it in a long-lived branch, we can do it in main, as part of the “next” version. Long-term, engineers will be able to write documentation for upcoming releases while developing the features, rather than having to wait until the release. 

## Work to be done

* Cut a new version of the docs through Docusaurus, to reflect what is currently “latest”. 
* Update docusaurus.config.js as in #910. 
* I don’t think there is any work needed re: redirects, but confirm this. The URLs for all old versions don’t change; neither do the URLs for the current version; there are new URLs for the “next” version, but they’re new, and don’t require redirects.

## POC

See #910 for a POC of this work.

## Background
This is Recommendation #2 in [the Optimize Multi-Instance Documentation Proposal](https://docs.google.com/document/d/1GpB7OcSzo5htcBHDAdEVAOgqMOAWrAaJELycz5Gqcko/edit#).

pepopowitz:
Woops, lost track! Closed by #1118.

PR #1118, opened 2022-08-03
feat: cut 8.0 version (and default to it); expose "next" version

pepopowitz:
Addresses #1115 

## What

This PR: 
1. Cuts a new 8.0 version
2. Configures docusaurus to use 8.0 as the default version
3. Exposes a "next" version

The vast majority of changes in this PR are to cut the 8.0 version -- this process clones the entire set of docs into a new versioned folder. The only exception is visible in https://github.com/camunda/camunda-platform-docs/commit/6772e2b6c7e7c32cf9ba560333820675ac311f8d -- it configures docusaurus to use a `Next` version, and default to `8.0` as the current version. 

## Why

This gives us the ability to iterate on docs for upcoming releases and content restructuring efforts. Rather than being forced to do them in a long-lived branch, we can do them in main, as part of the “next” version. 

More specifically, it sets us up to iterate toward giving the Optimize documentation its own set of versions, as described in #1116. 

## How does this impact users?

1. When viewing the docs for the current version, they'll see an `8.0` selected in the version selector, instead of `latest`. This will remain the default version when they come to docs.camunda.io. 
2. They'll be able to view a set of docs for the `Next` version, which they hadn't been able to do before.

#### Default version

By default, browsing to `docs.camunda.io` will serve the user the `8.0` docs. The URLs will look the same as they currently do for the `latest` version in production -- absent of a version (e.g. `/docs/guides/`). Because of this, **we do not need to add any redirect rules.**

The docs they're viewing for 8.0 will be sourced from the `/versioned_docs/version-8.0` source folder. 

image

#### `Next` version
If the user selects `Next` from the version dropdown, they'll browse the next set of docs. Their URL will include the `/next/` version. The docs they're viewing will be sourced from the `/docs` source folder.

image

## How does this impact content authors?

1. Updates to docs for our **current** release (8.0) will take place in the `/versioned_docs/version-8.0` source folder, instead of `/docs`
2. Updates to our **next** release (8.1?) will take place in the `/docs` source folder. These updates can be merged whenever, rather than waiting for the 8.1 release to take place. 
3. **Open PRs may need to be recreated.** I will follow up on each currently open PR with guidance on this. I'm willing to assist in whatever work needs to happen, too. 

## When will this be merged?

Because of the impact of this PR, especially on existing PRs, I don't intend to merge this immediately. I will be following up with all currently open PRs to develop a plan for minimizing merge headaches. If all goes well, early next week (August 8th) I intend to: 

1. re-generate version 8.0 in this PR so that it includes all recently merged changes, then
2. merge this PR, then
3. work with authors of open PRs that need to be recreated. 

## Background 

This work is in preparation of giving the Optimize documentation its own set of versions. It is the implementation of Recommendation #2 described in [the Optimize Multi-Instance Documentation Proposal](https://docs.google.com/document/d/1GpB7OcSzo5htcBHDAdEVAOgqMOAWrAaJELycz5Gqcko/edit#).


korthout (dismissed):
I think this is a wonderful idea! 💯 

👍 I like that we can just work in `/docs/` to add changes for the upcoming version
👍 I like that we can just merge to `main` when we have changes for the upcoming version
👍 I like that we don't have to keep a long-lived 'upcoming-release' branch
👍 I like that the upcoming version is labeled as `Next` (it has my preference over a concrete version like `8.1 🚧`, because its more clear to users that this is not yet available)
👍 I like that it only takes `12` removed lines of config to make this work

Keep up the great work @pepopowitz ❤️

pepopowitz:
I don't like to do this, but I'm bypassing branch protection to get this merged without an approval -- it was approved before I regenerated the 8.0 version, and I don't want to delay this merge another day (and risk a new batch of PRs in conflict).

PR #1128, opened 2022-08-08
[DO NOT MERGE] experiment: test auto-formatting in CI for forked PRs

pepopowitz:
This is test #2 for #1130 -- forked PRs.

With #1130 configured, if I push up a PR for a branch _from a fork_, does the GITHUB_TOKEN running the auto-format workflow have write access to the branch?

## Findings

**No**, it doesn't.

We can see it [_attempting_ to commit the format fix](https://github.com/camunda/camunda-platform-docs/runs/7734035953?check_suite_focus=true)...but it fails, with an ......

ugh it's not failing for the reason I think it is


pepopowitz:
Closing this because I am confused by the results and can't tell noise from signal

PR #1130, opened 2022-08-08
[DO NOT MERGE] experiment: configure auto-formatting for all PRs question mark?

pepopowitz:
This PR is an experiment to see if the `GITHUB_TOKEN` associated with this repo has the permissions it needs in order to write commits to forks. 

More specifically, I wanted to find out if the options I recommended in #1104 are even feasible. 

It configures a job in the `build-docs` workflow named `auto-format`. This job runs `npm run format` against the branch, then commits any updates. I followed [this article](https://mskelton.medium.com/auto-formatting-code-using-prettier-and-github-actions-ed458f58b7df) as a guide. 

## Findings

No luck.

The `GITHUB_TOKEN` does **not** have write access when running in forks. (Actually, it has a confusing amount of access when running in forks --  [See this PR/comment](https://github.com/camunda/camunda-platform-docs/pull/1140#issuecomment-1213307313) for more details on what we do/do not have permissions to do against a forked repo.).

And we don't have access to the setting in this repo to grant it write access, which suggests it's disabled at the org level.

PR #1140, opened 2022-08-10
[DO NOT MERGE] experiment: test auto-formatting in CI for forked PRs (version 2)

pepopowitz:
This is test #2 (actually #3 because [the first #2](https://github.com/camunda/camunda-platform-docs/pull/1128) was confusing me) for #1130 -- forked PRs.

With #1130 configured, if I push up a PR for a branch _from a fork_, does the GITHUB_TOKEN running the auto-format workflow have write access to the branch?


akeller:
Let me know if you want to talk through this tomorrow because it looks annoying. I can make time 😄

pepopowitz:
Yes, let's talk because IT'S SO CONFUSING?!!!!!?!!!!!!?!?!?!?!?!?!?!?!?!!


TL;DR 

1. [the workflow says it doesn't have access to push a commit to the fork](https://github.com/camunda/camunda-platform-docs/runs/7776376181?check_suite_focus=true), which makes sense
2. **!!!!!!BUT A COMMIT FROM THE BOT IS SOMEHOW APPEARING ON THE FORK'S BRANCH!!!!!!** 👇 
image

pepopowitz:
I'm going to close this, as I'm setting this work aside for now. 

## Recap

The concern we have in #876 is that the actions we run through this repo won't have permission to make changes to a forked branch. This experiment, and specifically this PR, attempted to answer that question. 

The [documentation from GitHub](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#enabling-workflows-for-forks-of-private-repositories) suggests that we should **not** have write access to a forked branch. We do not have the ability in this repo to check the "Send write tokens to workflows from pull requests" box, which would attach write permissions to the `GITHUB_TOKEN` used in the workflow. This is likely because that box is disabled at the org level, and "if a policy is disabled for an organization, it cannot be enabled for a repository."

But this PR shows mixed results. As [this workflow indicates](https://github.com/camunda/camunda-platform-docs/runs/7790657459?check_suite_focus=true), we do get a 403 error when attempting to push the committed changes to the forked repo -- this is expected. However, a commit magically appears in the forked repo after receiving that error, consistently: 

image

So it _does_ seem to have access to push...but I don't know if this is because of a bug, or because I don't have things configured correctly in the workflow definition. 

I'm setting this work aside for now. When I come back to it, I think I should assume that this approach -- pushing a commit to the forked repository -- will not work. The docs tell me it shouldn't....it _does_ push a commit, but that's unexpected, and I think I should assume that eventually a bug will be fixed.

PR #1141, opened 2022-08-11
feat(read only): add optimize config for entity editor authorization

RomanJRW:
relates to OPT-6335


PHWaechtler (approved):
👍 👍

PHWaechtler:
@RomanJRW This needs merged into `/docs` and removed from `3.8.0`

PR #1145, opened 2022-08-11
docs(identity): reference new base path variable in configuration options

Ben-Sheppard:
Resolves https://github.com/camunda-cloud/identity/issues/938

This PR adds the new configuration variable in Identity that allows a user to configure Identity to run using a subpath, for example: `http://adomain.com/identity`.

I've elected to use a `:::note` block to suggest to the user a caveat which is in place when using a subpath, this caveat is that there is a requirement to either be on localhost or to have a connection which is HTTPS.

I had initially anticipated a user guide addition but at this time I don't feel it would add any great value to the Identity docs so I have left out that change.

This change has been available since 8.1.0-alpha3 so it does not need to be backported to the 8.0 docs :)

### DIFF
--- a/docs/self-managed/identity/deployment/configuration-variables.md
+++ b/docs/self-managed/identity/deployment/configuration-variables.md
@@ -18,7 +18,7 @@ methods.
 | `IDENTITY_BASE_PATH`                 | Used to configure Identity to run on a subpath       |                                                     |
 
 :::note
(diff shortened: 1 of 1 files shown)


christinaausley (approved):
Added a single period to complete the sentence in the note. LGTM from a technical perspective!

btepe (approved):
Looks good, thank you Ben! 🎉

akeller:
> This change has been available since 8.1.0-alpha3 so it does not need to be backported to the 8.0 docs :)

Thank you for answering questions before we ask them! 🎉

Issue #444, opened 2021-11-17
Adding components/architecture overview


berndruecker:
I think, we should have a high-level architecture of product components graphically here: https://github.com/camunda-cloud/camunda-cloud-documentation/blob/master/docs/components/overview.md - probably related to https://docs.camunda.io/docs/components/zeebe/technical-concepts/architecture/ but also showing Operate, Tasklist and how it all connects/communicates.

Anecdote where I could make use of such a picture: https://forum.camunda.io/t/camunda-cloud-api/2602

berndruecker:
@akeller If you give me some hint on how we envision building such pictures (tooling, style guide), I am happy to do a first stab at it.

berndruecker:
Definitely interesting in the topic and I don't like the above graphic for a developer audience - I find it quite hard to process. I would envision it to be closer to 
![grafik](https://user-images.githubusercontent.com/1217596/165241504-fa3d2017-554e-4e5b-b2df-514d47171c44.png) or like https://docs.camunda.io/docs/components/best-practices/operations/reporting-about-processes/#history-architecture.

Let me know if I can be of any help. And yes - I think identity should be included in there - it is an important piece to understand (if you look at it from a deployment perspective for example).

berndruecker:
Generally yes - but most probably not going to happen this or next week

berndruecker:
OK - quicker than expected. Also needed (or wanted to have) it for a customer workshop - WDYT about this slide?
![c8-compoments-architecture](https://user-images.githubusercontent.com/1217596/165463352-efdd6a6f-9bb9-4d31-be77-3db68c72f0b8.png)

akeller:
I really like this @berndruecker!

berndruecker:
[ComponentsAndArchitecture.pptx](https://github.com/camunda-cloud/camunda-cloud-documentation/files/8575644/ComponentsAndArchitecture.pptx)

akeller:
![components-architecture](https://user-images.githubusercontent.com/1051363/165750723-0c071330-5cf4-4be5-ae06-04b18a07e0a4.png)

berndruecker:
@akeller I send out the link to the issue and the image draft quite often - any chance we get that into the docs properly?

akeller:
@berndruecker I can look into adding this sometime this week. Do you have an updated version with Connectors?

akeller:
@falko FYI

Here is the small update for the typo.
[...]

PR #1148, opened 2022-08-12
Add SaaS architecture diagram

akeller:
Partially addresses #444.


akeller:
@christinaausley would love your thoughts here beyond just a review! I thought the image was a bit strange above the text describing the components and then below the components I felt like it needed a little intro sentence.

This does violate some of my emerging thoughts about how to handle images accessibly, but I'm still unsure how to make architecture-like diagrams like this accessible.

christinaausley:
@akeller Are you looking for specific feedback on the diagram?

Based on what you've changed, I do agree we need an introductory sentence (which I've adjusted slightly in my latest commit) and this should definitely be placed as you have it after the components are textually introduced.

As far as accessibility goes, contrasting colors and clear alt text assist. However, here's where things get a little more complex with diagrams...

We do indeed have the option to create accessible PDFs where you write out a brief description of a chart/diagram and provide the accessible document link near the image/clickable on the image. However, if the content on the screen describes the exact same information presented in a diagram and someone who could not see the chart or diagram would not lose any information by not being able to see it, you do not need to provide an accessible document because you have already provided an accessible way of engaging with the information. This all, of course, depends on the level of accessibility we need/want to meet.

akeller:
Thanks for the adjustment to the introductory sentence. That's all I was looking for review-wise right now.

Interesting about the accessibility angle, maybe we can test out the accessible PDF with this image in a future PR 🤔

akeller:
Can I nudge you for an approval @christinaausley? 😄

christinaausley (approved):

PR #1155, opened 2022-08-15
update introduction to camunda ids

christinaausley:
Updating all "Introduction to Camunda" IDs to `introduction-to-camunda`.

Resolves https://github.com/camunda/camunda-platform-docs/issues/1111#

### DIFF
--- a/static/.htaccess
+++ b/static/.htaccess
@@ -10,23 +10,6 @@ Options -Indexes -MultiViews
 
 RewriteRule ^docs/components/optimize/userguide/combined-reports/(.*)$ /docs/components/optimize/userguide/combined-process-reports/$1 [R=301,L]
 
-RewriteRule ^docs/guides/introduction-to-camunda-platform-8/(.*)$ /docs/guides/introduction-to-camunda/$1 [R=301,L]
-RewriteRule ^docs/guides/introduction-to-camunda-platform-8$ /docs/guides/introduction-to-camunda/ [R=301,L]
-RewriteRule ^docs/8.0/guides/introduction-to-camunda-platform-8/(.*)$ /docs/8.0/guides/introduction-to-camunda/$1 [R=301,L]
-RewriteRule ^docs/8.0/guides/introduction-to-camunda-platform-8$ /docs/8.0/guides/introduction-to-camunda/ [R=301,L]
-RewriteRule ^docs/1.3/guides/introduction-to-camunda-cloud/(.*)$ /docs/1.3/guides/introduction-to-camunda/$1 [R=301,L]
-RewriteRule ^docs/1.3/guides/introduction-to-camunda-cloud$ /docs/1.3/guides/introduction-to-camunda/ [R=301,L]
-RewriteRule ^docs/1.2/guides/introduction-to-camunda-cloud/(.*)$ /docs/1.2/guides/introduction-to-camunda/$1 [R=301,L]
-RewriteRule ^docs/1.2/guides/introduction-to-camunda-cloud$ /docs/1.2/guides/introduction-to-camunda/ [R=301,L]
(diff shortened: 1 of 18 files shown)


akeller (dismissed):
Wow this looks like an annoying PR.

I'm not skilled enough write regex, so I hate to even suggest this... but it feels like there is a better way to do these redirects. 

@christinaausley did you test to confirm this corrects the incorrect behavior?

static/.htaccess
```diff
@@ -10,6 +10,23 @@ Options -Indexes -MultiViews
...
+RewriteRule ^docs/1.2/guides/introduction-to-camunda-cloud/(.*)$ /docs/1.2/guides/introduction-to-camunda/$1 [R=301,L]
+RewriteRule ^docs/1.2/guides/introduction-to-camunda-cloud$ /docs/1.2/guides/introduction-to-camunda/ [R=301,L]
+RewriteRule ^docs/1.1/guides/introduction-to-camunda-cloud/(.*)$ /docs/1.1/guides/introduction-to-camunda/$1 [R=301,L]
+RewriteRule ^docs/1.1/guides/introduction-to-camunda-cloud$ /docs/1.1/guides/introduction-to-camunda/ [R=301,L]
+RewriteRule ^docs/1.0/guides/introduction-to-camunda-cloud/(.*)$ /docs/1.0/guides/introduction-to-camunda/$1 [R=301,L]
+RewriteRule ^docs/1.0/guides/introduction-to-camunda-cloud$ /docs/1.0/guides/introduction-to-camunda/ [R=301,L]
+RewriteRule ^docs/0.26/guides/introduction-to-camunda-cloud/(.*)$ /docs/0.26/guides/introduction-to-camunda/$1 [R=301,L]
+RewriteRule ^docs/0.26/guides/introduction-to-camunda-cloud$ /docs/0.26/guides/introduction-to-camunda/ [R=301,L]
+RewriteRule ^docs/0.25/guides/introduction-to-camunda-cloud/(.*)$ /docs/0.25/guides/introduction-to-camunda/$1 [R=301,L]
+RewriteRule ^docs/0.25/guides/introduction-to-camunda-cloud$ /docs/0.25/guides/introduction-to-camunda/ [R=301,L]
```
pepopowitz, line 28: Are these redirects needed? All of the docs changed in this PR had their _`id`_ changed, but none had their `slug` change. The `id` is the one used by docusaurus to know where to route links, but the `slug` is the one that users see in their browser.

pepopowitz (approved):
Nice, thank you! My feedback is non-blocking.

static/.htaccess
christinaausley: You may be right here @pepopowitz -- I'll remove and see if it still build fine. 👍

christinaausley:
> Wow this looks like an annoying PR.
> 
> I'm not skilled enough write regex, so I hate to even suggest this... but it feels like there is a better way to do these redirects.
> 
> @christinaausley did you test to confirm this corrects the incorrect behavior?

Locally, this does correct the behavior on build. I'll double check this in stage 👍

static/.htaccess
christinaausley: Built clean. Thanks for the tip @pepopowitz ! @akeller In reference to [this comment](https://github.com/camunda/camunda-platform-docs/pull/1105#discussion_r940700209), Steve and I talked today and I'm going to go ahead and launch a PR to test and possibly drop any excess redirects with slugs like this. I might also go through and label the sections of redirects according to why they're there (C8 migration, "overview" naming changes, etc) as a sort of "htaccess clean up".

akeller: I _love_ this! Please add some comments in there!

akeller (approved):

PR #1159, opened 2022-08-15
clarify xml editing in desktop modeler

christinaausley:
In reference to https://github.com/camunda/camunda-platform-docs/issues/990, adding admonition to clarify usage of XML editing tab in Desktop Modeler rather than Web Modeler.


akeller (approved):
💯 this is a very good note!

PR #1160, opened 2022-08-16
fix(connectors): correctly use code inline formatting

MaxTru:


MaxTru:
Super minor fix which I just discovered reading docs. No priority.

akeller (approved):
Wow, good spot! I also double checked the `/docs/` for the `introduction.md` change it it looks like it was already there.

I'll get this merged but hold off on releasing so we can include some other changes.

PR #1166, opened 2022-08-16
feat: set up optimize docs instance and move one doc over

pepopowitz:
Part of #1116. 

- [x] Adds a second docs plugin for Optimize documentation
- [x] Moves what-is-optimize doc into Optimize docs plugin
- [x] Configures sidebars for cross-plugin navigation
- [x] Updates links in documentation for cross-plugin navigation (what-is-optimize doc only)
- [x] Updates mappings in expandVersionedUrl based on the fact that there are no optimize versions yet
- [x] Adds a script and some instructions for generating the `optimize_sidebars.js` file

## Scope

Only one doc is moved in this PR! But, it is _completely_ moved, and the PR includes all setup needed so that we can iteratively move docs in ensuing PRs. It also demonstrates what all needs to be done to move a doc, so we should be able to follow it as a guide. 

### Out of scope of this PR

There are several details that are not included in this PR which disrupt the experience of reading/navigating docs within the Optimize section:

* The "Components" top-nav item is not active within the Optimize section
    image
* The version selector shows main versions, instead of Optimize versions, in the Optimize section
    image
* The "this is unreleased documentation, see the latest version" warning does not appear in the Optimize section. (This one will likely take care of itself when I get to the stage of adding more Optimize versions)
    image
* The footer links to the next/previous documents are empty at the beginning/end of the Optimize section, and for the components surrounding Optimize the links skip over the Optimize section: 

These items are all in scope of #1116, the parent issue, but I've elected to exclude them from this PR. I think they're all important issues for us to resolve, but I don't think they are _required_ to move docs iteratively into the Optimize section. If you think I'm incorrect, and they are important to include in this PR, let me know!

## Proof that it includes all latest Optimize updates

Note the last updated timestamp [in the docs/components/optimize folder](https://github.com/camunda/camunda-platform-docs/tree/main/docs/components/optimize): 

image

These timestamps are all older than the timestamp for the first commit in which I moved the `what-is-optimize` doc: 

image
 

## Implementation notes

- The markdown links take advantage of the MDX plugin added in #1170, for a better experience linking across docs plugin instances. 

## Other notes

**Note to future us**: docusaurus has an item on their roadmap to [enable separate docs plugins to talk to each other](https://github.com/facebook/docusaurus/issues/6370), at which time we might be able to undo this cross-linking experience in favor of something built into docusaurus 🙌


docusaurus.config.js
```diff
@@ -32,6 +32,15 @@ module.exports = {
       },
     ],
     "./static/plugins/bpmn-js",
+    [
+      "@docusaurus/plugin-content-docs",
+      {
+        id: "optimize",
+        path: "optimize",
+        routeBasePath: "optimize",
+        // sidebarPath: require.resolve("./optimize_sidebars.js"),
```
pepopowitz, line 41: I'll remove this, it looks like it _assumes_ this same path anyway.

optimize/what-is-optimize.md
```diff
@@ -0,0 +1,33 @@
...
+New to Optimize? Visit our introductory guide to [Optimize](/docs/next/guides/improve-processes-with-optimize/) to get started.
+:::
+
+Camunda Platform 8 is built to handle three key aspects of process automation:
+
+- Design
+- Automate
+- Improve
+
+A user can design process flows through our [Modeler](/docs/next/components/modeler/about-modeler/). In a production scenario, you can deploy through Desktop Modeler, Web Modeler, or programmatically. A user can use [Tasklist](/docs/next/components/tasklist/introduction-to-tasklist/) to review and complete tasks, and [Operate](/docs/next/components/operate/operate-introduction) to view and analyze process instances.
```
pepopowitz, line 17: 👋 👀 

FYI I'm going to add an issue for the outcome of this very long note, but I also want to get feedback from people @christinaausley @akeller @korthout in case you think I'm wasting my time and being overly dramatic. 

This is ....kind of sucky. These links all had to change [from the originals](https://github.com/camunda/camunda-platform-docs/blob/bdf1dccec35bcad8aeb95d1f17d47c4f60b4c75c/docs/components/optimize/what-is-optimize.md?plain=1#L17). e.g. `[Operate](/components/operate/operate-introduction.md)` changed to `[Operate](/docs/next/components/operate/operate-introduction)`

## Why?

This is because the links cross from one docs plugin to the other (that's how we implement multi-versioning in docusaurus -- two docs plugins, one for each set of versions). 

Previously we were able to specify the doc ID and path via source....and docusaurus would magically find the doc you were linking to. We didn't need to know the URL, and we didn't have to specify a version because it used the same version you were browsing.

But now that they're cross-plugin, we have to specify the **full URL** of the target doc. This includes the version....which to some degree makes sense, because d-rex doesn't know that Optimize v3.8.0 is associated with C8 v8.0. 

## Why it's awful

Where it especially stinks though, is in the "next" and "current" docs. In the "next" docs, we'll want Optimize to link to C8 "next", and C8 to link to Optimize "next". When we cut versions 3.9.0/8.1 from that, we'll have to go through and replace every markdown link in those versions to remove the "/next" from the path. And then we'll have to go through the current 3.8.0/8.0 docs, and replace all _those_ markdown links to add "/3.8.0" or "8.0".

This feels like **a lot** of friction. I didn't catch this when building the POCs. It'll be a pain for people writing docs because they can't just copy/paste their changes from the next version into current (or vice versa). It'll be a pain for reviewers because we'll overlook it all the time. It'll be a pain when we cut new versions. 

There is also friction caused by sidebars needing to be expanded to include full URLs when crossing docs plugins. But I'm comfortable over-looking that friction because we don't modify the sidebars files very often. 

Introducing friction on every markdown link, though, seems like it'll be extremely un-fun.

## What I think we can do about it

So right now my theory is this: we could write [an MDX plugin](https://docusaurus.io/docs/markdown-features/plugins) for these docs so that we can write something like this in the markdown: 

```md
Hello I like Camunda especially [the modeler]($docs$/components/modeler/about-modeler)
```

and the mdx plugin would see the `$docs$` token, and replace it with the correctly mapped version path. e.g:

- if it was found in a doc from optimize/next, it would replace it with `/docs/next/`
- if it was found in a doc from optimize/3.7.0, it would replace it with `/docs/1.3/`
- if it was found in a doc from optimize/3.8.0, `/docs/` (at least right now....when 8.0 is current.)
- and same to go backwards, from links found in docs from c8 into optimize. 

## A caveat

I haven't worked with MDX plugins before. I'm looking forward to the fun of writing one, but I don't really know what is or isn't possible in the context of a docusaurus site. So I guess we'll find out if my theory (or something similar) is anywhere feasible.

optimize_sidebars.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+  Components: [
+    {
+      type: "link",
+      label: "Overview Components",
+      href: "/docs/next/components",
+    },
```
pepopowitz, line 7: This is a foreseen down-side to the multi-version docs -- [having to rewrite the sidebars files, and duplicate them, due to the two plugins not knowing about each others' docs](https://github.com/camunda/camunda-platform-docs/pull/906#discussion_r874097861).

sidebars.js
```diff
@@ -221,7 +221,11 @@ module.exports = {
         },
       ],
       Optimize: [
-        "components/optimize/what-is-optimize",
+        {
+          type: "link",
+          label: "What is Optimize?",
+          href: "/optimize/what-is-optimize",
+        },
```
pepopowitz, line 228: This is the other end of the foreseen down-side to the multi-version docs -- [having to rewrite the sidebars files, and duplicate them, due to the two plugins not knowing about each others' docs](https://github.com/camunda/camunda-platform-docs/pull/906#discussion_r874097861).

optimize/what-is-optimize.md
pepopowitz: two bits of context I should have shared, from a conversation with @RomanJRW: 

1. [This list of optimize releases](https://confluence.camunda.com/display/CO/Optimize+Release+Overview). 
2. Optimize has decided to align their release schedule with that of C8. (I don't have this in writing anywhere, but I think that's what Josh told me)

Those two facts, combined with the fact that we only have "Optimize" docs starting in the platform version 1.3 docs, lead me to believe that the versioning _is_ tightly coupled, with the specific versions you listed, and will be in the future.

korthout: @pepopowitz As part of the summerhackdays 2022, I recently created some MDX pages for Docusaurus that embed 2 custom build React components. This was rather simple TBH. I was surprised by the freedom we had in there. 

I think that MDX plugins work similarly, but are embedded in a different way. As I understand it, the plugins are configured in docusaurus.config.js, instead of imported and embedded in each markdown file directly.

docusaurus.config.js
```diff
@@ -32,6 +34,16 @@ module.exports = {
       },
     ],
     "./static/plugins/bpmn-js",
+    [
+      "@docusaurus/plugin-content-docs",
+      {
+        id: "optimize",
+        path: "optimize",
+        routeBasePath: "optimize",
+        beforeDefaultRemarkPlugins: [versionedLinks],
```
pepopowitz, line 43: 🔗 This configures the _optimize_ docs to use the MDX plugin that interprets link target URLs based on the source file directory.

docusaurus.config.js
```diff
@@ -207,6 +219,7 @@ module.exports = {
           // Please change this to your repo.
           editUrl:
             "https://github.com/camunda/camunda-platform-docs/edit/main/",
+          beforeDefaultRemarkPlugins: [versionedLinks],
```
pepopowitz, line 222: 🔗 This configures the _main_ docs to use the MDX plugin that interprets link target URLs based on the source file directory.

optimize/what-is-optimize.md
```diff
@@ -0,0 +1,33 @@
+---
+id: what-is-optimize
+title: What is Optimize?
+description: "Leverage process data and analyze areas for improvement."
+---
+
+:::note
+New to Optimize? Visit our introductory guide to [Optimize]($docs$/guides/improve-processes-with-optimize/) to get started.
```
pepopowitz, line 8: 🔗 This uses the MDX plugin I wrote in (TODO: PR number) to have the site interpret which version of the main docs to send the user to, based on the location of the source file.

optimize/what-is-optimize.md
pepopowitz: 🔗 See the other comment in this file for an example of escaping this linking hell!

pepopowitz (commented):
Adding more comments based on another PR I'm about to open, for an MDX plugin that allows us to use tokens to represent paths that will be expanded based on the directory of the source file. In other words, much-easier-to-maintain cross-instance docs linking.

optimize/what-is-optimize.md
pepopowitz: For posterity: it sounds like docusaurus has plans for v3 to [enable different plugins to see each other's stuff](https://github.com/facebook/docusaurus/issues/6370). When that happens we might be able to undo this work 🙌

docusaurus.config.js
```diff
@@ -32,6 +34,16 @@ module.exports = {
...
     ],
     "./static/plugins/bpmn-js",
+    [
+      "@docusaurus/plugin-content-docs",
+      {
+        id: "optimize",
+        path: "optimize",
+        routeBasePath: "optimize",
+        beforeDefaultRemarkPlugins: [versionedLinks],
+        sidebarPath: require.resolve("./optimize_sidebars.js"),
```
pepopowitz, line 44: @akeller this is the line I had commented out that caused me to go down a rabbit hole yesterday.........docusaurus wasn't using my sidebars because I wasn't telling docusaurus to use them. 😅 😬

hacks/generateOptimizeSidebars.md
```diff
@@ -0,0 +1,68 @@
...
+
+   - The last statement of the script copies the generated sidebars into your clipboard.
+
+4. Paste the results into the optimize_sidebars.js file
+
+   - The results do not include the statement `module.exports = `, so leave that part in 😅.
+
+   - Make sure the results get formatted, as the script generates an object that does not match prettier's desired format.
+
+5. Replace/revert any Optimize links that exist within the optimize docs
```
pepopowitz, line 45: As I move more and more docs over, this step might become enough of a pain that I figure out a way to make the script do doc-based links for already-moved docs.

optimize_sidebars.js
```diff
@@ -0,0 +1,1053 @@
+function docsComponentsLink(label, href) {
+  return {
+    type: "link",
+    label: label,
+    href: `/docs/next/${href}`,
+  };
+}
```
pepopowitz, line 7: Long term vision: 

I could have generated this file without this helper, but it would have resulted in `/docs/next/` littered throughout it. This seems like it'd be not fun when we generate a new version of optimize -- we'd have to do a find/replace and include all those changes in the version PR. 

Extracting the prefix of the paths to this one spot means we can just replace it in this one spot when we cut a new version. 

It might be YAGNI, but it made sense to me.

optimize_sidebars.js
```diff
@@ -0,0 +1,1053 @@
...
+              "components/operate/userguide/operate-feedback-and-questions/"
+            ),
+          ],
+        },
+      ],
+    },
+
+    {
+      Optimize: [
+        "what-is-optimize",
```
pepopowitz, line 642: This is the only link in this sidebar file that is doc-based instead of URL based (because it's the only doc moved over in this PR).

pepopowitz (commented):
Adding some more comments for my recent changes

docusaurus.config.js
akeller: Funny how that works 😬

hacks/generateOptimizeSidebars.md
akeller: You can always call in back up too if manually moving things gets the job done. I love the automated approach though.

pepopowitz:
Heads up that this is ready for final review. 

Also cc @christinaausley for visibility.

akeller (approved):
Nothing obviously concerning popping up for me. It makes sense the changes are only in `/next/`.

PR #1169, opened 2022-08-17
Include link to Slack docs regarding message formatting

superbeagle:
I added a line right after the compose image directing users to Slack's documentation on how to format messages using markdown


PreetiNirwal (dismissed):
Looks good. Thanks.

christinaausley (approved):
Adjusted the wording a bit and added to 8.0 versioned docs -- LGTM!

Issue #1167, opened 2022-08-16
Investigate ways to reduce friction in cross-plugin markdown links


pepopowitz:
Epic: https://github.com/camunda/product-hub/issues/231
Parent: #1116 

Markdown links that cross from one docs plugin to another [look like they're going to introduce a bunch of friction](https://github.com/camunda/camunda-platform-docs/pull/1166#discussion_r947018939). We'd spend a lot of time worrying about linking docs to the correct version/URL, at the authoring, reviewing, and versioning steps. 

We should investigate a way to remove the version from cross-plugin links. We know that C8 versions correspond to Optimize versions (8.0 = 3.8.0, 1.3 = 3.7.0, etc), so based on those rules we should be able to expand a link that indicates "the corresponding optimize version" correctly. 

### Implementation?

We might be able to use [a custom MDX plugin](https://docusaurus.io/docs/markdown-features/plugins) for this.

PR #1170, opened 2022-08-17
feat: MDX plugin to interpret link target URLs based on source file directory

pepopowitz:
Addresses #1167.

In #1166, we found that [markdown links across docs instances](https://github.com/camunda/camunda-platform-docs/pull/1166#discussion_r947018939) was going to introduce a ton of friction.

This PR introduces an MDX plugin that should remove this friction. Instead of needing to specify the version of the other docs instance to link to, we can now use a token like `$docs$` to have docusaurus interpret the version based on the source file. 

For example, if we have a link that goes to `$optimize$/what-is-optimize`, the target URL will be expanded to....
* `optimize/next/what-is-optimize` when the source file is in the `/docs/` folder
* `optimize/what-is-optimize` when the source file is in the `/versioned_docs/version-8.0` folder
* `optimize/3.7.0/what-is-optimize` when the source file is in the `/versioned_docs/version-1.3` folder

See the src/mdx/expandVersionedUrl.spec.js file in this PR for more examples of URL expansion.

My intention is to merge this PR separate from and prior to #1166, to make it easier to review in isolation. (That PR will get very large with all of the content moving around.)

## What it looks like in use

[Here's a video of the plugin in action](https://www.loom.com/share/511730f7dbec41e9ad6fb1d748da0041). 

In addition, #1166 demonstrates the usage:

* The plugin needs to be configured for both documentation plugins ([main](https://github.com/camunda/camunda-platform-docs/pull/1166/files#r948312736) and [optimize](https://github.com/camunda/camunda-platform-docs/pull/1166/files#r948312116))
* Markdown links use the tokens [`$docs$`](https://github.com/camunda/camunda-platform-docs/pull/1166/files#r948313279) or [`$optimize$`](https://github.com/camunda/camunda-platform-docs/pull/1166/files#r948311310) to signal that they should be interpreted based on the source location.

## Implementation notes

* This PR **does not** configure our docs to use this plugin. That will happen in #1166 instead, because there's no need to run this plugin until we have our optimize docs in a second instance. 
* This mapping from source directory to target path only works because Optimize versions are released at the same cadence as other C8 products, and there is a 1:1 mapping from Optimize version to main C8 version. 
* The `$docs$` and `$optimize$` tokens are completely arbitrary. If we want, we can use any other token. 
* The PR includes Jest tests, but no CI configuration to run these tests. I could set CI up to run tests in the future, if we'd like.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,9 @@
     "serve": "docusaurus serve",
     "format": "prettier --write .",
     "format:check": "prettier --check .",
-    "prepare": "husky install"
+    "prepare": "husky install",
+    "test": "jest",
+    "test:watch": "jest --watch"
   },
   "dependencies": {
     "@auth0/auth0-react": "^1.10.2",
@@ -46,6 +48,7 @@
   },
   "devDependencies": {
     "husky": "^8.0.1",
+    "jest": "^28.1.3",
     "lint-staged": "^13.0.3",
     "prettier": "2.7.1"
   },
--- a/package.json
+++ b/package.json
@@ -29,7 +29,8 @@
     "mixpanel-browser": "^2.45.0",
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
-    "react-player": "^2.10.1"
+    "react-player": "^2.10.1",
+    "unist-util-visit": "^4.1.0"
   },
   "browserslist": {
     "production": [
--- a/src/mdx/expandVersionedUrl.js
+++ b/src/mdx/expandVersionedUrl.js
@@ -1,20 +1,20 @@
-const versionMap = [
-  {
-    docs: "next",
-    optimize: "next",
-  },
-  {
-    docs: "8.0",
-    optimize: "3.8.0",
-  },
-  {
-    docs: "1.3",
-    optimize: "3.7.0",
-  },
+const docs = [
+  // these mappings are currently wrong because there are no optimize versions yet...but eventually, this is what we'll want.
+  { match: "/optimize/", expandTo: "/docs/next" },
+  { match: "/optimize_versioned_docs/version-3.8.0/", expandTo: "/docs" },
+  { match: "/optimize_versioned_docs/version-3.7.0/", expandTo: "/docs/1.3" },
 ];
 
-function expandVersionedUrl(url, cwd) {
-  return url.replace("$docs$", "/docs/next");
+function expandVersionedUrl(targetUrl, sourcePath) {
+  let expandedUrl = targetUrl;
+
+  docs.forEach((rule) => {
+    if (sourcePath.includes(rule.match)) {
+      expandedUrl = targetUrl.replace("$docs$", rule.expandTo);
+    }
+  });
+
+  return expandedUrl;
 }
 
 module.exports = expandVersionedUrl;
--- a/src/mdx/expandVersionedUrl.js
+++ b/src/mdx/expandVersionedUrl.js
@@ -5,12 +5,24 @@ const docs = [
   { match: "/optimize_versioned_docs/version-3.7.0/", expandTo: "/docs/1.3" },
 ];
 
+const optimize = [
+  { match: "/docs/", expandTo: "/optimize/next" },
+  { match: "/versioned_docs/version-8.0/", expandTo: "/optimize" },
+  { match: "/versioned_docs/version-1.3/", expandTo: "/optimize/3.7.0" },
+];
+
 function expandVersionedUrl(targetUrl, sourcePath) {
   let expandedUrl = targetUrl;
 
   docs.forEach((rule) => {
     if (sourcePath.includes(rule.match)) {
       expandedUrl = targetUrl.replace("$docs$", rule.expandTo);
+    }
+  });
+
+  optimize.forEach((rule) => {
+    if (sourcePath.includes(rule.match)) {
+      expandedUrl = targetUrl.replace("$optimize$", rule.expandTo);
     }
   });
 
--- a/src/mdx/expandVersionedUrl.js
+++ b/src/mdx/expandVersionedUrl.js
@@ -1,30 +1,41 @@
-const docs = [
-  // these mappings are currently wrong because there are no optimize versions yet...but eventually, this is what we'll want.
-  { match: "/optimize/", expandTo: "/docs/next" },
-  { match: "/optimize_versioned_docs/version-3.8.0/", expandTo: "/docs" },
-  { match: "/optimize_versioned_docs/version-3.7.0/", expandTo: "/docs/1.3" },
-];
-
-const optimize = [
-  { match: "/docs/", expandTo: "/optimize/next" },
-  { match: "/versioned_docs/version-8.0/", expandTo: "/optimize" },
-  { match: "/versioned_docs/version-1.3/", expandTo: "/optimize/3.7.0" },
+const tokens = [
+  {
+    token: "$docs$",
+    rules: [
+      // these mappings are currently wrong because there are no optimize versions yet...but eventually, this is what we'll want.
+      { match: "/optimize/", expandTo: "/docs/next" },
+      { match: "/optimize_versioned_docs/version-3.8.0/", expandTo: "/docs" },
+      {
+        match: "/optimize_versioned_docs/version-3.7.0/",
+        expandTo: "/docs/1.3",
+      },
+    ],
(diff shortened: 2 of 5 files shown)


src/mdx/versionedLinks.js
```diff
@@ -0,0 +1,22 @@
+const visitPromise = import("unist-util-visit");
+const expandVersionedUrl = require("./expandVersionedUrl");
+
+const resolve = async () => {
+  visit = (await visitPromise).visit;
+};
+
+resolve();
```
pepopowitz, line 8: This `import`/`resolve` nonsense irritates me, but I was led here due to the following: 

1. `require` is not supported. And in fact it suggest I use a dynamic `import()` instead:

```
[ERROR] Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/stevenhicks/sjh/dev/camunda/camunda-platform-docs/node_modules/unist-util-visit/index.js from /Users/stevenhicks/sjh/dev/camunda/camunda-platform-docs/src/mdx/versionedLinks.js not supported.
Instead change the require of index.js in /Users/stevenhicks/sjh/dev/camunda/camunda-platform-docs/src/mdx/versionedLinks.js to a dynamic import() which is available in all CommonJS modules.
```

I don't know enough about modules to know why this is, or if there's a way to make it accept `require(...)`. 

2. `await import()` is not supported, no top-level await: 
```
[ERROR] /Users/stevenhicks/sjh/dev/camunda/camunda-platform-docs/src/mdx/versionedLinks.js:1
const visit = await import("unist-util-visit");
              ^^^^^

SyntaxError: await is only valid in async functions and the top level bodies of modules
```

I thought that maybe I could [rename from `.js` to `.mjs` files to signal that they were modules](https://stackoverflow.com/a/65018201), but that didn't work.

I didn't want to spend a lot of time dealing with build issues, so I just accepted this code.

src/mdx/expandVersionedUrl.js
```diff
@@ -0,0 +1,43 @@
+const tokens = [
```
pepopowitz, line 1: These tokens and rules are defined in a way that _I think_ could come from configuration instead, in case we decide to share this plugin with the rest of the world. I'm intending on reaching out to the docusaurus community to see if this kind of thing is useful for others.

src/mdx/versionedLinks.js
```diff
@@ -0,0 +1,22 @@
...
+
+const versionedLinks = (options) => {
+  const transformer = async (ast, vfile) => {
+    await visit(ast, "link", (node) => {
```
pepopowitz, line 12: This visits every single markdown link in the docs. I think I've provided enough early-exits in the `expandVersionedUrl` function that we don't have to worry about builds slowing down....that also might even be premature optimization on my part. But if we do notice build times go way up, there might be some micro-optimizations we can make here?

akeller: We would see this on local builds too right? I may be the only person who builds locally anymore, but I tend to do that when I'm making significant link changes, particularly across multiple pages.

package.json
```diff
@@ -29,7 +31,8 @@
     "mixpanel-browser": "^2.45.0",
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
-    "react-player": "^2.10.1"
+    "react-player": "^2.10.1",
+    "unist-util-visit": "^4.1.0"
```
akeller, line 35: WAT!! I remember when I had to do this in vanilla JS to parse JSON objects.

src/mdx/expandVersionedUrl.js
```diff
@@ -0,0 +1,43 @@
...
+  {
+    token: "$optimize$",
+    rules: [
+      // these mappings are currently wrong because there are no optimize versions yet...but eventually, this is what we'll want.
+      { match: "/docs/", expandTo: "/optimize/next" },
+      { match: "/versioned_docs/version-8.0/", expandTo: "/optimize" },
+      { match: "/versioned_docs/version-1.3/", expandTo: "/optimize/3.7.0" },
+    ],
+  },
+];
```
akeller, line 23: Thinking ahead here, would we (DX) then add (and adjust) the rules for `8.1` or would we ask the Optimize team to do this? Whose responsibility is it? Feels like DX if we are the ones publishing the release.

src/mdx/expandVersionedUrl.spec.js
```diff
@@ -0,0 +1,57 @@
+const expandVersionedUrl = require("./expandVersionedUrl");
+
+describe("expandVersionedUrl", () => {
+  describe("unexpandable URLs", () => {
+    const sourcePath =
+      "/Users/monkeypants/camunda-platform-docs/optimize/what-is-optimize.md";
```
akeller, line 6: My favorite username 🐒 👖

src/mdx/versionedLinks.js
akeller: Huh, is this d-🦖?

akeller:
@pepopowitz this would require a 1 time update to all internal links in the docs? Or just all internal Optimize links?

akeller:
image

Any anticipated impact to the buttons for content in the sidebar "tree"? This is a super wonky way to describe this, but these buttons are generated by d-🦖.

src/mdx/versionedLinks.js
pepopowitz: I think so, but what's confusing is that the docs for writing custom MDX plugins in docusaurus use `require`. I'm planning to reach out via discord to see if I'm the only one experiencing this.

src/mdx/versionedLinks.js
pepopowitz: Yes, if there's a performance degradation we'd see it on local builds too.

For what it's worth, and anecdotally, developing locally with `npm start` doesn't appear any slower than it used to.

src/mdx/expandVersionedUrl.js
pepopowitz: I would anticipate DX owning this, but that's because I'm assuming we're the ones who cut new versions of the main docs when they're released, too. We'd have to run `docusaurus version` to generate the new version number in addition to changing these rules, and that feels like infrastructure work for _us_. 

Right?

pepopowitz:
> this would require a 1 time update to all internal links in the docs? Or just all internal Optimize links?

@akeller it would be a 1 time update to:

* any links that go from the main docs into the optimize docs
* any links that go from the optimize docs into the main docs

I would consider that all to be part of #1116.

pepopowitz:
> Any anticipated impact to the buttons for content in the sidebar "tree"? 

@akeller based on the way those buttons span across left-sidebar sections, my assumption is they're based on the sidebars files. This is again covered by #1116, which includes synchronizing the sidebars files for main docs with optimize, basically to preserve this (and the left nav) structure.

src/mdx/versionedLinks.js
christinaausley: Big proponent of building locally before pushing anything @akeller  🙏

src/mdx/expandVersionedUrl.js
akeller: Excellent, we are on the same page here. 

🔮 I foresee us updating the docs-on-docs to include clear responsibilities around these minor product releases vs. continuous product releases (when we need to run `docusaurus version` vs. just releasing). These should probably have a clear name too.

christinaausley (approved):
No comments on my end here. Nice work and keep me updated on how I can assist 🥳

src/mdx/versionedLinks.js
pepopowitz: Apparently there is a disclaimer in the docs that is intended to explain this:

> NOTE
> There's recently a trend in the Remark/Rehype ecosystem to migrate to ES Modules, a new JavaScript module system, which Docusaurus doesn't support yet. Please make sure your installed plugin version is CommonJS-compatible before we officially support ESM. Alternatively, you can read about using dynamic import() as a workaround in the tutorial of installing [rehype-katex](https://docusaurus.io/docs/markdown-features/math-equations#upgrading-rehype-katex-beyond-recommended-version).

If we want to use `require` instead of dynamic imports, [we need to install an older version of unist-util-visit](https://discord.com/channels/398180168688074762/867060369087922187/1011369318807765072): 

> You need to use unist-util-visit v2


I'm personally fine with this clunky import/await code so I don't intend on rewriting with that version.

PR #1174, opened 2022-08-18
ensure all sidebar labels are sentence case structure

christinaausley:
Adjusting all sidebars to sentence case format.

### DIFF
--- a/sidebars.js
+++ b/sidebars.js
@@ -299,19 +299,19 @@ module.exports = {
           ],
         },
       ],
-      "Integration framework": [
+      "Integration Framework": [
         "components/integration-framework/introduction-to-connectors",
         {
           Connectors: [
             "components/integration-framework/connectors/use-connectors",
             {
-              "Out-of-the-box connectors": [
+              "Out-of-the-box Connectors": [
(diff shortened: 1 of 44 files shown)


sidebars.js
```diff
@@ -311,7 +311,7 @@ module.exports = {
                 "components/integration-framework/connectors/out-of-the-box-connectors/sendgrid",
                 "components/integration-framework/connectors/out-of-the-box-connectors/slack",
               ],
-              "Custom-build Connectors": [
+              "Custom-built Connectors": [
```
akeller, line 314: Thank you for catching this one 😅

akeller (approved):
Thank you! And it looks like we've committed to "Connector" or "Connectors", but I just want to make that super clear based on the changes in this PR. This works for me, but the scope of this PR is just the titles, descriptions, and sidebar labels not the contents of the page.

christinaausley:
> Thank you! And it looks like we've committed to "Connector" or "Connectors", but I just want to make that super clear based on the changes in this PR. This works for me, but the scope of this PR is just the titles, descriptions, and sidebar labels not the contents of the page.

Correct!

PR #1175, opened 2022-08-18
lowercase secrets 

christinaausley:
Would you believe we have 1005 instances of "secret" in our docs?

You know that feeling you get when you've stared at a word so many times, it doesn't even look like that word anymore? This was one of those times. 😀 

BTW -- don't ask why my commit is called "glossary", my eyes were too glazed over to notice before I pushed the Enter key.

@akeller I'm also not sure why the google drive file was added to this commit -- should I just delete it?


akeller:
> @akeller I'm also not sure why the google drive file was added to this commit -- should I just delete it?

Makes sense to delete it since it's not on `main` anyway. Also your build-docs action may continue to fail because of a problem with [GitHub](https://www.githubstatus.com/).

christinaausley:
> > @akeller I'm also not sure why the google drive file was added to this commit -- should I just delete it?
> 
> Makes sense to delete it since it's not on `main` anyway. Also your build-docs action may continue to fail because of a problem with [GitHub](https://www.githubstatus.com/).

Done!

akeller (approved):
LGTM!

This might be a good term to add to the cheatsheet too.

Issue #1154, opened 2022-08-15
Add item to Documentation Board when PR needs docs review


akeller:
Goal: See things that need a "Docs review" show up on the Documentation Board in the "In Review" column.

Current way is manual - engineer assigns @christinaausley or @akeller when PR is "ready for docs review".

* https://github.com/actions/add-to-project
* Using labels for this?
  * Adding the label is the same thing as asking in a comment for a review
  * Author or PR owner should do a thing to initiate a review (comment, assign reviewer, add label)

Trigger from GitHub - reviewer was added - check for @christinaausley or @akeller 
Add to project Action

akeller:
🍐 / 🍐  sessions! Thanks @pepopowitz! And double thanks for getting it over the line and out the door!

PR #1183, opened 2022-08-19
feat: add PRs to Documentation board in In Review column when review requested from Amara or Christina

pepopowitz:
A collaboration with @akeller.

Addresses #1154 

This PR rewrites the `sync-board` workflow to:

1. Listen for reviewers to be assigned to a PR.
2. If the assigned reviewer isn't @akeller or @christinaausley, stop.
3. Otherwise, add the PR to the Documentation Board project, and set its status to In Review.

[Here is an example of a successful workflow run](https://github.com/camunda/camunda-platform-docs/actions/runs/2891812456).

## Alternatives considered

1. We initially attempted to use existing actions to do this:
    * [https://github.com/actions/add-to-project](actions/add-to-project) was able to add the PR to the project, but it does not support also setting the Status field to In Review.
    * [https://github.com/alex-page/github-project-automation-plus](alex-page/github-project-automation-plus) is supposed to be capable of setting the Status field to In Review, but we were unable to get it to work. We aren't sure why -- it may have been an issue with the scopes assigned to the personal access token we were using, but it seemed like we should have been covered. 

2. Using labels to trigger the workflow instead of review request. We initially did this because the actions/add-to-project job operated based on labels instead of reviewers. Once we got to the point of writing all of the GraphQL ourselves, it was no longer necessary to trigger on labels. Since both options would use `jobs.if` conditions to target specific PRs, and `jobs.if` conditions leave noise in the Actions tab for all skipped jobs, we chose review assignment over labels because it would likely cause fewer "skipped" actions.

## Edge cases

If the sync-board job were to fail, it would leave a PR in red for a confusing reason. We want to avoid this if possible. Here is a list of edge cases that we considered, and the outcome of each:

- if the PR is already on the documentation board in In Review, [the job succeeds](https://github.com/camunda/camunda-platform-docs/actions/runs/2891957669) without any changes. 
- if the PR is already on the documentation board in a different status, [the job succeeds](https://github.com/camunda/camunda-platform-docs/actions/runs/2891944563), but the status gets updated back to In Review. This might lead to things moving when we don't want them to, but I'd like to see it be a problem before we try to fix it.
- if review is requested from both @akeller and @christinaausley at the same time, the job runs twice ([here](https://github.com/camunda/camunda-platform-docs/actions/runs/2891920262) and [here](https://github.com/camunda/camunda-platform-docs/actions/runs/2891920478)). Both runs succeed.
- if review is requested from multiple people at the same time, one of them a docs reviewer and one not, the run with the docs reviewer would successfully add the PR to the board. The other run would be skipped.

### DIFF
--- a/.github/workflows/sync-board.yaml
+++ b/.github/workflows/sync-board.yaml
@@ -2,13 +2,12 @@ name: sync-board
 
 on:
   pull_request:
-    types:
-      - labeled
+    types: [review_requested]
 
 jobs:
   sync-board:
     # if: github....reviewer == "christina" or "amara"
-    if: contains(github.event.pull_request.labels.*.name, 'needs-docs-review')
+    # if: contains(github.event.pull_request.labels.*.name, 'needs-docs-review')
     runs-on: ubuntu-latest
     steps:
       - name: Dump GitHub context
--- a/.github/workflows/sync-board.yaml
+++ b/.github/workflows/sync-board.yaml
@@ -61,30 +61,30 @@ jobs:
             }' -f project=$PROJECT_ID -f pr=$PR_ID --jq '.data.addProjectV2ItemById.item.id')"
 
             echo 'ITEM_ID='$item_id >> $GITHUB_ENV
-      # - name: Set fields
-      #   env:
-      #     GITHUB_TOKEN: ${{ secrets.ADD_TO_PROJECT_PAT }}
-      #   run: |
-      #     gh api graphql -f query='
-      #       mutation (
-      #         $project: ID!
-      #         $item: ID!
-      #         $status_field: ID!
-      #         $status_value: String!
-      #       ) {
-      #         set_status: updateProjectV2ItemFieldValue(input: {
-      #           projectId: $project
-      #           itemId: $item
-      #           fieldId: $status_field
-      #           value: {
-      #             singleSelectOptionId: $status_value
-      #             }
-      #         }) {
-      #           projectV2Item {
-      #             id
-      #             }
-      #         }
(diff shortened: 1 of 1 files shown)


pepopowitz:
Heads up friends that this is ready for actual review!!!

akeller:
Is it worth mentioning that the PAT is technically associated with me?

akeller (approved):
I may even offer a reward for whoever finds `requested_reviewer` in the GitHub docs. Singular, not plural.

pepopowitz:
> Is it worth mentioning that the PAT is technically associated with me?

YES, good call. I can add a comment. What do you think about also adding the token to Keeper, with a comment about which scopes are included, in case we need to regenerate it after you win the lottery or quit the tech industry to start a garden rattlesnake extraction company?

.github/workflows/sync-board.yaml
```diff
@@ -4,12 +4,9 @@ on:
   pull_request:
     types: [review_requested]
 
-env:
```
akeller: It was a valiant attempt.

christinaausley (dismissed):
No comments from a technical review here 👍

akeller:
> YES, good call. I can add a comment. What do you think about also adding the token to Keeper, with a comment about which scopes are included, in case we need to regenerate it after you win the lottery or quit the tech industry to start a garden rattlesnake extraction company?

NO SNAKES

akeller:
I'll add this to Keeper today, but I need to regenerate the token.

pepopowitz:
@akeller latest commit adds a comment about the PAT -- let me know if there's more info you want from it.

akeller (approved):

PR #1186, opened 2022-08-22
feat: Identity read/write access

btepe:
Resolves https://github.com/camunda-cloud/identity/issues/894

### DIFF
--- a/docs/self-managed/identity/user-guide/managing-user-access.md
+++ b/docs/self-managed/identity/user-guide/managing-user-access.md
@@ -4,7 +4,7 @@ title: "Managing user access to Identity"
 sidebar_label: "Managing user access"
 ---
 
-In this guide, you will learn how to grant users different levels of access to the Identity application.
+In this guide, you will learn about the different access levels users can receive within the Identity application.
 
 ## Permissions supported by Identity
 
@@ -24,7 +24,7 @@ This gives the user access to the Identity application.
 
 ## Assign Identity permissions to a user
 
-Without an Identity permission, users can access Identity only to log in to other applications.
+Without at least one of the permissions listed above, users can access Identity only to log in to other applications.
 They are not able to access the Identity UI beyond the login page.
 
 To grant a user access to the UI, you need to assign at least one Identity permission as part of a role to the user. This can be achieved in one of the following ways:
@@ -34,10 +34,7 @@ To grant a user access to the UI, you need to assign at least one Identity permi
 When you use our Keycloak presets, Identity creates the `Identity` role automatically.
 The role contains the necessary permissions to give a user full read and write access to Identity.
 
-When Web Modeler is connected to Identity, Identity creates the `Web Modeler` role. This role contains the Identity
-`read:users` permission.
-
-[Assign one of the above roles to a user](assigning-a-role-to-a-user.md) to enable the user to access the Identity UI.
+[Assign the `Identity` role to a user](assigning-a-role-to-a-user.md) to enable the user to access the Identity UI.
 
 ### Create a custom role
 
--- a/docs/self-managed/identity/user-guide/managing-user-access.md
+++ b/docs/self-managed/identity/user-guide/managing-user-access.md
@@ -10,9 +10,9 @@ In this guide, you will learn how to grant users different levels of access to t
 
 Identity implements the following permissions:
 
-- `read`: Users can access all pages in Identity. They _cannot_ create, modify or delete any data.
-- `read:users`: Users can access only the "Users" page and related subpages.
-- `write`: Users have access to all pages. They can create, modify and delete data.
+- `read`: Users can access all pages in Identity. They _cannot_ create, modify, or delete any data.
+- `read:users`: Users can access only the **Users** page and related subpages.
+- `write`: Users have access to all pages. They can create, modify, and delete data.
 
 :::note
 If you are connecting Identity to your existing Keycloak installation, the `read:users` permission needs to be created
@@ -27,8 +27,7 @@ This gives the user access to the Identity application.
 Without an Identity permission, users can access Identity only to log in to other applications.
 They are not able to access the Identity UI beyond the login page.
 
-To grant a user access to the UI, you need to assign at least one Identity permission as part of a role to the user.
-This can be achieved in one of the following ways.
+To grant a user access to the UI, you need to assign at least one Identity permission as part of a role to the user. This can be achieved in one of the following ways:
 
 ### Using Keycloak presets
 
@@ -42,5 +41,5 @@ When Web Modeler is connected to Identity, Identity creates the `Web Modeler` ro
 
 ### Create a custom role
 
-When using an existing Keycloak instance or if you want to create your own set of permissions, follow our guides to
+When using an existing Keycloak instance, or if you want to create your own set of permissions, follow our guides to
 [create a new role](adding-a-role.md) and [assign it to users](assigning-a-role-to-a-user.md).
--- a/docs/self-managed/identity/user-guide/managing-user-access.md
+++ b/docs/self-managed/identity/user-guide/managing-user-access.md
@@ -10,33 +10,37 @@ In this guide, you will learn how to grant users different levels of access to t
 
 Identity implements the following permissions:
 
-- `read`: Users can access all pages in Identity. They cannot create, modify or delete any data.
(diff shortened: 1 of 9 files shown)


christinaausley (dismissed):
Minor adjustments in latest commit, including:
- Bold UI elements.
- Add Oxford comma
Nice work here!

docs/self-managed/identity/user-guide/managing-user-access.md
```diff
@@ -0,0 +1,45 @@
...
+
+Identity implements the following permissions:
+
+- `read`: Users can access all pages in Identity. They _cannot_ create, modify, or delete any data.
+- `read:users`: Users can access only the **Users** page and related subpages.
+- `write`: Users have access to all pages. They can create, modify, and delete data.
+
+:::note
+If you are connecting Identity to your existing Keycloak installation, the `read:users` permission needs to be created
+manually in Keycloak.
```
Ben-Sheppard, line 19: Do we have a guide on how to do this? In theory with 8.1 hopefully it won't be an issue as we will [adapt Identity to handle this for the user](https://github.com/camunda-cloud/identity/issues/983) but I can see users getting stuck with this task as we require the permissions to be configured in a specific way.

As we have a planned piece of work to negate this issue I'd be tempted to remove this note for now as its easier than adding in a temporary page on adding permissions in Keycloak :)

docs/self-managed/identity/user-guide/managing-user-access.md
```diff
@@ -0,0 +1,45 @@
+---
+id: managing-user-access
+title: "Managing user access to Identity"
+sidebar_label: "Managing user access"
+---
+
+In this guide, you will learn how to grant users different levels of access to the Identity application.
```
Ben-Sheppard, line 7: I'm not sure this guide tells the user _how_ to grant users different access but I feel it more describes to the user the different access levels available.

I wonder if an alternative could be 
"In this guide, you will learn about the different access levels available to users within the Identity application." - WDYT?

docs/self-managed/identity/user-guide/managing-user-access.md
```diff
@@ -0,0 +1,45 @@
...
+If you are connecting Identity to your existing Keycloak installation, the `read:users` permission needs to be created
+manually in Keycloak.
+:::
+
+You can [assign the above permissions to users as part of a role](assigning-a-role-to-a-user.md).
+This gives the user access to the Identity application.
+
+## Assign Identity permissions to a user
+
+Without an Identity permission, users can access Identity only to log in to other applications.
```
Ben-Sheppard, line 27: I wonder if this paragraph can be simplified to
"Users are unable to access the Identity UI without at least one of the permissions listed above. To grant a user access to the Identity UI ..."

docs/self-managed/identity/user-guide/managing-user-access.md
```diff
@@ -0,0 +1,45 @@
...
+They are not able to access the Identity UI beyond the login page.
+
+To grant a user access to the UI, you need to assign at least one Identity permission as part of a role to the user. This can be achieved in one of the following ways:
+
+### Using Keycloak presets
```
Ben-Sheppard, line 32: Are these the component presets?

docs/self-managed/identity/user-guide/managing-user-access.md
```diff
@@ -0,0 +1,45 @@
...
+
+When you use our Keycloak presets, Identity creates the `Identity` role automatically.
+The role contains the necessary permissions to give a user full read and write access to Identity.
+
+When Web Modeler is connected to Identity, Identity creates the `Web Modeler` role. This role contains the Identity
```
Ben-Sheppard, line 37: I think here we should be using consistent terms so instead of connected I would instead say something like "When Web Modeler is initialised in Identity, ..." as the users are told about component initialisation in other guides

docs/self-managed/identity/user-guide/managing-user-access.md
```diff
@@ -0,0 +1,45 @@
...
+`read:users` permission.
+
+[Assign one of the above roles to a user](assigning-a-role-to-a-user.md) to enable the user to access the Identity UI.
```
Ben-Sheppard, line 40: I think here we should only suggest that users really assign either Identity or a role that specifically is designed for access to Identity to be able to access the UI, I understand that the Web Modeler role has a permission for Identity that allows the user access to the UI but its primary purpose is actually to support behaviour within the Web Modeler component and not provide access to the Identity UI

Ben-Sheppard (commented):
Thank you for adding this guide! I really like how we're building up a nice ecosystem of pages and guides that link to each other to help our users navigate the Identity docs, I've just got some small comments

docs/self-managed/identity/user-guide/managing-user-access.md
btepe: Do you think it would be enough to link to the [Keycloak docs on managing permissions](https://www.keycloak.org/docs/16.1/authorization_services/index.html#_permission_overview)? I can imagine that we would at least need some basic explanations on which options to choose in Keycloak, so I would be fine removing the note.

Ben-Sheppard: Linking to that page would definitely work though the more I think about it I would personally opt for removing the note for now, we can always add something in if we don't manage to remove the requirement for the user to manually create the permission

docs/self-managed/identity/user-guide/managing-user-access.md
btepe: I initially wanted to write something similar. The reason behind mentioning the login (even though its functionality is part of Keycloak) was to make sure users understand that users can still log in using Identity. This is under the assumption that customers use our Keycloak with our theme in which case they will likely not know that it is part of Keycloak and not Identity. Wdyt?

I like that you referenced the permissions mentioned above, this makes it less ambiguous 👍

docs/self-managed/identity/user-guide/managing-user-access.md
btepe: Sounds good 👍

docs/self-managed/identity/user-guide/managing-user-access.md
btepe: In general, yes that was the intention. I'm not sure about the `Identity` role though. Would you consider it part of the component presets? 

I realise that the title and formulations in the text might be too vague. Perhaps a better way would be to write something like "Using our component presets" and "When you start Identity with our pre-configured Keycloak container...". Wdyt? We could also link the docs section on initialisation.

docs/self-managed/identity/user-guide/managing-user-access.md
btepe: 👍

docs/self-managed/identity/user-guide/managing-user-access.md
btepe: Good point! I removed the part about Web Modeler 👍

docs/self-managed/identity/user-guide/managing-user-access.md
Ben-Sheppard: Ah fair point, I think thats a fair suggestion to make actually as Users may not be aware that the login flows still happen via Identity its just the UI that is protected by permissions. An alternative suggestion could be 

"Users are always able to use Identity to login to the components however they are unable to access the Identity UI without at least one of the permissions listed above. To grant a user access to the Identity UI ..."

docs/self-managed/identity/user-guide/managing-user-access.md
Ben-Sheppard: For me Identity is separate to the component presets. Your alternative suggestions here sound good to me, though in the future even with a customer owned Keycloak the Identity role will automatically be created. I'd be happy with your suggestions now and we can always adjust if necessary :)

docs/self-managed/identity/user-guide/managing-user-access.md
btepe: That sounds reasonable 👍 I removed the note for now.

docs/self-managed/identity/user-guide/managing-user-access.md
btepe: I like your suggestion 👍

Ben-Sheppard (approved):
Thank you for addressing my comments :)

christinaausley (approved):

PR #1193, opened 2022-08-23
chore: remove vNext and optimize docs from sitemap temporarily

pepopowitz:
Related to #1116 

This PR removes the "next" version & the Optimize docs from the sitemap, temporarily. This is in preparation for moving a bunch of files around, iteratively, for #1116. 

### Why?

1. I'm not sure what effect having a bunch of files disappearing & moving from the sitemap will have for SEO
2. I think link-checking will break once I start submitting PRs that move docs from the `docs` instance to the `optimize` instance. Removing the entire next version from the sitemap would prevent link-checking from happening on these files that are being moved.

## Proof that it removes things

Here's a screenshot of me searching a generated sitemap for `/next/` -- with zero results. Note that it's pointless to include a search for the optimize documentation, because there is no optimize docs instance in this branch yet. 

image

## PR Checklist

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -216,6 +216,8 @@ module.exports = {
           //cacheTime: 600 * 1000, // 600 sec - cache purge period
           changefreq: "weekly",
           priority: 0.5,
+          // this is temporary. Ignore these from the sitemap while we move a lot of things around.
+          ignorePatterns: ["/docs/next/**", "/optimize/**"],
         },
       },
     ],


docusaurus.config.js
```diff
@@ -216,6 +216,8 @@ module.exports = {
           //cacheTime: 600 * 1000, // 600 sec - cache purge period
           changefreq: "weekly",
           priority: 0.5,
+          // this is temporary. Ignore these from the sitemap while we move a lot of things around.
+          ignorePatterns: ["/docs/next/**", "/optimize/**"],
```
pepopowitz, line 220: There are no `/optimize/**` matches in this branch, currently....but I'll be rebasing #1166 on main after this is merged, and there are matches there.

akeller:
This change makes me nervous that the Optimize docs may be hard to reach during the window this work will take to be completed. Can you do a docs release before merging these changes to `main`?

pepopowitz:
> This change makes me nervous that the Optimize docs may be hard to reach during the window this work will take to be completed. Can you do a docs release before merging these changes to main?

I definitely can. 

Digging deeper on this: I don't see the "window this work will take to be completed" being small. I think we're going to want to deploy the docs multiple times (for other reasons) while it's in progress. My intention is to make the `next`/`optimize` sections as isolated as possible, so that they can sit in a WIP state without affecting real users. Do you think there are any other steps I need to take in order to make that possible? 

- the sitemap won't render the next/optimize sections
- there should be no link-checking of those sections
- search results should not include those sections
- users (and crawlers) could still switch to version "next" in their browser and view these sections
    - Note that we can't mark a "section" or "version" of the docs as `noindex`....if we want to apply `noindex`, we have to do it on each individual page.

pepopowitz:
> Can you do a docs release before merging these changes to main?

Done: https://github.com/camunda/camunda-platform-docs/releases/tag/8.0.25

akeller:
> Digging deeper on this: I don't see the "window this work will take to be completed" being small. I think we're going to want to deploy the docs multiple times (for other reasons) while it's in progress. My intention is to make the `next`/`optimize` sections as isolated as possible, so that they can sit in a WIP state without affecting real users. Do you think there are any other steps I need to take in order to make that possible?
> 
> * the sitemap won't render the next/optimize sections
> * there should be no link-checking of those sections
> * search results should not include those sections
> * users (and crawlers) could still switch to version "next" in their browser and view these sections
>   
>   * Note that we can't mark a "section" or "version" of the docs as `noindex`....if we want to apply `noindex`, we have to do it on each individual page.

Optimize's release schedule has their next minor release coming in October. Will this still potentially be in progress by then? 

What you've outlined feels complete. Having `/next/` docs available seems like a "nice to have" from a user perspective, so I'm not too upset about them being a little harder to find during this process. 

I'm likely a bit paranoid about this because I know the majority of our traffic comes from Google searches, so removing it from the sitemap may be disruptive. But, it's `/next/` content, which is relatively new anyway. We can always be transparent about this and mention it on the forum so if anyone has feedback during this time, we point them there.

pepopowitz:
> Optimize's release schedule has their next minor release coming in October. Will this still potentially be in progress by then?

I told you this in person @akeller but for posterity no, I don't think this will still be in progress then.

A thing it took me a long time to recognize about myself is that I like to tackle the hardest stuff at the beginning of a problem, and set the rest of the project up to move quickly. This unfortunately means that it looks like nothing is happening for a long time....until all of a sudden everything is in place, and the changes come quickly. #1166 is nearing the end of the "putting things into place with little visible progress" stage, and once it's there, I should be able to iterate rapidly. 

Just not rapidly enough to get it all done in the scope of one docs release.

akeller (approved):
Unblocking you 😄

PR #1197, opened 2022-08-26
chore: add a gitattributes to normalize line endings, and normalize some previously added files

pepopowitz:

1. Adds a .gitattributes file to normalize line-endings. Contributors on Windows [have experienced issues with line-endings being updated when committing](https://camunda.slack.com/archives/C026U8GBNSW/p1661255681204399). In that conversation, @tmetzke resolved the issue by setting `autocrlf` to `input` in his .gitconfig, but I think that adding this .gitattributes file should prevent people from having to do that. 
2. Normalizes line-endings for a few files that were previously committed with `crlf`. 

I followed [this guide](https://www.aleksandrhovhannisyan.com/blog/crlf-vs-lf-normalizing-line-endings-in-git/#a-simple-gitattributes-config) to add the gitattributes file/setting and renormalize files. I'm not super strong with line-endings, so if you're wondering "why'd you do it this way" when you think there's a better way, it's probably because I don't know any better! So please speak up. 

## PR Checklist


akeller (approved):
Interesting that this only impacted a few Optimize docs in V1.3. I guess that's because they were copied over from raw C7 docs 🤷‍♀️ .

Confirmed this is a good default approach per the [GitHub docs](https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings) as well.
> text=auto Git will handle the files in whatever way it thinks is best. This is a good default option.

tmetzke (approved):
This will help a lot of Windows users, thanks @pepopowitz 👍

korthout (commented):
LGTM, but I'm not a windows user 😄

PR #1198, opened 2022-08-26
Remove second H1 on the home page

akeller:

There are two H1's on the home page and I'm embarrassed I did this, but more embarrassed that Bing Webmaster Tools caught it before any of us did... OOPS! The second one was changed to H2. This will need to be live for us to recheck in Bing Webmaster Tools, but you can also see it in the updated HTML.

No

Whenever, but this may be impacting SEO so sooner rather than later would be good.

## PR Checklist


pepopowitz (approved):
👍

PR #1216, opened 2022-08-31
Fix typo

tobiasschaefer:
## What is the purpose of the change

Make documentation more readable

## Are there related marketing activities

No, I don't intend to post this on social media 😉

## When should this change go live?

Up to you ...

SCNR

😃

PR #1219, opened 2022-08-31
Typo in Operate API docs (8.0)

akeller:
Typo - @tobiasschaefer caught this in #1216, but we should also fix it in the 8.0 docs.


christinaausley (approved):
Nice little catch 👍

PR #1238, opened 2022-09-07
feat (optimize/3.8.0): redirect rules for optimize/3.8.0 docs

pepopowitz:
Child of #1234

My intention is to merge this into #1236 (not #1234), then #1236 into #1234, then #1234 into `main`. It's kind of complicated but this PR doesn't make sense without the content all having been moved.

This PR adds redirects for all optimize/3.8.0 docs that have been moved. 

Included are a few pre-existing redirects against some Optimize docs, which have been updated to redirect to the correct optimize/3.8.0 location. 

## PR Checklist


static/.htaccess
```diff
@@ -306,6 +298,50 @@ RewriteRule ^docs/0.25/components/zeebe/tasklist-user-guide /docs/0.25/component
...
 RewriteRule ^docs/components/best-practices/?$ /docs/components/best-practices/overview/ [R=301,L]
 
+# optimize docs get their own section of the site
+## these rules previously existed, but are now updated to point at the /optimize section:
+RewriteRule ^docs/components/optimize/userguide/combined-reports/(.*)$ /optimize/userguide/combined-process-reports/$1 [R=301,L]
+RewriteRule ^docs/components/optimize/userguide/process-analysis/overview/(.*)$ /optimize/userguide/process-analysis/process-analysis-overview/$1 [R=301,L]
+RewriteRule ^docs/components/optimize/userguide/process-analysis/overview$ /optimize/userguide/process-analysis/process-analysis-overview/ [R=301,L]
+RewriteRule ^docs/components/optimize/userguide/decision-analysis/overview/(.*)$ /optimize/userguide/decision-analysis/decision-analysis-overview/$1 [R=301,L]
+RewriteRule ^docs/components/optimize/userguide/decision-analysis/overview$ /optimize/userguide/decision-analysis/decision-analysis-overview/ [R=301,L]
+RewriteRule ^docs/components/optimize/userguide/additional-features/filters/(.*)$ /optimize/userguide/process-analysis/filters/$1 [R=301,L]
```
pepopowitz, line 308: The only edit I made to these rules, all of which were moved from elsewhere in the file, is to update the "to" URL to be in the /optimize section instead of /docs/components/optimize.

static/.htaccess
```diff
@@ -306,6 +298,50 @@ RewriteRule ^docs/0.25/components/zeebe/tasklist-user-guide /docs/0.25/component
...
+## these rules didn't exist, but now point anything in docs/components/optimize to /optimize:
+RewriteRule ^docs/components/optimize/what-is-optimize(\/?)$ /optimize/what-is-optimize/ [R=301,L]
```
pepopowitz, line 310: Most of the redirects in this file have two rules per URL. In a few cases it's to handle child routes, but in the large majority _I think_ it's to handle two cases each: when the URL contains a trailing slash, and when it doesn't. 

These new rules only contain one line per URL, but the `(\/?)` at the end of each catches both cases -- trailing slash or no trailing slash. 

If I'm missing another reason for the dual rules, let me know!

(Here's an example of another rule in the file, which uses two sub-rules): 

```
RewriteRule ^docs/self-managed/optimize-deployment/setup/configuration(.*)$ /docs/self-managed/optimize-deployment/configuration/system-configuration [R=301,L]
RewriteRule ^docs/self-managed/optimize-deployment/setup/configuration$ /docs/self-managed/optimize-deployment/configuration/system-configuration [R=301,L]
```

akeller: > when the URL contains a trailing slash, and when it doesn't.

Nailed it. Child routes and trailing slashes. 

We had to go back and forth a bit on trailing vs. no trailing slashes. I think we've committed to trailing slashes at this point so if you have a good way to reduce these to ONE rule, I'm all for it.

akeller: Plus I'm terrible at anything regex or regex leaning.

akeller (approved):
I see nothing obvious with these. 

As a bonus, it would be great to add the `(\/?)` into our docs on docs.

christinaausley:
Thank you so much for executing these redirects. I'll plan to hold off on my "redirect clean up/pruning" until this is merged.

pepopowitz:
Merging this into the parent branch, then I'll deal with any rebases that I need to.

Issue #115, opened 2020-12-08
Automate screenshots


akeller:
For the long-term goal referenced in #110.

We should have an automated process for taking and updating screenshots. Let's use this to discuss relevant tools and methodologies.

christinaausley:
**Timeline**
- **By Nov 11**: Launch draft repository for Operate and Tasklist screenshots.
- **By Nov 18**: Standardize Zeebe screenshots based on approved proposal [here](https://docs.google.com/document/d/1gATpZXVxGiWEF4HfUbpUpQJ8ElyeZROERwiArPX5FRk/edit). Make necessary edits to Operate/Tasklist repo to have it fully functioning.
-  **By Dec 9**: Write out scripts for Operate/Tasklist screenshots.
- **By Dec 16**: Make any necessary final adjustments to Operate/Tasklist repo and screenshots.

akeller:
Thank you for the timeline @christinaausley! This looks great!

akeller:
@christinaausley in terms of the scope or this issue, is anything remaining? Do you have an updated timeline or do you want to create a new issue with a clearer scope about next (or remaining) steps?

PR #1269, opened 2022-09-15
Document inclusive gateway behavior changes

### LINKED ISSUE #1256: Document inclusive gateway behavior changes
We discovered an inconsistency between the inclusive gateway behavior in Zeebe and the BPMN spec, see:
- https://github.com/camunda/zeebe/issues/10310

Now that has been changed, we should also update the docs. Specifically, we should make the following changes:

- in the screenshots, the default flow's are labeled (somewhat indicating that there is a condition). We should remove the labels on the default flow
- it states that `If an inclusive gateway has multiple outgoing sequence flows, all sequence flows must have a condition to define when the flow is taken. Optionally, one of the sequence flows can be marked as the default flow.` which is no longer true. We should indicate that the default flow does not require a condition and that it would be ignored if it does have one
- the screenshot `inclusive-gateway-1` indicates that the condition of the default flow was evaluated. We need to change that
- the same is true for screenshot `inclusive-gateway-2`

korthout:

closes #1256 

This documents the changed behavior of the inclusive gateway:
- default flow is not expected to have a condition,
- and if a condition is still provided for the default flow then it is not evaluated.

This also clarifies the difference between flows with conditions and the default flow by only using names (aka labels) on the sequence flows with a condition.

New screenshots have been manually produced from Camunda Modeler and Operate.

Additionally, this introduces a new `media-pre-src` folder to store/share files used to produce media-sources like images. This should make it easier to make changes to change screenshots without having to remodel a process in the Camunda Modeler. It also paves the way towards
- #115 

Not that I'm aware of.

This feature (and the behavioral change) are part of Camunda Platform 8.1.

## PR Checklist


christinaausley (approved):
Nice. No changes on my end for this one.

PR #1270, opened 2022-09-15
fix?: attempt to fix mixpanel events

pepopowitz:

Mixpanel events haven't been sending since about the time that #1114 was published. This PR does two things: 

1. imports `mixpanel` because that line was removed in #1114 and shouldn't have been.
2. Adds some `console.log` statements for silently-caught errors, to help provide details when testing in production.

The big challenge with this is that the only way for us to test Mixpanel events is in production. (Maybe staging too? I'm not sure.) So in order to see if this fix works, we'll need to merge, and maybe also deploy to production. 

Immediately, so that I can test it in production.

## PR Checklist


akeller (approved):
Based on our conversation :shipit:

PR #1275, opened 2022-09-16
change io.zeebe to io.camunda

christinaausley:

Based on https://github.com/camunda/camunda-platform-docs/issues/1009. Renaming `io.zeebe` to `io.camunda` we seemed to rename the text but not the actual link, causing the links to send users to old versions of the Javadocs.

## PR Checklist

### DIFF
--- a/versioned_docs/version-1.1/components/clients/java-client/index.md
+++ b/versioned_docs/version-1.1/components/clients/java-client/index.md
@@ -65,7 +65,7 @@ Let's go over this code snippet line by line:
 3. Create the client by passing in the address of the cluster we want to connect to and the credentials provider from the step above.
 4. Send a test request to verify the connection was established.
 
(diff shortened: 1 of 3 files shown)


falko (approved):
The links seem to work.

A follow up feature request would be to link to the correct version based on which version of our docs one is browsing, e.g. https://javadoc.io/doc/io.camunda/zeebe-client-java/8.0.6/index.html

Knowing the latest patch version might be a little tricky, though. A workaround would be to go to the .0 releases only, e.g. 8.0.0, 8.1.0

8.1 and 8.2 are going to add new client features.

akeller (approved):
I'm good with this! 

And to @falko's comment, definitely a feature request on the version, not the patch. So when we cut a release, we could adjust the URLs to point to that version (8.0 points to 8.0, 8.1 points to 8.1, etc.). Since we don't cut patch releases and they just sit in `Next` this would be too much manual overhead unless we could automate (and auto PR) updates as the API version is released.

Issue #1217, opened 2022-08-31
Update `release-procedure.md`


akeller:
Documentation now supports continuously releasing components and components with two different versioning strategies or paths. We should update the docs for both release procedures:

* Simple GitHub release for patch version
* Release for C8 minor version
* [Future] Release for Optimize minor version

akeller:
The first two items were taken care of in #1281. 

@pepopowitz do you have a separate issue for your Optimize work or is it reflected in #1116?

pepopowitz:
Yes! 

> [Add documentation clarifying our multi-instance setup.](https://github.com/camunda/camunda-platform-docs/issues/1116)

pepopowitz:
But this is better defined & I'm going to link to it from #1116.

pepopowitz:
Sharing some findings about what it takes to do a release with our new multi-instance setup, for posterity but also in hopes of coming to a conclusion on one of them: 

## Things that need to happen to cut a new version

1. Edit `sidebars.js` and `optimize_sidebars.js` to remove `/next` from the helper at the top of each file. 
   - this is needed for docusaurus to generate the correct URLs in the new version's sidebar files. 
   - this is a temporary step and will not be committed.
   - depending on a discussion further down, this might be something we choose to automate.
2. Create new versions via docusaurus: 
   ```
   npm run docusaurus docs:version 8.1
   npm run docusaurus docs:version:optimize 3.9.0
   ```
3. Revert sidebars/optimize_sidebars changes from step 1
4. Update [`expandVersionedUrl`](https://github.com/camunda/camunda-platform-docs/blob/main/src/mdx/expandVersionedUrl.js) so that mappings are all correct
   - This is clunky. It's a lot to expect people to get these mappings all correct -- it's not just _adding new ones_, it's also editing existing ones. 
   - It also includes fixing tests that will be broken based on the mappings changing. 
   - I can (and will) make this better before the release. If we only have to maintain a list of mappings (e.g. `8.1:3.9.0`), it'll be easier, and we can just generate all the mappings in `expandVersionedUrl` from that. It would also remove the need to update tests.
5. Update previous version's sidebars.json files to add version number to the cross-instance URLs.
   - this is not fun. This is why I wrote the helper functions at the top of the sidebars.js files in the first place; unfortunately docusaurus not supporting .js files for previous versions bites us here. 
   - I could definitely write a regex to make this easier.
[...]

PR #1328, opened 2022-09-29
refactor: simplify version mapping logic

pepopowitz:
Part of #1217 

Refactors the version mapping logic (i.e. docs v8.0 <=> optimize v3.8.0) to simplify its external API. 

At time of cutting new versions, this logic would have required detailed changes that were prone to mistakes. With this refactor, updating this file involves nothing more than specifying the new docs & optimize versions. 

This PR also updates the tests associated with this logic based on the correct version mappings (I forgot to update them when I extracted v3.8.0 of Optimize!)

No

Should have no effect to production, but merging in the next couple days is probably best in terms of getting things ship-shape for the upcoming version release.

## PR Checklist


src/mdx/expandVersionedUrl.js
```diff
@@ -1,27 +1,71 @@
...
-  },
+// @ts-check
+
+/**
+ * @typedef {object} VersionMapping
+ * @property {string} docsVersion
+ * @property {string} optimizeVersion
+ */
+
+/** @type {Array<VersionMapping>} */
```
pepopowitz, line 9: These type definitions were useful for me when writing the code. They aren't actually checked in the build process because we don't have docusaurus configured for typescript (yet!), but in VS Code at least, they spit out errors when things aren't passing. 

I can remove them if we think they're noisy.

src/mdx/expandVersionedUrl.js
```diff
@@ -1,27 +1,71 @@
...
-      // TODO: introduce these mappings, as versions are added for optimize.
-      { match: "/docs/", expandTo: "/optimize/next" },
-      { match: "/versioned_docs/version-8.0/", expandTo: "/optimize" },
-      // { match: "/versioned_docs/version-1.3/", expandTo: "/optimize/3.7.0" },
-    ],
+    docsVersion: "8.0",
+    optimizeVersion: "3.8.0",
   },
+  { docsVersion: "1.3", optimizeVersion: "3.7.0" },
 ];
```
pepopowitz, line 17: This is the payoff of this PR. On the left side we would have had to align the source directories with URLs when adding a new version; on the right, all we have to do is add a new object that aligns the two versions.

src/mdx/expandVersionedUrl.js
```diff
@@ -1,27 +1,71 @@
...
+ * @typedef {object} TokenRule
+ * @property {string} match
+ * @property {string} expandTo
+ */
+
+/** @type {Token} */
+const docsToken = {
+  token: "$docs$",
+  rules: [
+    { match: "/optimize/", expandTo: "/docs/next" },
```
pepopowitz, line 35: I could add comments here if you think it'd be useful.

This first rule is for the "next" version, which will always have the same mapping....

src/mdx/expandVersionedUrl.js
```diff
@@ -1,27 +1,71 @@
...
+    {
+      match: `/optimize_versioned_docs/version-${currentVersions.optimizeVersion}`,
```
pepopowitz, line 37: ...This second rule maps the current version in the source to the non-versioned URL....

src/mdx/expandVersionedUrl.js
```diff
@@ -1,27 +1,71 @@
...
+      expandTo: "/docs",
+    },
+    ...remainingVersions.map((version) => {
```
pepopowitz, line 40: And for each of the remaining versions, both the source and target URLs have the version in them.

src/mdx/expandVersionedUrl.js
```diff
@@ -1,27 +1,71 @@
...
-    rules: [
-      // TODO: introduce these mappings, as versions are added for optimize.
-      { match: "/docs/", expandTo: "/optimize/next" },
-      { match: "/versioned_docs/version-8.0/", expandTo: "/optimize" },
-      // { match: "/versioned_docs/version-1.3/", expandTo: "/optimize/3.7.0" },
-    ],
+    docsVersion: "8.0",
+    optimizeVersion: "3.8.0",
   },
+  { docsVersion: "1.3", optimizeVersion: "3.7.0" },
```
pepopowitz, line 16: I haven't moved any docs for these versions yet, but having this mapping here shouldn't hurt anything (because no docs are using the `$docs$/$optimize$` shortcuts in v1.3 yet.)

src/mdx/expandVersionedUrl.spec.js
```diff
@@ -21,6 +21,7 @@ describe("expandVersionedUrl", () => {
         "/Users/monkeypants/camunda-platform-docs/optimize/what-is-optimize.md",
         "/docs/next/some/thing",
       ],
+
       [
         "/Users/monkeypants/camunda-platform-docs/optimize_versioned_docs/version-3.8.0/what-is-optimize.md",
```
pepopowitz, line 26: These tests *will* fail when we cut a new version. I think that's fine for now -- we don't actually run them as part of CI anyway, and it's less important to me that the tests are robust and more important that versioning is easier.

src/mdx/expandVersionedUrl.js
akeller: I don't have strong feelings about this, but please leave a brief comment in the code.

src/mdx/expandVersionedUrl.js
akeller: 🤩 so CLEAN

src/mdx/expandVersionedUrl.js
akeller: This seems clear enough not to need a comment.

src/mdx/expandVersionedUrl.js
```diff
@@ -1,27 +1,71 @@
...
+
+/**
+ * @typedef {object} VersionMapping
+ * @property {string} docsVersion
+ * @property {string} optimizeVersion
+ */
+
+/** @type {Array<VersionMapping>} */
+const versionMappings = [
+  // When cutting a new version, add a new mapping here!
```
akeller, line 11: 💯

akeller (dismissed):
This looks SO clean. I can't get over it! Thank you! 🚀

src/mdx/expandVersionedUrl.js
pepopowitz: Added in most recent commit!

akeller (approved):

PR #1345, opened 2022-10-06
[DO NOT MERGE]: POC: DocNavbarItems link across instances to the correct version

### LINKED ISSUE #1333: bug: non-Next versions of Optimize documentation link to incorrect version of main docs in header
Top-nav links (Guides, Components, etc) on all versions of Optimize docs link to the Next version of main documentation. This is correct for the Next version of Optimize docs, but incorrect for all other versions.

## Reproduction

Visit https://docs.camunda.io/optimize/apis-clients/optimize-api/optimize-api-authorization/

Expected result: top-nav links (Guides, Components, etc) go to **current** version of main documentation 
Actual result: top-nav links go to **Next** version of main documentation

## Implementation thoughts

I think this might be fixable in src/theme/NavbarItem/index.js, or another component inside NavbarItem. Like we did for the docsVersionDropdown in that file, we might need to alter some props we pass through to the link components when viewing optimize pages. 

cc @akeller for visibility

pepopowitz:

This is a POC PR, demonstrating how we _could_ fix #1333.

I say "could" because there is risk associated with this fix -- we're overriding multiple internal functions for this component, which is unsafe in the sense that docusaurus has no obligation to let us know when these things change/break. A future minor/patch version update of docusaurus could leave us with broken code, and/or we will likely spend time in the future rewriting our overrides based on the updated internals. 

I also don't know if we've determined how much of a problem #1333 really is - it certainly annoys me, but it might be an edge case that no one but me runs into.

Video demonstrating the problem, the source code responsible for it, and the fix: https://www.loom.com/share/ea910112df7643b4856ed60c93105157

### What's going wrong?

When browsing a document within Optimize, Docusaurus is trying to identify which version you want to see for the top nav links, but it's basing it on where you were within the main docs, instead of where you currently are. 

### How does this fix it?

This PR swizzles the DocNavbarItem component, so that we can write our own version of `useLayoutDoc`, so that we can write our own version of `useDocsVersionCandidates`. 

More detailed comments are in the code.

No

🤷 _should_ it go live? 

## PR Checklist

### DIFF
--- a/src/theme/NavbarItem/DocNavbarItem.js
+++ b/src/theme/NavbarItem/DocNavbarItem.js
@@ -9,38 +9,44 @@ import {
 } from "@docusaurus/theme-common/internal";
 import DefaultNavbarItem from "@theme/NavbarItem/DefaultNavbarItem";
 
-// Note that the original method is memoized, and we will likely want to do the same.
-//   (I can see logs from this function a dozen times per page load, due to it running for every nav link.)
-// Based on https://github.com/facebook/docusaurus/blob/abe545052693bfab98b347c14f3e00709bb2ac06/packages/docusaurus-theme-common/src/utils/docsUtils.tsx#L203-L216
-function useDocsVersionCandidatesAlt(docsPluginId, optimizeActiveDocContext) {
-  // 1. Let docusaurus determine the version candidates -- it orders them based on what it thinks the user wants.
-  //   Note that it's non-deterministic -- because it's based on what you've browsed recently.
-  const versions = useDocsVersionCandidates(docsPluginId);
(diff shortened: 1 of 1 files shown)


akeller (commented):
I don't love this because I don't like having to go into the guts of a platform to do anything. That said, if everything is "contained" to this single file, we know what to rip out if/when this breaks in the future. 

I want to drag my feet on this a bit and see if we get feedback first.

pepopowitz:
> I want to drag my feet on this a bit and see if we get feedback first.

I'm on board with that. I'll close this out and if we get feedback we can revisit.

PR #1367, opened 2022-10-10
chore: re-synchronize sidebars & optimize_sidebars in prep for v8.1

pepopowitz:

This PR regenerates the cross-instance sections in sidebars.js and optimize_sidebars.js, in preparation for cutting v8.1. 

#1116 introduced the need to keep these two files synchronized, but (a) we forgot to document that it was needed, and (b) it was easy to miss this synchronization in PRs. Especially with a flurry of new docs leading up to this release, the two files drifted. 

In the future I'd like to make it impossible to edit one without the other.

Note that I did not regenerate anything for v8.0 and previous versions -- those might also have drifted, but I think that most (if not all) of the new docs added recently were in vNext. 

no

Either before or at the same time as v8.1.

## PR Checklist

### DIFF
--- a/sidebars.js
+++ b/sidebars.js
@@ -263,7 +263,7 @@ module.exports = {
         },
       ],
       Optimize: [
-        optimizeLink("What is Optimize?", "components/what-is-optimize"),
+        optimizeLink("What is Optimize?", "components/what-is-optimize/"),
         {
           "User guide": [
             optimizeLink(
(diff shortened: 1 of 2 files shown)


sidebars.js
```diff
@@ -263,7 +263,7 @@ module.exports = {
         },
       ],
       Optimize: [
-        optimizeLink("What is Optimize?", "components/what-is-optimize"),
+        optimizeLink("What is Optimize?", "components/what-is-optimize/"),
```
pepopowitz, line 266: All of the changes in this file are adding slashes to the end of paths. This is unnecessary, because they go the right place even without the slash, but I included them because the generation script includes them.

akeller (approved):

akeller:
This wasn't as bad as I was expecting 🙌

pepopowitz:
> This wasn't as bad as I was expecting 🙌

Issue #1240, opened 2022-09-07
sync-board job fails for forked PRs


pepopowitz:
The `sync-board` job fails for forked PRs with the following message:

```
gh: To use GitHub CLI in a GitHub Actions workflow, set the GH_TOKEN environment variable. Example:
  env:
    GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

Example: https://github.com/camunda/camunda-platform-docs/runs/8213953649?check_suite_focus=true

PR #1371, opened 2022-10-11
feat(modeler/bpmn): document candidate users for Assignments

### LINKED ISSUE #1370: I can set candidate users on a user task
As a user I would like to specify task candidate-users for user tasks. (issue: https://github.com/camunda/zeebe/issues/10346, pr: https://github.com/camunda/zeebe/pull/10459).

lzgabel:
As a user I would like to specify task candidate-users for user tasks.

closes #1370 

No.

No need to hurry. Maybe 8.2?

## PR Checklist

- [x] My changes require an [Engineering review](https://github.com/camunda/camunda-platform-docs/blob/main/howtos/documentation-guidelines.md#review-process), and I've assigned @korthout as a reviewer.

### DIFF
--- a/docs/components/modeler/bpmn/user-tasks/user-tasks.md
+++ b/docs/components/modeler/bpmn/user-tasks/user-tasks.md
@@ -38,7 +38,7 @@ attributes can be specified simultaneously:
 - `candidateUsers`: Specifies the users that the task can be assigned to.
 - `candidateGroups`: Specifies the groups of users that the task can be assigned to.
 
-Typically, the assignee, candidate users and candidate groups are defined as static values (e.g. `some_username`, `some_username, another_username` and
+Typically, the assignee, candidate users, and candidate groups are defined as static values (e.g. `some_username`, `some_username, another_username` and
 `sales, operations`), but they can also be defined as
 [expressions](/components/concepts/expressions.md) (e.g. `= book.author` and `= remove(reviewers, book.author)` and `= reviewer_roles`). The expressions are evaluated on activating the user task and must result in a
 `string` for the assignee and a `list of strings` for the candidate users and a `list of strings` for the candidate groups.
--- a/docs/components/modeler/bpmn/user-tasks/user-tasks.md
+++ b/docs/components/modeler/bpmn/user-tasks/user-tasks.md
@@ -35,12 +35,13 @@ This can be used to define which user the task can be assigned to. One or both o
 attributes can be specified simultaneously:
 
 - `assignee`: Specifies the user assigned to the task. [Tasklist](/components/tasklist/introduction-to-tasklist.md) will claim the task for this user.
+- `candidateUsers`: Specifies the users that the task can be assigned to.
 - `candidateGroups`: Specifies the groups of users that the task can be assigned to.
 
-Typically, the assignee and candidate groups are defined as static values (e.g. `some_username` and
+Typically, the assignee, candidate users and candidate groups are defined as static values (e.g. `some_username` and
 `sales, operations`), but they can also be defined as
-[expressions](/components/concepts/expressions.md) (e.g. `= book.author` and `= remove(reviewers, book.author)`). The expressions are evaluated on activating the user task and must result in a
-`string` for the assignee and a `list of strings` for the candidate groups.
+[expressions](/components/concepts/expressions.md) (e.g. `= book.author` and `= reviewers` and `= remove(reviewers, book.author)`). The expressions are evaluated on activating the user task and must result in a
+`string` for the assignee and a `list of strings` for the candidate users and a `list of strings` for the candidate groups.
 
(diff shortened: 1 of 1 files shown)


lzgabel:
Hi @korthout, @christinaausley. Please take a moment to check this out. ❤️

docs/components/modeler/bpmn/user-tasks/user-tasks.md
```diff
@@ -35,12 +35,13 @@ This can be used to define which user the task can be assigned to. One or both o
 attributes can be specified simultaneously:
 
 - `assignee`: Specifies the user assigned to the task. [Tasklist](/components/tasklist/introduction-to-tasklist.md) will claim the task for this user.
+- `candidateUsers`: Specifies the users that the task can be assigned to.
 - `candidateGroups`: Specifies the groups of users that the task can be assigned to.
 
-Typically, the assignee and candidate groups are defined as static values (e.g. `some_username` and
+Typically, the assignee, candidate users and candidate groups are defined as static values (e.g. `some_username` and
```
korthout, line 41: 🔧 The examples of the static values were mentioned respectively of the different properties (first the assignee, then the candidate groups). Now that 'candidate users' is placed in between, we should add an additional example in between.

Something like:
> (e.g. `some_username`, `some_username, another_username`, and `sales, operations`)

docs/components/modeler/bpmn/user-tasks/user-tasks.md
```diff
@@ -35,12 +35,13 @@ This can be used to define which user the task can be assigned to. One or both o
...
 `sales, operations`), but they can also be defined as
-[expressions](/components/concepts/expressions.md) (e.g. `= book.author` and `= remove(reviewers, book.author)`). The expressions are evaluated on activating the user task and must result in a
-`string` for the assignee and a `list of strings` for the candidate groups.
+[expressions](/components/concepts/expressions.md) (e.g. `= book.author` and `= reviewers` and `= remove(reviewers, book.author)`). The expressions are evaluated on activating the user task and must result in a
```
korthout, line 43: I notice now that the example for a 'candidate groups' expression does not make sense semantically. Actually, it seemed to express 'candidate users' much better.

🔧 I propose to use the existing `= remove(reviewers, book.author)` as 'candidate users' expression and add a simple expression for the 'candidate groups' instead. For example, `= reviewer_roles` (sounds like a variable name that contains a list of different candidate groups).

korthout (dismissed):
Thanks @lzgabel 🚀 

LGTM 👍  But please consider my suggestions. Feel free to ignore them if you don't agree.

lzgabel:
Hi @korthout. I'm sorry to interrupt your review task. :bow:

docs/components/modeler/bpmn/user-tasks/user-tasks.md
```diff
@@ -35,12 +35,13 @@ This can be used to define which user the task can be assigned to. One or both o
 attributes can be specified simultaneously:
 
 - `assignee`: Specifies the user assigned to the task. [Tasklist](/components/tasklist/introduction-to-tasklist.md) will claim the task for this user.
+- `candidateUsers`: Specifies the users that the task can be assigned to.
 - `candidateGroups`: Specifies the groups of users that the task can be assigned to.
 
-Typically, the assignee and candidate groups are defined as static values (e.g. `some_username` and
+Typically, the assignee, candidate users and candidate groups are defined as static values (e.g. `some_username`, `some_username, another_username` and
```
christinaausley, line 41: Just one minor comment/nit pick here -- need an Oxford comma, so:

"Typically, the assignee, candidate users**,** and candidate groups are defined as static values..."

christinaausley (approved):
Just one non-blocking comment and minor fix 👍

korthout:
The `sync-board` workflow fails with a message about the `GITHUB_TOKEN`. I think this is because this PR comes from a fork. 

Looking at [the workflow itself, it runs on `pull_request`](https://github.com/camunda/camunda-platform-docs/actions/runs/3239658530/workflow#L6). If this workflow is safe to run for untrusted code (i.e. PRs from forks), then we could change this to [`pull_request_target`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target). Note that this changes the context in which the workflow is executed.

@pepopowitz Do you think this is a good solution?

christinaausley (approved):

pepopowitz:
> Is this workflow blocking the merge? 

No! It is a housekeeping type workflow, and its failure shouldn't be considered blocking. Also the squash/merge button looks enabled to me, so I think it's also non-blocking from GitHub's perspective.

> And do you think this could be a good solution?

We have this issue tracked at #1240, and hopefully will get to it soon! It doesn't happen often, but it's friction that shouldn't be there. My understanding of how GitHub handles tokens on forked PRs is that we can't grant forked PRs the access they need to make changes to a GitHub project from a fork; but reading more about `pull_request_target` I'm not so sure my understanding is correct! I will experiment with this, thanks for the tip.

PR #1391, opened 2022-10-14
Update Guide changes related to 8.1.1 import issue

johanwelgemoed:

Updated warning text in Update Guide to now point users to 8.1.1 version to use when updating from 8.0


menski (approved):
LGTM

PR #1394, opened 2022-10-17
docs(identity): add step to set keycloak realm when connecting to an existing keycloak instance

Ben-Sheppard:
Whilst working recently with @mschoe I realised that the documentation around connecting to an existing Keycloak instance was a little unclear with regards to setting the realm variable, this variable is an important consideration for this task and could easily cause confusion if not mentioned.

Given this, I have decided to add a step specifically to handle the realm setting as it should be a part of the defined process.

None :) 

This functionality is already live therefore this can be released whenever

## PR Checklist


dlavrenuek (dismissed):
👍

christinaausley (approved):
Nice 👍  I just added a commit with a couple changes, and removed space in between numerical list of steps so we wouldn't run into issues where it accidentally goes from step 12 to step 1, etc.

PR #1398, opened 2022-10-17
adjust modeler references

christinaausley:

Based on [this conversation](https://github.com/camunda/camunda-platform-docs/pull/1356#discussion_r996423390), Desktop Modeler, Web Modeler, and Camunda Modeler should be referred to as such, without "the" article.


akeller (approved):
@christinaausley this is a good item to surface with the Modeling teams and remind them to not use "the".

PR #1400, opened 2022-10-17
(retro) technical review web modeler self managed

christinaausley:

Based on documentation added as part of https://github.com/camunda/camunda-platform-docs/pull/1356 -- **these are changes being made retroactively on an already-merged PR** as I was on FTO during the original request for review.


akeller (changes requested):
Thank you for going back and reviewing this content @christinaausley. 

Can you also update the versioned docs for 8.1?

akeller (dismissed):

akeller (approved):

PR #1421, opened 2022-10-21
feat: add ability to run local Apache server to test htaccess redirects

pepopowitz:

Adds docker configuration & documentation for running a local Apache server so that .htaccess rules can be tested in an environment that simulates production.

### Why? 

In #1386, I wrote some rewrite rules that passed on https://htaccess.madewithlove.com/ but failed during CI checks. I spent a lot of time trying to figure out why they weren't passing CI, and if I'd been able to spin up a local server I'd have discovered the problem sooner. 

And in fact, the way I _did_ figure out what was wrong with those rules was to spin up a local Apache server with Docker. So this PR is really just capturing what I went through.

## PR Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,19 @@
+# For local development purposes only
+
+FROM httpd:alpine
+
+RUN rm -r /usr/local/apache2/htdocs/*
+COPY /build /usr/local/apache2/htdocs/
+
+# Remove SSL requirement
+RUN sed -i \
+        '/SERVER_PORT/d;/SERVER_NAME/d' \
+        /usr/local/apache2/htdocs/.htaccess
+
+# Enable the rewrite module
+RUN sed -i \
+        -e 's/^#\(LoadModule .*mod_rewrite.so\)/\1/' \
+        /usr/local/apache2/conf/httpd.conf
+
+# Enable rewrites
+RUN sed -i '/<Directory "\/usr\/local\/apache2\/htdocs">/,/<\/Directory>/ s/AllowOverride None/AllowOverride All/' /usr/local/apache2/conf/httpd.conf
--- a/howtos/documentation-guidelines.md
+++ b/howtos/documentation-guidelines.md
@@ -125,8 +125,7 @@ When linking internally from one document to another, follow these guidelines:
 3. If the doc is [in one of the shared sections](#synchronization-of-sidebars), update [the opposite instance's corresponding sidebars file(s)](#sidebar-navigation).
 4. Add necessary redirect/rewrite rules to the top of `.htaccess`.
 
-> **Note**
-> The redirects/rewrite rules added to `.htaccess` will not work when running the documentation locally. You can use online tooling to help with this (e.g. https://htaccess.madewithlove.com/).
+See [Redirect rules](#redirect-rules) for information on testing `.htaccess` rules.
 
 ## Remove an existing page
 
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,19 @@
+# For local development purposes only
```
pepopowitz, line 1: Is this comment obvious and dumb to include? I added it because there is a line in here that removes the requirement for https, and I don't want anyone to think they can use this dockerfile to build a production server without removing that.

Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+
+FROM httpd:alpine
+
+RUN rm -r /usr/local/apache2/htdocs/*
+COPY /build /usr/local/apache2/htdocs/
```
pepopowitz, line 6: I made a decision here that I'm not 100% confident on. 

This is copying what's in your build folder and running the server from that copy. 

I could have instead made it run with a mapped volume, so that it is actually running off of your build folder directly. In which case you could change things in there, and you'd see them in the running instance. 

I chose to do the copy instead of the mapped volume for two reasons: 

1. Part of the configuration is removing the SSL requirement, which means editing the .htaccess file, and I don't like the idea of editing your build folder directly. 
2. I _think_ I'd need to restart Apache on the running instance after making changes to .htaccess. That isn't itself a reason not to choose this method, but it makes this method less appealing.

howtos/documentation-guidelines.md
```diff
@@ -135,8 +134,25 @@ When linking internally from one document to another, follow these guidelines:
...
+   Tools like https://htaccess.madewithlove.com/ apply a set of redirect rules to a specific URL. Note that there are edge cases where this tool doesn't give the same results as a published environment.
+
+2. Use `docker compose` to spin up a locally-running Apache webserver.
+   This repo includes Docker configuration ([Dockerfile](../Dockerfile) and [docker-compose.yml](../docker-compose.yml)) to spin up a local environment that better simulates a published environment. Redirect rules can then be tested directly in a browser.
+
+   The local server is based on the contents of your `./build` folder. To start this local server:
+
+   1. Build the docs with `npm run build`.
+   2. Start the server with `docker compose up`.
+   3. Browse http://localhost:3001 and test redirects.
```
pepopowitz, line 155: Two other bits I considered mentioning: 

1. For Docker newbs like me, `docker compose down` will clean all this stuff up. But that feels like Docker guidance and not part of _our_ docs.
2. Browsers hold onto 301 redirects tightly, and it can be tricky to figure out if a redirect is happening because that's what the rule says now or that's what the rule said the last time you ran it. 

Let me know if you think either of those are worth including.

Dockerfile
pepopowitz: Oh, and I should add that the down-side to running from a copy instead of the build folder directly is that if I want to make changes to my htaccess and see it reflected in the running instance, I need to rebuild with Docker, which makes it a little tedious. 

But with having to restart Apache after making changes when running directly from build, I think it's a bit tedious either way.

Dockerfile
akeller: I appreciate forward-looking CYA and don't consider this dumb at all.

Dockerfile
akeller: I don't think you can get out of rebuilding.  I'm ultimately unsure about restarting Apache. What is your goal here? Ideally, avoid rebuilding, restarting, and modifying the build folder? Or maybe minimize tediousness? 

In our build workflow, we do the following steps (which you may also not like, but have tolerated, knowingly or unknowingly):

```
- name: Build
   run: npm run build
   env:
        NODE_OPTIONS: --max_old_space_size=4096
- name: Remove https redirect
   run: |
        sed -i '/SERVER_PORT/d;/SERVER_NAME/d' build/.htaccess
- name: Run web server
    run: |
       docker run -d --name webserver -v "$PWD/build":/app -p 8888:8080 bitnami/apache:latest
       echo "container_ip=$(docker inspect -f '{{range.NetworkSettings.Networks}}{{.IPAddress}}{{end}}' webserver)" >> $GITHUB_ENV
```

So that's build > _modify_ build file > run web server. 

We are running into a situation where compromising (and documenting said compromise) is the best thing to do.

@menski thoughts here?

howtos/documentation-guidelines.md
akeller: Given that we'll likely not have everyone working in Docker regularly (or regularly with docs), we should add it to save ourselves later.

For 2, is the guidance to do a hard reload, or something equivalent, in the browser?

Dockerfile
akeller: I may also not fully understand this because of my mediocre understanding of Apache.

pepopowitz: I'm thinking specifically of this feedback loop, which is the reason one would want to run this server locally:

* .htaccess redirects are failing in CI but we can't figure out why
* I start up the server locally
* I enter a URL in the address bar that I think should redirect
* It goes to the wrong place
* I want to edit my .htaccess to make an adjustment
* I enter the URL in the address bar 
* ➿ 

It'd be great if there was nothing between editing the .htaccess and entering something in the address bar...but I think regardless of which of the methods I described above, there's a step where I have to bounce something before I can see the changes.

pepopowitz: > In our build workflow, we do the following steps (which you may also not like, but have tolerated, knowingly or unknowingly):

wowww I feel like I must have known that, but somehow have not noticed or remembered it.

howtos/documentation-guidelines.md
pepopowitz: my preference is to only test things in an incognito window, because then I don't have to clear anything, but honestly I am usually confused about what I have to clear to get rid of 301 redirects so I don't have good guidance.

pepopowitz: Both these items were added in my latest commit.

Dockerfile
akeller: Do you have an issue in the backlog for the follow-up work?

akeller (approved):
Note - I did not step through this and run it myself.

PR #1424, opened 2022-10-21
(retro) process instance modification review

christinaausley:

Retroactive technical review on changes merged in https://github.com/camunda/camunda-platform-docs/pull/1330 as I was OOO during the original request for review.


huygur (approved):
thanks for the fixes 👍

PR #1493, opened 2022-11-11
Explicitly mention variable creation behaviour

jonathanlukas:
When variables are coming with a start-event of an event-subprocess, the variable creation works in the default way. However, this might be confusing. So we should actually mention it.

Give a better insight on engine behaviour

no

asap

## PR Checklist

### DIFF
--- a/docs/components/modeler/bpmn/event-subprocesses/event-subprocesses.md
+++ b/docs/components/modeler/bpmn/event-subprocesses/event-subprocesses.md
@@ -26,7 +26,7 @@ If an event subprocess is triggered, its containing scope is not completed until
 
 Unlike a boundary event, an event subprocess is inside the scope. Therefore, it can access and modify all local variables of its containing scope. This is not possible with a boundary event because a boundary event is outside of the scope.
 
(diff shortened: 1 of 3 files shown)


christinaausley:
@jonathanlukas Do you want to apply these changes to only the `next` version of our docs or also the current and `8.0` versions?

christinaausley (dismissed):

jonathanlukas:
@christinaausley I am not sure yet. This is why I converted this into a draft. Engineering is still wrapping their heads around this topic.

saig0 (dismissed):
@jonathanlukas thank you for improving the docs. :cookie: 

I'm okay with the changes. :+1: 

I see more potential in improving this documentation. For example, mention explicitly that it's a good practice to use input mappings for non-interrupting event subprocesses to avoid overwriting of variables. But we can improve later.

jonathanlukas:
@christinaausley as this behaviour existed since forever, I would also apply this change to `8.0` and `8.1`.

christinaausley:
@jonathanlukas Are you ready to merge this? Happy to do so but just wanted to double check 👍

jonathanlukas:
Yeah sure, I was not sure whether the docs team pushes the button or if I can do that as well :)

PR #1511, opened 2022-11-16
Updated cluster sizing to latest SaaS

berndruecker:

We adjusted our cluster size in SaaS due to learnings - this change brings those updates to this page.

No

no specific date needed, can go live asap


pepopowitz (dismissed):
👍 

If these changes should also apply to future versions, they should also be introduced to the `/docs/` folder.

akeller (approved):

PR #1545, opened 2022-11-29
docs(go-client): add missing task header detail and refer to Operate instead of Zeebe Monitor

tstegemann:

1. Went through the "Go client - Getting started guide" and in the end I noticed that my result looked slightly different as one  value was missing in my output. That value is there in the screenshots in the docs, however it is nowhere mentioned to (and how to) set it. Added that detail in all the relevant versions

missing-task-header

2. As of now in the section https://docs.camunda.io/docs/next/apis-clients/go-client/go-get-started/#see-the-process-in-action it is mentioned to use Operate. Later in the text there still was a reference to "Zeebe Monitor" which was never mentioned before in the guide and should be replaced with Operate, especially as it already was introduced before in the guide.

No.

Not urgent. As soon as possible.

## PR Checklist


christinaausley (approved):
No changes here 👍

Issue #1171, opened 2022-08-17
Add regression testing


pepopowitz:
We should add some level of regression testing. 

* #1087 suggests that, assuming we provided clear feedback about why regression tests are failing, engineers would not be strongly opposed to it.
* #1101 illustrates that some level of regression testing could help prevent bugs from impacting main functions of the site (in that example, website navigation)
* https://github.com/camunda/camunda-platform-docs/pull/1087#issuecomment-1196062011 illustrates a use case for testing popular content

My personal takeaways from #1087:

* People aren't begging for regression testing, but they also don't seem opposed to adding it as part of the workflow
* It seems like a straightforward implementation of tests in CI, without a fancy service, would at least initially meet our needs

akeller:
Moving to ready for Q4.

pepopowitz:
Note for myself: [trigger another workflow upon completion success](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-a-workflow-based-on-the-conclusion-of-another-workflow). (If we decide to trigger the regression testing workflow to run only after a production deploy.)

akeller:
@pepopowitz thinking about our focus in Q1 - can you provide a status update for this card and then move it to the most appropriate column?

PR #1547, opened 2022-11-29
feat: initialize & configure playwright, and write a single regression/smoke test

pepopowitz:

Part of #1171.

Installs & configures playwright, and introduces a single smoke test against the production site. 

For now the test would not be triggered by anything; but I'm PR'ing it in this state so that I can set this work aside for now. It's progress, and I don't want to throw it away.

### What it looks like when I run the test locally

image

## Follow-up work (not included in this PR)

* Schedule test(s) to run after a production deploy
* Write more tests to cover things described in #1171 
* Get the URL from an environment variable?

## PR Checklist

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -1,0 +1,108 @@
+import type { PlaywrightTestConfig } from "@playwright/test";
+import { devices } from "@playwright/test";
+
+/**
+ * Read environment variables from file.
+ * https://github.com/motdotla/dotenv
+ */
+// require('dotenv').config();
+
+/**
+ * See https://playwright.dev/docs/test-configuration.
+ */
+const config: PlaywrightTestConfig = {
+  testDir: "./spec",
+  /* Maximum time one test can run for. */
+  timeout: 30 * 1000,
+  expect: {
+    /**
+     * Maximum time expect() should wait for the condition to be met.
+     * For example in `await expect(locator).toHaveText();`
+     */
+    timeout: 5000,
+  },
+  /* Run tests in files in parallel */
+  fullyParallel: true,
+  /* Fail the build on CI if you accidentally left test.only in the source code. */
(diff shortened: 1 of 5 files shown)


akeller:
Can you use the `hold` label?

pepopowitz:
Heads up @akeller that this is ready for actual review.

playwright.config.ts
```diff
@@ -0,0 +1,108 @@
+import type { PlaywrightTestConfig } from "@playwright/test";
+import { devices } from "@playwright/test";
+
+/**
+ * Read environment variables from file.
+ * https://github.com/motdotla/dotenv
+ */
+// require('dotenv').config();
```
akeller, line 8: Will you use a config file in the future, is that why this is here?

> Get the URL from an environment variable?

Anything else we would want to pop in there?

spec/regression/smoke.spec.ts
```diff
@@ -0,0 +1,12 @@
+import { test, expect } from "@playwright/test";
+
+test("homepage has title and links to intro page", async ({ page }) => {
+  await page.goto("/");
+
+  await expect(page).toHaveTitle(
+    /Camunda Platform 8 Docs \| Camunda Platform 8 Docs/
```
akeller, line 7: Now I know why you were mentioning this 😆

akeller (approved):
Nothing terribly interesting in my review comments. I like this and think it makes sense to move it forward.

playwright.config.ts
pepopowitz: > is that why this is here?

wellllll it _was put there_ by playwright when I initialized it....I _left it there_ because it seemed useful, if we decide to use it.

akeller: Sounds good. Just remember to comment it back in when you try to use it later 😅

PR #1605, opened 2023-01-05
(WIP) Move project items to done when their PR merges

### LINKED ISSUE #1539: Automagically move items to "Done" on Developer Experience board when merged
Most items move manually across the [Developer Experience board](https://github.com/orgs/camunda/projects/27).

If a PR is merged or an issue is closed, the item should appear in the "Done" column. 

## Implementation Suggestions

- See [these GitHub docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-when-a-pull-request-merges) on how to trigger a workflow when a PR is merged.
- ~[This GitHub doc](https://docs.github.com/en/actions/managing-issues-and-pull-requests/moving-assigned-issues-on-project-boards) _might_ describe an existing action we can use to move the card over, although when we originally set up [the sync-board workflow](https://github.com/camunda/camunda-platform-docs/blob/main/.github/workflows/sync-board.yaml) we didn't use this, and I don't remember why.~ 
   - @christinaausley and I dug into this a bit to see if we could get a quick win, and we were reminded why we can't use this workflow: 
      - It [looks for a project based on the current repository or the repository's owner](https://github.com/alex-page/github-project-automation-plus/blob/main/src/generate-mutation-query.js#L14-L18), and the Developer Experience project does not exist in either of those places. See screenshot for graphql query that simulates what the code in the workflow is trying to find (and note that neither of the collapsed results are the Developer Experience project):
      image
      - This results in the error `Could not find the column "✅ Done" or project "Developer Experience"` when [running the workflow](https://github.com/camunda/camunda-platform-docs/actions/runs/3743622327/jobs/6356000154).
- If we need to write the action ourselves, see [the sync-board workflow](https://github.com/camunda/camunda-platform-docs/blob/main/.github/workflows/sync-board.yaml) for hints on how to update the status on a board item. More specifically:
   - [Define the org and project](https://github.com/camunda/camunda-platform-docs/blob/main/.github/workflows/sync-board.yaml#L17-18)
   - [Get the project, status field, and status IDs](https://github.com/camunda/camunda-platform-docs/blob/main/.github/workflows/sync-board.yaml#L20-L43)
   - [This part](https://github.com/camunda/camunda-platform-docs/blob/main/.github/workflows/sync-board.yaml#L49-L58) will need to be rewritten - in sync-board it returns the item ID from a mutation; we'll need to _query_ for the item, instead. See specific query below.
   - [Set the status field](https://github.com/camunda/camunda-platform-docs/blob/main/.github/workflows/sync-board.yaml#L63-L82).

### Query for getting the item ID of a PR within a project

`$prID` can be retrieved [from the workflow environment](https://github.com/camunda/camunda-platform-docs/blob/main/.github/workflows/sync-board.yaml#L47) with `github.event.pull_request.node_id`

```graphql
query pr($prID: ID!) {
  node(id: $prID) {
    ... on PullRequest {
      title
      author {
        login
      }
      projectItems(first: 10) {
        nodes {
          id
          project {
            number
          }
          fieldValues(first: 10) {
            nodes {
              ... on ProjectV2ItemFieldSingleSelectValue {
                optionId
                name
              }
            }
          }
        }
      }
    }
  }
}
```

An example response: 

```json
{
  "data": {
    "node": {
      "title": "[WIP] feat: set up regression tests",
      "author": {
        "login": "pepopowitz"
      },
      "projectItems": {
        "nodes": [
          {
            "id": "PVTI_lADOACVKPs4ADm-7zgDqwE0",
            "project": {
              "number": 27
            },
            "fieldValues": {
              "nodes": [
                {},
                {},
                {},
                {},
                {
                  "optionId": "e34d6107",
                  "name": "🏗 In progress"
                }
              ]
            }
          }
        ]
      }
    }
  }
}
```

The value we'll want to capture from the response is `data.node.projectItems.nodes[].id`.

To be cautious, we can confirm that it's the `data.node.projectItems.nodes[]` with `project.number === 27` (which is the number of the DX board project). The sync-board job defines this value as an environment variable.

pepopowitz:
collaboration with @christinaausley 

WIP: attempting to resolve #1539 by building a workflow that uses the GraphQL API to move project items to done.

_Briefly describe the change you are making, this helps the reviewer by providing an overview._

## PR Checklist

### DIFF
--- a/.github/workflows/move-to-done.yaml
+++ b/.github/workflows/move-to-done.yaml
@@ -1,0 +1,84 @@
+name: move-to-done
+
+# FYI secrets.ADD_TO_PROJECT_PAT is a personal access token belonging to @akeller.
+
+on:
+  pull_request:
+    # sjh: `labeled`/`unlabeled` for now! but when this is ready to go, change to `closed`.
+    types: [labeled, unlabeled]
+
+jobs:
+  move-to-done:
+    # sjh: uncomment this if when we change the trigger to `closed`
+    # if: github.event.pull_request.merged == true
+    runs-on: ubuntu-latest
+    steps:
+      - name: Get project IDs for later steps
+        env:
+          GITHUB_TOKEN: ${{ secrets.ADD_TO_PROJECT_PAT }}
+          ORGANIZATION: camunda
+          PROJECT_NUMBER: 27
+        run: |
+          gh api graphql -f query='
+            query($org: String!, $number: Int!) {
+              organization(login: $org){
+                projectV2(number: $number) {
+                  id
+                  fields(first:20) {
+                    nodes {
+                      ... on ProjectV2SingleSelectField {
+                        id
+                        name
+                        options {
+                          id
+                          name
+                        }
+                      }
+                    }
+                  }
+                }
+              }
+            }' -f org=$ORGANIZATION -F number=$PROJECT_NUMBER > project_data.json
+
+          echo 'PROJECT_ID='$(jq '.data.organization.projectV2.id' project_data.json) >> $GITHUB_ENV
+          echo 'STATUS_FIELD_ID='$(jq '.data.organization.projectV2.fields.nodes[] | select(.name== "Status") | .id' project_data.json) >> $GITHUB_ENV
+          echo 'DONE_OPTION_ID='$(jq '.data.organization.projectV2.fields.nodes[] | select(.name== "Status") | .options[] | select(.name=="✅ Done") |.id' project_data.json) >> $GITHUB_ENV
+      # - name: Add PR to project
+      #   env:
+      #     GITHUB_TOKEN: ${{ secrets.ADD_TO_PROJECT_PAT }}
+      #     PR_ID: ${{ github.event.pull_request.node_id }}
+      #   run: |
+      #     item_id="$( gh api graphql -f query='
+      #       mutation($project:ID!, $pr:ID!) {
+      #         addProjectV2ItemById(input: {projectId: $project, contentId: $pr}) {
+      #           item {
+      #             id
+      #           }
+      #         }
+      #       }' -f project=$PROJECT_ID -f pr=$PR_ID --jq '.data.addProjectV2ItemById.item.id')"
+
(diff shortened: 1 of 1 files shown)


.github/workflows/move-to-done.yaml
```diff
@@ -0,0 +1,80 @@
+name: move-to-done
```
pepopowitz, line 1: We initially thought about combining this with the `sync-board` workflow, since that is nicely generically named, but it would have basically been two completely different workflows in one file. 

If you're interested in us renaming the `sync-board` workflow to be _less_ generic, since it can't be used as universally as we initially thought, let us know!

.github/workflows/move-to-done.yaml
```diff
@@ -0,0 +1,80 @@
...
+
+# FYI secrets.ADD_TO_PROJECT_PAT is a personal access token belonging to @akeller.
+
+on:
+  pull_request:
+    types: [closed]
+
+jobs:
+  move-to-done:
+    if: github.event.pull_request.merged == true
```
pepopowitz, line 11: This prevents us from running this action against PRs that are `closed` instead of `merged`. Is that correct? Should we move the closed ones to Done also?

.github/workflows/move-to-done.yaml
akeller: I think renaming makes a ton of sense here. Do you have any suggestions? `sync-review`? Maybe @christinaausley has a good idea.

.github/workflows/move-to-done.yaml
akeller: Good question. I move things to `Done` when there is no remaining action for that item. With that thinking, `closed` should be moved to `Done` as well. That said, we don't have many that are closed these days other than things you happen to be testing. In that sense, I would say we can leave this as a todo, or if it's relatively quick, make the change in this PR. Up to you!

akeller (approved):
> When we finished testing, we changed the trigger so that it will only execute on PRs that have just been merged.

Neat!

I left some comment on the inline comments (thank for those @pepopowitz!) but nothing is blocking.

PR #1610, opened 2023-01-09
Updated sizing recommendation and relaxed binding to SaaS

berndruecker:

self-managed and SaaS figures differed because of various updates. As more updates are expected to occur over time, I relaxed the binding and no longer say the self-manged example is exactly "S". Also I removed the sum, as it makes adjusting the table much easier - waiting for push back on this.

no

whenever / asap


Sijoma (dismissed):
LGTM. Although I would stick with the old requested values.

Sijoma (approved):

Issue #1659, opened 2023-01-31
Spike: do we have enough info in a PR's workflow to build only a portion of the c8 docs?


pepopowitz:
Related to https://github.com/camunda/product-hub/issues/499

The vast majority of `build-docs` workflow executions are for PRs running checks before GitHub decides if they're mergeable. And the vast majority of those PRs are modifying no more than 1 to 3 versions of the documentation.

We should research if it's feasible to use the `build-docs` workflow context to decide which versions are affected by a PR, and modify the docusaurus config to only include those versions before building. If it's possible, this would allow us to merge PRs sooner. It would also likely be significantly less work than archiving old versions of the docs completely.

*Timebox*: a half day
*Output of this spike*: details added to https://github.com/camunda/product-hub/issues/499 describing the feasibility of reducing the scope of a `build-docs` build by excluding versions not changed in the PR.

## Concerns/Questions I'd like to answer

- ✅ ❌ do we have access to file paths changed in the workflow context? This would be required to determine which versions are affected.
   - TL;DR: we can get them all for `pull_request` trigger; I haven't figured out how to get them all for `push` trigger. But we might be able to remove the `push` trigger entirely to simplify the workflow _and_ unlock build time improvements.
   - See #1673 for full details.
- ✅ what about link-checking? It runs in each `build-docs` workflow, and checks _all_ production URLs.
   - TL;DR we can limit link-checking to a subset of versions in a way that it will still pass. 
   - See #1675 for full details, including the answers to most of these questions:
      - does the build for a PR with only one version fail on link-checking? 
      - can we include only a subset of URLs in the link-checking, based on the built versions?
        - link check against only the recent X versions?
        - or current version and next?
        - Only supported versions? (Might still be too much)
      - what is going wrong with daily link-checking? If we reduce the link-checking coverage of individual PRs, I'd want to rely on daily link-checking to accurately report problems.
      - can we decouple link-checking from the `build-docs` workflow in a reasonable manner?
      - and if so, would we just be shifting the merge check time to a new separate link-checking workflow?
   - note: we can remove link-checks against tags (tags were brought in by Best Practices and laziness)
- ✅  ~note that the full build will still need to occur for deployments. I think those might be using the same workflow, in which case we'd want to bifurcate it 🍴.~ this is not true, there are separate workflows for publishing.

pepopowitz:
Investigation completed!

PR #1673, opened 2023-02-03
[HOLD] Spike: can we identify all changed files in a PR?

pepopowitz:
DO NOT MERGE! I'm messing around.

Part of #1659. 

## Findings

1. ✅ The first two commits indicate that for the `pull_request` trigger, [we can see all the files changed in this PR](https://github.com/camunda/camunda-platform-docs/actions/runs/4087332378/jobs/7047706462). 
2. ❌ The third commit indicates that for the `push` trigger, [we only see the files changed in the most recent commit](https://github.com/camunda/camunda-platform-docs/actions/runs/4087332378/jobs/7047706462). 

* I might not be using the `fetch-depth` property of https://github.com/actions/checkout correctly. We might be able to fix number 2 above by passing the correct value of that field. I'm not finding very clear docs on what values for that field will accomplish what I'm looking for.
* I don't really know why we need to run `build-docs` on `push`. The workflow was configured that way _before_ we added the `pull_request` trigger, but it's possible that we can remove the `push` trigger now that we're also running on `pull_request`.
   * I have a vague recollection of someone once telling me that some people like to see the results of `build_docs` _before_ they open a PR. That's the only scenario I can think of where we would still need to run on `push`. 
   * I've scheduled a post to #ask-documentation for Monday to find out if there are people who like to do that. I kind of suspect they don't exist, because the build takes so dang long these days and I can't imagine they'd wait for green before opening the PR. 
   * This could also be a case of moving cheese for a few people for the sake of bettering the overall experience. If removing the `push` trigger enables us to speed up build times for everyone, that's a pretty significant net gain that might exceed the needs and wants of a couple people. 

3. ✅ [This PR](https://github.com/camunda/camunda-platform-docs/pull/1674) demonstrates that the `pull_request` trigger, on forked PRs, would also give [the full list of files changed in the PR](https://github.com/camunda/camunda-platform-docs/actions/runs/4087550133/jobs/7048196117).

### DIFF
--- a/.github/workflows/report-changed-files.yaml
+++ b/.github/workflows/report-changed-files.yaml
@@ -1,10 +1,17 @@
 name: report-changed-files
 
 on:
+  push:
+    branches:
+      - "**"
+      - "!main"
+    tags-ignore:
+      - "*"
   pull_request:
 
 jobs:
   report-changed-files:
+    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.fork
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/report-changed-files.yaml
+++ b/.github/workflows/report-changed-files.yaml
@@ -1,0 +1,20 @@
+name: report-changed-files
+
+on:
+  pull_request:
+
+jobs:
+  report-changed-files:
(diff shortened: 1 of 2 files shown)


pepopowitz:
Closing this as it was for research purposes.

PR #1675, opened 2023-02-03
[HOLD] Spike: link-checking with excluded versions

pepopowitz:
DO NOT MERGE!!!

Part of #1659 

## Questions answered

1. ❌  Will a link-check of a build that doesn't include all versions pass?
   No, it will not pass. [Here is a build of only the latest versions](https://github.com/camunda/camunda-platform-docs/actions/runs/4087680639/jobs/7048479399), and [here is a build of only older versions](https://github.com/camunda/camunda-platform-docs/actions/runs/4087717101/jobs/7048555825). (see commits 1 and 2 in this PR.)
2. ❌ If I reduce the sitemap to be just the versions included, will that pass?
   Not likely, because we also check product-links.txt, and that also includes a bunch of URLs.
   It _might_ pass if the build includes the current version, too. 
3. ✅ Will a link-check pass if the build includes only the current version and affected versions, and the sitemap links are limited to the current/affected versions?
   [This workflow](https://github.com/camunda/camunda-platform-docs/actions/runs/4088291137/jobs/7049809747) demonstrates that a build that includes the next, current, and one old version will pass when it is link-checked for only the next, current, and one old version URLs. (see commits 3 and 4 in this PR.)
4. ✅ Why are cron-based link-checks failing regularly? If these were more stable I'd feel more comfortable about doing partial link-checking on PRs. 
   Lots of connection errors: see [this one](https://github.com/camunda/camunda-platform-docs/actions/runs/4069120227/jobs/7008489640) and [that one](https://github.com/camunda/camunda-platform-docs/actions/runs/4058753002/jobs/6986054023). 
   There is actually a flag we can pass to the link-checking tool [to report connection failures as warnings instead of errors](https://github.com/filiph/linkcheck/blob/master/lib/linkcheck.dart#L198-L199). I've PR'ed a change for this at #1676.


pepopowitz:
Closing this as it was for experimentation only.

PR #1676, opened 2023-02-03
ci(linkcheck): use github action instead of downloading tar, and ignore connection failures for cron-based workflow

pepopowitz:

1. Swaps a tarball download of https://github.com/filiph/linkcheck for its github action
   I noticed some comments in the actions that indicate we were waiting for a specific release to switch back to the github action. That release has been shipped, so we should switch back to the github action to simplify the workflow definitions.
2. On the cron-based `link-check` workflow, reports connection failures as warnings instead of errors
   Our cron-based link-check workflow fails more often than it succeeds, and it appears to be mostly a result of "connection failures" for certain pages. Since we just ignore those failing workflows anyway, we can tell the linkcheck tool not to bother reporting them as errors, and hopefully get to more successful runs than failed ones.

## PR Checklist

### DIFF
--- a/.github/workflows/build-docs.yaml
+++ b/.github/workflows/build-docs.yaml
@@ -46,10 +46,10 @@ jobs:
         with:
           url: "http://${{ env.container_ip }}:8080/"
       - name: Check internal links
-        uses: filiph/linkcheck@v2.0.23
+        uses: filiph/linkcheck@2.0.23
         with:
           arguments: "-i sitemap.prod.txt"
       - name: Check product links
-        uses: filiph/linkcheck@v2.0.23
+        uses: filiph/linkcheck@2.0.23
         with:
           arguments: "-i product-links.prod.txt"
--- a/.github/workflows/build-docs.yaml
+++ b/.github/workflows/build-docs.yaml
@@ -46,13 +46,10 @@ jobs:
         with:
           url: "http://${{ env.container_ip }}:8080/"
       - name: Check internal links
-        # disable action until https://github.com/tylerbutler/linkcheck-bin/issues/1 is released
-        #uses: filiph/linkcheck@v2.0.20
-        #with:
-        #  arguments: '-i sitemap.prod.txt'
-        run: |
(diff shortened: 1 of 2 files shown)


pepopowitz:
Well......I'm not sure that flag does what it claims to do.

[This workflow](https://github.com/camunda/camunda-platform-docs/actions/runs/4088441799/jobs/7050117538) was run with that flag enabled. It appears to have failed on a connection error!!!

[This workflow](https://github.com/camunda/camunda-platform-docs/actions/runs/4088541857/jobs/7050329979) proves to myself that the way I passed the argument was correct. This one is failing because I ran it with garbage instead of that flag, and it failed on not recognizing the option. 

I ran the workflow with the correct flag again and [I think it's going to succeed](https://github.com/camunda/camunda-platform-docs/actions/runs/4088441799/jobs/7050426559)....but note that neither step outputs any warnings (yet -- I'm signing off before it finishes.). I'd expect to see warnings if it had any connection failures. 

I'm not sure what to do with this one. At the very least the flag doesn't _hurt_ anything.....it just might not actually _do_ anything.

pepopowitz:
> Well......I'm not sure that flag does what it claims to do.
> 
> ....
> 
> I'm not sure what to do with this one. At the very least the flag doesn't _hurt_ anything.....it just might not actually _do_ anything.

Having thought about this more, I'd like to proceed with reviewing/merging this PR. It doesn't hurt anything. It _might_ make the link-checking succeed more frequently. I'd like to get it out there, and see what the effect is with a larger sample size.

pepopowitz:
More notes:

1.[ Here's another failed workflow, where all that appears in the output are warnings](https://github.com/camunda/camunda-platform-docs/actions/runs/4088441799/jobs/7084401079). 
2. I opened [this issue](https://github.com/filiph/linkcheck/issues/116) with the linkcheck project. 
3. I suspect the problem might be that when there are warnings it's still spitting out a bunch of messages, but I don't know enough about dart & github actions to fiddle with it and find out.

akeller (approved):
No issues with this. Read through your thoughts and it sounds like this is the preferred way we want to go and it doesn't seem to break more than was already broken.

PR #1698, opened 2023-02-13
style: use orangemunda as primary color

korthout:

The red color of the fonts, links, navigation, search, etc. used in the Docs is not the color I would expect to be used at Camunda. I would expect that we use Orangemunda `#FC5D0D` as our primary color.

It is unclear to me why `#ff1f00` was used as the primary color before. It might be related to contrast to improve accessibility. However, it may also simply be a leftover from the start of the camunda-platform-docs. I propose we switch, so the Docs fit our branding, unless we confirm we need increased contrast. In that case I wonder if there are other colors that might fit better. 

This pull request switches Infima's (the design system used by Docusaurus) primary colors to shades of `Orangemunda`. The shades are produced using Docusaurus' own tool for producing different shades.

How it is now (red)How it will be after once merged (orangemunda)
Left: How it is now (red) - Right: How it will be after once merged (orangemunda)

Orangemunda > Red

## PR Checklist

### DIFF
--- a/src/css/custom.css
+++ b/src/css/custom.css
@@ -5,16 +5,18 @@
  * work well for content-centric websites.
  */
 
-/* You can override the default Infima variables here. */
+/*
+These colors are produced using #fc5d0d (orange-munda) in the primary color shades tool.
+See: https://docusaurus.io/docs/styling-layout#styling-your-site-with-infima
+*/
 :root {
-  --ifm-color-primary: #ff1f00;
-  --ifm-color-primary-dark: #ff3800;
-  --ifm-color-primary-darker: #ff4700;
-  --ifm-color-primary-darkest: #ff5400;
-  --ifm-color-primary-light: #ff7000;
-  --ifm-color-primary-lighter: #ff7d00;
-  --ifm-color-primary-lightest: #ff8b00;
-  --ifm-code-font-size: 95%;
+  --ifm-color-primary: #fc5d0d;
+  --ifm-color-primary-dark: #ec5103;
+  --ifm-color-primary-darker: #de4c03;
+  --ifm-color-primary-darkest: #b73f02;
+  --ifm-color-primary-light: #fc6f27;
+  --ifm-color-primary-lighter: #fc7734;
+  --ifm-color-primary-lightest: #fd925c;
 }
 
 .docusaurus-highlight-code-line {


akeller (approved):
🚀 🚢 :shipit: 

Thank you @korthout!

pepopowitz:
## Re: Contrast

### TL;DR

I think we should open another PR with a slightly darkened value of orange-munda, maybe #C54302, which meets a WCAG minimum contrast value of 4.5:1 as normal text on a light gray background.

Open to feedback on my conclusion & thought process, though!

### The long and winding path that got me there

1. [The WCAG minimum for contrast](https://webaim.org/articles/contrast/) is a ratio of 4.5:1.

2. orange-munda on white _does not_ meet WCAG standards for minimum contrast, for normal text: 

image

3. Actually, neither does the red that we were previously using. Though it was a little closer to the minimum: 

image

4. Those numbers were on a white background. There are a few places where we have orange text on a light gray background: 

image

Those contrast numbers are, as expected, even lower:

image

5. It doesn't take too much (#CF4602) to shift the orangemunda to a tone that meets minimum contrast values, at least for normal text on a white background.
image

6. #C54302 gets us to a tone that meets minimum contrast for normal text on the gray background. This is what it looks like:

image

7. I was torn between the idea of using this darkened value of orange-munda, and finding another brand color that met minimum contrast levels, until I looked for a brand guide in confluence, and found [this](https://confluence.camunda.com/display/PD/Brand+elements+in+UI): 

> Do not use brand colors in UI, due to poor accessibility

Which led me to the conclusion that we should not use orange-munda, and we would be okay to find a suitable non-brand color. 

And basing the non-brand color on a brand color, but slightly darkening it, makes more sense to me than choosing a random red color.

akeller:
@pepopowitz, I thought this would be a quick win, but thank you for digging into this.

Since I got too merge happy and merged this in, how would you like to proceed? Revert to the previous color or go with your suggestion? I'm good with either.

When I've brought up accessibility internally in the past, I was told to wait until we Carbonized all the product UIs. Then we would launch a product scan, which docs could opt into for accessibility concerns. You can connect with web marketing to see if they have any suggestions, although I think most of their text is white on black, black on offwhite/grey, or green on white.

pepopowitz:
Since it sounds like there are no objections to a darker shade of orangemunda, I'll open a PR later today to shift to that. It's not much more work than reverting to the red, and I think if we can get to something that meets the standards, it's worth this small effort. I also much prefer something that resembles a brand color, over a random red that we don't know why we're using (thanks @korthout for kicking this off).

And when product does the accessibility sweep, we can change it again to whatever it recommends.

korthout:
@pepopowitz Would it perhaps be useful to discuss this with Branding? They might have some ideas about dealing with this, especially as they have already remarked:
>Do not use brand colors in UI, due to poor accessibility

pepopowitz:
@korthout I'm concerned that will open a can of worms, and I don't really want to prolong this issue. Given we were already not using Camunda branding, I feel comfortable moving forward with the darker orange.

PR #1701, opened 2023-02-15
Document that temporal values can be stored as variables

### LINKED ISSUE #650: Document that temporal values can now be stored in variables
**Current situation**
https://docs.camunda.io/docs/components/concepts/expressions/#temporal-expressions
>The current date and date-time can be accessed using the built-in functions today() and now(). To store the current date or date-time in a variable, convert it to a string using the built-in function string().

This is no longer true from 1.4 onwards, due to https://github.com/camunda-cloud/zeebe/pull/8842.

**Wanted situation**
We should make sure the documentation is clear about what happens to the result of temporal expressions. This includes checking the documentation for other places where similar things are stated.

remcowesterhoud:

_Briefly describe the change you are making, this helps the reviewer by providing an overview._
From version 1.4 and higher we can store temporal values as variables. Before we needed the string() function as a workaround. As this is no longer required we can remove this from the docs.

With version 1.4 😄 But there is not rush, it can go with the next one.

closes #650

## PR Checklist

### DIFF
--- a/docs/components/concepts/expressions.md
+++ b/docs/components/concepts/expressions.md
@@ -180,7 +180,14 @@ More functions for string values are available as [built-in string functions](/c
 
 ### Temporal expressions
 
-The current date and date-time can be accessed using the built-in functions `today()` and `now()`. To store the current date or date-time in a variable, convert it to a string using the built-in function `string()`.
+The current date and date-time can be accessed using the built-in functions `today()` and `now()`.
+Storing a temporal value as a variable will store it as a timezone extended ISO 8601 formatted string.
+
+:::note
+The timezone in the ISO 8601 format differs from the FEEL specific format. Storing a date-time as a process variable means
+you can not use it in temporal expressions later on in the process. To circumvent this we recommend storing
+date-times in the FEEL specific format using the built-in `string()` function.
+:::
 
 ```feel
 now()
--- a/versioned_docs/version-8.0/components/concepts/expressions.md
+++ b/versioned_docs/version-8.0/components/concepts/expressions.md
@@ -180,7 +180,14 @@ More functions for string values are available as [built-in string functions](/c
 
 ### Temporal expressions
 
-The current date and date-time can be accessed using the built-in functions `today()` and `now()`. To store the current date or date-time in a variable, convert it to a string using the built-in function `string()`.
+The current date and date-time can be accessed using the built-in functions `today()` and `now()`.
+Storing a temporal value as a variable will store it as a timezone extended ISO 8601 formatted string.
+
+:::note
+The timezone in the ISO 8601 format differs from the FEEL specific format. Storing a date-time as a process variable means
+you can not use it in temporal expressions later on in the process. To circumvent this we recommend storing
+date-times in the FEEL specific format using the built-in `string()` function.
+:::
 
 ```feel
 now()
--- a/versioned_docs/version-8.1/components/concepts/expressions.md
+++ b/versioned_docs/version-8.1/components/concepts/expressions.md
@@ -180,7 +180,14 @@ More functions for string values are available as [built-in string functions](/c
 
 ### Temporal expressions
 
-The current date and date-time can be accessed using the built-in functions `today()` and `now()`. To store the current date or date-time in a variable, convert it to a string using the built-in function `string()`.
+The current date and date-time can be accessed using the built-in functions `today()` and `now()`.
+Storing a temporal value as a variable will store it as a timezone extended ISO 8601 formatted string.
+
+:::note
+The timezone in the ISO 8601 format differs from the FEEL specific format. Storing a date-time as a process variable means
+you can not use it in temporal expressions later on in the process. To circumvent this we recommend storing
+date-times in the FEEL specific format using the built-in `string()` function.
+:::
 
 ```feel
 now()


docs/components/concepts/expressions.md
```diff
@@ -180,7 +180,14 @@ More functions for string values are available as [built-in string functions](/c
 
 ### Temporal expressions
 
-The current date and date-time can be accessed using the built-in functions `today()` and `now()`. To store the current date or date-time in a variable, convert it to a string using the built-in function `string()`.
+The current date and date-time can be accessed using the built-in functions `today()` and `now()`.
+Storing a temporal value as a variable will store it in the ISO 8601 format.
```
korthout, line 184: > store it in the ISO 8601 format.

❓ Is it clear to users that it will be stored as a string? Perhaps it is, but we could also choose to document it explicitly. WDYT?

> store it in the ISO 8601 format.

❌ Actually, this concerns FEEL's extension of the ISO-8601 representation that adds support for a Zone ID. The ISO-8601 only supports the UTC/Z and time offsets. We should clarify this. See the snippet from DMN 1.3 Spec (10.3.4.1) below. You can also look at 10.3.2.3.4, 10.3.2.3.6 for more details about FEEL's usage of the UTC/Z and time offsets.

> • **time string** – either a string value in the lexical space of the time datatype specified by XML Schema Part 2 Datatypes; or a string value that is the extended form of a local time representation as specified by ISO 8601, followed by the character "@", followed by a string value that is a time zone identifier in the IANA Time Zones Database (http://www.iana.org/time-zones)
• **date time string** – a string value consisting of a date string value, as specified above, optionally followed by the character "T" followed by a time string value as specified above.

docs/components/concepts/expressions.md
```diff
@@ -180,7 +180,14 @@ More functions for string values are available as [built-in string functions](/c
...
+
+:::note
+The timezone in the ISO 8601 format differs from the FEEL specific format. Storing a date-time as a process variable means
+you can not use it in temporal expressions later on in the process. To circumvent this we recommend storing
+date-times in the FEEL specific format using the built-in `string()` function.
```
korthout, line 189: ❓ Is this true? I [thought the variables contain exactly the same string representation](https://github.com/camunda/zeebe/pull/8842/files#diff-73d88a5d343f1206fdecd30b6d473ca53a1c966d53a6c8934b23f5f842188f92R105-R139) that can be used to create the FEEL temporal values in the first place. Why can't they be used in later temporal expressions?

korthout (changes requested):
Thanks for this @remcowesterhoud 

Please have a look at my comments/questions.

docs/components/concepts/expressions.md
remcowesterhoud: I based this on this comment on the issue 🤷 
https://github.com/camunda/camunda-platform-docs/issues/650#issuecomment-1344075449

docs/components/concepts/expressions.md
remcowesterhoud: > Actually, this concerns FEEL's extension of the ISO-8601 representation that adds support for a Zone ID. The ISO-8601 only supports the UTC/Z and time offsets. We should clarify this

How do you propose I write this down? Looking at the snippet there is no easy word to use here. I can write `store it in an extended form of a local time representation as specified by ISO 8601` but I feel like that's vague and confusing.

remcowesterhoud: @korthout I'm actually not convinced what you say here it correct. If I read [this issue](https://github.com/camunda/zeebe/issues/10307) it contains some examples which seem to verify that it's stored in an ISO 8601 format.

remcowesterhoud: Turned it into:
`Storing a temporal value as a variable will store it as a timezone extended ISO 8601 formatted string.`

docs/components/concepts/expressions.md
korthout: I see. After reading that issue, it sounds like we should just store the temporal value in FEEL format, not in ISO 8601. That seems like the cleanest solution, IMO. 

Perhaps we should document it immediately in that way then 🤔 But then we document the behavior that we don't have because of a bug... Now I'm not sure about documenting this before fixing that bug 😆 WDYT?

docs/components/concepts/expressions.md
korthout: Perhaps we should write that we store it in the FEEL format (an extended ISO 8601 format, see https://docs.camunda.io/docs/next/components/modeler/feel/language-guide/feel-data-types/#time for details)

docs/components/concepts/expressions.md
remcowesterhoud: I think I want to be done with this issue 😄

From what I can see what I've documented now is correct. When we fix the bug we can fix the documentation as well 🤷‍♂️

I don't like documenting that something works a specific way if we know it doesn't. That's only going to cause confusion and issues.

korthout: > When we fix the bug we can fix the documentation as well 🤷‍♂️

The problem with this is that we typically try not to change the behavior written in the docs. If we have bugs, then we should fix those, but the documented behavior should not change.

>I don't like documenting that something works a specific way if we know it doesn't. That's only going to cause confusion and issues.

Agreed. So, I feel that we should either not document it yet, or document it as we want it, with a clear not on which part doesn't work yet ([only using timezone offset seems to work right now](https://github.com/camunda/zeebe/issues/10307#issuecomment-1433373633))

christinaausley (approved):
No stylistic comments on this one so approving from that end, just need to resolve @korthout's comment.

korthout:
@remcowesterhoud How do you want to continue with this? Do we close the PR, document it referencing the bug, or perhaps you have another idea?

remcowesterhoud:
@korthout I will close the PR. As it's not clear how we will solve the bug it's unclear how to document it properly. We can document it after the bug is solved.

PR #1720, opened 2023-02-21
feat(modeler/desktop-modeler): use windows path example

nikku:

This proposes to using a Windows path to document `zeebe-ssl-certificate` flag, as it is apparently hardest to configure.

Related to https://github.com/camunda/camunda-modeler/issues/3466

No.

No specific time slot (any time).

## PR Checklist

### DIFF
--- a/versioned_docs/version-8.1/components/modeler/desktop-modeler/flags/flags.md
+++ b/versioned_docs/version-8.1/components/modeler/desktop-modeler/flags/flags.md
@@ -74,11 +74,13 @@ To display a custom version information in the status bar of the app, configure 
 
 > :information_source: The Modeler will read trusted certificates from your operating system's trust store.
 
-Provide additional certificates to validate secured connections to a Camunda Platform 8 installation. Configure your `flags.json` like this:
+Provide additional certificates to validate secured connections to a Camunda Platform 8 installation. 
+
+Configure your `flags.json` like this:
 
 ```js
 {
-    "zeebe-ssl-certificate": "/Users/local/trusted-custom-root.pem"
+    "zeebe-ssl-certificate": "C:\\path\\to\\certs\\trusted-custom-roots.pem"
 }
 ```
 


nikku:
@smbea We'd want to backport that change to `versioned_docs` before mergin.

nikku:
Yes. I usually backport to _recent_ versions, i.e. `8.x`.

nikku:
Did we remove them from `next`? 

I'd expect the documentation to be present in 8.1, 8.0 _and_ [next](https://docs.camunda.io/docs/next/guides/) but cannot see it being changed in next in this PR.

christian-konrad (changes requested):
To what @CatalinaMoisuc mentioned:

I think we should expose an example both for windows and unix-based systems.
We shouldn't hide those simple examples just because we assume those who read it are familiar enough with such details. The support request regarding this was one good example that even professionals forget about such things, and we can help them.

nikku:
> I think we should expose an example both for windows and unix-based systems.

Why? Unix paths are very easily represented in JSON, hence the issue at hand does not occur there if you use a standard unix path name. I'd love to ship this change as is and await further feedback.

The alternative is that we make this example 100% bullet-proof and extremely verbose, which means we need to cater for many different perspectives:

* flags vs CLI usage
* unix vs macos vs windows (each showcasing real, reasonable paths)

christian-konrad:
@nikku I am not sure if the reader will have the same understanding of this doc page as you have.

We can make this change as you propose, and wait for the next support request, and if it occurs, do another change, and then learn from it. It's ok for me.

But, the cost of making this change now is extremely cheap, or what's the hidden price? I don't see it. We don't need to cater for all cases, then. 
I will unblock the merge now

christian-konrad (approved):

nikku:
> But, the cost of making this change now is extremely cheap, or what's the hidden price?

The hidden price is that we stuff the documentation with (potentially) unnecessary details. Already by providing a windows path name we educate people about "how to properly escape backslashes" in JSON strings. I'd love to keep the reference as short as possible ("kiss").

If we get further feedback we can expand + iterate.

PR #1726, opened 2023-02-22
fix: stop failing `move-to-done` when there's no associated board item

### LINKED ISSUE #1707: `move-to-done` action fails when a PR hasn't been added to the DX board
## What

We should update the `move-to-done` action so that in the case of no associated DX board item, the workflow succeeds without updating anything.

## Why

The vast majority of PRs have @akeller or @christinaausley added as reviewers. When that happens, the `add-to-project` action adds an item to the DX board for that PR. 

When neither of them are added as reviewers, that action does not add a corresponding item to the DX board. When the `move-to-done` action runs, after the PR is merged, there is no associated project item to look up, and the `move-to-done` action fails with the message `Could not resolve to a node with the global id of ''`.

Example: https://github.com/camunda/camunda-platform-docs/actions/runs/4186555903

pepopowitz:

Resolves #1707 

I experienced a flurry of `move-to-done` failures on dependency PRs, and was irritated by them.

### Confirming the syntax of the condition

Since the edited workflow only runs when a PR is merged, I did some testing in this branch with a temporary workflow that ran on `pull_request`. 

Here's a workflow using the same condition as the final form of this PR, but on `pull_request` trigger, with no associated board item: https://github.com/camunda/camunda-platform-docs/actions/runs/4247780459/jobs/7386208183

And here's a workflow using the same condition as the final form of this PR, but on `pull_request` trigger, *with* an associated board item: https://github.com/camunda/camunda-platform-docs/actions/runs/4247780459/jobs/7386254382

Note that both complete with green, even though only the second prints out the project item ID.

## PR Checklist


.github/workflows/move-to-done.yaml
```diff
@@ -58,23 +58,25 @@ jobs:
...
+                  fieldId: $status_field
+                  value: {
+                    singleSelectOptionId: $status_value
+                    }
+                }) {
+                  projectV2Item {
+                    id
+                    }
+                }
+              }' -f project=$PROJECT_ID -f item=$PROJECT_ITEM_ID -f status_field=$STATUS_FIELD_ID -f status_value=${{ env.DONE_OPTION_ID }}
```
pepopowitz, line 81: most of this is indentation noise. only the `if` and `fi` lines wrapping this segment are actually new.

akeller (approved):
I didn't test this, but the new `if` looks good to me.

christinaausley (approved):
Thank you! This makes sense to me.

PR #1737, opened 2023-02-23
docs(optimize): call out that the readiness endpoint does not want an Authorization header

pepopowitz:

Adds an admonition to vNext and v3.9.0 Optimize API docs readiness endpoint, calling out that this endpoint does not work with an `Authorization` header.

This is in response to [my API exploration findings](https://github.com/camunda/developer-experience/issues/31#issuecomment-1438778116), where I got stuck for a bit because I was trying to send the header in a `/readyz` request, and getting a confusing error.

### What it looks like

image

## PR Checklist

### DIFF
--- a/optimize/apis-clients/optimize-api/health-readiness.md
+++ b/optimize/apis-clients/optimize-api/health-readiness.md
@@ -6,6 +6,10 @@ description: "The REST API to check the readiness of Optimize."
 
 The purpose of Health-Readiness REST API is to return information indicating whether Optimize is ready to be used.
 
+:::note
+The Health-Readiness REST API does not require an [`Authorization` header](./optimize-api-authorization.md), and rejects requests that include one.
+:::
+
 ## Method & HTTP target resource
 
 GET `/api/readyz`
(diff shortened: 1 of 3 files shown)


RomanJRW (dismissed):
Thanks for doing this @pepopowitz 👍 

Should we also backport this further than the 3.9 docs? I believe this endpoint has existed for 3.7 and 3.8 too

pepopowitz:
> Should we also backport this further than the 3.9 docs? I believe this endpoint has existed for 3.7 and 3.8 too

My instinct is that applying this to current & future versions is probably enough to help the people most likely to run into this, who are first exploring this API. If anyone has strong feelings about applying to 3.7 and 3.8, I can be convinced, though.

akeller:
> If anyone has strong feelings about applying to 3.7 and 3.8, I can be convinced, though.

Optimize docs include C7 and C8 users today, so I would recommend backporting for C7 users on an older minor version.

pepopowitz:
I've ported the changes to v3.8.0. The Optimize API docs didn't exist for v3.7.0. Requesting re-review!

akeller (approved):
Looks good! 

Question to @RomanJRW if the API docs should exist in 3.7 or not. But that can be another PR if we want to close this out by merging and following up later.

PR #1757, opened 2023-03-01
Add odd replication factor recommendation

### LINKED ISSUE #1743: Make a note about replication factor
I often discovered no in the forum that users are running with an even replication factor, which is problematic.

I recently added a warning to our start-up process https://github.com/camunda/zeebe/pull/11831 but we should also add a note to the docs to make it clearer.

For example, posts where it occurred:

 * https://forum.camunda.io/t/issue-in-connecting-gateway/42841/4
 * https://forum.camunda.io/t/zeebe-startup-issue/36622/4
 * I think there were several on slack

Potential note could be similar as our new warning:

```java
 "Expected to have odd replication factor, but was even ({}). Even replication factor has no benefit over "
          + "the previous odd value and is weaker than next odd. Quorum is calculated as:"
          + " quorum = floor(replication factor / 2) + 1. In this current case the quorum will be"
          + " quorum = {}. If you want to ensure high fault-tolerance and availability,"
          + " make sure to use an odd replication factor.";
```

Or something similar as here https://stackoverflow.com/questions/58823341/why-is-it-recommended-to-create-clusters-with-odd-number-of-nodes

akeller:

Closes #1743 - adds a note to Zeebe technical concepts and glossary about even replication factors.

No

No specific date to go live.

## PR Checklist


akeller:
@Zelldon, I made the initial changes in `next` so please add or modify, and then we can get the final changes added to 8.1.

docs/components/zeebe/technical-concepts/partitions.md
```diff
@@ -48,6 +48,16 @@ A partition is a persistent append-only event stream. Initially, a partition is
 
 For fault tolerance, data in a partition is replicated from the **leader** of the partition to its **followers**. Followers are other Zeebe broker nodes that maintain a copy of the partition without performing event processing.
 
+We recommend an even replication factor. Odd replication factors can be problematic and may trigger this warning:
```
Zelldon, line 51: It's the other way around 😃

Zelldon: Due we want to just add the warning log here or maybe we bake it in as text?

docs/reference/glossary.md
```diff
@@ -166,7 +166,9 @@ Replication is the act of copying data in a partition from a leader to its follo
 
 This is the number of times data in a partition are copied. This depends on the number of brokers in a cluster. A cluster with one leader and two followers have a replication factor of three, as data in each partition needs to have three copies.
 
-- [Clustering](/components/zeebe/technical-concepts/clustering.md#raft-consensus-and-replication-protocol)
+We recommend running an even replication factor.
```
Zelldon, line 169: Here also the other way around :)

Zelldon (commented):
Thanks @akeller for the quick changes :) I have some comments

akeller:
My brain thought one thing and typed another. 🤪

docs/components/zeebe/technical-concepts/partitions.md
```diff
@@ -48,6 +48,20 @@ A partition is a persistent append-only event stream. Initially, a partition is
...
 
+We recommend an odd replication factor. Even replication factors can be problematic and may trigger this warning:
+
+```
+ "Expected to have odd replication factor, but was even ({}). Even replication factor has no benefit over "
+          + "the previous odd value and is weaker than next odd. Quorum is calculated as:"
+          + " quorum = floor(replication factor / 2) + 1. In this current case the quorum will be"
+          + " quorum = {}. If you want to ensure high fault-tolerance and availability,"
+          + " make sure to use an odd replication factor.";
+```
```
Zelldon, line 59:
```suggestion
We recommend an odd replication factor. 
```

Zelldon: I would just remove this, since we explain it later anyway.

Wdyt?

Zelldon (approved):
Thanks @akeller I have one comment, maybe you can take a look at it before merging. :) Thanks again for doing the work.

docs/components/zeebe/technical-concepts/partitions.md
akeller: I included it in case someone copies and pastes it into external or internal search. It is a bit redundant.

We can remove it for now and add it in the future if we want to leverage the SEO benefits.

christinaausley (approved):
Adjusted some of the spacing and layout in my latest commit 👍

PR #1759, opened 2023-03-01
Update Migration Guide

jonathanlukas:

This change should reflect changes being made to the tooling that should support migration.

no

asap

## PR Checklist

### DIFF
--- a/docs/guides/migrating-from-camunda-platform-7.md
+++ b/docs/guides/migrating-from-camunda-platform-7.md
@@ -212,9 +212,11 @@ In general, **workflow engine data** is harder to migrate to Camunda Platform 8:
 
 The [Camunda Platform 7 to Camunda Platform 8 migration tooling](https://github.com/camunda-community-hub/camunda-7-to-8-migration), available as a community extension, contains two components that will help you with migration:
 
-1. [A Desktop Modeler plugin to convert BPMN models from Camunda Platform 7 to Camunda Platform 8](https://github.com/camunda-community-hub/camunda-7-to-8-migration/tree/main/modeler-plugin-7-to-8-converter). This maps possible BPMN elements and technical attributes into the Camunda Platform 8 format and gives you warnings where this is not possible. This plugin might not fully migrate your model, but should give you a jump-start. It can be extended to add your own custom migration rules. Note that the model conversion requires manual supervision.
+1. [A converter available in different flavors (webapp, cli) to convert BPMN models from Camunda Platform 7 to Camunda Platform 8](https://github.com/camunda-community-hub/camunda-7-to-8-migration/tree/main/backend-diagram-converter). This maps possible BPMN elements and technical attributes into the Camunda Platform 8 format and gives you warnings where this is not possible. The result of a conversion is a model with mapped implementation details as well as hints on what changed, needs to be reviewed or adjusted in order to function properly in Camunda Platform 8.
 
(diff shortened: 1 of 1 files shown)


docs/guides/migrating-from-camunda-platform-7.md
```diff
@@ -530,16 +548,12 @@ Here's an example of a good Java Delegate:
...
-    @Autowired
-    private ObjectMapper objectMapper;
-
     @Autowired
     private CrmFacade crmFacade;
 
     public void execute(DelegateExecution execution) throws Exception {
         // Data Input Mapping
-        String customerDataJson = (String) execution.getVariable("customerData");
-        CustomerData customerData = objectMapper.readValue(customerDataJson, CustomerData.class);
```
berndruecker, line 542: Why did you remove the ObjectMapper - how would the data mapping work in C7? As we typically discourage object serialization and Spin - I think most realistic is to have a JSON String and transform it yourself - that's why I added it there

berndruecker (approved):
Great changes and looks like good progress 👍

docs/guides/migrating-from-camunda-platform-7.md
jonathanlukas: I see, however, Java objects can be serialized in `application/json` directly (which is using Spin implicitely). 

At the same time, saving a complex object as String is an anti-pattern (I also removed a similar snippet below)

christinaausley (approved):
Minor adjustments in latest commit for styling 👍

akeller:
@jonathanlukas do you also want these changes reflected in the current docs? If so, I can add them to 8.1.

jonathanlukas:
@akeller I would actually like to have them in 8.1, yes. 

However, I am not yet ready with @berndruecker in some points.

For ease, I would copy the changes over to the versioned docs when this PR is ready to merge.

docs/guides/migrating-from-camunda-platform-7.md
jonathanlukas: @berndruecker should we discuss this?

berndruecker: I would not fight this battle now due to bandwith reasons - but I still think using Spin is a bad idea and I would prefer storing JSON as String variable (or byte[] - wasn't there a difference in size? I already forgot :-)) and do serialization myself (what ends up in the database is the same - isn't it?). 

But as we are not advising people that this is a best practice thing, but give guidance on migrating existing things, it might not matter too much any way (nobody would base a decision on how to do their C7 program on this code example)

berndruecker: @jonathanlukas So just go ahead if this is the consulting viewpoint on it 👍

jonathanlukas: @berndruecker in general yes. 

To remove the reference to Spin from the example, we could retrieve the Java object directly and rely on the serialisation defaults of the engine (which should be `application/json`).

And yes, there is a problem with the text2 column in the process variable (which will allow 4000 characters) that some customers were already facing.

This is why I do not like the approach of using a String... without this limitation, I would go for this solution as well.

akeller:
@jonathanlukas, just a quick check-in here, do you anticipate having this ready for the minor release next week?

christinaausley (approved):

akeller:
@christinaausley do you plan to visit this PR and add it to 8.2 too?

PR #1762, opened 2023-03-01
docs(dependencies): add query string and fix OOTB connectors headings

akeller:

Add query string on dependencies page for `next` and 8.1. 

Fix heading for OOTB connectors on dependencies page for `next`.

Internal excitement about deep linking is real.

Not urgent, but the dependencies page sidebar looks cluttered with all the connectors headings.

## PR Checklist


akeller:
@bastiankoerber I added you as a reviewer to confirm the Connector headings change. The hierarchy should reflect how they appear in the Connectors docs, correct?

docs/reference/dependencies.md
```diff
@@ -10,7 +10,7 @@ A complete list of all dependencies and third-party libraries for all the compon
 import Tabs from "@theme/Tabs";
 import TabItem from "@theme/TabItem";
 
-<Tabs groupId="dependencies" defaultValue="zeebe" values={
+<Tabs groupId="dependencies" defaultValue="zeebe" queryString values={
```
akeller, line 13: @pepopowitz this was almost _too_ easy.

I did notice because we have a trailing slash in our docs the queryString that's added ends up looking like `https://docs.camunda.io/docs/next/reference/dependencies/?dependencies=zeebe` which is fine, but looks a bit silly. I'm just excited for deep linking!

bastiankoerber (approved):
looks good to me

docs/reference/dependencies.md
pepopowitz: yeah, that trailing slash thing has always been annoying to me since the first time I generated a static site, but I think it's the price you pay when every friendly markdown filename turns into a nicely named folder with an index.html.

pepopowitz (approved):
This is awesome! :shipit: 

--- 

non-blocking, and it turns out not an issue at all: 

I was curious about redirect rules, and I couldn't remember how they by default handle query strings. Basically, I was wondering if suddenly passing someone a URL that had a querystring param in it would match our rewrite rules and pass them through, or if it would not match our rewrite rules and result in suddenly-broken URLs. I found two things: 

1. This specific change doesn't correspond to any existing rewrite rules, so it would not be a concern anyway.
2. Even if it was, our rewrite rules handle querystring values as I'd hope, which is to say (a) they match the URL regardless of querystring, and (b) they pass the querystring through to the target URL. 

No action to be taken here, I'm just documenting it in case anyone was wondering the same thing I was.

PR #1764, opened 2023-03-02
Add section about message expiration

ingorichtsmeier:

Add a section to describe what happens, if a message could not be correlated to a process during its time-to-live.

Resolves https://github.com/camunda/zeebe/issues/11410

No

With the next release of the docs.

## PR Checklist

- [X] My changes apply to an already released version, and I have added them to the relevant `/versioned_docs` directory, or they are not for an already released version.
- [X] My changes apply to future versions, and I have added them to the main `/docs` directory, or they are not for future versions.
- [X] My changes require an [Engineering review](https://github.com/camunda/camunda-platform-docs/blob/main/howtos/documentation-guidelines.md#review-process), and I've assigned an engineering manager or tech lead as a reviewer, or my changes do not require an Engineering review.

### DIFF
--- a/docs/components/concepts/messages.md
+++ b/docs/components/concepts/messages.md
@@ -49,6 +49,13 @@ zbctl publish message "Money collected" --correlationKey "order-123" --ttl 1h
    </p>
  </details>
 
+## Message expiration
+
+When a buffered message could not be correlated to a process instance in the
+given time-to-live, it gets removed from the buffer afterwards. Internally, the
+MessageObserver runs once a minute and exports an event, that the message
+expired.
+
 ## Message cardinality
 
 A message is correlated only _once_ to a process (based on the BPMN process id), across all versions of this process. If multiple subscriptions for the same process are opened (by multiple process instances or within one instance,) the message is correlated only to one of the subscriptions.
--- a/docs/components/concepts/messages.md
+++ b/docs/components/concepts/messages.md
@@ -54,6 +54,9 @@ zbctl publish message "Money collected" --correlationKey "order-123" --ttl 1h
 When a buffered message could not be correlated to a process instance in the
 given time-to-live, it gets removed from the buffer afterwards.
 
+As the expired messages are cleaned up in batches, there will be a configurable
+delay before the expiration is available.
+
 More details about the expired messages can be found in the exporters chapter
 of the Self-Managed section. (Link to be done).
 
--- a/docs/components/concepts/messages.md
+++ b/docs/components/concepts/messages.md
@@ -52,9 +52,10 @@ zbctl publish message "Money collected" --correlationKey "order-123" --ttl 1h
 ## Message expiration
 
 When a buffered message could not be correlated to a process instance in the
-given time-to-live, it gets removed from the buffer afterwards. Internally, the
-MessageObserver runs once a minute and exports an event, that the message
-expired.
+given time-to-live, it gets removed from the buffer afterwards.
+
+More details about the expired messages can be found in the exporters chapter
+of the Self-Managed section. (Link to be done).
 
 ## Message cardinality
 
--- a/versioned_docs/version-8.1/components/concepts/messages.md
+++ b/versioned_docs/version-8.1/components/concepts/messages.md
@@ -49,6 +49,13 @@ zbctl publish message "Money collected" --correlationKey "order-123" --ttl 1h
    </p>
  </details>
 
+## Message expiration
+
+When a buffered message could not be correlated to a process instance in the
+given time-to-live, it gets removed from the buffer afterwards. Internally, the
+MessageObserver runs once a minute and exports an event, that the message
+expired.
+
 ## Message cardinality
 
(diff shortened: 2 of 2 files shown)


docs/components/concepts/messages.md
```diff
@@ -49,6 +49,13 @@ zbctl publish message "Money collected" --correlationKey "order-123" --ttl 1h
    </p>
  </details>
 
+## Message expiration
+
+When a buffered message could not be correlated to a process instance in the
+given time-to-live, it gets removed from the buffer afterwards. Internally, the
```
korthout, line 55: ❌ This is duplicate because it is already discussed in the `Message buffering` section directly above this.

>A message has a **time-to-live** (**TTL**) which specifies for how long it's buffered. Within this time, the message can be correlated to a process instance.

docs/components/concepts/messages.md
```diff
@@ -49,6 +49,13 @@ zbctl publish message "Money collected" --correlationKey "order-123" --ttl 1h
...
+MessageObserver runs once a minute and exports an event, that the message
+expired.
```
korthout, line 57: ❓ Do we want to document this in concepts? 

IMO, this is an implementation detail. Perhaps we should change the description of the TTL in the `Message Buffering` section to 
> "A message has a **time-to-live** (**TTL**) which specifies the minimum duration that it's buffered".

We don't and cannot guarantee how soon after the TTL expires, the message is discarded. The MessageTimeToLiveChecker's interval is just an implementation detail currently in place. In my opinion, it is not part of our API, so we should not discuss it in the documentation, as we may change this later.

Note that I'm currently working on allowing users to configure the MessageTimeToLiveChecker's interval. We can describe that users can adjust this in the config file. There it should be clear what the effect of that adjustment is.

❌ I'm also not convinced of event exporting in the Concepts section, as it is a Self-Managed-only feature.

korthout (changes requested):
Thanks @ingorichtsmeier, for initiating this 👍 

👍 I think it's a good idea to improve the documentation around Buffering.

❌ I would like to propose a different approach. Please have a look.

docs/components/concepts/messages.md
ingorichtsmeier: Hi @korthout, from my developer perspective it is important to know, that at some point in time I can get information about the expired message. 

I discussed it with a customer who wants to react on this expired message (send it again or announce the expiration to somebody else). Eventually, the message expiration event should be available via Operate API.

At the moment, I don't care about the minute. Could be longer or shorter or configurable.

For me, the important learning of the underlying issue was, that the event happens after some delay, and it is worth to mention the delay and the source where it is coming from.

Even if I can't change the minute by myself right now, it may be useful to look up the implementation and estimate the effort, that a feature request to make in configurable would take. With this argument, the section gives me all information that I was looking for.

korthout: Sorry @ingorichtsmeier, I missed your reply.

>it is important to know, that at some point in time I can get information about the expired message

I find it difficult to document it like this. This functionality only exists in Self-Managed (exporters are not available to SaaS users), so documenting it as a feature in the main Concepts would be confusing SaaS users. If we would document it there, it would require a special Self-Managed-only notice.

>Eventually, the message expiration event should be available via Operate API.

That would be a great addition 👍 

>a feature request to make in configurable 

👍 I'm currently implementing this, it is part of the [message expiration performance improvements](https://github.com/camunda/zeebe/issues/11591). Since that is considered a bug, we will backport these changes to all supported minor versions.

>With this argument, the section gives me all information that I was looking for.

❓ Do I understand correctly that you're happy with the configuration approach as alternative documentation?

ingorichtsmeier:
Just a short summary of our in-person-discussion:

1. It would be the best to keep the concepts as concepts and just add a reference to the exporter chapter of the SAAS section.
2. Improve the elasticsearch exporter section with more details about the exported documents. Details can be explored with the help of
  - https://github.com/camunda/zeebe/wiki/How-to-test-workflow-engine-changes-manually#inspect-data-in-elasticsearch
  - https://github.com/camunda/camunda-platform#kibana
  - https://github.com/camunda/zeebe/tree/main/exporters/elasticsearch-exporter/src/main/resources

I will improve my proposal and @korthout agreed to review.

korthout:
@ingorichtsmeier I just talked with @saig0 about message TTL.

I previously stated:

> We don't and cannot guarantee how soon after the TTL expires, the message is discarded. 

Apparently, this only concerns when we write the `Message:EXPIRED` event on the log (and export it). **It does not** concern at what moment the message can be correlated to the process from the buffer. 

So, while the message may exist in the buffer longer than its requested TTL, the message will not be correlated anymore to the process once it surpasses its deadline. The engine safeguards against this.

Example:
- we publish a message with TTL 1 minute
  - the published message receives a deadline of `15:46:05` (current time `15:45:05` (h:m:s) + 1 minute)
  - the message can only be correlated to a process until `15:46:05`
- the TTL checker detects it at `15:46:55` and writes `Message:EXPIRE` command
- the engine processes the `Message:EXPIRE` command at `15:47:10` (it was busy with other stuff) and writes `Message:EXPIRED` event
  - the message is now removed from the buffer

korthout:
@ingorichtsmeier I'm closing this PR for now 🙇

PR #1794, opened 2023-03-09
Docs apis tools styling

christinaausley:

**Paired work with @pepopowitz on Swizzling techniques.**

This is a second iteration to the APIs/Tools rework.

## What has changed

This PR kickstarted [Swizzling](https://docusaurus.io/docs/swizzling) techniques to accomplish two things:

- Add a border to beta releases (`cardContainerBetaBorder`).
Screen Shot 2023-03-09 at 12 08 32 PM

- Lengthen cards to update descriptions (I just removed the `text--truncate` here.)
Screen Shot 2023-03-09 at 12 08 08 PM

Yielding:

Screen Shot 2023-03-09 at 11 59 31 AM

## Pending changes before merge

@pepopowitz I know we had roughly discussed this in the last PR, but what on your thoughts on making it possible to include a description for the Optimize API? What can I test here?

## What still needs to be considered in future iterations

We've tossed around the idea of adjusting the icons, but this is a bit of a rabbit hole in terms of which way we could go. @pepopowitz feel free to add comments here.

Screen Shot 2023-03-09 at 12 04 26 PM

We could potentially create an `svg` of the preferred icon (GitHub, for example), but I think this does require more work outside of the scope of this PR.

## PR Checklist


christinaausley:
**Note that once this branch is merged in, I will open a new PR to officially adjust the APIs & Clients tab to APIs & Tools 👍**

src/theme/DocCard/index.js
```diff
@@ -0,0 +1,82 @@
...
+import Link from "@docusaurus/Link";
+import {
+  findFirstCategoryLink,
+  useDocById,
+} from "@docusaurus/theme-common/internal";
+import isInternalUrl from "@docusaurus/isInternalUrl";
+import { translate } from "@docusaurus/Translate";
+import styles from "./styles.module.css";
+function CardContainer({ href, children, border }) {
+  clsx("card padding--lg", styles.cardContainer);
```
pepopowitz, line 12:
```suggestion
```

I think this line was left behind accidentally during some copy/paste/refactor.

src/theme/DocCard/index.js
```diff
@@ -0,0 +1,82 @@
...
+  var className = clsx("card padding--lg", styles.cardContainer);
+  if (border == "alt") {
```
pepopowitz, line 14:
```suggestion
  if (border == "highlight") {
```

non-blocking: What do you think about using the value `highlight` instead of `alt` to signify which items should be....highlighted? 😎 

(Note that if we make this change, we also need to make it in the item descriptions)

src/theme/DocCard/index.js
```diff
@@ -0,0 +1,82 @@
+import React from "react";
```
pepopowitz, line 1:
```suggestion
// Why is this swizzled?
//   1. To eliminate unnecessary text truncation on card descriptions
//   2. To allow for highlighting specific cards with a different color border
// Swizzled from version 2.3.1.

import React from "react";
```

Blocking:

One thing I forgot to mention is that I would like us to specify in each swizzled component: 

1. Why it's swizzled, and
2. Which version of docusaurus it was swizzled from. 

This is helpful when we do docusaurus updates -- it helps identify which swizzled components need to be rewritten during an update, and when they do, what functionality needs to be preserved.

Feel free to rewrite the lines I suggested. You can also see [another swizzled component](https://github.com/camunda/camunda-platform-docs/pull/1794) for an example.

src/theme/DocCard/styles.module.css
```diff
@@ -0,0 +1,32 @@
+.cardContainer {
```
pepopowitz, line 1:
```suggestion
/* 
  Why is this swizzled?
    See the associated DocCard/index.js for details. There are no customizations in this file.
*/

.cardContainer {
```

Blocking: 

I _think_ this message is accurate, and that you didn't make any changes in this file. Regardless, we should also specify what's custom in this file to make updates easier.

src/theme/DocCardList/index.js
```diff
@@ -0,0 +1,27 @@
...
+  return (
+    <section className={clsx("row", className)}>
+      {filteredItems.map((item, index) => (
+        <article key={index} className="col col--6 margin-bottom--lg">
+          <DocCard item={item} />
+        </article>
+      ))}
+    </section>
+  );
+}
```
pepopowitz, line 27: Blocking: 

If I remember correctly, there are no changes in this file. If that's the case, this file is unnecessary and we should delete it!

If I'm not remembering correctly, and there are customizations in here, we should capture that with a comment at the top, like I described in DocCard/index.js.

pepopowitz (changes requested):
This is awesome, thank you! 

A few changes I'm requesting are blocking.

pepopowitz:
> @pepopowitz I know we had roughly discussed this in the last PR, but what on your thoughts on making it possible to include a description for the Optimize API? What can I test here?

The issue here is that we have two docs instances, and the Optimize API is in the opposite instance from this file. This instance has no access to that Optimize doc.

The one thing I can think, if we really want a message there, is to specify the description in a property on the item, and pass it through as far as it needs to go, similar to how we did the `border` property. The `CardLink` component could then use that `description` property as a fallback if there is no known doc.

src/theme/DocCard/index.js
christinaausley: Agreed here. Resolving in latest commit 👍

src/theme/DocCard/index.js
christinaausley: Thank you for this!

src/theme/DocCardList/index.js
christinaausley: Removed in my latest commit 👍

src/theme/DocCard/styles.module.css
christinaausley: Thank you for noting this!

docs/apis-clients/operate-api/index.md
```diff
@@ -1,7 +1,7 @@
 ---
 id: index
 title: Operate API (REST)
-description: "Operate API is a REST API and provides searching, getting, and changing Operate data. Requests and responses are in JSON notation."
```
akeller, line 4: Notation notation 😉 good catch!

pepopowitz (approved):
This is awesome, thanks for making the changes!!!

PR #1819, opened 2023-03-22
docs(identity): update supported environments

dlavrenuek:

Related to https://github.com/camunda-cloud/identity/issues/1515

We have added support for Keycloak 20 in Identity, which will be effective in the next 8.2.0 release. Although we only test with the latest patch of each major version, backwards compatibility for other patch/minor releases should be seen as given by semver design. Therefore I made the supported versions we have tested more generic (19.x instead of 19.0.3). This should ensure the users who run different patch versions of Keycloak, that running c8 with their environment will not break anything.

None are known

With the 8.2 release

## PR Checklist

- [ ] ~My changes apply to an already released version, and I have added them to the relevant `/versioned_docs` directory, or they are not for an already released version.~
- [ ] ~My changes require an [Engineering review](https://github.com/camunda/camunda-platform-docs/blob/main/howtos/documentation-guidelines.md#review-process), and I've assigned an engineering manager or tech lead as a reviewer, or my changes do not require an Engineering review.~


christinaausley (approved):

christinaausley:
@dlavrenuek To confirm, this is only for the `next` version of docs?

dlavrenuek:
@christinaausley exactly, the support for Keycloak 20 will be included in the next minor release

PR #1820, opened 2023-03-22
Pepopowitz/dockerfile improvement

pepopowitz:

Improves the experience when testing htaccess rules locally. 

1. Adds a step to the Dockerfile that restarts the Apache server, to make sure it picks up changes to the server configuration.
2. Better clarifies how to re-start the server after making changes to the htaccess file.

I'm not totally convinced item 1 needs to happen, because I haven't needed it in the past.....but for some reason, when testing with @christinaausley, I had to go into the running instance and manually restart Apache. I don't think there's any harm in adding it, even if it isn't necessary.

## PR Checklist


christinaausley (approved):
Stellar. Thank you for this!

akeller (approved):
Great add! Thank you and thanks for supporting each other 💪

PR #1841, opened 2023-03-24
fix: rename all apis-clients/working-with-apis-clients docs to apis-tools/working-with-apis-tools so that docusaurus top nav item is consistent across all versions

pepopowitz:

Renames the entry page of the APIs section to apis-tools/working-with-apis-tools for all versions of the docs. 

Docusaurus only allows one page ID for each top nav menu item, so when one differs across versions, it results in top nav items not always working. 

Prior to this change, if you were to browse the `next` docs and click "APIs & Clients" in the top nav, it would take you to an unexpected page. After this change, clicking that top nav item will always take you to the currently browsed version of the same APIs & Clients page.

Note that this page did not exist prior to version 1.2, so no versions prior to 1.2 were updated.

## Video evidence that the nav item is consistent across versions 1.2 through next

https://user-images.githubusercontent.com/1627089/227625372-d338b6e4-c776-4210-af93-4a93521595d7.mov

## Follow-up work

- [ ] introduce this change into #1811 

## PR Checklist


docs/apis-clients/working-with-apis-clients.md
```diff
@@ -1,5 +1,5 @@
 ---
-id: working-with-apis-tools
+id: working-with-apis-clients
```
pepopowitz, line 2: I could have left the id as "working-with-apis-tools", but I feel like consistency between the ID and filename is less likely to lead to confusion.

akeller (dismissed):
Thank you for the quick work here @pepopowitz. I knew this change was going to be disruptive at the top-nav, but I didn't fully realize how it would break the `next` experience.

static/.htaccess
```diff
@@ -14,7 +14,8 @@ RewriteRule ^docs(/(8.0|next))?/components/modeler/desktop-modeler/element-templ
 # Added Self-Managed Modeler redirect
 RewriteRule ^docs/next/self-managed/web-modeler/installation/(.*)$ /docs/next/self-managed/modeler/web-modeler/installation/$1 [R=301,L]
 
-RewriteRule ^docs/next/apis-clients/working-with-apis-clients/?$ /docs/next/apis-clients/working-with-apis-tools/ [R=301,L]
+# this looks inverted, but it's because we had a redirect from clients -> tools, and we're changing it back from tools -> clients
+RewriteRule ^docs/next/apis-clients/working-with-apis-tools/?$ /docs/next/apis-clients/working-with-apis-clients/ [R=301,L]
```
pepopowitz, line 18: This is probably the weirdest part of the PR -- we pushed out a redirect already to send them from apis-clients -> apis-tools, so if they saved that URL we want to send them _back_ to apis-clients. I think this could result in a redirect loop for some users -- if their browser had cached the apis-clients -> apis-tools redirect, and now they get an apis-tools -> apis-clients redirect. 

I don't know a way around this. I think it will affect very very few people. Clearing browser cache would resolve it. I'm not sure if that's something we'll need to communicate somehow, or where to communicate it.

akeller: C8 Slack and Forum post for externals. ask-documentation for internal. Keep ears open 🤷‍♀️ 

As much as we try to avoid significant disruptions, we may have them.

pihme (approved):

static/.htaccess
pepopowitz: The more I think about this, the more I think the correct move is to move all the old versions to apis-tools. 

If we revert now, we'll end up re-reverting in the near future, to point everything _back_ at apis-clients again. And then we'd have to warn people about redirect loops _again_. 

Or, we put everything in the end-state of apis-tools, and we wash our hands of this. It's a tiny bit more work (I think) than what I already did in this PR, but I think it's the smarter move long-term. 

I'm going to work on that now -- if it turns into a rabbit hole, I will bail quickly, and merge this PR as-is instead.

pepopowitz: Ok, I just pushed up the inverted scenario. Now everything is changed to `/apis-tools/working-with-apis-tools`.

docs/apis-tools/img/ComponentsAndArchitecture_SaaS.png
pepopowitz, line 1: This image is only referenced by the working-with-apis-tools page, so it moved with it.

docs/apis-tools/working-with-apis-tools.md
pepopowitz, line 1: Since this moved folders, many links had to be updated to travel across folders.

docusaurus.config.js
```diff
@@ -84,7 +84,7 @@ module.exports = {
         },
         {
           type: "doc",
-          docId: "apis-clients/working-with-apis-clients",
+          docId: "apis-tools/working-with-apis-tools",
```
pepopowitz, line 87: This configures the item in the top nav.

product-links.txt
```diff
@@ -9,7 +9,7 @@ https://docs.camunda.io/docs/components/concepts/job-workers/
 https://docs.camunda.io/docs/guides/model-your-first-process
 https://docs.camunda.io/docs/guides/setup-client-connection-credentials
 https://docs.camunda.io/docs/guides/message-correlation
-https://docs.camunda.io/docs/apis-clients/working-with-apis-clients/
+https://docs.camunda.io/docs/apis-clients/working-with-apis-tools/
```
pepopowitz, line 12: Oops, I think I got this wrong.

```suggestion
https://docs.camunda.io/docs/apis-tools/working-with-apis-tools/
```

src/pages/index.js
```diff
@@ -32,6 +32,7 @@ const features = [
   {
     title: "APIs & Clients",
     imageUrl: "img/apisclients.png",
+    // TODO: update this URL to apis-tools once that page exists in production
```
pepopowitz, line 35: These are the 5 "feature" blocks on the homepage. They are currently hardcoded to production. I'm going to try updating these to be relative paths as part of this PR. 

(Otherwise I'll have to leave this one pointed at a URL that is incorrect, and rely on a redirect correcting it when the PR is deployed.)

static/.htaccess
```diff
@@ -128,12 +131,12 @@ RewriteRule ^docs/components/console/troubleshooting$ /docs/components/console/c
 RewriteRule ^docs/components/best-practices/overview/(.*)$ /docs/components/best-practices/best-practices-overview/$1 [R=301,L]
 RewriteRule ^docs/components/best-practices/overview$ /docs/components/best-practices/best-practices-overview [R=301,L]
 
-RewriteRule ^docs/1.3/apis-clients/overview/(.*)$ /docs/1.3/apis-clients/working-with-apis-clients/$1 [R=301,L]
-RewriteRule ^docs/1.3/apis-clients/overview$ /docs/1.3/apis-clients/working-with-apis-clients [R=301,L]
-RewriteRule ^docs/1.2/apis-clients/overview/(.*)$ /docs/1.2/apis-clients/working-with-apis-clients/$1 [R=301,L]
-RewriteRule ^docs/1.2/apis-clients/overview$ /docs/1.2/apis-clients/working-with-apis-clients [R=301,L]
-RewriteRule ^docs/apis-clients/overview/(.*)$ /docs/apis-clients/working-with-apis-clients/$1 [R=301,L]
-RewriteRule ^docs/apis-clients/overview$ /docs/apis-clients/working-with-apis-clients [R=301,L]
+RewriteRule ^docs/1.3/apis-clients/overview/(.*)$ /docs/1.3/apis-tools/working-with-apis-tools/$1 [R=301,L]
```
pepopowitz, line 134: I corrected these older rules based on the new target path.

product-links.txt
pepopowitz: Oh wait, I _did_ get it wrong, but in a different way -- I shouldn't touch this file at all, since it's keeping track of links from the product. I'll revert this line!

static/.htaccess
akeller: This looks like it's working when I test locally. LMK how else I can support.

akeller (approved):

Issue #1749, opened 2023-02-28
Document complete Zeebe configuration


npepinpe:
Most of the self-managed applications currently have a page where all configuration is documented. The Zeebe broker does not. Instead, we document a few important settings, and leave the rest out.

All configuration for Zeebe _is documented_ in the form of YAML files packaged along with Zeebe itself. That's great for the dev team and for contributors, but it's a poor UX for users:

- Hard to search
- Hard to discover
- Relatively sparse, as any documentation focuses on the specific setting and not how it relates to the application in general

As a user, I would always look at the official docs to find such things. See for example the [Kafka documentation](https://kafka.apache.org/documentation/#brokerconfigs). While still somewhat overwhelming and hard to search, it's better than our current status quo.

npepinpe:
Deepthi has also brought up (and I agree) that the table format is not optimal for this. The description column is very narrow, and if you have long names, the table has to be scrolled. The Kafka template is already a bit better, imo, but I can't say if there isn't something better than that.

akeller:
DevEx might be able to help if you want a custom table experience that's easier to view. Or maybe we can find something that isn't a table to display this info better. 🤷‍♀️ 

If you want to go that route, pull together the info you want to be displayed (either in a doc or a PR), then DevEx can iterate on the docs design. We can have a separate issue but pull you and Deepthi in for collaboration and review.

megglos:
Potential first scope:
convert yaml files to a table in the docs to start from.

@megglos check-in with Amara if this is something DevEx could do for kicking-this off

megglos:
@megglos link the relevant config files, a first iteration could be that we create doc pages for these to start with

From there on the zeebe engineers can enhance the descriptions going forward.

megglos:
@akeller I linked the relevant YAML files in the issue description, could you take it from here?
@npepinpe wdyt how we should treat the [standalone-broker case](https://github.com/camunda/zeebe/blob/main/dist/src/main/config/broker.standalone.yaml.template)? Would it be enough to mention inside the Broker Configuration that the Gateway config can just be nested to setup and embedded Gateway?

npepinpe:
It's probably enough initially...except that some of the config gets overwritten when it's embedded 😅 

It's probably good enough for the first iteration, and we can also specify which settings specifically are ignored/overwritten when the gateway config is embedded. I'm at QCon now, I can look them up after, unless you do it 🙃

akeller:
[...]

PR #1865, opened 2023-03-28
Add broker and gateway YAML to docs

akeller:

Addresses #1749, possibly not completely. 

No.

AFAIK, there is no hard requirement but this would be a nice to have in line with 8.2.

## PR Checklist


akeller:
I'm about to break for the day, but I wanted to push my changes up to a remote branch.

The [template YAML](https://github.com/camunda/zeebe/blob/main/dist/src/main/config/broker.yaml.template) as-is was hard to consume. Nearly everything is commented with `#` ([except flush](https://github.com/camunda/zeebe/blob/c3881f795f7c894588b45b7d8ee1b28fdf975236/dist/src/main/config/broker.yaml.template#L370)?) I had to start by copying the file locally and removing the comments for it to be valid YAML.

From there, I took the lazy(ish) approach and used https://github.com/ishswar/yaml-to-table, which didn't work without some changes. Instead of copying the descriptions, it added ipsum... ugh.

I removed one of the columns to match the [Tasklist experience](https://docs.camunda.io/docs/next/self-managed/tasklist-deployment/tasklist-configuration/#settings-to-connect-and-import). I plan to match the experience in Tasklist, separating out the giant table and using the dot notation (may not be the technical term) instead of the spacing you would usually see in the YAML. I'll also add the raw YAML snippets, but not the full YAML, because I want that to live on GitHub.

@npepinpe @megglos, will this work for a V1? A table with name, description, and default/example value.

npepinpe:
This is a great improvement! I wouldn't worry too much about getting it perfect, we can always improve it later. Already separating it per section like this is :rocket:

akeller:
As a V1, this gets everything into an HTML table, which doesn't look great but is better than nothing.

Once I get a clean build, I'll start converting the HTML into a markdown table so things actually render better, like in-line code. 

I think I may have found a few other oversights in the YAML. I'll do a PR over there 😄

akeller:
Markdown tables implemented! I left one HTML table specifically because it included a list and that wasn't working in an MD table.

I also intentionally did not document the experimental section or most of the deprecated config. The experimental section I'm confident should not be included, but I wasn't sure what the schedule was for deprecating. 

@megglos @npepinpe take a look and see if there is anything incorrect or something you want to be adjusted. Feel free to add a commit.

akeller:
We might want to add some additional changes, but I consider this a good V1 and mergable set of changes.

docs/self-managed/zeebe-deployment/configuration/broker.md
```diff
@@ -0,0 +1,609 @@
...
+```yaml
+exporters:
+  debuglog:
+    className: io.camunda.zeebe.broker.exporter.debug.DebugLogExporter
+    args:
+      logLevel: debug
+      prettyPrint: false
+```
+
+### zeebe.broker.exporters.debugHttp (Debug HTTP Export)
```
npepinpe, line 438: :x: Let's remove the debug exporters in general, they aren't really meant for production usage.

docs/self-managed/zeebe-deployment/configuration/broker.md
```diff
@@ -0,0 +1,609 @@
...
+
+#### YAML snippet
+
+```yaml
+exporters
+  jarPath:
+  className:
+```
+
+### zeebe.broker.exporters.debuglog (Debug Log Exporter)
```
npepinpe, line 421: :x: Let's remove the debug exporters in general, they aren't really meant for production usage.

npepinpe (dismissed):
:rocket: 

This is an amazing first version! As you said, there's stuff we can improve, but I think having this will make it much easier for everyone to contribute to that :+1: 

I've marked two things to remove, the debug exporters, but I don't think you need another review from me on that. In case you disagree, please address the comments before merging.

docs/self-managed/zeebe-deployment/configuration/gateway.md
```diff
@@ -0,0 +1,256 @@
+---
+id: gateway-config
+title: "Gateway configuration"
+sidebar_label: "Gateway configuration"
+description: "Let's analyze how to configure the Zeebe gateway"
+---
+
+A complete gateway configuration template is available in the [Zeebe repo](https://github.com/camunda/zeebe/blob/main/dist/src/main/config/gateway.yaml.template).
```
megglos, line 8: 💭 eventually the versioned configuration docs and the content this link will diverge, given main will be 8.3 anytime now. Do we have similar cases, what could be the options to keep them sync? (copy the yaml for a version into the docs? link to the stable branches based on version and only to main from `next`?)

docs/self-managed/zeebe-deployment/configuration/broker.md
```diff
@@ -0,0 +1,576 @@
...
+exporters
+  jarPath:
+  className:
+```
+
+### zeebe.broker.exporters.elasticsearch
+
+An example configuration for the elasticsearch exporter is below.
+
+These setting can also be overridden using the environment variables "`ZEEBE*BROKER_EXPORTERS_ELASTICSEARCH*`..."
```
megglos, line 425:
```suggestion
These setting can also be overridden using the environment variables `ZEEBE_BROKER_EXPORTERS_ELASTICSEARCH_*`
```

docs/self-managed/zeebe-deployment/configuration/broker.md
```diff
@@ -0,0 +1,576 @@
...
+        enabled: false
+        minimumAge: 30d
+        policyName: zeebe-record-retention-policy
+```
+
+### zeebe.broker.exporters.opensearch (Opensearch Exporter)
+
+An example configuration for the opensearch exporter.
+
+These setting can also be overridden using the environment variables "`ZEEBE*BROKER_EXPORTERS_OPENSEARCH*`..."
```
megglos, line 494:
```suggestion
These setting can also be overridden using the environment variables `ZEEBE_BROKER_EXPORTERS_OPENSEARCH_*`
```

docs/self-managed/zeebe-deployment/configuration/broker.md
megglos: the first asterisk got me confused, I assumed this might be just a formatting issue? Or am I missing something 😅

docs/self-managed/zeebe-deployment/configuration/broker.md
```diff
@@ -0,0 +1,576 @@
...
+| processing         | When the free space available is less than this value, this broker rejects all client commands and pause processing. This setting can also be overridden using the environment variable `ZEEBE_BROKER_DATA_DISK_FREESPACE_PROCESSING`                                                                                                                                                                                                                                                                                        | 2GB           |
+| replication        | When the free space available is less than this value, broker stops receiving replicated events. This value must be less than freeSpace.processing. It is recommended to configure free space large enough for at least one log segment and one snapshot. This is because a partition needs enough space to take a new snapshot to be able to compact the log segments to make disk space available again. This setting can also be overridden using the environment variable `ZEEBE_BROKER_DATA_DISK_FREESPACE_REPLICATION` | 1GB           |
```
megglos, line 171: ❓ these two are sub-properties of `freeSpace` would it make sense to indicate that in the table e.g. by naming them `freeSpace.processing` ? (other similar cases further down as well)

docs/self-managed/zeebe-deployment/configuration/gateway.md
```diff
@@ -0,0 +1,256 @@
...
+        </tr>
+        <tr>
+            <td>jarPath</td>
+            <td>Path (relative or absolute) to the JAR file containing the interceptor class and its dependencies. All classes must be compiled for the same language version as Zeebe or lower. This setting can also be overridden using the environment variable `ZEEBE_GATEWAY_INTERCEPTORS_0_JARPATH`.</td>
+            <td></td>
+        </tr>
+        <tr>
+            <td>className</td>
+            <td>Entry point of the interceptor, a class which must:
+              <li>implement [io.grpc.ServerInterceptor](https://grpc.github.io/grpc-java/javadoc/io/grpc/ServerInterceptor.html)</li>
```
megglos, line 239: this markdown link doesn't render for me, also needs to be a html link then? 😢

megglos (approved):
I really like the sectioned approach, allows to navigate easily!  Thank you so much for getting onto this that fast! 🚀 

>  I left one HTML table specifically because it included a list and that wasn't working in an MD table.

Well that's a pity 😅 I noticed the markdown link in the table doesn't render, so needs to get converted too?

Other than that just some questions.

docs/self-managed/zeebe-deployment/configuration/gateway.md
akeller: Good point. We don't have a process for this currently. Looking at the tags in the Zeebe repo, there are many tag options (8.1.x, for example) per minor. 

Linking to main in `next` seems easiest. But every time we cut a new release, should we point to 8.x.0? We could add a note that users could look for other versions in the repo.

docs/self-managed/zeebe-deployment/configuration/broker.md
akeller: `      # These setting can also be overridden using the environment variables "ZEEBE_BROKER_EXPORTERS_[exporter name]_..."`

Above was pulled straight from the YAML, so I think it's a formatting issue.

docs/self-managed/zeebe-deployment/configuration/broker.md
akeller: Oh shoot! Good catch. The spacing was difficult to understand when removing the # from the template. I'll fix it.

akeller:
Reminder to myself to add this to the docs-on-docs - https://tableconvert.com/markdown-to-markdown

docs/self-managed/zeebe-deployment/configuration/gateway.md
akeller: fixed ✅

docs/self-managed/zeebe-deployment/configuration/gateway.md
```diff
@@ -0,0 +1,277 @@
...
+
+A complete gateway configuration template is available in the [Zeebe repository](https://github.com/camunda/zeebe/blob/main/dist/src/main/config/gateway.yaml.template).
+
+## Conventions
+
+Take the following conventions into consideration when working with the gateway configuration.
+
+### Byte sizes
+
+For buffers and others must be specified as strings and follow the following format: "10U" where U (unit) must be replaced with KB = Kilobytes, MB = Megabytes or GB = Gigabytes. If unit is omitted, the default unit is bytes.
```
christinaausley, line 16: "For buffers and others must be specified as strings..." -- what do we mean by "others" here?

docs/self-managed/zeebe-deployment/configuration/gateway.md
```diff
@@ -0,0 +1,277 @@
...
+#### YAML snippet
+
+```yaml
+longPolling:
+  enabled: true
+```
+
+### zeebe.gateway.interceptors
+
+Consider reading our documentation on interceptors first.
```
christinaausley, line 234: Should likely link to the interceptors documentation here.

docs/self-managed/zeebe-deployment/configuration/gateway.md
```diff
@@ -0,0 +1,277 @@
...
+
+Each interceptor should be configured with the values described below:
+
+<table name="interceptors" id="interceptors">
```
christinaausley, line 238: Is there a reason why this table is structured differently from the others?

christinaausley (approved):
Technical review in latest commit, with a few comments for clarification!

docs/self-managed/zeebe-deployment/configuration/gateway.md
megglos: For every supported minor release we have a `stable/8.x` branch, e.g. the newest https://github.com/camunda/zeebe/tree/stable/8.2, these could be used.

docs/self-managed/zeebe-deployment/configuration/gateway.md
akeller: Laziness and the list elements did not easily convert in markdown.

docs/self-managed/zeebe-deployment/configuration/gateway.md
megglos: from my understanding that is a general wildcard for anything that requires some data size, e.g. also size limits such as [`maxMessageSize`](https://github.com/camunda/zeebe/blame/main/dist/src/main/config/gateway.yaml.template#L53).
Buffers is essentially just an example, given this do you have a suggestion on how to best rephrase?

akeller: Maybe something like "For buffers and other data values that include sizing..."

akeller: @christinaausley do you have a suggestion for phrasing here? My suggestion still feels a bit awkward.

christinaausley: Perhaps "Buffers and data values referencing sizing must be specified as strings..."

akeller, line 16:
```suggestion
Buffers and data values referencing sizing must be specified as strings and follow the following format: "10U" where U (unit) must be replaced with KB = Kilobytes, MB = Megabytes or GB = Gigabytes. If unit is omitted then the default unit is simply bytes.
```

akeller:
@christinaausley can I get a review to push this through?

christinaausley (approved):

PR #1881, opened 2023-03-31
Java Client code example not working

jenshb:
Code example is not working because zeebeApi is not defined. By trial and error figured out, that the Zeebe Adress without Port has to be passed here.

Working code example

No

No

## PR Checklist


akeller:
@berndruecker is this behavior correct? The address doesn't include the port?

jenshb:
Hi @berndruecker,

I also thought about providing examples, an example says more than thousand words ;-)
In the code example, the zeebeAPI is written into audience and gatewayAddress.
When I set the zeebeAPI to the ZEEBE_ADDRESS provided by Operate on creating a new client (in my case 
ZEEBE_ADDRESS='0f6517db-6973-4604-af18-4f7f2583042a.dsm-1.zeebe.camunda.io:443')
I get an HTTP return code 401. 

As far as I can see, this is because the audience should not contain the port but has to be 
"0f6517db-6973-4604-af18-4f7f2583042a.dsm-1.zeebe.camunda.io". gatewayAddress works with "0f6517db-6973-4604-af18-4f7f2583042a.dsm-1.zeebe.camunda.io:443"

christinaausley:
My renaming work with APIs/clients will likely cause conflicts here. Once you've completed the change, please loop me in if you need assistance with a clean build here.

berndruecker:
Not sure if @jenshb is still working on this? I think we can merge the current proposal, still I personally would prefer an example - not the rule "gateway address without port" - but it is an improvement anyway

jenshb:
I introduced a new variable for _audience_, because it can differ from zeebeApi, and added examples. Maybe it's better now?

akeller:
@berndruecker @jenshb is this still being worked?

berndruecker:
I will merge it now - it makes total sense. @jenshb: Thanks for the improvement of the docs!

berndruecker (approved):

christinaausley:
@jenshb I think this change also needs to be applied to the next version of docs, then we can merge. Let me know if you need any assistance with this @jenshb!

christinaausley:
@jenshb @berndruecker Do we just need to add this change to versioned docs? Happy to assist with this and merge, but let me know if you were planning to make any other changes here.

jenshb:
@christinaausley I added the changes from the /version_docs directory to the /docs directory. I hope that's what you intended?

christinaausley (approved):
Thanks all! I'll enable auto-merge and we can wrap this up.

PR #1884, opened 2023-03-31
remove version banner from 8.0

akeller:

OOTB Docusaurus puts up a banner stating older docs versions are "unmaintained," which is not entirely true. 

This PR removes the banner from 8.0 since OOTB, we only have 3 options - no banner, "unmaintained," or "unreleased" (as in `next`).  

Shoutout to @remcowesterhoud for starting this discussion 😄! 

No.

This should not be merged until feedback is collected internally.

## PR Checklist

### DIFF
--- a/howtos/release-procedure.md
+++ b/howtos/release-procedure.md
@@ -65,7 +65,18 @@ To create the new versions:
 
    - Add the new mapping in the first position of the array.
 
-5. Create a PR with the changes and merge to `main`. Confirm no build issues before moving to the release steps.
+5. Ensure the "unmaintained" banner does not appear for supported versions. We currently support all versions of Camunda Platform 8 since none are older than 18 months.
+
+```javascript
+// 👋 When cutting a new version, remove the banner for maintained versions by adding an entry. Remove the entry to versions >18 months old.
+   versions: {
+      "8.0": {
+         banner: "none",
+      },
+   },
+```
+
+6. Create a PR with the changes and merge to `main`. Confirm no build issues before moving to the release steps.
 
 ### Release the new version
 
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -223,8 +223,8 @@ module.exports = {
           beforeDefaultRemarkPlugins: [versionedLinks],
           versions: {
             "8.0": {
-                banner: 'none',
-            }
+              banner: "none",
+            },
           },
         },
         blog: false,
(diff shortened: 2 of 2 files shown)


akeller:
This will require specifying exactly what older versions are still maintained (removing the banner).

This doesn't give a great visual cue to the user that they are on an older version. We would need to explore something custom.

https://user-images.githubusercontent.com/1051363/229162555-e5e179ac-75a2-4d97-8e21-da09a328c813.mov

akeller:
@pepopowitz I added you as a reviewer for the implementation, but also for the ongoing overhead. Every new versioned release we'll need to make an adjustment to include the previous 18 month versions.

We'll also need to do something similar with Optimize, but I'd need to check with the team how many versions or months they support.

pepopowitz (commented):
Re: ongoing maintenance -- we should at least add content to https://github.com/camunda/camunda-platform-docs/blob/main/howtos/release-procedure.md#perform-a-minor-release describing the need to also update this configuration when doing a release. If we wanted to get fancy about it, we could update [the versioning script](https://github.com/camunda/camunda-platform-docs/blob/main/hacks/cutNewVersions.sh#L4-L8)  to make the change for us. 

@akeller -- would you mind adding a bit to that versioning guide about what needs to be done & why, and I'll look into adding it to the script as I prepare for the upcoming release?

howtos/release-procedure.md
```diff
@@ -65,7 +65,18 @@ To create the new versions:
...
+5. Ensure the "unmaintained" banner does not appear for supported versions. We currently support all versions of Camunda Platform 8 since none are older than 18 months.
+
+```javascript
+// 👋 When cutting a new version, remove the banner for maintained versions by adding an entry. Remove the entry to versions >18 months old.
+   versions: {
+      "8.0": {
+         banner: "none",
+      },
+   },
+```
```
pepopowitz, line 77: ❤️ ❤️ ❤️

pepopowitz (approved):
💪 Thank you!!!!!

akeller:
I was going to keep this in draft, but based on all the feedback I've received internally, no one can definitively tell me if this is confusing customers or not.

I'd like to move forward with removing the banner and then plan to remove the banner for 8.1 when 8.2 goes lives. This will clearly show 8.0, 8.1, and 8.2 are all supported and maintained (as intended).

PR #1893, opened 2023-04-04
Document Signal events

remcowesterhoud:

This PR documents the top-level signal start event and adds a short description about the Signals concept.

closes https://github.com/camunda/zeebe/issues/11198

Release 8.2

## PR Checklist


docs/components/concepts/signals.md
```diff
@@ -0,0 +1,19 @@
+---
+id: signals
+title: "Signals"
+description: "Broadcast signals to trigger signal events"
+---
+
+Signals can be broadcasted within Zeebe. These broadcasts are caught by [signal catch events](../modeler/bpmn/signal-events/signal-events.md).
+Once a signal is caught the process instance will respond accordingly.
```
korthout, line 8: 🔧 Let's start by introducing what signals are: e.g. Signals are similar to Messages. But they are also different because they trigger all matching signal events, where messages are correlated to a specific process instance.

🔧 Let's also avoid referring to Camunda Platform 8 instead of Zeebe, because Signals also exist in Web and Desktop Modeler, Operate, and Optimize.

korthout (changes requested):
Great work @remcowesterhoud ✍️ 

There are a few things that should be adjusted. I also wrote a bunch of suggestions. Please have a look.

sidebars.js
```diff
@@ -73,6 +73,7 @@ module.exports = {
         "components/concepts/workflow-patterns",
         "components/concepts/process-instance-modification",
         "components/concepts/data-retention",
+        "components/concepts/signals",
```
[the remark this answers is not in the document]
remcowesterhoud: I wouldn't want to merge it to be honest. I think there are too much differences and the page would become a mess explaining what is relevant for message and what is relevant for signals. I can definitely move it closer though!

docs/components/concepts/signals.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Signal subscriptions
+
+Signals work using subscriptions. When a process encounters a signal catch event it will create a new signal subscription.
+Ths process instance will wait until a signal with a matching name is broadcasted in Zeebe. You can define the signal name
+in the process definition.
```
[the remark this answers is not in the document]
remcowesterhoud: > While this is correct, we don't yet support this behavior. Currently, only the top-level signal start event is supported, which behaves differently. A signal subscription is opened when a process is deployed with a top-level signal start event.

I'm not sure I agree that this is a problem. We document the concept of Signals on this page. We shouldn't care about the elements and if they are support or not. That's why we have these different pages.

> We should also document what happens when a new process version is deployed (subscription is closed).

This is specific to the top-level start event. That's why it's documented on that page, instead of on this one. As I said above I think the concept of signals is separate from the events and what you mention here is in my opinion an event detail, not a concept detail.

We also don't document these things in the concept page of messages.

korthout:
@remcowesterhoud You probably have to rebase on top of main and copy all the changes to the `version-8.2` versioned docs 😢

remcowesterhoud:
> @remcowesterhoud You probably have to rebase on top of main and copy all the changes to the version-8.2 versioned docs 😢

Not sure where you're looking, but this version doesn't exist yet. I'd assume this won't be created until the 8.2 release.

korthout:
> Not sure where you're looking, but this version doesn't exist yet. I'd assume this won't be created until the 8.2 release.

Sorry for the confusion. I had seen a comment about this, but the 8.2 docs haven't been cut yet.

christinaausley (approved):
Technical review in latest commit 👍

akeller:
@remcowesterhoud @korthout, you are good to merge, but based on the conversation in the PR I wasn't sure if this was ready 😅

docs/components/concepts/signals.md
```diff
@@ -4,13 +4,15 @@ title: "Signals"
...
+process instance, whereas signals can trigger _all_ the matching signal events with a single broadcast.
+Depending on the type of [signal catch events](../modeler/bpmn/signal-events/signal-events.md) the process instance will
+respond accordingly.
 
 ## Signal subscriptions
 
 Signals work using subscriptions. When a process encounters a signal catch event it will create a new signal subscription.
-Ths process instance will wait until a signal with a matching name is broadcasted in Zeebe. You can define the signal name
+Ths process instance will wait until a signal with a matching name is broadcasted. You can define the signal name
 in the process definition.
```
korthout: ❓ Should we add a remark that a signal subscription is also opened for a deployed process with signal start event? Just for completeness.

docs/components/modeler/bpmn/signal-events/signal-events.md
```diff
@@ -21,6 +21,13 @@ Signal subscriptions only exists for the latest version of a process definition.
 process (based on the BPMN process id) will delete the old signal subscription. A new subscription is opened for the
 new deployed process definition.
 
+## Variable mappings
+
+When broadcasting a signal you can pass along variables. These variables are propagated to the process instance that's
+created and will be available at the process instance scope.
```
korthout, line 27: ❌ We need to expand this by explaining that the merging behavior can be customized using output mappings.

> When broadcasting a signal you can pass along variables. By default, all signal variables are merged into the process instance. This behavior can be customized by defining an output mapping at the signal catch event.

korthout (approved):
Thanks @remcowesterhoud and @christinaausley 

Please have a look at my comments before merging. I have one last change request and a question, but I see no reason why this could not be merged.

LGTM 👍

remcowesterhoud:
@akeller Could I ask you for an approval here? All the feedback has been discussed and processed, but the approval I had became stale after a new commit 😞 

Not a big fan of commits removing approvals 😛

akeller (approved):
> Not a big fan of commits removing approvals 😛

Same. Need to see if we can get a better balance after this minor release.

PR #1894, opened 2023-04-04
Add use cases for failing a job with variables

### LINKED ISSUE #1508: Document use case for fail job with variables
Since `8.2.0-alpha1` users can [pass variables along when failing a job](https://github.com/camunda/zeebe/issues/9404). This is also documented in the job-worker concept where we discuss [completing or failing jobs](https://docs.camunda.io/docs/next/components/concepts/job-workers/#completing-or-failing-jobs).

> When Completing or failing jobs with [variables](https://docs.camunda.io/docs/next/components/concepts/variables/), the variables are merged into the process at the job's associated task.
>
> - When Completing a job the variables are propagated from the scope of the task to its higher scopes.
> - When Failing a job the variables are only created in the local scope of the task.\

Currently, that documentation only describes the behavior. Not why a user would want to do this. We should clarify why a user might want to send along variables when failing a job:
- may help users to analyze the incident when failing the job with `retries: 0` by allowing them to provide additional details along with the fail job command
- may help users to continue where they left off when retrying the job. I.e., the job was completed partially, and the job worker is able to skip parts of the work but needs to know where to continue.

koevskinikola:

Document different examples of why failing a job with variables may be useful to users.

## PR Checklist

## Additional context
- The related feature is present only on version 8.2, so we don't need to backport the changes to older versions.

closes #1508


docs/components/concepts/job-workers.md
```diff
@@ -77,6 +77,15 @@ When `Completing or failing jobs` with [variables](components/concepts/variables
...
 - When `Failing a job` the variables are only created in the local scope of the task.
 
+:::tip Failing a job with variables
+There are several advantages when failing a job with variables. Consider the following use cases:
+
+- You can fail a job and raise an incident by setting the job `retries` to zero. In this case, it would be useful to
+  provide some additional details through a variable when the incident is analyzed.
+- You can partially complete a job and fail it. In this case, you can store store some information on the completed work
+  in a variable, so job execution may continue from where it left off.
+  :::
```
korthout, line 87: ❌ The tip ammonition must be closed correctly

```suggestion
:::
```

Screenshot 2023-04-04 at 18 05 28

docs/components/concepts/job-workers.md
```diff
@@ -77,6 +77,15 @@ When `Completing or failing jobs` with [variables](components/concepts/variables
 - When `Completing a job` the variables are propagated from the scope of the task to its higher scopes.
 - When `Failing a job` the variables are only created in the local scope of the task.
 
+:::tip Failing a job with variables
+There are several advantages when failing a job with variables. Consider the following use cases:
+
+- You can fail a job and raise an incident by setting the job `retries` to zero. In this case, it would be useful to
+  provide some additional details through a variable when the incident is analyzed.
+- You can partially complete a job and fail it. In this case, you can store store some information on the completed work
```
korthout, line 85: ❓ I wonder whether we should refer to `partially complete a job and fail it` differently because it could imply that partially completing a job is something the job worker should do, just like completing a job. But I don't know what wording would be more fitting.

docs/components/concepts/job-workers.md
```diff
@@ -77,6 +77,15 @@ When `Completing or failing jobs` with [variables](components/concepts/variables
 - When `Completing a job` the variables are propagated from the scope of the task to its higher scopes.
 - When `Failing a job` the variables are only created in the local scope of the task.
 
+:::tip Failing a job with variables
```
korthout, line 80: 👍 I like that you made a tip out of this!

korthout (changes requested):
Thanks @koevskinikola 👏 

Please have a look at my comments

christinaausley (dismissed):
Technical review in latest commit -- this is only for the `next` version of docs, correct?

docs/components/concepts/job-workers.md
christinaausley: I agree here. Thoughts @koevskinikola? Happy to assist based on what you are trying to say here.

docs/components/concepts/job-workers.md
christinaausley: Adjusted in latest commit 👍

docs/components/concepts/job-workers.md
koevskinikola: That's a good point. Instead of:
> You can partially complete a job and fail it.
how about?
> You can fail a job if it can't be successfully completed.

It's a bit difficult to write something that doesn't insinuate that the job worker should be able to do partial work.

@christinaausley the idea is that in some use cases, the work can be split into more standalone parts. For example, the work may include a calculation that is then used for some other step that couldn't be done at that point in time. The user can then store this intermediate data in a variable and retrieve it when they retry the job. Of course, this "partitioning" of the job execution is completely up to how the user implements their job worker, so we shouldn't indicate in our documentation that we expect that in any way.

korthout: How about something like:

```
If your job worker can split the job into smaller pieces and finish some but not all of these, then it can fail the job with variables to indicate which parts of the job were successfully finished and which parts weren't. Such a job should be failed with a positive number of retries so another job worker can pick it up again and continue where the other job worker left off. The job can be completed when all parts are finished by a job worker successfully.
```

christinaausley: Nice @korthout!

akeller:
> Technical review in latest commit -- this is only for the `next` version of docs, correct?

From PR description:
> The related feature is present only on version 8.2, so we don't need to backport the changes to older versions.

docs/components/concepts/job-workers.md
akeller: Was this thread resolved?

christinaausley: I added this in my latest commit. Happy to go back and edit if you have additional comments @koevskinikola. Merging this in!

christinaausley (approved):

PR #1908, opened 2023-04-05
[DO NOT MERGE] RFC: Embedded OpenAPI docs for Operate API

pepopowitz:
Addresses https://github.com/camunda/developer-experience/issues/47

This PR is an RFC (Request For Comments).

It will not be merged as-is. It is intended to provide a real-life code sample and working website, to prove the concept of embedding OpenAPI docs into our C8 documentation, with the goal of soliciting feedback. 

The PR will be closed when I feel I've gathered enough feedback and information to proceed with this line of work. 

Introduces the [redocusaurus plugin](https://github.com/rohit-gohri/redocusaurus), to provide generated API docs for the Operate API. 

A [preview site is available on Netlify](https://fanciful-basbousa-18e705.netlify.app/docs/next/apis-clients/operate-api/openapi-withnav/).

* Two formats are provided:
   1. [A page that includes the left sidebar](https://fanciful-basbousa-18e705.netlify.app/docs/next/apis-clients/operate-api/openapi-withnav/). Note that there is an overflow thing happening with the _right_ sidebar (which shows input/output details), but I can only make that look nice in either this format _or_ the next format:
   2. [A page that does not include the left sidebar](https://fanciful-basbousa-18e705.netlify.app/docs/next/apis-clients/operate-api/openapi-nonav/).
* The only API currently implemented is the Operate API. 

## What it looks like

But really, you're better off visiting the preview site.

image

## Implementation notes

1. This plugin **does not** provide interactive functionality. If we like this plugin, it would be an end-state in our goal toward improving API docs, _or_ we would need to start over with another approach to move further. 
   * Note that I tried to integrate [the most recommended plugin](https://github.com/PaloAltoNetworks/docusaurus-openapi-docs) in the community, which _does_ include interactivity, [but was unsuccessful](https://github.com/camunda/developer-experience/issues/47#issuecomment-1490507483).
   * We should think about this PR as a level of improvement. It might not provide interactivity, but does it make our API docs better? If so, that's valuable, and we should move forward with it, and iterate on further improvement afterward.

## Feedback I'm looking for

1. What, generally, do you think of this?
2. Are these reference docs an improvement over [the current documentation](https://docs.camunda.io/docs/next/apis-clients/operate-api/)?
3. Do you have a preference between the docs with or without the navigation sidebar? 
   * With the sidebar gives us nested left sidebars, which is not very friendly.
   * Without the sidebar makes it easier to get lost while viewing the docs.
   * I can add content to the top of the page that doesn't have the nav sidebar, if you think that would be a useful way to retain context for the user.
   * We could also consider adding a top-nav link directly into these API docs, if that's useful.
4. Do you think these API reference docs should live within the current API sections of the site? Would they make more sense under their own "API" section, like [in the docusaurus docs](https://docusaurus.io/docs/2.3.1/api/plugins/@docusaurus/plugin-content-docs)?
5. Are there any absolutely-blocking UX issues you see with this solution?

## Next steps

If feedback is positive, I will implement this for all our RESTful APIs, in the "next" version. 

If it isn't, I'll pursue other approaches.


docusaurus.config.js
```diff
@@ -57,7 +82,7 @@ module.exports = {
       isCloseable: true,
     },
     prism: {
-      additionalLanguages: ["java", "protobuf"],
+      additionalLanguages: ["java", "protobuf", "scala"],
```
pepopowitz, line 85: I don't know _why_ I had to add this to prevent the plugin from crashing. [It was an issue in previous versions](https://github.com/rohit-gohri/redocusaurus/issues/183), but it appears that the issue was resolved. 🤷

docusaurus.config.js
```diff
@@ -222,11 +247,11 @@ module.exports = {
...
-          versions: {
-            "8.0": {
-              banner: "none",
-            },
-          },
+          // versions: {
+          //   "8.0": {
+          //     banner: "none",
+          //   },
+          // },
```
pepopowitz, line 254: commented out because I removed all versions except the most recent ones (8.1 / 3.9.0)

redocly.yaml
```diff
@@ -0,0 +1,24 @@
...
+    theme:
+      colors:
+        primary:
+          main: "#c54302"
+      typography:
+        fontFamily: "IBM Plex Sans, -apple-system, blinkmacsystemfont, Segoe UI, roboto, oxygen-sans, ubuntu, cantarell, Helvetica Neue, sans-serif;"
+      sidebar:
+        activeTextColor: "#c54302"
+        groupItems:
+          fontWeight: bold
```
pepopowitz, line 24: We can do _some_ customization here. Anything [on this page](https://redocly.com/docs/api-reference-docs/configuration/theming/) marked with "Supported in Redoc CE".

optimize_versions.json
```diff
@@ -1 +1 @@
-["3.9.0", "3.8.0", "3.7.0"]
+["3.9.0"]
```
pepopowitz, line 1: Removed old versions just to speed up the build.

docusaurus.config.js
akeller: 😵 So weird but 🤷‍♀️

Sijoma:
1. What, generally, do you think of this?
A: I really like it. Especially the version without the nested sidebar.

2. Are these reference docs an improvement over [the current documentation](https://docs.camunda.io/docs/next/apis-clients/operate-api/)?
A: Yes. As a developer I immediately see "ahh this is what I need".
 
3. Do you have a preference between the docs with or without the navigation sidebar?
A: Without. The nesting is just too much and there is not enough space to consume the information of the API with the sidebar. 
   
4. Do you think these API reference docs should live within the current API sections of the site? 
A: What are "API reference docs" exactly? Does that mean "embedding OpenAPI docs" == "API reference docs"?
     If this is meant then I think they should live in the current API section. The OpenAPI docs are IMHO the most useful thing a dev can consume to integrate with an API.

akeller:
I _think_ I like the no navigation experience better because you don't have nested sidebars. The nested sidebars feel like an iframe experience and I'd rather not have the sidebar context and use the top nav than deal with iframes. 

> We should think about this PR as a level of improvement. It might not provide interactivity, but does it make our API docs better? If so, that's valuable, and we should move forward with it, and iterate on further improvement afterward.

I cannot stress this enough. We are looking for levels of improvement or increments. I have no hard requirement that V1 _must_ have interactivity. I hesitate to call this an experiment because I don't want it to seem like it's throw away work, but you can see how it resonates with our community and customers while you look at other levels of improvement.

I *do* think the value here is in API references docs looking and feeling like I expected. OpenAPI or SwaggerUI-based experiences >>> text-based experiences. Things are color coded for GET/POST or errors. I can see all request info, required or not, and responses, including errors.

pepopowitz:
> 4. Do you think these API reference docs should live within the current API sections of the site?
>    A: What are "API reference docs" exactly? Does that mean "embedding OpenAPI docs" == "API reference docs"?
>    If this is meant then I think they should live in the current API section. The OpenAPI docs are IMHO the most useful thing a dev can consume to integrate with an API.

@Sijoma good catch -- I changed my name for it halfway through the body of my PR 😅 . Yes -- by "API reference docs", I meant the same thing as "embedding OpenAPI docs".

pepopowitz:
@christinaausley when we hang out tomorrow, I'd like to talk about this RFC a bit.

I think my plan with this is to implement it without the docusaurus sidebar navigation. When we meet, I'd like to brainstorm ideas on how we can retain some sort of context for the user, since it can be easy to get lost with no sidebar nav.

pepopowitz:
A few thoughts about maintaining context, from conversation with @christinaausley : 

1. Can we put a breadcrumb component at the top of the page?
2. Can we do a dropdown nav item for APIs & Clients that includes the main APIs?
3. We can reach out for feedback from community after shipping to see if they like/dislike the experience

pepopowitz:
I'm going to close this PR/RFC, as I think I have enough feedback to move forward.

My intent is to open a new PR with [the experience that contains only the API spec sidebar](https://fanciful-basbousa-18e705.netlify.app/docs/next/apis-clients/operate-api/openapi-nonav/).

I'll also investigate [the couple of ideas](https://github.com/camunda/camunda-platform-docs/pull/1908#issuecomment-1507629359) from Christina about maintaining context. I'll decide later whether those are part of the initial implementation PR, or if they'll come as follow-ups.

Thanks all for your feedback!!!

pepopowitz:
> A few thoughts about maintaining context, from conversation with @christinaausley :
> 
> 1. Can we put a breadcrumb component at the top of the page?
> 2. Can we do a dropdown nav item for APIs & Clients that includes the main APIs?
> 3. We can reach out for feedback from community after shipping to see if they like/dislike the experience

1. Not really, not without a significant amount of overriding base Docusaurus components. Docusaurus is not interested in showing _some_ contextual navigation without showing _all_ contextual navigation. In this case, it means we have to show the sidebar if we want the breadcrumb.
2. Yes, this is much easier! Here's an example: 

   image

   I generated this by updating the docusaurus.config.js, navbar.items property. Here's my local diff:

   image

   Not surprisingly, I think the part we'll need to give the most thought and consideration, whether we choose to do this or not, is the information architecture. Where should these specs fit in the sidebar navigation, file system, and top nav? This will be a fun thing for me and @christinaausley to work out.

PR #1911, opened 2023-04-06
docs: added connectors dependency onto helm charts diagram

igpetrov:

docs: added connectors dependency onto helm charts diagram

TBD

## PR Checklist


aabouzaid (changes requested):
@igpetrov 
It looks great but the font changed.
we are using `IBM Plex Sans` font.

and when I open the diagram it loaded correctly on my browser (so the XML of the diagram has it)
so probably it should just be imported to diagrams.net :thinking:

aabouzaid (approved):
Looks better now, thank you :+1:

akeller (approved):

PR #1918, opened 2023-04-07
8.1 apis tools rename

christinaausley:

APIs/Clients -> APIs/Tools rename for `8.1` docs. @pepopowitz Should I also make this change in the `3.9.0` version of Optimize, or would you recommend handling Optimize adjustments in a separate PR?

## PR Checklist


pepopowitz:
> Should I also make this change in the 3.9.0 version of Optimize, or would you recommend handling Optimize adjustments in a separate PR?

@christinaausley I think making the changes in optimize as part of this PR makes the most sense -- I don't think making an incremental transition will save you any work, in fact I think it will result in more.

akeller (commented):
Building this locally looks and works like I expect, however I want to mention the following:

* The top nav still says "APIs & Clients."
  * I assume that's intended in this PR.
* The sidebar and overview page title still say APIs & Clients
  * Feels like this should be changed in this PR(?)
* The URL is `http://localhost:3001/docs/apis-tools/working-with-apis-tools/`, as intended

I clicked around in Optimize docs and didn't see anything unexpected.

christinaausley:
Bad news bears on this one @pepopowitz -- will I just need to adjust all of the URL-based links here?

static/.htaccess
```diff
@@ -8,6 +8,11 @@ Options -Indexes -MultiViews
 
 # Redirect pages - https://httpd.apache.org/docs/current/mod/mod_rewrite.html#rewriterule
 
+# 8.1 apis/clients to apis/tools change
+RewriteRule ^docs/apis-clients(.*)$ /docs/apis-tools$1 [R=301,L]
+RewriteRule ^optimize/apis-clients(.*)$ /optimize/apis-tools$1 [R=301,L]
+RewriteRule ^docs/apis-clients/working-with-apis-clients/?$ /docs/apis-tools/working-with-apis-tools/ [R=301,L]
```
pepopowitz, line 14:
```suggestion
RewriteRule ^docs/apis-clients/working-with-apis-clients/?$ /docs/apis-tools/working-with-apis-tools/ [R=301,L]
RewriteRule ^docs/apis-clients(.*)$ /docs/apis-tools$1 [R=301,L]
RewriteRule ^optimize/apis-clients(.*)$ /optimize/apis-tools$1 [R=301,L]
```

The current link-checking issue is because the more generic rule is matching, before it can get to the more specific rule! So `/docs/apis-clients/working-with-apis-clients/` is being redirected to `/docs/apis-tools/working-with-apis-clients/` -- the new folder is correct, but the new filename isn't changing. 

If we move the more-specific rule first, then that one will be caught instead.

static/.htaccess
```diff
@@ -22,8 +27,6 @@ RewriteRule ^docs(/(8.0|next))?/components/modeler/desktop-modeler/element-templ
 # Added Self-Managed Modeler redirect
 RewriteRule ^docs/next/self-managed/web-modeler/installation/(.*)$ /docs/next/self-managed/modeler/web-modeler/installation/$1 [R=301,L]
 
-# Renamed the folder and the file name. This page didn't exist before 1.2. 
-RewriteRule ^docs(/(8.0|1.3|1.2|next))?/apis-clients/working-with-apis-clients/?$ /docs$1/apis-tools/working-with-apis-tools/ [R=301,L]
```
pepopowitz, line 26: I _think_ we're fine removing this rule in this PR, but we might find that we need to put it back in for other versions until they're all merged. If we do, we could do one of two things: 

1. Leave the rule as-is, because in this version it will never be hit, because the rules we added above would capture it first, or 
2. Rewrite this rule to require a version (remove the `?` after the list of versions). 

Let's do nothing for now, but just in case other versions start failing after this merges, this might be the issue!

docusaurus.config.js
```diff
@@ -7,7 +7,7 @@ module.exports = {
   url: "https://docs.camunda.io",
   // baseUrl: "/camunda-cloud-documentation/",
   baseUrl: "/",
-  onBrokenLinks: "throw",
+  onBrokenLinks: "warn",
```
akeller, line 10: Just noticing this since it's different from the other PRs. Do we want to keep this config like this in the PR?

christinaausley: Do we want this on @pepopowitz? I know it was being weird on build.

pepopowitz: We turned this on on-purpose! I can't remember why. There was a conversation about some pages that were failing the build (pre-link-checking), and we had intended to clean them up later.

I don't see that conversation in this PR -- does anyone remember it, and know where it is? Or, @christinaausley do you know if the pages in question were moved to the correct place, and we can now remove this? 

I'll also try running a build locally with it set back to "error" and see if this one still runs fine, in which case I'd say YES to setting back to "error".

christinaausley: This was in [Slack](https://camunda.slack.com/archives/C026U8GBNSW/p1680893718374489?thread_ts=1680867949.253929&cid=C026U8GBNSW)!

akeller: I think we want to revert this to "throw".

akeller: As in throw the error.

akeller, line 10:
```suggestion
  onBrokenLinks: "throw",
```

pepopowitz: It runs locally with no issue when reverted to `throw`, so I think whatever was originally inspired us to change this has since been resolved.

akeller (approved):

docusaurus.config.js
christinaausley: Ah, @pepopowitz we may have to keep as-is tentatively as this failed.

pepopowitz (approved):

PR #1941, opened 2023-04-11
dx: speed up link-checking in build-docs workflow

pepopowitz:

Introduces [lychee](https://github.com/lycheeverse/lychee) for link-checking in the build-docs workflow, to decrease build time and give more predictable results.

### Why?

As we continue to add content and versions, we've basically maxed out the capabilities of the `linkcheck` tool we were using. It gives connection errors more often than not, resulting in flappy builds.

We basically can't move forward with our release without doing one of the following: 

1. Merge without a green build
2. Remove link-checking entirely
3. Replace link-checking with something faster, even it isn't as thorough

### Alternatives considered

* I looked a lot at [htmltest](https://github.com/wjdp/htmltest). I actually liked it better in some ways -- much more configuration options, especially.
  * In the end I settled on lychee because I liked that it emits [a summary on green runs](https://github.com/camunda/camunda-platform-docs/actions/runs/4662970162/jobs/8253881148). [Htmltest succeeds almost silently](https://github.com/camunda/camunda-platform-docs/actions/runs/4662937125/jobs/8253815246), which is fine but sometimes makes me wonder if it's actually doing anything.
  * Lychee is significantly faster than htmltest (~10s vs 1m30s) on green runs. On red runs, it depends on how many pages fail - lychee seems to slow down when it has a lot of info to emit. I saw it [as slow as about 2m](https://github.com/camunda/camunda-platform-docs/actions/runs/4663276481/jobs/8254459662), for a build that had about 2000 broken links. 
    * I think this occasional slower build, which is still significantly faster than our current situation, is worth it for the much-faster green builds.
  * I also had to do a bit of trickery with htmltest, [dumping the sitemap links as anchors](https://github.com/camunda/camunda-platform-docs/blob/f292bd67ed11c1d50026323b83269e3db06f8737/.github/workflows/build-docs.yaml#L40-L43) into an html page, because it requires an html file to run. Avoiding that trickery will be easier to maintain.

* There's a nice list of alternatives [here](https://github.com/marketplace/actions/hyperlink-link-checker#alternatives). I was looking specifically for speed, so I focused on the faster options.
  * muffet is an option listed there that seemed worth investigating, but it doesn't accept multiple URLs, and has to be called once per URL. That was a dealbreaker for me.
 
### But what are we losing by switching?

Yeah, this is the stinky part. Neither of the tools I looked at _crawl_ the URLs. They only hit the URLs you give them. This is part of the reason they're so much faster.

I think this tradeoff is acceptible, because: 

1. Right now, _we can't build anything_. We aren't getting reliable crawling anyway.
2. If we can get markdown-link-checking re-enabled soon, and then also add automated enforcement that links target .md/.mdx files instead of URLs, we'll catch the majority of links _within_ the site. 
3. The one place not covered by 2 (that I can think of) is links that cross documentation instances -- from optimize to docs, and vice versa. Those links _can't_ target by .md/mdx file, because the instances don't know about each other's docs. 
4. We should leave the scheduled check-links workflow running daily. This might help us notice broken links in 3. 
5. Longer-term, we can look into something that crawls the site, but faster and without crashing the build. I'm sure there's something out there that I haven't looked at yet.

### What I think we should do with this PR.

1. Merge it into `main` soon. 
2. Merge it into #1940 and then release a version.
3. Merge it into #1938, and re-enable errors on broken markdown links, before merging that to `main`. This would get us to a relatively safe place for future development.
4. Not necessarily immediately, figure out how to _crawl_ faster and without crashing the build.

## PR Checklist


akeller (approved):
Thanks for this @pepopowitz and so sorry you had to work into your evening ☹️ 

This makes the most sense both for today and with what you've outlined.

PR #1944, opened 2023-04-11
chore(zeebe): update of dependencies and licenses

megglos:

This updated the license books for 8.0, 8.1 and 8.2.0.
I created them manually via running:
`mvn org.codehaus.mojo:license-maven-plugin:2.0.0:aggregate-download-licenses -Dlicense.excludedScopes=test,system,provided -Dlicense.excludedGroups=io.camunda -Dlicense.sortArtifactByName=true -B --fail-at-end`

and converting the produced XML to markdown with some WIP code that I'll take forward into automatic this for future releases.

Closes https://github.com/camunda/camunda-platform-docs/issues/1639

No

Either with or right after the 8.2.0 release.

## PR Checklist


oleschoenburg:
@megglos The command looks fine to me :+1: 

I'm a little bit confused how you derived the markdown from the generated licenses.xml file though. If I checkout `8.2.0` and run `mvn org.codehaus.mojo:license-maven-plugin:2.0.0:aggregate-download-licenses -Dlicense.excludedScopes=test,system,provided -Dlicense.excludedGroups=io.camunda -Dlicense.sortArtifactByName=true -B --fail-at-end`, the licenses.xml contains many more dependencies that are not present here: https://github.com/camunda/camunda-platform-docs/blob/cdb4fae0c48bf386283c25b57476e3a1304ebbad/docs/reference/dependencies.md
Just as an example, we apparently have a transitive dependency on `org.apiguardian:apiguardian-api:1.1.2` :
```
    <dependency>
      <groupId>org.apiguardian</groupId>
      <artifactId>apiguardian-api</artifactId>
      <version>1.1.2</version>
      <licenses>
        <license>
          <name>The Apache License, Version 2.0</name>
          <url>http://www.apache.org/licenses/LICENSE-2.0.txt</url>
          <file>the apache software license, version 2.0 - license-2.0.txt</file>
        </license>
      </licenses>
    </dependency>
```

This is not listed as a dependency. Looking at the tree with `mvn dependency:tree -Dverbose=true` it _looks_ like this is a test only dependency. Did you do some additional filtering on `licenses.xml` to remove test-only dependencies? 

I'm a bit confused why we would get different results here.

megglos:
> I'm a bit confused why we would get different results here.

I missed to mention I only ran the command on the `dist` module, assuming this is what we deliver.

oleschoenburg:
> I missed to mention I only ran the command on the dist module, assuming this is what we deliver.
Might that cause the diff you noticed?

Oh yeah, I guess that's it. Keep in mind we're also delivering other modules though, e.g. the java client, the exporter api etc.

oleschoenburg (approved):

christinaausley (approved):

Issue #1876, opened 2023-03-30
Detect improperly formatted links


pepopowitz:
Part of #1742 

Currently, it is possible (and easy) to introduce a link within a doc that violates [our stated preferences](https://github.com/camunda/camunda-platform-docs/blob/main/howtos/documentation-guidelines.md#internal-links). 

Example: [in this PR](https://github.com/camunda/camunda-platform-docs/pull/1871/files#diff-46beea7ce44c9e12931cfee403d37a9b0a94f2c16c31e1fb9fddeffe7d63526bR63), we are introducing links that are hardcoded to the "current" version, despite the changes living in the "next" version. 

We should implement a CI check that confirms that any links within changes follow [the stated guidelines](https://github.com/camunda/camunda-platform-docs/blob/main/howtos/documentation-guidelines.md#internal-links).

pepopowitz:
Implemented in #2675

PR #1973, opened 2023-04-17
dx: use hyperlink to crawl site looking for broken links

pepopowitz:

During last week's release debacle, we [replaced our very-slow link-checking with fast link-checking](https://github.com/camunda/camunda-platform-docs/pull/1941/)....but in doing so, we lost the checking of links within the site. The replacement tool does not crawl the site, it only looks directly at the URLs you give it. 

This PR re-introduces crawling of the site to find broken links, using the [`hyperlink`](https://github.com/untitaker/hyperlink) tool. 

This is much faster than our previous crawling link-checker. One of the reasons is that it does not use HTTP to identify broken links -- it looks at the .html source files directly. This is slightly less accurate, because it does not account for server redirects....but assuming it doesn't uncover a ton of internal links that were depending on redirects, I think the slight loss in accuracy is worth it for the immense speed gains.

## PR Checklist

### DIFF
--- a/.github/workflows/build-docs.yaml
+++ b/.github/workflows/build-docs.yaml
@@ -55,7 +55,7 @@ jobs:
         with:
           args: "product-links.prod.txt"
           fail: true
-      - name: crawl site
+      - name: Check internal links
         uses: untitaker/hyperlink@0.1.27
         with:
           args: build/
--- a/.github/workflows/build-docs.yaml
+++ b/.github/workflows/build-docs.yaml
@@ -55,3 +55,7 @@ jobs:
         with:
           args: "product-links.prod.txt"
           fail: true
+      - name: crawl site
+        uses: untitaker/hyperlink@0.1.27
+        with:
+          args: build/


akeller:
> it looks at the .html source files directly

Do you have examples of what kinds of URLs fail and what sneak by? I saw the one failure, which prompted me to look at my file, where I saw I didn't fully clean up the fully qualified URLs 🙄 which the link-check didn't catch, but my own eyeballs did. This would help me approve since I don't have the time to test malformed URLs myself.

The one it did catch in the failed build is sneaky, so that made me feel good about what it DOES catch!

pepopowitz:
> Do you have examples of what kinds of URLs fail and what sneak by? I saw the one failure, which prompted me to look at my file, where I saw I didn't fully clean up the fully qualified URLs 🙄 which the link-check didn't catch, but my own eyeballs did. This would help me approve since I don't have the time to test malformed URLs myself.
> 
> The one it did catch in the failed build is sneaky, so that made me feel good about what it DOES catch!

My understanding is that it checks all .html files in the built site for **internal** links only, and checks the file system to make sure a .html file exists at that linked location. 

It doesn't check **any** external links, whether ours or someone else's. It doesn't make HTTP requests at all. 

Re: the fully qualified URLs -- long-term, I'd expect those to be caught by #1876 instead of this crawler.

akeller:
Ok, this makes sense. We do have some paces with slugs that may or may not match the id/file name - 

filename - create-account.md
id - create-account
slug - /guide/getting-started/

I believe we could reference this file with the filename or the slug, but I corrected many of these recently because something wasn't working for Docusaurus. Maybe it's nothing, but can you have a look at how the linking works with this?

pepopowitz:
> Ok, this makes sense. We do have some paces with slugs that may or may not match the id/file name -
> 
> filename - create-account.md id - create-account slug - /guide/getting-started/
> 
> I believe we could reference this file with the filename or the slug, but I corrected many of these recently because something wasn't working for Docusaurus. Maybe it's nothing, but can you have a look at how the linking works with this?

I'm unclear whether you're asking about this link-checker, or other linking issues you've experienced with docusaurus not finding documents. Do you have a specific example you can point me to for clarity?

akeller:
I don't understand if we need multiple internal link-checking tools or if this and the Docusaurus one are sufficient. Without HTTP, what should we look for in reviews, or should we just wait for someone to say a link failed if it wasn't caught?

`/guides/create-account.md` vs. `/docs/guides/getting-started/` 

This link occurs in `zeebe-installation.md`. It works, so it appears to be in a valid format, although I would also expect it to work with the slug shortened to `/guides/getting-started/`, but it doesn't because clearly, that's not how slugs work. Docusaurus catches this on build. I've answered my own question with this.

Looking at the failure on this PR, the link is currently `/docs/apis-tools/java-client/index/`, which Docusaurus did **not** catch, but this link check did. This led me to think - do we have any idea what malformed links may be missed by both, or do we care?

pepopowitz:
The docusaurus docs do a relatively thorough job of explaining the difference between linking via file path vs linking via URL: https://docusaurus.io/docs/2.0.1/markdown-features/links.

As recommended by them, our goal should be to use file-path linking whenever possible. (I wish there was an anchor to send you to on that page, but the bulleted list at the end of the doc explains why.)

File-path linking _should_ be possible in most cases, with two notable exceptions: 

1. links that leave the docs site
2. As the red warning at the bottom of that doc describes, when crossing plugin instances. This is significant to us, because of our two docs instances -- one for the main docs, and one for the optimize docs. Those links that cross instances must be URL-based, because the instances can't see each other's stuff.

#1876 is intended to force all links _not_ described by those two exceptions to use file path referencing. Docusaurus would validate all of those links for us during build.

This PR will primarily catch links for exception number 2 above -- crossing instances from /docs/ to /optimize/, or vice versa.

The links described by exception number 1 (external) won't be validated at all. I am comfortable with that because (a) that's not very many links, and (b) that doesn't seem like a thing we get wrong often.


Does that make sense?

akeller:
Thank you @pepopowitz, I'm following now.

akeller (approved):

akeller (approved):

PR #2005, opened 2023-04-20
identity user guide restructure

christinaausley:

Resolves https://github.com/camunda/developer-experience/issues/50. Based on conversation and sync with @Ben-Sheppard. Currently, we have a single and very long "User guides" section which requires organization for an appropriate user experience.

See workshopped workshopped [Google Doc](https://docs.google.com/document/d/1wgOrQqLFhsvhwMB5lzZs-U5_NPylaWK_RDGTjArhxYQ/edit?usp=sharing) in collaboration with @Ben-Sheppard and @dlavrenuek.

## At a glance

Previously:
Screen Shot 2023-04-20 at 3 17 28 PM

Currently:
Screen Shot 2023-04-20 at 3 18 01 PM
Screen Shot 2023-04-20 at 3 18 28 PM

## Questions to consider

- Do we link how things are currently titled? Would we rename headers something else?
- Do we like that configuring an external identity provider and logging have been combined?

**This can go live anytime, but note that this is only for `next`. Once approved, identical changes will be launched in `8.2` in a separate PR.**

## PR Checklist


Ben-Sheppard:
Thank you for progressing this set of changes on @christinaausley it definitely feels like we're moving in the right direction! I like the way the pages are grouped, I think that there is a natural flow to the tasks. The only questions I would have would be around the naming of the sidebar/pages. I wonder if a title of "Getting started with groups" is too broad, I need to go into the page to see what subjects are actually covered, if I take the [Docker documentation](https://docs.docker.com/get-started/) as as example the titles often describe the task to me so I can decide without accessing the page if it may be relevant to me.

Screenshot 2023-04-21 at 13 40 01

Title such as "Persist the DB" or "Use bind mounts" feel quite useful, I wonder if we could apply a similar logic here in our titles, so, one task that I would achieve with a role is giving application access to a set of users so the structure could maybe be something like `Roles` -> `Assigning application access`, WDYT? (IMO we already use this approach in the `Configuration` section, i.e. `Connect to an existing Keycloak instance`)

Ben-Sheppard:
> > Title such as "Persist the DB" or "Use bind mounts" feel quite useful, I wonder if we could apply a similar logic here in our titles, so, one task that I would achieve with a role is giving application access to a set of users so the structure could maybe be something like `Roles` -> `Assigning application access`, WDYT? (IMO we already use this approach in the `Configuration` section, i.e. `Connect to an existing Keycloak instance`)
> 
> @Ben-Sheppard So essentially, you would like to break up the single pages within roles and groups into separate tasks/pages with more informative titles? I'm good with this, just wanted to clarify.

Yea I think so, rather that the path titles being generic and broad there is a maybe more focused page (and therefore title)

christinaausley:
@akeller Given how large this restructure is, I'm going to launch `8.2` changes in a second PR.

Ben-Sheppard (approved):
I think this change is a very big step in the right direction, As per your questions to consider, I'm not 100% sure that I would combine configuring an external provider with logging however I think that is something that maybe could be resolved later on?

PR #2008, opened 2023-04-20
add node sdk to next

akeller:

A nod to the Node SDK. I threw this change together at the very end of my day, so I'm confident it needs more eyes and word-smithing. 

Maybe link to this in the blog?

Certainly when the blog is out this should be live.

## PR Checklist

### DIFF
--- a/docs/apis-tools/working-with-apis-tools.md
+++ b/docs/apis-tools/working-with-apis-tools.md
@@ -107,6 +107,6 @@ Collections and APIs are manually updated and not all API functionality may be a
 
 ### Community SDKs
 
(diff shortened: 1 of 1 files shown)


akeller:
I'm also not sure I love the ordering on the page, so I'm certainly open to feedback and changes there too.

christinaausley (approved):

christinaausley:
> I'm also not sure I love the ordering on the page, so I'm certainly open to feedback and changes there too.

I think this largely is based on the order in which we want the users to step through these resources, or the ones we would recommend first. Thoughts?

akeller:
@christinaausley Totally get it - I was struggling with the ordering of Postman vs. SDKs specifically, although I wasn't clear about this. I'm good to ship it as-is and if we get clear feedback we can reorder it.

PR #2018, opened 2023-04-24
[DO NOT MERGE] experiment: get production deployments working against AWS

pepopowitz:
This is in no state to merge, and it will never be!!!

I'm using this PR to experiment with getting our deployments working against a new host, this time for the production publish step.


akeller:
We are running on AWS in prod, I will close this! Thanks @pepopowitz!

PR #2026, opened 2023-04-25
identity restructure current version of docs

christinaausley:

Resolves https://github.com/camunda/developer-experience/issues/50. Based on conversation and sync with @Ben-Sheppard. Currently, we have a single and very long "User guides" section which requires organization for an appropriate user experience.

Related to https://github.com/camunda/camunda-platform-docs/pull/2005.

See workshopped [Google Doc](https://docs.google.com/document/d/1wgOrQqLFhsvhwMB5lzZs-U5_NPylaWK_RDGTjArhxYQ/edit?usp=sharing) in collaboration with @Ben-Sheppard and @dlavrenuek.

## Things I did

- Condensed files into separate sub-sections.
- Condensed short pages into one (as can bee seen in the Roles and Groups sections, for example.)
- Removed screenshots where UI can speak for itself. Will keep an eye on this based on user experience moving forward.
- Updated broken links.
- Incorporated redirects.

**This can go live anytime, but note that this is only for `8.2` and `3.10.0`.**

## PR Checklist


Ben-Sheppard (dismissed):
This is quite a large change but I have checked the code out locally to have a click around and things seem reasonable to me!

akeller (commented):
Some general feedback and questions which may not be completely in scope of this PR. Please triage them as appropriate. - 

In the User Guide sidebar, is this generally the order we would expect users to navigate through this content? I think the answer to this is yes, but I'm not a seasoned Identity user.

Roles, authorizations, and permissions feel synonymous and I'm happy to see them defined in the Concepts section, but I feel like something is still missing. Will the Concepts pages be more elaborate in another PR? Some of the pages read more like a glossary definition.

Would users benefit from a quick start guide? Or maybe its better to think about it in terms of setting up a development environment vs. a production environment. 

You've made great progress on the Identity User Guide looking much more approachable.

Ben-Sheppard:
Hey @akeller!

> In the User Guide sidebar, is this generally the order we would expect users to navigate through this content? I think the answer to this is yes, but I'm not a seasoned Identity user.

That is the intention however we don't really have empirical data on that so its more of a "this feels like a reasonable ordering"

> Roles, authorizations, and permissions feel synonymous and I'm happy to see them defined in the Concepts section, but I feel like something is still missing. Will the Concepts pages be more elaborate in another PR? Some of the pages read more like a glossary definition.

This is actually something that I discussed with @christinaausley recently, I raised the same point regarding them moving to the glossary instead as thats really what they feel like. I'm not convinced right now that anything is going to be added to those pages. This is to be addressed after the restructure.

> Would users benefit from a quick start guide? Or maybe its better to think about it in terms of setting up a development environment vs. a production environment.

There is a [first steps](https://docs.camunda.io/docs/next/self-managed/identity/getting-started/install-identity/) page in the docs already which tries to get a user going but maybe its not visible enough? Or maybe doesn't quite fit your suggestion here.

akeller:
@Ben-Sheppard @christinaausley it sounds like the two of you have this handled and I want to keep the momentum going, so I'm going to be brief and give this an approve 👍 .

The install page you mentioned is exactly what I had in mind - what's the fastest way to get up and running or what are the clearly outlined steps.

akeller (approved):

PR #2055, opened 2023-05-02
docs(connectors): clarify element templates entry

igpetrov:

- Clarification of an entry
- Bugfix

No

Any time

## PR Checklist


versioned_docs/version-8.0/guides/update-guide/connectors/060-to-070.md
```diff
@@ -0,0 +1,17 @@
+---
+id: 060-to-070
+title: Update 0.6 to 0.7
+description: "Review which adjustments must be made to migrate from Connector SDK 0.6.x to 0.7.0."
+---
+
+import Tabs from "@theme/Tabs";
+import TabItem from "@theme/TabItem";
+
+With the Connector SDK [0.7.0 release](https://github.com/camunda/connector-sdk/releases/tag/0.7.0), we made
```
christinaausley, line 10: Does this breaking change need to be noted anywhere else in the documentation for awareness?

christinaausley (commented):
I noticed some changes are made in some versions but not in others, is this intentional?

Technical review in latest commit 👍

akeller (commented):
This is a good start, but I'm still missing some things.

First, @bastiankoerber would need to clarify the naming - Connector templates vs. Connector element templates. 

Second, this page is missing the end user perspective. For new users to Camunda (and Connectors), what should we recommend? Creating a new Connector template from scratch? Or modifying an existing example? When would I need to invest time in learning more about element templates?

Using Connectors and managing Connectors is on two separate pages, but managing Connectors sits in the Modeler section. I think these should be closer and all within Connectors. This would keep the general knowledge of Connectors together. (We can consider this out of scope for this PR. I may incorporate it into the broader restructuring PR.)

The note you added is a really good catch - I think there is some confusion in how things are named in the Connectors space. "Templates" is similar to the concern I identified with "types" in my comment in the larger restructuring PR - https://github.com/camunda/camunda-platform-docs/pull/2017#issuecomment-1528055592.

versioned_docs/version-8.0/guides/update-guide/connectors/060-to-070.md
igpetrov: Yes, I clarified here and similar files.

igpetrov:
Thank you, @akeller for your review. I agree, that some terms are currently vague and/or have unclear meaning. Happy to wait for @bastiankoerber's input. Just few thoughts.

There is only one "Connector template" - it is a boilerplate code. It is already pre-configured with Connectors SDK, so a developer can build their own connector without extra configuration efforts.

> Second, this page is missing the end user perspective. For new users to Camunda (and Connectors), what should we recommend? Creating a new Connector template from scratch? Or modifying an existing example? When would I need to invest time in learning more about element templates?

We cannot make a recommendation because it is pretty much depends on users use-case. This page only explains that "we have Connectors element templates and here's an example how to use it". To address your concern, I think that we could modify [this paragraph](https://docs.camunda.io/docs/components/connectors/custom-built-connectors/connector-sdk/#connector-template) to bring more clarification when user should use element templates only, and when to build from scratch. Maybe adding entirely new page (e.g., `Building own connector`) under `Component -> Custom Connectors`. WDYT?

> Using Connectors and managing Connectors is on two separate pages, but managing Connectors sits in the Modeler section. I think these should be closer and all within Connectors. This would keep the general knowledge of Connectors together. (We can consider this out of scope for this PR. I may incorporate it into the broader restructuring PR.)

IDK, it would be really hard to split using and managing. For me, `Components -> Connectors -> Use Connectors -> *`, `Components -> Connectors -> Out-of-the-box Connectors -> *` are for No-, Low-code users - mostly around usage, while `Components -> Connectors -> Custom Connectors -> *` are for Low-, Pro-code users. But I don't have a strong opinion here.

In addition to that, we have an [Ops part](https://docs.camunda.io/docs/self-managed/connectors-deployment/install-and-start/) for connections. If we're restructuring, those, maybe we can do something with that as well.

igpetrov:
Also, setting `on hold` label. I will also add `inbound` connector template in this one.

akeller:
> There is only one "Connector template" - it is a boilerplate code. It is already pre-configured with Connectors SDK, so a developer can build their own connector without extra configuration efforts.

Are template-only connectors something different, then? 

I might be struggling with interpreting @bastiankoerber's vision for documentation, but as he describes it we have custom connectors made with the SDK and template-only connectors, both of which offer a customizable experience.

How do you build template-only connectors? With the connector template? Which is a special kind of element template?

igpetrov:
> Are template-only connectors something different, then?

Yes, that's correct. We also call those protocol-connectors - meaning they use Camunda-supported protocols - REST, GraphQL. Example of (element) template-only template is Asana. As you can see, there's only an [element template](https://github.com/camunda/connectors-bundle/tree/main/connectors/asana) and no `src/` folder where actual Java code resides for several other connectors (e.g., [Google Drive](https://github.com/camunda/connectors-bundle/tree/main/connectors/google-drive)). (Element) template-only connectors are targeted to low-code users.

The Connector Template is a [repository](https://github.com/camunda/connector-template) that has pre-configured Java source code that serves as foundation to build pro-code connectors.

> I might be struggling with interpreting @bastiankoerber's vision for documentation, but as he describes it we have custom connectors made with the SDK and template-only connectors, both of which offer a customizable experience.

Right. I am not entirely sure what is meant by the "custom" connectors but my best guess would be those connectors we develop with SDK, like Google Drive. If, say, customer wants to build a (custom) connector for their system, that we don't support out-of-the-box, they can do it with SDK, and customize it in whatever way they want: from introducing special business logic to applying exotic security practices.

> How do you build template-only connectors? With the connector template? Which is a special kind of element template?

We build (element) template-only connector by composing JSON. This is how I did [Asana connector](https://github.com/camunda/connectors-bundle/blob/main/connectors/asana/element-templates/asana-connector.json) before. In the nearest future, @bastiankoerber will roll out a public repository with primers and examples how to build own (element) template-only connector. For now, users can refer our [Awesome List](https://github.com/camunda-community-hub/camunda-8-connectors#out-of-the-box-protocol-connectors) for examples.

akeller:
I'm going to merge these changes into the connectors-restructure branch. My goal will be to resolve any conflicts, get the build (and redirects) working, then merge everything to `main` so the changes will be available on stage.docs.camunda.io for review.

akeller:
I've merged this into `connectors-restructure` and the #2017 PR.

Issue #1990, opened 2023-04-19
Update the "start instance" page in the Modeler docs to explain how to run or publish a process


christian-konrad:
We should update the "start instance" page in the Modeler docs so that they explain how to run or publish a process.

The current documentation (c.f. https://docs.camunda.io/docs/components/modeler/web-modeler/start-instance/) is insufficient and does not describe all the options a user has to run or publish a process.
Even without the new options recently added (webhook, public form...), the documentation does not really tell the user what deploy/start instance means to them. We also renamed "start instance" to "run", which isn't yet reflected in the documentation.

To let the documentation follow the PDLC phases, it makes sense to rename the page to "Run or publish your process", and add multiple sections to cover the following options to run/publish a process:

- Publish via Webhook
- Schedule via timer
- Listen to messages or signals
- Run manually from Modeler
- Dry run using Play mode
- Publish via link
- Publish into Tasklist

We should also merge the "Save and deploy" page into it, since it only contains a sentence saying something like "To deploy, click **deploy**"; and mentions autosave, which is easy to discover in the product itself and does not necessarily require separate documentation. We can also remove the "New Web Modeler" page, since it's now more than a year since Web Modeler is available. This simplifies the docs and helps readers to understand what's possible with Modeler and C8.

Old menu:

<img width="279" alt="Bildschirmfoto 2023-04-19 um 16 23 00" src="https://user-images.githubusercontent.com/56470362/233105503-e72a5098-ca80-4716-a173-ca298e19b07e.png">

Proposal:

<img width="281" alt="Bildschirmfoto 2023-04-19 um 16 23 50" src="https://user-images.githubusercontent.com/56470362/233105740-d6993814-980f-44da-a699-62cf85bebbc2.png">

PR #2058, opened 2023-05-03
docs(modeler): describe run and publish options

christian-konrad:

Implements #1990

* [ ] We could derive a blog post

* Soon

## PR Checklist


christian-konrad:
@akeller I need help with this. Could you please tell me how to run a style check locally only for my files, and what is required in terms of formatting to get a green light for the **`build-docs / build-docs (push) `** check? 

Running `npm run format:check`, I receive a warning for every single file in the project, and then it fails with a non-helpful error.

Bildschirmfoto 2023-05-03 um 14 44 21

christinaausley:
@christian-konrad I just ran `npm run format`, and then a `git commit -a -m "run prettier"` to get the formatting correct here 👍

christian-konrad:
> I just ran npm run format, and then a git commit -a -m "run prettier" to get the formatting correct here 👍

@christinaausley thank you! Should we document this in somewhere in https://github.com/camunda/camunda-platform-docs/blob/main/howtos/setup.md or similar?

christinaausley:
> > I just ran npm run format, and then a git commit -a -m "run prettier" to get the formatting correct here 👍
> 
> @christinaausley thank you! Should we document this in somewhere in https://github.com/camunda/camunda-platform-docs/blob/main/howtos/setup.md or similar?

I could have sworn this was documented somewhere, but I may be wrong/it may not be visible enough. I'll do some digging and launch a PR where necessary to make this clear.

christinaausley:
Looks like you just need a redirect for the former `start-instance` file -- let me know if you need any assistance with this.

static/.htaccess
```diff
@@ -8,6 +8,10 @@ Options -Indexes -MultiViews
 
 # Redirect pages - https://httpd.apache.org/docs/current/mod/mod_rewrite.html#rewriterule
 
+# Modeler start instance and deploy page renaming and merging
+RewriteRule ^docs/components/modeler/web-modeler/start-instance/(.*)$ /docs/components/modeler/web-modeler/run-or-publish-your-process/$1 [R=301,L]
+RewriteRule ^docs/components/modeler/web-modeler/save-and-deploy/(.*)$ /docs/components/modeler/web-modeler/run-or-publish-your-process#deploy-a-process/$1 [R=301,L]
```
akeller, line 13: @pepopowitz @christinaausley can we redirect to page section headers? I remember this wasn't working previously.

pepopowitz:
> > > I just ran npm run format, and then a git commit -a -m "run prettier" to get the formatting correct here 👍
> > 
> > 
> > @christinaausley thank you! Should we document this in somewhere in https://github.com/camunda/camunda-platform-docs/blob/main/howtos/setup.md or similar?
> 
> I could have sworn this was documented somewhere, but I may be wrong/it may not be visible enough. I'll do some digging and launch a PR where necessary to make this clear.

It's in https://github.com/camunda/camunda-platform-docs/blob/main/CONTRIBUTING.MD#code-formatting

> You can auto-format your code locally by running `npm run format`.

static/.htaccess
pepopowitz: I'm not sure, I've never tried it! I will try locally and see what happens.

pepopowitz: Ok, a few things about these rules that I learned!

1. I'm guessing the reason we haven't been able to redirect to page section headers is that the server encodes the `#` character as `%23`. I noticed this when running the redirects locally. 
2. But I learned that [we can prevent this by adding the `NE` flag to the rule](https://serverfault.com/a/987814). (`NE` stands for "no escape" which is very relatable 😅 ) 
3. These redirects are both incorrect for the current PR, as the changes are only made in the `next` version. So they should both be matching and targeting the `next` URLs, instead of the current version (which does not specify a version in the path). 
4. There are also issues with the wildcard matching. For the first rule, the rule works, though it could be improved to eliminate one redirect. The rule matches `start-instance/(.*)`, which means the start-instance page, _and any of its children_. This page doesn't have any children, so that's a minor point, it will just never affect anything but the main page. But there's no need to capture this group, or preserve it.

pepopowitz, line 13:
```suggestion
RewriteRule ^docs/next/components/modeler/web-modeler/start-instance/?$ /docs/next/components/modeler/web-modeler/run-or-publish-your-process/ [R=301,L]
RewriteRule ^docs/next/components/modeler/web-modeler/save-and-deploy/?$ /docs/next/components/modeler/web-modeler/run-or-publish-your-process/#deploy-a-process [R=301,L,NE]
```

Ok, a few things about these rules that I learned! This is a lot, but I want to share _why_ I'm making these recommendations.

1. I'm guessing the reason we haven't been able to redirect to page section headers is that the server encodes the `#` character as `%23`. I noticed this when running the redirects locally. 
2. But I learned that [we can prevent this by adding the `NE` flag to the rule](https://serverfault.com/a/987814). (`NE` stands for "no escape" which is very relatable 😅 ) 
3. These redirects are both incorrect for the current PR, as the changes are only made in the `next` version. So they should both be matching and targeting the `next` URLs, instead of the current version (which does not specify a version in the path). 
4. There are also issues with the wildcard matching. For the first rule, the rule works, though it could be improved to eliminate one redirect. The rule matches `start-instance/(.*)`, which means the start-instance page, _and any of its children_. This page doesn't have any children, so that's a minor point, it will just never affect anything but the main page. But there's no need to capture this group, or preserve it. 

   In reality, what we probably mean to do, is capture an optional `/` at the end of the path -- so for either `start-instance/` or `start-instance` (no `/`), redirect to the new path, with a `/` at the end. This saves the user a hop when they request `start-instance` (no `/`). Instead of getting redirected to `start-instance/` before getting redirected to `run-or-publish-your-process`, they'd go directly to the target right away. 

   All this to say, the left side should change from `...start-instance/(.*)$` to `...start-instance/?$`, and the right side to `...run-or-publish-your-process/`. 
5. The second rule is more problematic, due to including an anchor. It has the same problems described in number 4 above, but even with the `NE` flag, it redirects to an invalid URL -- `run-or-publish-your-process/#deploy-a-process/`. Note the extra slash at the end. That URL doesn't exist.

   But if we make the same changes as in number 4 -- make the left side end in `/?$` and remove the `$1` from the right side -- it goes to the correct place. A page named `run-or-publish-your-process`, and an anchor named `deploy-a-process`. 
6. This raises one last question from me - should that first rule _also_ be redirecting to a page anchor? If so, that's something I'm not accounting for in these recommendations.

Sorry for being so wordy, thanks for coming to my TED talk!

pepopowitz: See my novel at https://github.com/camunda/camunda-platform-docs/pull/2058#discussion_r1187957017

christian-konrad: @pepopowitz this is great! Thank you. It will save a lot of thought work for future restructurings. Would it make sense to document your findings in the documentation guide?

christian-konrad: Ah, it seems like you already did! https://github.com/camunda/camunda-platform-docs/pull/2097

akeller (commented):
We support versions for 18 months, which I understand makes changes to continuously releasing components or areas like Web Modeler heavy handed. 

To me, this really only makes sense for our Self-Managed users but it looks like Web Modeler is available in Self-Managed for Enterprise customers based on this admonition/note:

> Web Modeler Self-Managed is available to [enterprise customers](https://docs.camunda.io/docs/next/reference/licenses/#web-modeler) only.

Should these changes be backported to 8.0 and 8.1 as well?

akeller:
> We support versions for 18 months, which I understand makes changes to continuously releasing components or areas like Web Modeler heavy handed.
> 
> To me, this really only makes sense for our Self-Managed users but it looks like Web Modeler is available in Self-Managed for Enterprise customers based on this admonition/note:
> 
> > Web Modeler Self-Managed is available to [enterprise customers](https://docs.camunda.io/docs/next/reference/licenses/#web-modeler) only.
> 
> Should these changes be backported to 8.0 and 8.1 as well?

I also realize we may not have enforced this previously, so please don't view this as blocking.

christian-konrad:
@akeller good catch! 

> > Web Modeler Self-Managed is available to [enterprise customers](https://docs.camunda.io/docs/next/reference/licenses/#web-modeler) only.
> Should these changes be backported to 8.0 and 8.1 as well?

However, Web Modeler Self-Managed is only available as GA since 8.2

christian-konrad:
> start can be done (1) explicitly, on blank start event or (2) implicitly when an event happens

@nikku thanks, this makes sense! However, I would like to get this merged, and add this change via a separate PR.

Could you or @akeller please approve the PR so I can merge?

nikku (approved):

PR #2065, opened 2023-05-04
Make wording on supported OpenSearch version more specific

johanwelgemoed:

The decision was made to update the wording to make it more explicit that we support the AWS OpenSearch 1.3 version in 8.2.

no

as soon as possible

## PR Checklist

### DIFF
--- a/docs/reference/supported-environments.md
+++ b/docs/reference/supported-environments.md
@@ -32,14 +32,14 @@ We highly recommend running Camunda Platform 8 Self-Managed in a Kubernetes envi
 
 Requirements for the components can be seen below:
 
-| Component   | Java version | Other requirements                                                                                                                    |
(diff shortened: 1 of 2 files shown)


akeller:
@johanwelgemoed should we add this to our style guide and qualify all references of OpenSearch as AWS OpenSearch?

johanwelgemoed:
@akeller I don't think that is needed as this is only for this instance and version combination. The intent is that we want to make sure SM customers don't start using 1.3.x (which is going out of support in couple months), but then we need to keep supporting it. Making this AWS specific we don't need to worry about it as Amazon will enforce it for their customers.
Going forward we will support 2.x so for that documentation the distinction between AWS and non-AWS versions should not matter.

christinaausley (approved):

Issue #2047, opened 2023-04-28
Web Modeler API docs are duplicated


akeller:
**Duplicate content is an SEO risk.** This may inadvertently lead to both pages being de-indexed from search engines like Google and only reachable through the docs navigation.

**8.2**
* https://docs.camunda.io/docs/self-managed/modeler/web-modeler/api/
* https://docs.camunda.io/docs/apis-tools/web-modeler-api/

**next**
* https://docs.camunda.io/docs/next/self-managed/modeler/web-modeler/api/
* https://docs.camunda.io/docs/next/apis-tools/web-modeler-api/

Remove the page from the Self-Managed section and instead link to the content in APIs & Tools. If the Web Modeler API is only available for Enterprise customers, please add the banner as it appears on the Self-Managed page today.

akeller:
@christinaausley Can you follow up with the Web Modeler team on this? I'll start by posting in the channel, but include you to keep things moving.

christinaausley:
@wollefitz is going to adjust the content, and it sounds like the banner can be removed entirely.

Once the PR is launched, I'll adjust any links across docs and incorporate redirects appropriately 👍

PR #2089, opened 2023-05-09
docs(web-modeler): document Amazon Aurora PostgreSQL support

wollefitz:
Closes https://github.com/camunda/web-modeler/issues/4457

Documents Web Modeler Amazon Aurora PostgreSQL support (see https://github.com/camunda/web-modeler/issues/4457).

No.

When you release next.

## PR Checklist

- [x] My changes apply to an already released version, ~and I have added them to the relevant `/versioned_docs` directory, or they are not for an already released version.~
- [x] My changes apply to future versions, and I have added them to the main `/docs` directory, ~or they are not for future versions.~
- [x] My changes require an [Engineering review](https://github.com/camunda/camunda-platform-docs/blob/main/howtos/documentation-guidelines.md#review-process), and I've assigned an engineering manager or tech lead as a reviewer, ~or my changes do not require an Engineering review.~
- [x] My changes require a [technical writer review](https://github.com/camunda/camunda-platform-docs/blob/main/howtos/documentation-guidelines.md#review-process), and I've assigned @christinaausley as a reviewer, ~or my changes do not require a technical writer review.~


docs/self-managed/modeler/web-modeler/configuration/configuration.md
```diff
@@ -1,81 +1,24 @@
...
-
-- Provide client certificates at these locations:
-  - `myClientCertificate.pk8 -> ~/.postgresl/postgresql.pk8`
-  - `myClientCertificate.crt -> ~/.postgresl/postgresql.crt`
-- Provide root certificate at this location: `myCA.crt -> ~/.postgresql/root.crt`
-- Modify database URL: `jdbc:postgresql://[DB_HOST]:[DB_PORT]/[DB_NAME]?ssl=true`
-
-Furthermore, configure the database server to verify client certificates:
-[PostgreSQL documentation](https://www.postgresql.org/docs/current/ssl-tcp.html)
+Please refer to the [Advanced Database Configuration Guide](./database.md) for details on how to configure Web Modeler's database connection.
```
wollefitz, line 21: @jfriedenstab I decided against repeating the table with environment variables for configuring the database connection to not run into duplication issues like in #2047.

@christinaausley From a technical writer perspective, is it fine to transfer some parts of the configuration to sub-pages or would this make the guide harder to understand?

docs/self-managed/modeler/web-modeler/configuration/database.md
```diff
@@ -0,0 +1,89 @@
...
+
+Web Modeler requires a PostgreSQL database as persistent data storage (other database systems are currently not supported).
+Use the following environment variables of the `restapi` component to set up your database connection:
+
+| Environment variable                  | Description                                           | Example value                                            |
+| ------------------------------------- | ----------------------------------------------------- | -------------------------------------------------------- |
+| `SPRING_DATASOURCE_URL`               | JDBC URL of the database                              | `jdbc:postgresql://postgres.example.com:5432/modeler-db` |
+| `SPRING_DATASOURCE_USERNAME`          | Database user name                                    | `modeler-user`                                           |
+| `SPRING_DATASOURCE_PASSWORD`          | Database user password                                | \*\*\*                                                   |
+| `SPRING_DATASOURCE_DRIVER_CLASS_NAME` | [optional]<br/>Java class name of the database driver | `software.amazon.jdbc.Driver`                            |
```
jfriedenstab, line 19: Since `SPRING_DATASOURCE_DRIVER_CLASS_NAME` is optional you could add a new column "Default value" with the value `org.postgresql.Driver`.

docs/self-managed/modeler/web-modeler/configuration/database.md
```diff
@@ -0,0 +1,89 @@
...
+### AWS IAM Authentication
+
+If you'd like to use AWS Identity and Access Management (IAM) database authentication with your Amazon Aurora PostgreSQL
+instance, in addition to the adjustments described [above](#running-web-modeler-on-amazon-aurora-postgresql), follow these steps:
+
+1. Modify the `SPRING_DATASOURCE_URL` environment variable as follows: `jdbc:aws-wrapper:postgresql://[DB_HOST]:[DB_PORT]/[DB_NAME]?wrapperPlugins=iam`
+2. Remove the `SPRING_DATASOURCE_PASSWORD` environment variable
+
+Please ensure that the database user you're connecting with has the necessary permissions to authenticate via AWS IAM as
+described in the [Amazon Aurora documentation](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/UsingWithRDS.IAMDBAuth.DBAccounts.html#UsingWithRDS.IAMDBAuth.DBAccounts.PostgreSQL).
```
jfriedenstab, line 89: Do we also need to mention the `SPRING_DATASOURCE_USERNAME` here? If I remember correctly you mentioned once that it should be the IAM service account name.

docs/self-managed/modeler/web-modeler/configuration/configuration.md
wollefitz: > Hm, but you wouldn't need to repeat the table, would you?

Ah true, haven't thought about that. I now moved the table to the "Overview" page again 👍

docs/self-managed/modeler/web-modeler/configuration/database.md
wollefitz: This was already covered by "Please ensure that the database user you're connecting with has the necessary permissions to authenticate via AWS IAM ..." but you're right, explicitly mentioning `SPRING_DATASOURCE_USERNAME` here makes the configuration clearer.

docs/self-managed/modeler/web-modeler/configuration/database.md
wollefitz: Technically that's not correct though, if `SPRING_DATASOURCE_DRIVER_CLASS_NAME` is not set, Spring Boot automatically derives the driver from the JDBC URL (via [org.springframework.boot.jdbc.DatabaseDriver](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/jdbc/DatabaseDriver.java). 

I'd refrain from adding a default value here in order not to entice users to add an explicit value and thus override Spring Boot's autoconfiguration mechanism. Wdyt?

jfriedenstab: OK, good point. Then let's don't add a default value. Most users won't need to care about `SPRING_DATASOURCE_DRIVER_CLASS_NAME` anyway.

versioned_docs/version-8.2/self-managed/modeler/web-modeler/configuration/configuration.md
jfriedenstab, line 1: I just noticed that in `8.2`, the sub page is also called "Configuration" (whereas it's "Overview" in `Next`).

jfriedenstab (commented):
Looks good to me now! :) Just one minor comment.

christinaausley (approved):
We may need to apply redirects here for moved/renamed content. Let me know if you have questions on this, but it looks like the build is running clean 👀

versioned_docs/version-8.2/self-managed/modeler/web-modeler/configuration/configuration.md
wollefitz: Good find, fixed this

wollefitz:
> We may need to apply redirects here for moved/renamed content. Let me know if you have questions on this, but it looks like the build is running clean 👀

I just tested this locally and it seems like no additional redirects are necessary. http://localhost:3000/docs/self-managed/modeler/web-modeler/configuration/ (same path as before this PR) correctly leads to the "Configuration" page, same applies to the versioned docs.

christinaausley (approved):

PR #2097, opened 2023-05-09
feat: extract & expand a Guide to Writing Redirect Rules

pepopowitz:
!! This PR is currently based on an open PR, to avoid noise in the diff. We should probably merge that PR before this one, and then rely on GitHub to point it at the correct base. !!

Inspired by [a recent comment](https://github.com/camunda/camunda-platform-docs/pull/2058#discussion_r1187957017), I thought it'd be useful to more formally provide guidance for writing redirect rules.

It's very possible this is too much detail. I tried my best to avoid discussing things that are explained better elsewhere (linking to Apache docs especially), and focus on the things that are specific to our docs, while providing enough detail to be informative. Let me know if I totally missed the mark!

## PR Checklist

### DIFF
--- a/howtos/redirect-rules.md
+++ b/howtos/redirect-rules.md
@@ -1,0 +1,160 @@
+# Guide to Writing Redirect Rules
+
+Since the docs are always changing, it's possible for users to bookmark or share links to pages that move. To ensure that these users can still find their content, we use redirect rules to map old URLs to new URLs.
+
+Redirect rules are written in the [Apache .htaccess format]. The rules are applied to the production and staging sites, but not to the local development environment. All redirect rules are stored in the [.htaccess](../static/.htaccess) file.
+
+The `build-docs` workflow of each PR runs a step to verify that all links present in the production sitemap are still valid. When your build fails on a link validation step, it likely means you moved a doc and did not add a redirect rule that matches the original path.
+
+## Redirect rule guidelines
+
+### Basic structure of a redirect rule
+
+Redirect rules are written in the [Apache .htaccess format]. The basic structure of a redirect rule is:
+
+```apache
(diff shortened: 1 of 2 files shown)


pepopowitz:
> Following the steps for Docker, this would setup the full web-server, right? So you'd have to sit through the build process too? What's the advantage of this over pushing a commit up to a remote branch and letting the GitHub Action run?

* I actually didn't make any changes to that section in this PR! I just moved it from the other file into this one. I'll slap a comment on that section noting that.
* The initial `npm run build` takes as much time as pushing up a commit; but once you've built, you're able to edit the .htaccess file within the `build` folder and re-compose, which takes almost no time at all. This is a much tighter feedback loop if you're editing rules.
* The other advantage is that you can actually hit the website in your browser, allowing you to debug why a rule might not work, and see what's _actually_ happening instead of what you _think_ is happening. When CI fails for a missed redirect, it tells you that the redirect is missing; debugging locally can tell you _why_ it's missing. So maybe the word "debug" needs to be added to the docker section?

howtos/redirect-rules.md
```diff
@@ -0,0 +1,160 @@
...
+   RewriteRule ^old/path/?$ /new/path [R=301,L]
+   ```
+
+   Example of a redirect that will not double-hop:
+
+   ```apache
+   RewriteRule ^old/path/?$ /new/path/ [R=301,L]
+   ```
+
+## Testing redirect rules
```
pepopowitz, line 129: This whole section was just ported from the documentation-guidelines file, with no changes.

pepopowitz:
I realized that the changes in #2092 are actually completely unrelated to this PR, so I'm going to do some rebasing of this branch.

akeller (approved):

Issue #1273, opened 2022-09-15
Remove `/next/` from sitemap and link-checking


akeller:
The `/next/` docs are considered "under construction". We should remove them from the sitemap, and if necessary also exclude them from link-checking, as those artifacts presume content stability.

## Origin

> If we want to stop worrying about redirects for `/next/` docs, we need to update our link-checking steps to ignore URLs if they match a certain pattern. I _think_ we are looking for the previous option just to get to green, and then I should add an issue for me to exclude `/next/` URLs from link-checking in a follow-up.

_Originally posted by @pepopowitz in https://github.com/camunda/camunda-platform-docs/issues/1241#issuecomment-1248505776_

akeller:
@pepopowitz We also determined that `/next/` was in the sitemap, or pieces of it. Should we modify this issue to include removing `/next/` from the sitemap, which should also remove it from link-checking?

pepopowitz:
I think those are two different questions. 

#### Should we remove `/next/` docs from link-checking?

link-checking confirms that people/bots with old URLs are redirected. By removing `/next` from link-checking we're saying "we don't care if people/bots get redirected when things move". 

I think that's correct, because `/next/` is technically unstable and I don't think we should be concerned with people/robots getting lost. 

#### Should we remove `/next` from the sitemap? 

URLs in the sitemap support the ability to find the page in search. The presence in the sitemap doesn't _guarantee_ that it can be found, but it increased the likelihood. Removing `/next/` from the sitemap says "we don't care if the `/next/` docs are searchable". 

I _think_ that we _do_ want the `/next/` docs to be searchable....but I guess I don't have a strong feeling either way. I suspect that the people helped the most by the `/next/` docs being searchable are internal...though most updates come right before the release, so maybe that doesn't matter.

What do you think about that second question?

akeller:
> What do you think about that second question?

I'm worried `/next/` and the current docs version are competing with each other SEO-wise. The content is too similar and may get deindexed. `/next/` should be available for internal searching, but anyone coming into docs externally should come into the latest numbered version and navigate to older number versions for `/next/` manually.

pepopowitz:
That makes sense. I'm convinced; I'll update the body of this issue.

akeller:
`ignorePatterns` doesn't look like it worked... 😢 

I also realized I made a typo and added the same version twice. Oops! Will correct this in a PR shortly.

[...]

PR #2112, opened 2023-05-15
add ignorePatterns to sitemap config

akeller:

Partially addresses #1273 w/ https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-sitemap#ignorePatterns

No.

This is one of those things we can likely only test in production, and only after Google recrawls the site with the updated sitemap live.

## PR Checklist


akeller:
I purposefully did not add supported versions 8.0+, but I could be convinced to add them unless we want to do that in a second iteration just to "test" to see if these actually do get removed from the sitemap.

pepopowitz (approved):

PR #2128, opened 2023-05-19
[DO NOT MERGE] Prototype: archived versions, the `main` end

pepopowitz:
## What is the purpose of the change

Demonstrates the "active docs" end of #1173. See #2129 for the "archived version" end. 

This PR won't be merged, but it serves to provide actual code for a review of the approach & details. 

### What's happening on the "active docs" end? 

* Archived versions are not built in this site. In the real implementation, we could remove the source for all archived versions.
* Rewrite rules are updated to send all 1.3 requests to the archived 1.3 website.
* A link is added to the navbar for each archived version: 

   image

Clarifying comments will also be added in-line.

### A video demonstrating the relationship between this and the active docs 

An updated recording is at https://www.loom.com/share/e8d5c9d736554f71b66bbcf774de09e0

Issue #1173, opened 2022-08-17
🛑 BLOCKED 🛑  Archive older versions to external static sites


pepopowitz:
We should archive older versions of our docs, into separate static sites.

## Why?

Each time we add a new version to our docs, it impacts our build time -- it's a bunch more files that docusaurus needs to turn into pages. The recommendation from docusaurus, and what their docs do, is to archive old versions to separate static sites. You can see this in their version selector, which contains a section of "archived versions": 

<img width="1120" alt="image" src="https://user-images.githubusercontent.com/1627089/185242936-6a37dc55-bd8d-4f2e-8ef2-eb6aca47c929.png">

## Implementation considerations

* We should call older versions something other than "archived" in our UI.
   * Decision: we shall call them "unsupported."
* We support releases for 18 months, so anything older than that is fair game for archiving. 
* Do we want to build old versions from source files, in case we need to update in the case of security exploits or other important reasons?
   * Decision: we will build them from source, with each unsupported version in its own branch, fully isolated from the other versions.
* If we keep the source, can we leave them in the repo but have docusaurus not build them for every build?
   * See above decision.
* Docusaurus does not hang on to old source, they remove it. They rely on the fact that [Netlify creates immutable unique URLs for every build, and they capture a URL for older versions](https://docusaurus.io/docs/versioning#keep-the-number-of-versions-small).

## Progress

```[tasklist]
### Tasks
- [x] https://github.com/camunda/camunda-platform-docs/pull/2128
- [x] https://github.com/camunda/camunda-platform-docs/pull/2129
- [x] https://github.com/camunda/camunda-platform-docs/pull/2155
- [x] https://github.com/camunda/camunda-platform-docs/pull/2160
- [x] https://github.com/camunda/camunda-platform-docs/pull/2181
- [ ] https://github.com/camunda/camunda-platform-docs/issues/2183
- [ ] #2223
- [ ] https://github.com/camunda/camunda-platform-docs/issues/2184
- [ ] https://github.com/camunda/camunda-platform-docs/issues/2185
- [ ] https://github.com/camunda/camunda-platform-docs/issues/2186
- [ ] https://github.com/camunda/camunda-platform-docs/issues/2188
- [ ] https://github.com/camunda/camunda-platform-docs/issues/2189
- [ ] https://github.com/camunda/camunda-platform-docs/issues/2475
```

akeller:
This will likely get promoted to a Product Hub Epic.

akeller:
[...]

PR #2129, opened 2023-05-19
[DO NOT MERGE] Prototype: archived versions, the *archived* end

pepopowitz:

Demonstrates the "archived version" end of #1173. See #2128 for the "active docs" end.

This PR won't be merged, but it serves to provide actual code for a review of the approach & details. 

### What's happening on this "archived version" end?

* All archived versions of the docs would be deployed to the URL `https://unsupported.docs.camunda.io`, with each version running as its own web app, and accessible at a path containing its version. So this version would be available at the URL `https://unsupported.docs.camunda.io/1.3/`. 
* Because version number appears in the URL as part of the web app name, any links should no longer specify the version number. They're cleaned up in the sidebars files, as well as a couple random links that were probably incorrect to begin with.
* It builds *only* version 1.3. In the actual implementations, the source for all versions but the current archived version will be _deleted_. I didn't do that in this PR because it would be really noisy.
* There is no method in these docs to link back to the active docs. I think that's fine. 
* There are two hints for the user that they're viewing outdated documentation: the announcement banner message is updated and is no longer dismissable, and the yellow banner is displayed on every doc. Screenshots below!
* No sitemap is generated, and all pages are marked with `noindex`. A robots.txt is also added to prevent crawlers from bothering with this site.
* Redirects are updated so that only rules that were added for v1.3 remain. They're slightly modified, to account for the fact that the site would be running within the `/1.3` web app.

Clarifying comments will also be added in-line.

#### Outdated but preserved for history
* ~This version of the docs would be deployed to a 1.3-specific subdomain, like `1_3.docs.camunda.io`.~
* ~Due to the version being in the host name, this PR removes the `1.3` version from all documentation URLs.~

### Hints to the user that they're viewing unsupported docs

image

### A video demonstrating the relationship between this and the active docs 

An updated recording is at https://www.loom.com/share/e8d5c9d736554f71b66bbcf774de09e0

☝️ this video is out of date based on some recent changes, and will be updated soon.


docusaurus.config.js
```diff
@@ -39,18 +39,33 @@ module.exports = {
       {
         id: "optimize",
         path: "optimize",
+        includeCurrentVersion: false,
```
pepopowitz, line 42: this setting prevents docusaurus from building a "next" version.

docusaurus.config.js
```diff
@@ -39,18 +39,33 @@ module.exports = {
...
         routeBasePath: "optimize",
         beforeDefaultRemarkPlugins: [versionedLinks],
-        sidebarPath: require.resolve("./optimize_sidebars.js"),
+        sidebarPath: require.resolve(
```
pepopowitz, line 45: rather than moving the archived sidebars file into the current version location, I pointed the current version at the archived version's sidebars file.

docusaurus.config.js
```diff
@@ -39,18 +39,33 @@ module.exports = {
...
+          "./optimize_versioned_sidebars/version-3.7.0-sidebars.json"
+        ),
         editUrl: "https://github.com/camunda/camunda-platform-docs/edit/main/",
+        lastVersion: "3.7.0",
```
pepopowitz, line 49: tells docusaurus that optimize v3.7.0 is the version that shouldn't include a number in the URL.

docusaurus.config.js
```diff
@@ -39,18 +39,33 @@ module.exports = {
...
         versions: {
-          "3.9.0": {
-            banner: "none",
-          },
-          "3.8.0": {
-            banner: "none",
+          "3.7.0": {
+            label: "3.7.0",
+            path: "/",
+            noIndex: true,
```
pepopowitz, line 57: `noindex` adds the `noindex` meta tags to every page of this version

docusaurus.config.js
```diff
@@ -59,7 +74,7 @@ module.exports = {
     announcementBar: {
       id: "camunda8",
       content:
-        '📣 <b><a target="_blank" rel="noopener noreferrer" href="https://signup.camunda.com/accounts?utm_source=docs.camunda.io&utm_medium=referral&utm_content=banner">Sign-Up</a></b> for a free account to start orchestrating business processes today.',
+        '📣 <b><a target="_blank" rel="noopener noreferrer" href="https://signup.camunda.com/accounts?utm_source=1_3.docs.camunda.io&utm_medium=referral&utm_content=banner">Sign-Up</a></b> for a free account to start orchestrating business processes today.',
```
pepopowitz, line 77: 🤷 I updated these referral URLs because i noticed them. Not sure if we care.

docusaurus.config.js
```diff
@@ -75,7 +90,7 @@ module.exports = {
       },
       items: [
         {
-          type: "docsVersionDropdown",
+          type: "docsVersion",
```
pepopowitz, line 93: turns it from a dropdown list in the navbar to one version only

docusaurus.config.js
```diff
@@ -224,29 +239,42 @@ module.exports = {
...
+          //   "./versioned_sidebars/version-1.3-sidebars.json"
+          // ),
+          versions: {
+            1.3: {
+              label: "1.3",
+              path: "/",
+              noIndex: true,
+            },
+          },
+          includeCurrentVersion: false,
```
pepopowitz, line 253: same thing as mentioned above for optimize, but now for the main docs.

optimize_versioned_sidebars/version-3.7.0-sidebars.json
```diff
@@ -3,50 +3,50 @@
     {
       "type": "link",
       "label": "Overview Components",
-      "href": "/docs/1.3/components/"
```
pepopowitz, line 6: This is not fun, but could be automated with `sed`. These URLs are all cross-docs-instance. Since I removed 1.3 from the path of all pages, I needed to remove it in the sidebars too.

optimize_versions.json
```diff
@@ -1 +1 @@
-["3.10.0", "3.9.0", "3.8.0", "3.7.0"]
```
pepopowitz, line 1: only includes the specific archived version! Associated with this change would be removing all the source files for those other versions.

src/mdx/expandVersionedUrl.js
```diff
@@ -19,18 +19,6 @@
...
-    optimizeVersion: "3.10.0",
-  },
-  {
-    docsVersion: "8.1",
-    optimizeVersion: "3.9.0",
-  },
-  {
-    docsVersion: "8.0",
-    optimizeVersion: "3.8.0",
-  },
```
pepopowitz, line 33: By removing these, the logic maps 1.3/3.7.0 to the non-versioned URLs when crossing instances.

src/pages/index.js
```diff
@@ -174,7 +174,9 @@ function Home() {
                   "button button--outline button--secondary button--lg button--hero get-started-use-case",
                   styles.getStarted
                 )}
-                to={useBaseUrl("docs/guides/orchestrate-microservices/")}
+                to={useBaseUrl(
+                  "docs/guides/getting-started-orchestrate-microservices/"
```
pepopowitz, line 178: This page isn't versioned, so these URLs were pointing at 8.2 pages. I modified them to point at similar 1.3 pages.

src/pages/index.js
```diff
@@ -183,18 +185,11 @@ function Home() {
...
-              </Link>
-              <Link
-                className={clsx(
-                  "button button--outline button--secondary button--lg button--hero get-started-use-case get-started-use-case-2",
-                  styles.getStarted
+                to={useBaseUrl(
+                  "docs/guides/getting-started-orchestrate-human-tasks/"
                 )}
-                to={useBaseUrl("docs/guides/orchestrate-api-endpoints/")}
               >
```
pepopowitz, line 191: But one of the buttons didn't have a 1.3 counterpart, so I removed that button completely.

src/theme/NavbarItem/index.js
```diff
@@ -12,7 +12,7 @@ export default function NavbarItemWrapper(props) {
   const { pathname } = useLocation();
 
   const childProps = { ...props };
-  if (type === "docsVersionDropdown") {
+  if (type === "docsVersionDropdown" || type === "docsVersion") {
```
pepopowitz, line 15: This tricks the navbar into displaying the correct version for the single version navbar item. (It was already doing this for the version dropdown.)

I'd make this change in `main`, because all archived versions will want this.

static/.htaccess
```diff
@@ -8,7 +8,7 @@ Options -Indexes -MultiViews
 
 # Redirect pages - https://httpd.apache.org/docs/current/mod/mod_rewrite.html#rewriterule
 
-# deduplication of Web Modeler API
+<!-- # deduplication of Web Modeler API
```
pepopowitz, line 11: I don't think these HTML-style comments actually work in an htaccess, but that's what VS Code did for me when I tried to comment out an entire block.

The point of these comments is that in the real implementation, I'd remove *all* of these redirects that don't pertain to 1.3. I commented them out here, instead of removing them, because removing would have been super noisy.

static/.htaccess
```diff
@@ -112,10 +112,10 @@ RewriteRule ^docs/apis-clients/working-with-apis-clients/?$ /docs/apis-tools/wor
 RewriteRule ^docs/apis-clients(.*)$ /docs/apis-tools$1 [R=301,L]
 RewriteRule ^optimize/apis-clients(.*)$ /optimize/apis-tools$1 [R=301,L]
 RewriteRule ^docs/8.0/apis-clients(.*)$ /docs/8.0/apis-tools$1 [R=301,L]
-RewriteRule ^optimize/3.8.0/apis-clients(.*)$ /optimize/3.8.0/apis-tools$1 [R=301,L]
-RewriteRule ^docs/1.3/apis-clients(.*)$ /docs/1.3/apis-tools$1 [R=301,L]
+RewriteRule ^optimize/3.8.0/apis-clients(.*)$ /optimize/3.8.0/apis-tools$1 [R=301,L] -->
+RewriteRule ^docs/apis-clients(.*)$ /docs/apis-tools$1 [R=301,L]
```
pepopowitz, line 116: The more important thing to notice here is that the rules that were created for 1.3 are redefined, without the version in the URL.

versioned_docs/version-1.3/apis-tools/go-client/get-started.md
```diff
@@ -214,7 +214,7 @@ Now, we want to do some work within our process. Follow the steps below:
 - Set the **type** of the second task to `fetcher-service`.
 - Set the **type** of the third task to `shipping-service`.
 
-5. Additionally, for the service task `Collect Money` set a [**task-header**](/docs/1.3/components/modeler/bpmn/service-tasks/#task-headers) with the key `method` and the value `VISA`. This header is used as a configuration parameter for the payment-service worker to hand over the payment method.
+5. Additionally, for the service task `Collect Money` set a [**task-header**](../../components/modeler/bpmn/service-tasks/service-tasks.md/#task-headers) with the key `method` and the value `VISA`. This header is used as a configuration parameter for the payment-service worker to hand over the payment method.
```
pepopowitz, line 217: There were a few of these stragglers -- links that were crossing versions, because of the way they were written. We'll want to update each of these for each archived version.

versioned_sidebars/version-1.3-sidebars.json
```diff
@@ -215,71 +215,71 @@
         {
           "type": "link",
           "label": "What is Optimize?",
-          "href": "/optimize/3.7.0/components/what-is-optimize/"
+          "href": "/optimize/components/what-is-optimize/"
```
pepopowitz, line 218: Similar to the other sidebars file -- removing the version from the paths, everywhere that the sidebars cross docs instances.

docusaurus.config.js
akeller: I would say let's remove the banner, or use it as a clear visual aid that you are in old, old, old docs.

akeller (approved):
This looks FANTASTIC! Smart to add the robots.txt file 🧠 

Non-blocking -> Should we also disable "Edit this page"?

docusaurus.config.js
pepopowitz: I like the idea to use it as a clear reminder that you are in old docs. I'll play with this.

pepopowitz:
> This looks FANTASTIC! Smart to add the robots.txt file 🧠
> 
> Non-blocking -> Should we also disable "Edit this page"?

Yes, I love that. I'll work that into this prototype so we have everything captured here.

docusaurus.config.js
akeller: I gave you a code snippet in the other [PR](https://github.com/camunda/camunda-platform-docs/pull/2128#issuecomment-1557783802). 🙈  You can have it here too

```
announcementBar: {
      id: "unsupported",
      content:
        '🚨 These docs are unsupported and unmaintained. Looking for <b><a target="_blank" rel="noopener noreferrer" href="https://docs.camunda.io">supported versions?</a></b> 🚨',
      backgroundColor: "#FFC600",
      textColor: "#434343",
      isCloseable: false,
    },
```

docusaurus.config.js
```diff
@@ -4,9 +4,9 @@ module.exports = {
   title: "Camunda Platform 8 Docs",
   tagline: "Documentation for all components of Camunda Platform 8",
   // url: "https://camunda-cloud.github.io",
-  url: "https://docs.camunda.io",
+  url: "https://unsupported.docs.camunda.io",
```
pepopowitz, line 7: I actually don't know what this URL is used for. I think it might be the sitemap; though we aren't generating a sitemap, so I don't think we see this show up anywhere.

docusaurus.config.js
```diff
@@ -4,9 +4,9 @@ module.exports = {
...
   // baseUrl: "/camunda-cloud-documentation/",
-  baseUrl: "/",
+  baseUrl: "/1.3/",
```
pepopowitz, line 9: When running locally, this will serve everything at `localhost:3000/1.3/docs/....`. When deployed, it will be `unsupported.docs.camunda.io/1.3/docs/....`. 

Changing this property to reflect this enables docusaurus to generate all links correctly within the docs.

docusaurus.config.js
```diff
@@ -39,18 +39,32 @@ module.exports = {
       {
         id: "optimize",
         path: "optimize",
+        includeCurrentVersion: false,
         routeBasePath: "optimize",
         beforeDefaultRemarkPlugins: [versionedLinks],
-        sidebarPath: require.resolve("./optimize_sidebars.js"),
-        editUrl: "https://github.com/camunda/camunda-platform-docs/edit/main/",
```
pepopowitz, line 45: Per feedback, removing the "Edit this page" link for both optimize (here) & main docs (further down)

docusaurus.config.js
```diff
@@ -59,10 +73,10 @@ module.exports = {
     announcementBar: {
       id: "camunda8",
       content:
-        '📣 <b><a target="_blank" rel="noopener noreferrer" href="https://signup.camunda.com/accounts?utm_source=docs.camunda.io&utm_medium=referral&utm_content=banner">Sign-Up</a></b> for a free account to start orchestrating business processes today.',
+        '⚠️ This version of Camunda Platform 8 is no longer actively maintained. For up-to-date documentation, see <b><a target="_blank" rel="noopener noreferrer" href="https://docs.camunda.io">the latest version</a></b>.',
```
pepopowitz, line 76: Updated to notify users that these docs are not supported.

docusaurus.config.js
```diff
@@ -224,29 +238,40 @@ module.exports = {
...
         blog: false,
         theme: {
           customCss: require.resolve("./src/css/custom.css"),
         },
         sitemap: {
           //cacheTime: 600 * 1000, // 600 sec - cache purge period
-          changefreq: "weekly",
-          priority: 0.5,
+          // exclude everything from sitemap
+          ignorePatterns: ["**"],
```
pepopowitz, line 272: ignores *everything* from sitemap

src/pages/contact.md
```diff
@@ -11,7 +11,7 @@ There are various channels where you can reach us.
 
 - We welcome your **bug** reports and **feature requests** through our community channels mentioned above.
 
-- For **security-related issues**, review our [Security notices](../../docs/reference/notices) for the most up-to-date information on known issues and steps to report a vulnerability so we can solve the problem as quickly as possible. Do not use GitHub for security-related issues.
+- For **security-related issues**, review our [Security notices](../docs/reference/notices) for the most up-to-date information on known issues and steps to report a vulnerability so we can solve the problem as quickly as possible. Do not use GitHub for security-related issues.
```
pepopowitz, line 14: I actually think this URL is incorrect in production, too...but it works out fine, because if you go two up from one level deep, you still just end up at the root. 

This will be fixed in `main`, not in the archive branches.

src/pages/index.js
```diff
@@ -68,9 +68,9 @@ function Feature({ imageUrl, url, title, description }) {
     <div className={clsx("col col--4 component-block", styles.feature)}>
       {imgUrl && (
         <div className="text--center">
-          <a href={url}>
+          <Link to={useBaseUrl(url)}>
```
pepopowitz, line 71: Two changes here, both of them will be done in main instead of the archive branches: 

1. use `Link` instead of `a`, to take advantage of client-side routing. The transition looks much snappier.
2. `useBaseUrl` to wrap the `url`, so that the `1.3/` gets applied to the front of the path. (In a non-archived branch, this would still result in a valid URL, because the baseURL in main is `/`.)

src/theme/AnnouncementBar/index.js
pepopowitz, line 1: Removing this file, because we want the announcement bar to **always** show in the archived versions. (Because it now displays a hint that the version is no longer supported.)

static/.htaccess
pepopowitz, line 1: The only rules remaining are the ones that were configured for version 1.3.

They've been updated, so that: 

1. The left side no longer includes the `1.3`, because the inbound URLs don't include them at this stage.
2. The right side includes the `1.3`, because the outbound URLs _need_ them in order to go to the correct archived version.

static/.htaccess
```diff
@@ -8,503 +8,21 @@ Options -Indexes -MultiViews
...
-RewriteRule ^docs/8.2/components/connectors/out-of-the-box-connectors/graphql/(.*)$ /docs/8.2/components/connectors/protocol/graphql/$1 [R=301,L]
-RewriteRule ^docs/8.2/components/connectors/out-of-the-box-connectors/http-webhook/(.*)$ /docs/8.2/components/connectors/protocol/http-webhook/$1 [R=301,L]
-RewriteRule ^docs/8.2/components/modeler/web-modeler/advanced-modeling/manage-connector-templates/(.*)$ /docs/8.2/components/connectors/manage-connector-templates/$1 [R=301,L]
-RewriteRule ^docs/components/connectors/out-of-the-box-connectors/rest/(.*)$ /docs/components/connectors/protocol/rest/$1 [R=301,L]
-RewriteRule ^docs/components/connectors/out-of-the-box-connectors/graphql/(.*)$ /docs/components/connectors/protocol/graphql/$1 [R=301,L]
-RewriteRule ^docs/components/connectors/out-of-the-box-connectors/http-webhook/(.*)$ /docs/components/connectors/protocol/http-webhook/$1 [R=301,L]
-RewriteRule ^docs/components/modeler/web-modeler/advanced-modeling/manage-connector-templates/(.*)$ /docs/components/connectors/manage-connector-templates/$1 [R=301,L]
+RewriteRule ^docs/components/overview/(.*)$ /1.3/docs/components/components-overview/$1 [R=301,L]
```
pepopowitz, line 16: It will be important to test these redirects out thoroughly for archived versions, because as this example proves, they might no longer be correct! (This one redirects to a 404.)

pepopowitz (commented):
Adding some context for my most recent updates. I think all my changes are here, but I still plan to (a) update the video in the body of the PR, and (b) make associated updates in #2128.

versioned_docs/version-1.3/apis-tools/go-client/get-started.md
akeller: In more recent versions of the docs, I've tried to catch and correct these. All of this to say, it should get better.

src/pages/contact.md
akeller: How are you finding these links? Are they appearing as build issues or through a search?

static/.htaccess
akeller: I would assume most redirects for unmaintained versions can be deleted. Or am I missing something?

pepopowitz: I suppose we could just delete them because they're _unsupported_, but it doesn't seem like much effort to me to preserve them, if it means helping some people who were holding on to links before we archived. Is this blocking?

akeller: Nope!

akeller (approved):

pepopowitz:
Closing this, as the work is starting to happen.

PR #2133, opened 2023-05-23
Update overview.md

jwulf:
Change localhost URL to example SaaS URL

Update example URL as per https://github.com/camunda/developer-experience/issues/16

Whenever we do a push.

## PR Checklist

- [X] My changes apply to an already released version, and I have added them to the relevant `/versioned_docs` directory, or they are not for an already released version.


akeller:
I was expecting this PR to include other components and introduce some standardization. 

For example, just within the Tasklist API reference overview pages we have two URLs:

`https://${REGION}.tasklist.camunda.io:443/${CLUSTER_ID}/graphql` and `https://${base-url}/swagger-ui/index.htm`

Web Modeler includes a nod to both SaaS and SM:

> A detailed API description is available as [OpenAPI](https://www.openapis.org/) specification at https://modeler.cloud.camunda.io/swagger-ui/index.html for SaaS and at http://localhost:8070/swagger-ui.html for Self-Managed installations.

Can I use the SaaS Swagger UI with my instance/region and expect to use the same URL? With what is available in the docs today, I'm not sure.

Will this be addressed in another PR or issue scope?

jwulf:
Hmmm, I see. Things have changed since we did the review that this PR is based on. 

I've taken a look at the docs based on what you point out, and hopefully this normalises them. I've essentially copied the pattern used by the Modeler team to the Tasklist docs. 

To this question:


> Can I use the (Modeler) SaaS Swagger UI with my instance/region and expect to use the same URL? 

Modeler is a singleton, like Console. There is a single endpoint, and the auth token that you use determines which organisation you are interacting with. 

Modeler > Organisations > Clusters > Components

The instance/region details are per-cluster, but with the Modeler API you are connecting at the Modeler level, and the routing to the appropriate organisation is done automatically based on your authentication credentials. 

Do we need to make that explicit in the docs? 

By using the Web Modeler you figure out where it sits in the hierarchy. It's per-organisation. From a model in the Modeler you can deploy to any cluster in your org.

akeller:
@jwulf thanks for splitting the instructions to include a sample for SaaS and SM. 

Re: Modeler, it looks like the info for SaaS vs. SM is already in the docs.

I think that completes the scope of this issue.

akeller:
Updating the branch to resolve merge conflicts.

PR #2140, opened 2023-05-23
refine PR template

akeller:

The PR Template, as it exists today, is working against everyone.

It's unclear why we ask for this information, no one is aligned on what "marketing activities" mean, and I see the same questions being asked about what should/shouldn't be in a PR. The template should help us with this.

No, but I hope to one day brag on LinkedIn that we've figured out the best way to support components that release twice yearly and ones that release constantly/continuously. 

After a healthy debate about whether people like this change or will actually use/read it.

## PR Checklist


.github/pull_request_template.md
```diff
@@ -1,18 +1,22 @@
...
 ## When should this change go live?
 
-_Does this change need to be released before a certain date or milestone? If so, when?_
+<!-- PRs merged go to stage.docs.camunda.io first and must be manually released to docs.camunda.io. -->
+<!-- Help the DevEx team prioritize our work (reviews, merges, etc.) by opening PRs sooner. -->
+
+- [ ] There is no urgency with this change.
+- [ ] This change or page is part of a marketing blog, conference talk, or something else on a schedule.
+- [ ] This functionality is already available but undocumented.
+- [ ] This is a bug fix or security concern.
```
christinaausley, line 13: Maybe something along the lines of "This change is not yet live and should not be merged until X (apply hold label)"?

.github/pull_request_template.md
```diff
@@ -1,18 +1,22 @@
-## What is the purpose of the change
+## Description
```
christinaausley, line 1: Should we encourage users to apply appropriate labels and assign themselves as well?

christinaausley (approved):
Non-blocking comments -- THANK YOU for this!

.github/pull_request_template.md
akeller: Great idea! I'll add a comment.

.github/pull_request_template.md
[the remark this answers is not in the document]
RomanJRW: Not necessarily, but I also don't think I have a full picture. Something that might be worth doing is either having an agreed understanding of what urgency means or having the author clarify on urgency in the event that the first checkbox is left unchecked. I can imagine any of the following could be considered subjectively urgent, depending on the author:

- It needs releasing ASAP as it is security related
- It closes an L2 support ticket so helps maintain SLAs
- The feature is being released tomorrow, so the docs should be too

But only in the first scenario would I personally consider this "urgent"

pepopowitz (approved):
These are great changes 💪 🙌

.github/pull_request_template.md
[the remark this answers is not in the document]
akeller: Each team seems to have its own definition of urgency, which I'm trying to respect here. With these specific questions I'm trying to get more at the _why_ and the _what_. The vast majority of PRs come in with "ASAP," "urgent", "immediately," etc. when in reality they seem to fall into these buckets. 

All of this to say, I'm ok with subjectivity as we balance between teams that release continuously and twice a year. 😄

akeller: Added!

christinaausley (approved):

.github/pull_request_template.md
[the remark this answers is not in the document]
RomanJRW: ha, sounds good. Thanks for clarifying! 💪

PR #2143, opened 2023-05-24
The HTTP Webhook Connector is a template vor an event, not a task

jenshb:
The HTTP Webhook Connector is a template vor an event, not a task


bastiankoerber (approved):
that's right.

PR #2151, opened 2023-05-30
Add blurb about not requiring patches on update path

### LINKED ISSUE #571: Upgrading Minor Versions Doesn't require patchs
I found that if you wanted to upgrade your zeebe cluster from lets say 1.2.0 to 1.3.0 you don't actually need to follow the upgrade path of 1.2.1 -> 1.2.2 -> ..... -> 1.3.0 

You can run a single upgrade when moving from one minor version to another. 
The docs on the upgrading don't seem to mention this is, it would be good to have this explicitly stated.

akeller:

Closes #571.

No.

Not urgent or critical.

## PR Checklist


akeller:
@christinaausley, this probably needs some wordsmithing. I want to ensure we include the terms minor, update, and patch.

christinaausley (approved):

megglos:
Just a heads-up this guide affects all C8 components, we need to check if this is fine for:

- [x] Zeebe - it's fine to update from one minor patch to another patch of the next minor
- [ ] Operate - @johanwelgemoed 
- [ ] Tasklist - @marcosgvieira 
- [ ] Identity - @Ben-Sheppard / @dlavrenuek

Ben-Sheppard:
From the Identity perspective: as a component we always consider the impacts of past compatibility when introducing features/fixes so in a general sense I can't see anything wrong with applying the statement to the Identity component. 

We of course may have additional considerations when moving from minor to minor (for example the 8.3 release is planned to introduce Keycloak v21+ support) however this is handled within the specific migration guides I know already exist.

RomanJRW:
For Optimize, we typically try to avoid schema migrations for patch releases unless unavoidable. In any case, we write updates in such a way that allows it to be possible to update from any patch from the previous minor to any patch of the next one

akeller:
@johanwelgemoed can we get your perspective here on Operate?

johanwelgemoed:
@akeller Sorry for delay. I was actually looking at this yesterday and though we think it should be ok from Operate side we would like to do a sanity check and confirm it. The plan is to do a migration like this next week as part of our release testing to confirm. Is that ok?

johanwelgemoed:
@akeller From Operate side, it should be ok. The only caveat is that you should not skip minors. (eg. if someone is on 8.0.x, they would need to go to 8.1.x first before goidn to 8.2.x. ie. they cannot go from 8.0.x directly to 8.2.x skipping 8.1.x). Other than that it will work. Does this help?

akeller:
Based on the feedback, it looks like we can move forward with this, but I'll add a small tweak to make sure users know they should **not** skip minors.

christinaausley (approved):
I've bolded the different versions here, to be extra clear. Good with wording 👍

akeller:
> I've bolded the different versions here, to be extra clear. Good with wording 👍

Nice add! I like it.

PR #2155, opened 2023-05-31
assorted: Archive pre-work

pepopowitz:
## Description

Adjusts a few things in preparation for #1173:

1. Corrects a link on the Contact page, which was incorrectly traveling one ancestor too far. This has no effect on the current docs, but will have an effect on the archived versions.
2. Updates the "feature" links on the home page to do two things: 
   1. Use the `<Link>` element instead of `<a>`, which enables docusaurus to navigate with the client-side router. To the user, this looks like more seamless transitions when they click on the feature.
   2. Wrap the `url` of each link in a `useBaseUrl()` call. This tells docusaurus to use the `baseUrl` property of its configuration when generating the link. This has no effect on the current docs, but will have an effect on the archived versions. 
3. Broadens the scope of the `docsPluginId`-sniffing logic in the `NavbarItem` component, so that archived versions will be able to display the correct version for Optimize documentation. 

## PR Checklist

### DIFF
--- a/src/pages/contact.md
+++ b/src/pages/contact.md
@@ -11,7 +11,7 @@ There are various channels where you can reach us.
 
 - We welcome your **bug** reports and **feature requests** through our community channels mentioned above.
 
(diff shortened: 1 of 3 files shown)


pepopowitz:
fyi, I'm merging this as-is, without a review, so that I can move forward! I'm happy to make any changes retroactively next week.

PR #2167, opened 2023-06-02
chore: archival of version 0.25, part 3: manual steps

pepopowitz:
## Description

Part of #1173. Follows #2166. 

More steps to isolate a site for version 0.25, including:

- [x] Update the docusaurus.config.js
- [x] Update the theme components
- [x] Update CI workflows
- [x] Update any broken links
- [x] Update htaccess rules

## Not included in this PR, but coming in follow-ups

- I forgot to remove all the `optimize` folders! That will create another large PR, so not including it here.

## PR Checklist


.github/workflows/build-docs.yaml
```diff
@@ -30,32 +30,3 @@ jobs:
...
-          fail: true
-      - name: Validate product links
-        uses: lycheeverse/lychee-action@v1.6.1
-        with:
-          args: "product-links.prod.txt"
-          fail: true
-      - name: Check internal links
-        uses: untitaker/hyperlink@0.1.27
-        with:
-          args: build/
```
pepopowitz, line 61: This is potentially controversial. The sitemap & product link checks are definitely not something we want to run, because this is an isolated site containing only one old version. But the internal links of the app is probably still worth checking, if it's not much work.

Unfortunately it isn't "not much work"! I'd have to alter the steps above that run the webserver to run the app in a /0.25 folder, or else [the internal link-checking would fail](https://github.com/camunda/camunda-platform-docs/actions/runs/5159889016/jobs/9295356005) because of the baseURL of /0.25.

My feeling is that it's fine to remove this link-checking. It's an unsupported version of the docs, and if an internal link gets broken over time, I think we just say forget it.

But if you feel strongly that I should get this internal link-checking working, let me know!!!

.github/workflows/linkcheck.yaml
pepopowitz, line 1: Removing this workflow entirely, as we won't want to run it against the 0.25 site.

.github/workflows/publish-prod.yaml
```diff
@@ -3,7 +3,7 @@ name: publish-prod
 on:
   push:
     tags:
-      - "[0-9]+.[0-9]+.[0-9]+"
+      - "0.25.[0-9]+"
```
pepopowitz, line 6: I think this is the right way to do this? Trigger a 0.25 production build only when we create a 0.25.X release.

🚨 We'll have to make sure we update `main` to exclude the archived tags before we publish a 0.25.x release! Otherwise I think `main` will trigger a new prod publish, and deploy the 0.25 site to docs.camunda.io!!! (I think that's what would happen?)

.github/workflows/publish-prod.yaml
```diff
@@ -37,7 +37,7 @@ jobs:
         with:
           switches: -avzr --delete
           path: build/
-          remote_path: ${{ secrets.AWS_PROD_PUBLISH_PATH }}
+          remote_path: ${{ secrets.AWS_PROD_PUBLISH_PATH_UNSUPPORTED }}/0.25
```
pepopowitz, line 40: I've already added this new key to the project secrets (and to the Keeper record that contains these secrets).

.github/workflows/publish-stage.yaml
```diff
@@ -43,7 +39,7 @@ jobs:
         with:
           switches: -avzr --delete
           path: build/
-          remote_path: ${{ secrets.AWS_STAGE_PUBLISH_PATH }}
+          remote_path: ${{ secrets.AWS_STAGE_PUBLISH_PATH_UNSUPPORTED }}/0.25
```
pepopowitz, line 42: I've already added this new key to the project secrets (and to the Keeper record that contains these secrets).

.github/workflows/publish-stage.yaml
```diff
@@ -3,9 +3,7 @@ name: publish-stage
 on:
   push:
     branches:
-      - "main"
-    tags-ignore:
-      - "*"
+      - "unsupported/0.25"
```
pepopowitz, line 6: I'm more confident about this trigger than the production one. 

I removed `tags-ignore` because I don't know what effect it has? I should probably research it instead of blindly removing.

.github/workflows/publish-stage.yaml
```diff
@@ -20,10 +18,8 @@ jobs:
           node-version: 16
       - name: Install Dependencies
         run: npm ci
-      - name: Disable Indexing
```
pepopowitz, line 23: Indexing is already disabled in the config itself.

.github/workflows/publish-stage.yaml
```diff
@@ -20,10 +18,8 @@ jobs:
...
-        run: "sed -i 's/noIndex: false/noIndex: true/g' docusaurus.config.js"
       - name: Update URL
-        run: 'sed -i ''s!url: "https://docs.camunda.io"!url: "https://stage.docs.camunda.io"!g'' docusaurus.config.js'
+        run: 'sed -i ''s!url: "https://unsupported.docs.camunda.io"!url: "https://stage.unsupported.docs.camunda.io"!g'' docusaurus.config.js'
```
pepopowitz, line 22: New URLs

docusaurus.config.js
```diff
@@ -34,35 +32,16 @@ module.exports = {
       },
     ],
     "./static/plugins/bpmn-js",
-    [
```
pepopowitz, line 37: No optimize docs for this version.

docusaurus.config.js
```diff
@@ -185,7 +152,7 @@ module.exports = {
             },
             {
               label: "Release cycle",
-              to: "docs/reference/release-policy",
+              to: "https://docs.camunda.io/docs/reference/release-policy",
```
pepopowitz, line 155: These are footer links. A few of these pages don't exist in the 0.25 site, so I pointed them at production instead.

src/pages/contact.md
```diff
@@ -11,7 +11,7 @@ There are various channels where you can reach us.
 
 - We welcome your **bug** reports and **feature requests** through our community channels mentioned above.
 
-- For **security-related issues**, review our [Security notices](../docs/reference/notices) for the most up-to-date information on known issues and steps to report a vulnerability so we can solve the problem as quickly as possible. Do not use GitHub for security-related issues.
+- For **security-related issues**, review our [Security notices](https://docs.camunda.io/docs/reference/notices) for the most up-to-date information on known issues and steps to report a vulnerability so we can solve the problem as quickly as possible. Do not use GitHub for security-related issues.
```
pepopowitz, line 14: No page here in 0.25! It feels better to send them somewhere, since it's security stuff.

src/pages/index.js
```diff
@@ -32,7 +32,7 @@ const features = [
   {
     title: "APIs & Tools",
     imageUrl: "img/apisclients.png",
-    url: "/docs/apis-tools/working-with-apis-tools/",
+    url: "/docs/components/clients/build-your-own-client/",
```
pepopowitz, line 35: These are the homepage features. This URL translates pretty well.

src/pages/index.js
```diff
@@ -41,7 +41,7 @@ const features2 = [
   {
     title: "Self-Managed",
     imageUrl: "img/self-managed.png",
-    url: "/docs/self-managed/about-self-managed/",
+    url: "/",
```
pepopowitz, line 44: This URL has nothing related in 0.25! Instead of removing the entire feature block, because that makes it look funny, I left it in and pointed it at the home page. It's "unsupported" so that's safe????

src/pages/index.js
```diff
@@ -164,41 +164,6 @@ function Home() {
             </div>
           </section>
         )}
-        <div className={clsx("hero hero--secondary", styles.heroBanner)}>
```
pepopowitz, line 167: None of these three buttons have a 0.25 page! (Technically one of them does but it just says "TODO"!) 

So I removed this whole green block.

static/.htaccess
pepopowitz, line 1: The only rules remaining are specific to 0.25, and they've been updated so that the left side does not include 0.25 (because the app itself won't receive the 0.25 part of the URL), and the right side _does_ include 0.25 (because otherwise it will redirect outside of the 0.25 app).

.github/workflows/publish-stage.yaml
pepopowitz: ah, I bet the tags-ignore is there so that it doesn't re-deploy to staging when we do a production deploy (via tagging the branch). 

I'm going to put the tags-ignore back!

.github/workflows/publish-prod.yaml
pepopowitz: PR to prevent accidental 0.25 deployment to docs.camunda.io is up, at https://github.com/camunda/camunda-platform-docs/pull/2172

PR #2180, opened 2023-06-06
chore(docs): warn about engine historic data clean up time

RomanJRW:
relates to OPT-6890

## Description

From a support ticket, we should clarify that engine data needs to remain for enough time for Optimize to import it, or it will be lost

## PR Checklist

### DIFF
--- a/optimize/self-managed/optimize-deployment/configuration/system-configuration-platform-7.md
+++ b/optimize/self-managed/optimize-deployment/configuration/system-configuration-platform-7.md
@@ -10,7 +10,9 @@ to import data from. Each engine configuration should have a unique alias associ
 with it and represented by `${engineAlias}`.
 
 Note that each connected engine must have its respective history level set to `FULL` in order to see all available data
-in Optimize. Using any other history level will result in less data and/or functionality within Optimize.
+in Optimize. Using any other history level will result in less data and/or functionality within Optimize. Furthermore,
+history in a connected engine should be configured for long enough for Optimize to import it. If data is removed from an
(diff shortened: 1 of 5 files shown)


optimize/self-managed/optimize-deployment/configuration/system-configuration-platform-7.md
```diff
@@ -10,7 +10,9 @@ to import data from. Each engine configuration should have a unique alias associ
 with it and represented by `${engineAlias}`.
 
 Note that each connected engine must have its respective history level set to `FULL` in order to see all available data
-in Optimize. Using any other history level will result in less data and/or functionality within Optimize.
+in Optimize. Using any other history level will result in less data and/or functionality within Optimize. Furthermore,
+history in a connected engine should be configured for long enough for Optimize to import it. If data is removed from an
```
akeller, line 14: "long enough for Optimize to import it"

Can we be more specific, or does it depend on the installation and config?

RomanJRW: I wish we could 😢 
Unfortunately it depends on a lot of things so we tend to avoid being specific with configuration like this in our docs. Normally the aggressiveness with which users delete their engine data is determined more by their engine performance than Optimize downstream. I feel like the best we can realistically do is warn them of the effects this might have in cases where the import can't keep up for whatever reason

akeller: Fair enough. Maybe one day 🙂

akeller (approved):

PR #2270, opened 2023-06-28
Zeebe: Clarify usage of advertised host

jonathanlukas:
## Description

Currently, the advertised host parameter is not clear to everybody.

After some try-and-error, everybody could face out what it is, but we should make it more obvious.

Just a detail that was unclear for some customers

## PR Checklist

I would to have this reviewed as I am not sure whether my wording is completely correct.


akeller (dismissed):
This is a very simple change, but I was curious if we want to change it directly in the broker template a well.

oleschoenburg (dismissed):
:wrench: The same setting exists for the gateway too so we might want to change that too.

:upside_down_face: Personally, I find "forwarded" to be a bit confusing in this context and would rather use "advertised". It's sort of a term of art and clearer than "forwarded". For example, searching for "advertised address" and "advertised host" brings up documentation for the same concept in similar software such as Kafka. Searching for "forwarded address" shows results for actual paper mail, not software.

I'd suggest something like "Controls the advertised host (the contact point advertised to other brokers)"

Either way, this is probably splitting hairs so I'm approving anyway.

akeller:
@jonathanlukas any additional thoughts here? We can always revise this language in the future if it causes more confusion. Feel free to merge if you are ready.

jonathanlukas:
Thank you @oleschoenburg for your suggestion.

I have to admit that I found it hard to create a proper wording. I changed it to what you proposed as on reading again, I found the usage of 2 different terms for the same thing confusing.

oleschoenburg (approved):

PR #2278, opened 2023-06-30
Add create account btn to navbar

jpguisa:
## Description

Add create account btn to navbar


jpguisa:
@akeller I had to create a new branch since I had some troubles with the author of some commits. Not the "create account" button is showing as you requested. Let me know if you need anything else

akeller (approved):
Approving this but I noticed the following behavior:

* The link disappears about 1245px
* The link reappears around 996px

This could be corrected in a future PR so we can move forward with this experiment.

jpguisa:
@akeller That behavior was made on purpose, on that screen size having the link there makes the left navigation breaks lines but we could find another solution

akeller:
@jpguisa I will merge and do a release later today and tomorrow to kick off this experiment. Thanks!

PR #2316, opened 2023-07-11
Clarify region value in SaaS

### LINKED ISSUE #2304: Clarify which region value is needed for the Operate Swagger UI for SaaS
There are two possible region values when crafting the URL for the Operate Swagger UI for SaaS - 

Region on the Cluster connection information shows the AWS data center region, so us-central1, for example.

Region on the client credential connection information shows a different region value, dsm-1, when I tested. 

On both UIs, the cluster id, also needed in crafting the URL for the Operate Swagger UI, shows the same value. Having "region" be the same name for two different values is confusing. We can mitigate this in the docs, but there should be a change in the product as well.

I'm listing this as a bug because it was confusing enough to me that I assumed I was stuck without assistance. The error when using the other region value (us-central1 instead of dsm-1) wasn't intuitive or helpful.

akeller:
## Description

Closes #2304.

## PR Checklist


docs/apis-tools/operate-api/overview.md
```diff
@@ -18,6 +18,10 @@ A detailed API description is also available as Swagger UI at `https://${base-ur
 
 For SaaS: `https://${REGION}.operate.camunda.io/${CLUSTER_ID}/swagger-ui.html`, and for Self-Managed installations: `http://localhost:8080/swagger-ui.html`.
 
+:::note
+Find your region and cluster id under connection information in your client credentials.
```
christinaausley, line 22: Is there a bolded UI element we should be looking for here?

christinaausley (commented):
Just one comment for clarity 👍

docs/apis-tools/operate-api/overview.md
akeller: Unfortunately, that's part of the problem. **Region** has to distinct values depending on where in the UI you find it. I've proposed changing this in the UI to "region id" - https://camunda.slack.com/archives/C0589TC88UU/p1689175806936099?thread_ts=1689000734.350649&cid=C0589TC88UU

PR #2422, opened 2023-08-09
feat(archive): automate update of docusaurus.config.js

pepopowitz:
## Description

Part of #1173.

This PR automates the updates of docusaurus.config.js during a version archival. 

See the changes in #2421, specifically within the `docusaurus.config.js` file, for an example of the changes this step generates.

## Alternatives considered

The file that's being edited is a javascript file, though the large majority of the contents are a single JSON object. I considered two approaches other than the `sed` commands in this PR: 

1. Use [jq](https://jqlang.github.io/jq/) to edit the JSON directly, since it's a tool designed for parsing/editing JSON. The issue here was that the file is _JavaScript_, not JSON, so jq can't parse it. I also couldn't _convert it_ from JavaScript to JSON, because (a) I think docusaurus only supports JavaScript for the config, probably because (b) there are run-time `require` statements in the object.
2. Renaming the existing `docusaurus.config.js` to something like `docusaurus.config.original.js`, creating a second `docusaurus.config.overrides.js` which would contain only properties that would be overridden, and creating a new `docusaurus.config.js` that merged those two sources together at runtime. I decided against this because it seemed like it wouldn't be significantly more maintainable than these `sed` commands, but would probably present tricky edge cases which would make it difficult to build correctly. (Example: some properties need to be replaced, some need to be deleted, some need only one or two child properties changed.)

In the end I chose the `sed` commands and tried to simplify and document them as much as possible, for the inevitable future where something doesn't work and we are pulling our hair out trying to figure out why.

## PR Checklist

### DIFF
--- a/hacks/isolateVersion/7-updateDocusaurusConfig.sh
+++ b/hacks/isolateVersion/7-updateDocusaurusConfig.sh
@@ -1,0 +1,65 @@
+notify "Updating docusaurus.config.js..."
+
+# Remove unused versionedLinks dependency
+sed -i '' '/const versionedLinks/d' docusaurus.config.js
+
+# Update `url` to include `unsupported`
+sed -i '' -e "s/docs.camunda.io/unsupported.docs.camunda.io/" docusaurus.config.js
+
+# Update `baseUrl` to specify isolated version
+sed -i '' -e "s/baseUrl: \"\\/\"/baseUrl: \"\/$ARCHIVED_VERSION\/\"/" docusaurus.config.js
+
+# Remove optimize docs plugin.
+#   1. Find the block that starts with the plugin declaration, and the closing braces afterward, and delete it.
+sed -i '' '/^      "@docusaurus\/plugin-content-docs"/,/^      },/d' docusaurus.config.js
+#   2. Format the config file, to collapse the empty brackets left behind (e.g. `  [\n  ]` -> `  []`)
+npx prettier --write docusaurus.config.js
+#   3. Remove the empty array
+sed -i '' '/^    \[\],/d' docusaurus.config.js
+
+# Add `announcementBar` to `themeConfig`
+#   Find the line containing `themeConfig` and append the new settings to it.
+sed -i '' "/themeConfig/a\\
+    announcementBar: {\\
+      id: \"camunda8\",\\
+      content:\\
+        '🚨 This version of Camunda Platform 8 is no longer actively maintained. For up-to-date documentation, see <b><a target=\"_blank\" rel=\"noopener noreferrer\" href=\"https://docs.camunda.io\">the latest version</a></b>.',\\
+      backgroundColor: \"#FFC600\",\\
+      textColor: \"#434343\",\\
+      isCloseable: false,\\
+    },\\
+" docusaurus.config.js
+
+# Convert version dropdown in top nav to static version number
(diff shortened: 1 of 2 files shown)


hacks/isolateVersion/7-updateDocusaurusConfig.sh
```diff
@@ -0,0 +1,65 @@
...
+sed -i '' '/const versionedLinks/d' docusaurus.config.js
+
+# Update `url` to include `unsupported`
+sed -i '' -e "s/docs.camunda.io/unsupported.docs.camunda.io/" docusaurus.config.js
+
+# Update `baseUrl` to specify isolated version
+sed -i '' -e "s/baseUrl: \"\\/\"/baseUrl: \"\/$ARCHIVED_VERSION\/\"/" docusaurus.config.js
+
+# Remove optimize docs plugin.
+#   1. Find the block that starts with the plugin declaration, and the closing braces afterward, and delete it.
```
pepopowitz, line 13: I tried to build this as one single `sed` command, but expanding beyond the matched block was difficult, and this seemed like a simpler way to accomplish the task.

hacks/isolateVersion/7-updateDocusaurusConfig.sh
```diff
@@ -0,0 +1,65 @@
...
+# Replace the `sitemap`` block with one that only specifies a blanket ignorePattern.
+sed -i '' '/^        sitemap: {/,/^        },/c\
+        sitemap: {\
+          \/\/ exclude everything from sitemap\
+          ignorePatterns: [\"**\"],\
+        },\
+' docusaurus.config.js
+
+# git add docusaurus.config.js
+# git commit -m "archiving: update docusaurus config"
```
pepopowitz, line 65: Oops! Gonna fix this. These shouldn't be commented out.

akeller (dismissed):
Thanks for the great explanation of alternatives considered. I really like your approach aspects, specifically maintainability and being more maintainable than what we have today.

hacks/isolateVersion/7-updateDocusaurusConfig.sh
```diff
@@ -0,0 +1,65 @@
+notify "Updating docusaurus.config.js..."
+
+# Remove unused versionedLinks dependency
+sed -i '' '/const versionedLinks/d' docusaurus.config.js
+
+# Update `url` to include `unsupported`
+sed -i '' -e "s/docs.camunda.io/unsupported.docs.camunda.io/" docusaurus.config.js
+
+# Update `baseUrl` to specify isolated version
+sed -i '' -e "s/baseUrl: \"\\/\"/baseUrl: \"\/$ARCHIVED_VERSION\/\"/" docusaurus.config.js
```
pepopowitz, line 10: These commands don't need the `-e`! That flag is used to give sed multiple commands, but these are broken into individual statements. 

I'm going to remove those flags. ChatGPT is a dumb robot that keeps putting them in for no good reason (and I a dumb human who doesn't catch it).

pepopowitz: (I know that that's super minor, but these commands are all so cryptic and confusing that I want to eliminate any unnecessary noise.)

akeller: 🙏 thank you! Future me and future you thank you too!

akeller (approved):

PR #2454, opened 2023-08-17
feat: introduce modeler categories

nikku:
This introduces relevant categories for parent pages. As a result:

* Breadcrumbs are meaningful
* Unnecessary "About" pages and nesting in the sitebar is avoided

----

capture K4kgls_optimized

capture 3bIp92_optimized

## Description

Adds necessary meta-data to turn sections into categories, fix the breadcrumbs, and avoid unnecessary nesting + about pages in the sidebar.

## PR Checklist

----

Related to https://github.com/camunda/developer-experience/issues/109

### DIFF
--- a/docs/components/modeler/reference/sidebar-schema.js
+++ b/docs/components/modeler/reference/sidebar-schema.js
@@ -1,17 +1,20 @@
 module.exports = {
-  Reference: {
-    "Modeling guidance": [
-      "components/modeler/reference/modeling-guidance/index",
-      {
-        Rules: [
-          "components/modeler/reference/modeling-guidance/rules/called-element",
-          "components/modeler/reference/modeling-guidance/rules/element-type",
-          "components/modeler/reference/modeling-guidance/rules/error-reference",
-          "components/modeler/reference/modeling-guidance/rules/escalation-reference",
-          "components/modeler/reference/modeling-guidance/rules/feel",
-          "components/modeler/reference/modeling-guidance/rules/message-reference",
-        ],
+  Reference: [
+    {
+      type: "category",
+      label: "Modeling guidance",
+      link: {
+        type: "doc",
+        id: "components/modeler/reference/modeling-guidance/index",
       },
-    ],
-  },
+      items: [
+        "components/modeler/reference/modeling-guidance/rules/called-element",
+        "components/modeler/reference/modeling-guidance/rules/element-type",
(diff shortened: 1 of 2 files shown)


pepopowitz (approved):
👍 this looks great!

I'm going to add another comment about an observation I have, but it is non-blocking for this PR!

pepopowitz:
An observation that is related but should have no impact on this PR -- I noticed from the gif in the body that there is an unordered list on the Modeling Guidance category page, enumerating the Rules within the category. In other docs, [we've used the built-in docusaurus `<DocCardList>` component](https://github.com/camunda/camunda-platform-docs/blob/e72f25f10119cd8650bd5adb7b6d179f4b623842/docs/self-managed/platform-deployment/helm-kubernetes/platforms/platforms.md?plain=1#L14) to enumerate the children without hardcoding them. 

There is an obvious visual difference between these approaches -- a [purely text-based unordered list](https://stage.docs.camunda.io/docs/next/components/modeler/reference/modeling-guidance/) (image 1 below), versus [a list of cards](https://stage.docs.camunda.io/docs/next/self-managed/platform-deployment/helm-kubernetes/platforms/) (image 2 below). Throw in another variant, [this tabular presentation](https://docs.camunda.io/docs/next/components/modeler/forms/form-element-library/forms-element-library/) (image 3 below). 

My question for @christinaausley -- from a consistency standpoint, how do you feel about these variations? 

And for @nikku -- the cards are very large, and take up a lot of space without adding a ton of information, and I could imagine that influencing your decision to use an unordered list instead. If the cards were more condensed, would you be more inclined to use them?

And for @nikku -- if you feel strongly about using the unordered list, would it be useful to have a component that renders all child pages, without hardcoding them, as an unordered list instead of a list of cards?

### Images

1: unordered list

image

2: card list

image

3. table

image

nikku:
Indeed we could optimize our modeling guidance landing page as you mention in https://github.com/camunda/camunda-platform-docs/pull/2454#issuecomment-1682670842, I don't see it as a big priority though:

* If we go for tiles, then descriptions must be on point. 
* Either way, this page is only there for completeness, and not as a main entry point. Main entry is via individual pages, through the modeler apps.

nikku:
Changed to use `DocCardList` via https://github.com/camunda/camunda-platform-docs/pull/2462. 

We can iterate on this in the future. For now, it ensures that all rules are consistently linked, which is :+1:.

Issue #2432, opened 2023-08-10
Release versions 8.3/3.11.0


pepopowitz:

akeller:
Per my conversation with @toco-cam, Optimize 3.11 isn't the next version. Instead, we will fast-forward Optimize to 8.3.

See: https://camunda.slack.com/archives/C05G6R155MK/p1692099046612449

Product Hub epic (for docs): https://github.com/camunda/product-hub/issues/1247

akeller:
Update that voids my above comment ☝ 

https://github.com/camunda/product-hub/issues/1247#issuecomment-1735492171

**Optimize will release 3.11 in October.**

- 3.x releases for Optimize will continue
  - There are changes we started making for 8.x (removing 3.11), which we need to revert
  - We will likely need to support the team with careful reviews or PRs to make these changes in docs
- Docs dropdown will show 8.x and 3.x (this hasn’t changed since we last spoke)
- We should include a compatibility matrix to clarify what 8.x and 3.x versions map. This reinforces other visual clues in the dropdown and the Optimize footer.

akeller:
@pepopowitz do you want to add archiving 3.7 to this tasklist as well?

pepopowitz:
> @pepopowitz do you want to add archiving 3.7 to this tasklist as well?

As far as I can tell, archiving 3.7 would still be blocked by 3.11 not supporting OpenSearch? Or at least, I don't see anywhere that it says 3.11 includes OpenSearch support.

If anything, I'd remove the archival of 8.0/3.8, as I don't want to do that until 1.3/3.7 archival is already archived. Before I do that, I want to confirm that 8.3/3.11 doesn't yet support OpenSearch.

akeller:
@toco-cam @RomanJRW will 3.7 be supported past October? @pepopowitz does archiving (moving docs that are unsupported versions) coupled with releases.

toco-cam:
@akeller Yes, we do have to support it until April 2024

akeller:
Is archival the only thing that remains here @pepopowitz? Anything else still open from 8.3?

pepopowitz:
There are two items remaining under "follow-up tasks" - archiving 8.0 and fixing vale errors. Neither should block this from being marked complete, so this is DONE.

Issue #2189, opened 2023-06-06
Archive v1.3/v3.7.0


pepopowitz:
This work can proceed as planned with the minor release in October - https://github.com/camunda/camunda-docs/issues/2189#issuecomment-2299647095

---

Archive version 1.3.

## Progress

```[tasklist]
### Previous tasks
- [ ] #2477 
- [ ] #2480
- [ ] #2494
- [ ] #2500 
- [x] Isolation step 4: [publish 1.3.20 release](https://github.com/camunda/camunda-platform-docs/releases/tag/1.3.20)
```
```[tasklist]
### Tasks that must wait until https://github.com/camunda/product-hub/issues/997 is complete
- [ ] #4474 
- [ ] https://github.com/camunda/camunda-docs/pull/4483
- [ ] #4485
- [ ] #4512 
- [x] publish new 1.3 release (https://github.com/camunda/camunda-docs/releases/tag/1.3.22)
- [ ] #4475
- [ ] #4479
```

akeller:
This remains blocked see https://camunda.slack.com/archives/C02AWA0RF8A/p1710317097178729?thread_ts=1703089659.451659&cid=C02AWA0RF8A

akeller:
Optimize specific OpenSearch issue - https://github.com/camunda/issues/issues/393

akeller:
No longer blocked, ya? Shipping with the October Minor 🚀

pepopowitz:
> No longer blocked, ya? Shipping with the October Minor 🚀

Was there an announcement somewhere? How can you tell? Everything I look at in the issues you linked above suggests there is still a long ways to go.

akeller:
https://camunda.slack.com/archives/C070H8AE2BZ/p1723479767179909

> We proceed with archiving the 3.7 docs from the C8 docs as planned for the October release. The docs are still accessible beyond this date, but will be marked as belonging to a deprecated version

pepopowitz:
I don't know that I want to remove the BLOCKED until we have seen it ship.

akeller:
@pepopowitz I updated the description, but left the BLOCKED status and title.

akeller:
This can continue to proceed as planned - https://camunda.slack.com/archives/C070H8AE2BZ/p1724946689758439?thread_ts=1723479767.179909&cid=C070H8AE2BZ

PR #2502, opened 2023-08-28
archive (1.3) exclude: remove versions 1.3 and 3.7.0, part 1: reviewable steps

pepopowitz:
🛑 BLOCKED 🛑 

- [x] The unsupported/1.3 site is not yet in production.
- [x] The version dropdown is confusing when viewing an Optimize doc! I'm intending to resolve in this PR, see comments.
- [ ] OpenSearch is not yet supported in C8 Optimize (OpenSearch support is tracked at a platform level at https://github.com/camunda/product-hub/issues/997. Optimize support is tracked at https://github.com/camunda/product-hub/issues/1908.)

We should check back when #2432 is complete, to see if OpenSearch support has been included in Optimize. If it has, we can complete this issue. If it has not, we should schedule a reminder to check again a month or so later.

## Description

Part of #2189.

Includes the smaller items for removing version 1.3 from the main docs, to keep the PR in a reviewable state.

## Included in this PR

- Remove v1.3 & Optimize v3.7.0 from the version manifest files, so that they're not included in the deployed site
- Redirect all 1.3 / 3.7.0  traffic to `unsupported.docs.camunda.io/1.3` (or `stage.unsupported.docs.camunda.io/1.3` for the staging build)
- Move 1.3 to the "Unsupported versions" sub-list in the top nav version selector, redirecting to `unsupported.docs.camunda.io/1.3`. 
   - The label for this version combines both the main version & Optimize version. This combined version label shows when viewing both a main doc (shown below on the left side) and an Optimize doc (shown below on the right side): 

      image

## PR Checklist


docusaurus.config.js
```diff
@@ -87,7 +87,7 @@ module.exports = {
               className: "dropdown-unsupported-versions",
               value: "<b>Unsupported versions</b>",
             },
-            ...["1.2", "1.1", "1.0", "0.26", "0.25"].map((version) => ({
+            ...["1.3", "1.2", "1.1", "1.0", "0.26", "0.25"].map((version) => ({
```
pepopowitz, line 90: This array turns into the “unsupported versions” in the versions dropdown. It leads to some confusing information when viewing Optimize docs. 

We’ve customized the versions dropdown to display different _supported_ versions, depending on if you’re browsing the main docs or Optimize docs. This is done by [swizzling a component, and passing metadata to its children based on the current URL](https://github.com/camunda/camunda-platform-docs/blob/7b84279bba09848a4ef978af6164e21d1a1bf96d/src/theme/NavbarItem/index.js#L16-L18).

Unfortunately, the rendering of the static unsupported versions is a lot more simplistic than the rendering of the supported versions. Thus, when I’m viewing an Optimize doc, the static/unsupported versions show the _main docs_ versions, because that’s all that’s configured to show there. Like this: 

image

## How much effort do we want to put into this?

I am pretty sure I can swizzle one or two components to make that static list toggle between two different sets, depending on which docs instance you’re viewing. But I also wonder if this is necessary. This is somewhere between a “small” and “medium” amount of effort — closer to small, but there is risk that I can’t accomplish it. 

I can also think of at least two other solutions that are even smaller, but less correct. Since these are “unsupported” versions, we’ve already made decisions along the way guided by the idea that we don’t care about them as much as supported versions. Maybe they are good enough? 

## Possible solutions

1. A more complicated solution, in which we swizzle some components to make the “static list of unsupported versions” toggle based on the doc you’re viewing. This is a "small" to "medium" effort. I am unsure how the pre-Optimize-extraction versions should appear in this dropdown when viewing an Optimize page, if we go this route.
2. Just let the incorrect unsupported versions display when viewing an optimize doc (0 effort, what you see in my screenshot above).
3. Similar to the work for the upcoming release, we could display both the main version and the Optimize version in the dropdown, separated by a slash (e.g. `1.3 / 3.7.0`). This is a "small" effort. This is what it'd look like: 
   image
If we choose this option, the `/ 3.7.0` would also show in the unsupported versions when viewing a main docs page.

@akeller is there anything unclear here, and do you have any opinions on which route meets the "good enough" bar?

akeller: I agree on spending a small amount of time/effort here. Spending no effort sounds like it will create confusion and may even block people from finding what they need. Ideally it doesn't because people are running one of the 3 supported versions, but we can't ding people for not being on our schedule.

I like your 3rd option, particularly because we have similar upcoming work. Showing in the main docs page feels natural.

pepopowitz: My most recent commits address this. Just for clarity (and I'll add this to the body of the PR), 

This is what the version dropdown looks like when viewing most docs:

image

This is what the version dropdown looks like when viewing an Optimize doc:

image

howtos/release-procedure.md
```diff
@@ -63,7 +63,7 @@ To create the new versions:
    ];
    ```
 
-   - Add the new mapping in the first position of the array.
+   - Order of the items in the array is important. Add the new mapping in the first position of the array, because the first item is treated as the "current" version mapping.
```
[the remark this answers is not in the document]
akeller: Solid add!

akeller:
@pepopowitz are you aware of the status of Optimize 1.3? Will this be included with the January minor release?

pepopowitz:
> @pepopowitz are you aware of the status of Optimize 1.3? Will this be included with the January minor release?

OpenSearch for Optimize is tracked at https://github.com/camunda/product-hub/issues/1908. It is targeting 8.5.

(I'll update this issue to include a link to the Optimize-specific Product Hub issue.)

akeller:
This remains blocked. See internal update - https://camunda.slack.com/archives/C02AWA0RF8A/p1710317097178729?thread_ts=1703089659.451659&cid=C02AWA0RF8A

akeller:
@pepopowitz, is your plan to resurrect this PR or open a new one?

pepopowitz:
we should definitely close this and re-create later.

PR #2516, opened 2023-08-30
Stadtler best practices 82 hints

till-stadtler:
## Description

This PR removes caution hints on Camunda 8 not supporting Escalation Events, Link Events, OR Gateways, and Conditional Events. This is no longer the case as of version 8.2. The changes were done for doc version 8.2 and /docs.

## PR Checklist


akeller (approved):
Thanks again!

PR #2525, opened 2023-09-01
All static BPMN files converted to C8 and redesigned

till-stadtler:
## Description

I converted all static/bpmn/... BPMN files to the Camunda 8 format. This affects various areas in the docs, most prominently the best practices section. At the same time I redesigned the workflows in regards to the following aspects:
* all elements placed on grid
* consistent distance between elements
* old elements (smaller) replaced
* introduced more explicit modeling
* improved placement and flow to improve readability

There are two changes to the text:
* modeling best practices -> avoid using too many data objects: a message icon on a message flow is no longer possible. The BPMN file and respective text is updated under /docs and all versions
* following the customer success path: fixed callouts under /docs and all versions

It has to be evaluated if images that contain BPMN processes need to recreated as well, e.g., if major inconsistencies arise from the redesign of all BPMN files.

Thinks to consider when reviewing:
* do the redesigns improve readability from a BPMN perspective?
* can they be improved?
* are the displayed processes too small because they have to be resized because of their width?

## PR Checklist

- [x] I additionally requested a review by Bernd Ruecker


christinaausley:
Nice work @till-stadtler! This is a massive effort.

I'm not the best for a technical review of the BPMN changes, so you may want to wait on a review from @berndruecker. Otherwise, looks good to me!

akeller:
I saw some "problems" detected in the BPMN. If we were to make these downloadable (or link to the .bpmn file in GitHub) from the docs, should we resolve any of these problems? Or since we do not expect people to implement these directly, these "problems" should be expected? (I want to make sure no one perceives this as an actual problem with the BPMN.)

akeller (commented):
None of my feedback is blocking, just questions about how we want to review net new BPMN diagrams in the future. Thanks for bringing this topic up @till-stadtler! Great work here and attention to detail.

* For diagrams with "Task A" style naming, I don't see a need to follow strict guidelines. This means the diagram may not have a start event or end event with a name (or other missing element names). These are not meant to be implemented and what information they do include is necessary to teach a particular topic.
  * Tasks with name "..." are also allowed.
* For diagrams we use as examples, use cases, or in the Guides section, we should follow strict guidelines. All elements should be named. 
* Modeling guidelines want to join incoming flows. IMO, this reduces readability in some cases. @till-stadtler I left a comment directly on the BPMN on this. WDYT?

till-stadtler:
Hi @akeller,
having the diagrams downloadable (or being able to open them in Web Modeler via deep link) would be great! My colleague @shaarmann and I have proposed using deep links to @christian-konrad here: https://camunda.slack.com/archives/C05HA0PD0PM/p1694010294724199
Maybe enabling linking to github or downloading them is a quick win, compared to involving the modeler team to set up the deep links.

I did not model everything explicitly. My personal criteria to model implicitly are:
* no merging XOR gateway directly after the start event (the "error-type" example fits here)
In my opinion, a merging XOR gateway directly after the start event impacts readability negatively. 
* use merging XOR gateway if the flow follows a "block structure"
* avoid multiple merging XOR gateways next to each other
https://modeler.cloud.camunda.io/diagrams/994f8e8a-fddb-4660-ad43-59be5b430250--avoiding-multi-page-sequence-flows?v=753,206,0.861
In this example, I do not want to have the sequence flow after the link catch event to merge in the same XOR gateway, as it would destroy the "block structure". An additional merging XOR gateway would result in two merging XOR gateways next to each other and an increased width of the diagram. Of course, there is room for discussion here :)
In the end, some "problems" will remain, and we will have to accept that. Maybe we can add tooltips (similar to what is available in the templates) to explain such reasoning in the models.

We should make sure to add some dummy technical configuration so they can be run by Play in the future.

akeller (dismissed):
@till-stadtler thank you for the thorough feedback on this topic.

> Maybe enabling linking to github or downloading them is a quick win, compared to involving the modeler team to set up the deep links.

Agreed, reflected in an issue here - https://github.com/camunda/camunda-platform-docs/issues/2586.

I responded in the channel you linked re: opening files in Web Modeler directly via deep link. I think that's super cool! 

I'm good to merge this as a V1, since like you mentioned, adding dummy tech config so they can run in Play would be nice.

till-stadtler:
Hi @akeller,
I am currently collecting some feedback on the more tricky redesigns in the #bpmn channel. I'll also meet with Bernd next Tuesday! Thus, I propose to wait a bit. I assume, we'll get some helpful feedback!

till-stadtler:
Hi @akeller,
I incorporated my colleagues' feedback and also added dummy technical configuration. Processes that can potentially be run with Play are set to executable. Processes with Call Activities or unsupported BPMN elements are set to not executable. Some of the processes are not so easy to play through, because process variables need to be set on start or as result of a task.

Merging would be fine with me. If Bernd has more feedback next Tuesday, we can open a new PR for that. What do you think?

berndruecker:
+1 for not waiting for me and merge right away

akeller (approved):

akeller:
Merging this! Thanks for the collaboration and continued conversations on this topic. Great initiative @till-stadtler 🚀

PR #2536, opened 2023-09-04
docs(self-managed): Optimize and Helm chart 8.2.9 compatibility

aabouzaid:
## Description

Document Optimize and Helm chart 8.2.9 compatibility.

ASAP.

## PR Checklist

- [x] I have added changes to the relevant `/versioned_docs` directory.
- [x] I have added changes to the main `/docs` directory (aka `/next/`).
- [x] My changes require an [Engineering review](https://github.com/camunda/camunda-platform-docs/blob/main/howtos/documentation-guidelines.md#review-process).

### DIFF
--- a/docs/self-managed/platform-deployment/helm-kubernetes/upgrade.md
+++ b/docs/self-managed/platform-deployment/helm-kubernetes/upgrade.md
@@ -131,6 +131,19 @@ zeebe:
     runAsUser: 0
 ```
 
+### v8.2.9
+
+#### Optimize
+
+In Optimize 3.10.1, a new environment variable introduced for redirection URL. However, the change was not compatible with Camunda Platform Helm chart until it is fixed in 3.10.3 (and Helm chart 8.2.9). For that reason, those versions are coupled to certain Camunda Platform Helm chart versions.
+
+| Optimize version                  | Camunda Platform Helm chart version |
+| --------------------------------- | ----------------------------------- |
+| Optimize 3.10.1 - Optimize 3.10.2 | 8.2.0 - 8.2.8                       |
+| Optimize 3.10.3                   | 8.2.9+                              |
+
+No action is needed if you use Optimize 3.10.3 (shipped with this Helm chart version by default), but that Optimize version cannot be used out of the box with previous Helm chart versions.
+
 ### v8.2.3
 
 #### Zeebe Gateway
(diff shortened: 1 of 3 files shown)


docs/self-managed/platform-deployment/helm-kubernetes/upgrade.md
```diff
@@ -131,6 +131,19 @@ zeebe:
     runAsUser: 0
 ```
 
+### v8.2.9
+
+#### Optimize
+
+In Optimize 3.10.1, a new environment variable introduced for redirection URL. However, the change was not compatible with Camunda Platform Helm chart until it is fixed in 3.10.3 (and Helm chart 8.2.9). For that reason, those versions are coupled to certain Camunda Platform Helm chart versions.
```
christinaausley: A comment for my own understanding, but can you define redirection URL for me?

christinaausley (dismissed):
One non-blocking comment to clarify a sentence for me.

akeller:
This feels very similar to the conversation on https://github.com/camunda/camunda-platform-docs/pull/2413. See changes in that PR to add content to **Announcements** and **Supported environments**. Do we need to do something similar here?

PHWaechtler (dismissed):
Lgtm as far as the Optimize version goes - I do not know the helm chart versioning so cannot say whether those are correct. From the ticket it also says this is fixed n 3.9.5 and 3.11.0-alpha3 as well as 3.10.3. I dont know if that needs added somewhere too.

akeller:
@aabouzaid, I need some help understanding this change - is this considered a breaking change? See my comment above - https://github.com/camunda/camunda-platform-docs/pull/2536#issuecomment-1707306625

I'm worried about the visibility since the docs are only versioned by minors (not alphas or patches). This looks like a change that impacts Optimize for 3.10+ and 3.9, but the 3.9 impact is not documented.

docs/self-managed/platform-deployment/helm-kubernetes/upgrade.md
aabouzaid: With the users logging into one of the applications, it sends them to the Identity service (namely, Keycloak).
Once the users logged in, they are redirected back to the application. For security, certain redirection URL are allowed (so no one can manipulate them and redirects the users to non-trusted pages).

aabouzaid:
@akeller this is not a breaking change per se, it's a bug/fix from the Optimize, which has side effects and the distro team implemented a workaround in the Helm chart till the final fix from the Optimize team.

I've added that to the announcement page, please check if it's as expected.

akeller (approved):
I'm good with this. We may rework it in the future, but I'd rather go ahead and get it merged and released.

Issue #1143, opened 2022-08-11
Investigate SEO for current docs version


akeller:
In https://github.com/camunda/product-hub/issues/232 we learned all sorts of things about SEO & versioned documentation. We should apply those findings to the C8 docs, to avoid reliving that painful experience.

## Acceptance criteria

* Only the current version (`8.1` for now) should be indexed and searchable from Google. 
  * #1905 describes [an example URL](https://docs.camunda.io/docs/0.26/reference/bpmn-workflows/multi-instance/)
* Internal search should clearly mark what version, if multiple versions show up in the search results.
* Identified work is based on best practices learned in https://github.com/camunda/product-hub/issues/232.

## Included work
- [x] https://github.com/camunda/camunda-platform-docs/pull/2112: Sitemap should include only the supported version(s) of docs
- [x] https://github.com/camunda/camunda-platform-docs/pull/2121: tag, links, and unsupported versions removed from sitemap
- [x] #2577: Old versions of docs should emit a `link rel=canonical` pointing at the current version of the docs (the URL with no version in it)

## Background

`camunda 8 bpmn coverage` in Google returns `https://docs.camunda.io/docs/0.25/components/zeebe/bpmn-workflows/bpmn-coverage/` as the first hit. (update: this example now links to the versionless URL, which is correct.)

Algolia docsearch returns `https://docs.camunda.io/docs/0.26/reference/bpmn-workflows/bpmn-coverage/` with similar search terms (although I could not reproduce this).

akeller:
Today, much to my chagrin, I found Google linking to a `1.3` page.

We'll need to spend some time user-designating canonicals before we end up in a situation like our C7 docs.

akeller:
Thanks for your updates here @pepopowitz!

akeller:
#2112 may fix this

pepopowitz:
> #2112 may fix this

It didn't, but 🤞 that #2121 does!

akeller:
#2121 is live and the sitemap correctly shows only supported versions of the product.

But Amara, this would still be incorrect, wouldn't it?

Amara and Steven have some paranoia about accidentally doing something to completely sabotage the SEO, so Amara started pairing down what was in the sitemap by removing unsupported/unmaintained versions first. 

[...]

PR #2577, opened 2023-09-11
[WIP] feat: properly identify canonicals for non-latest docs

pepopowitz:
WIP/Draft -- I'm pushing it up early for feedback on the logic and mechanics of assigning/determining canonicals 

## Description

Part of #1143.

Adds support for proper canonicals on non-latest docs.

## Background

As discovered in our C7 docs, we should specify canonicals for all non-latest docs, and we should always point them at their latest version counterpart. 

Docusaurus applies self-referential canonicals to all pages, which [Google has identified as having no effect on SEO](https://www.youtube.com/watch?v=TepFVYrBVg0&t=968s) when the canonical URL matches the browsed URL exactly. 

Docusaurus _does_ provide the ability to override the canonical via the `<Head>` component, which is utilized in this PR. 

Docusaurus also provides the ability to set custom frontmatter in any markdown document, but it does not currently have any effect on the canonical emitted by a page. This ability to set custom frontmatter properties is also utilized in this PR.

## Proposed architecture for specifying individual canonicals

As seen in this PR: 

1. Individual markdown files can specify canonical information in their frontmatter.
2. A swizzled `Metadata` component determines the proper canonical for each page. 
   1. If canonical information is specified in the frontmatter for a page, that is considered.
   2. If no canonical information is specified, a document in the latest version is potentially identified as canonical.

## Proposed logic for determining canonicals

There is a spec file in this PR that lays out the rules I propose for specifying/identifying canonical URLs. Summarizing here, but see extra contents in the file diff: 

1. when the current doc has a `canonicalUrl` property in its frontmatter
   1. when that URL exists in a newer version
      * it returns the value of the `canonicalUrl` property
   2. when that URL does not exist in a newer version
      * it throws an exception
         because it would be a canonical pointing at a 404. it probably means you moved a doc, and you should go back and point the old ones at the new location.
  2. when the current doc has a `canonicalId` property in its frontmatter
     1. when the latest version has a doc with that ID
        * it returns the URL of the matching latest version doc, which is the one without a version number in it
     2. when the latest version does not have a doc with that ID
        * it throws an exception
           it probably means you moved a doc, and you should go back and point the old ones at the new location
   3. when the current doc has no canonical frontmatter
      1. when there are newer docs with with the same id
         1. when one version newer has the same id
            * it returns the versioned URL of that one version newer
               because it's better for us to group as many canonicals as we can
               (instead of pointing them at dead URLs)
         1. when all newer versions have the same id
            * it returns the latest version's URL
               which is the one without a version number in it
      2. when there are no newer docs with the same id
         1. when the current doc is the latest version
            * it returns the URL for the current doc
         2. when the current doc is a non-latest version 
            * it returns the URL with the version removed
               because we have any way to link it to a known document.
               and we're going to hope that redirects send it to the right place, and that google is okay with that.

## PR Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
   "devDependencies": {
     "@playwright/test": "^1.32.2",
     "@swc/core": "^1.3.49",
+    "@types/jest": "^29.5.4",
     "husky": "^8.0.3",
     "jest": "^29.6.4",
     "lint-staged": "^14.0.1",
--- a/optimize/components/what-is-optimize.md
+++ b/optimize/components/what-is-optimize.md
@@ -2,6 +2,8 @@
 id: what-is-optimize
 title: What is Optimize?
 description: "Leverage process data and analyze areas for improvement."
+canonicalUrl: "/optimize/next/components/what-is-optimize"
+canonicalId: "components/what-is-optimize"
 ---
 
 :::note
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,6 +23,7 @@
       "devDependencies": {
         "@playwright/test": "^1.32.2",
         "@swc/core": "^1.3.49",
+        "@types/jest": "^29.5.4",
         "husky": "^8.0.3",
         "jest": "^29.6.4",
         "lint-staged": "^14.0.1",
@@ -4900,6 +4901,16 @@
       "dev": true,
       "dependencies": {
         "@types/istanbul-lib-report": "*"
+      }
+    },
+    "node_modules/@types/jest": {
+      "version": "29.5.4",
+      "resolved": "https://registry.npmjs.org/@types/jest/-/jest-29.5.4.tgz",
+      "integrity": "sha512-PhglGmhWeD46FYOVLt3X7TiWjzwuVGW9wG/4qocPevXMjCmrIc5b6db9WjeGE4QYVpUAWMDv3v0IiBwObY289A==",
+      "dev": true,
+      "dependencies": {
+        "expect": "^29.0.0",
+        "pretty-format": "^29.0.0"
       }
     },
     "node_modules/@types/json-schema": {
@@ -29421,6 +29432,16 @@
       "dev": true,
       "requires": {
         "@types/istanbul-lib-report": "*"
+      }
+    },
+    "@types/jest": {
+      "version": "29.5.4",
+      "resolved": "https://registry.npmjs.org/@types/jest/-/jest-29.5.4.tgz",
+      "integrity": "sha512-PhglGmhWeD46FYOVLt3X7TiWjzwuVGW9wG/4qocPevXMjCmrIc5b6db9WjeGE4QYVpUAWMDv3v0IiBwObY289A==",
+      "dev": true,
+      "requires": {
+        "expect": "^29.0.0",
+        "pretty-format": "^29.0.0"
       }
     },
     "@types/json-schema": {
--- a/src/theme/DocItem/Metadata/determineCanonical.js
+++ b/src/theme/DocItem/Metadata/determineCanonical.js
@@ -17,6 +17,13 @@
  *  * @param {Doc} doc
  */
 function determineCanonical(doc) {
+  // const {
+  //   metadata: {
+  //     unversionedId,
+  //     frontMatter: { canonicalUrl, canonicalId },
+  //   },
+  // } = doc;
+
   if (doc.frontMatter.canonicalUrl) {
     return doc.frontMatter.canonicalUrl;
   }
--- a/src/theme/DocItem/Metadata/determineCanonical.js
+++ b/src/theme/DocItem/Metadata/determineCanonical.js
@@ -1,0 +1,27 @@
+// @ts-check
+// Note: these type definitions are only checked within an editor. We aren't set up
+//   for type-checking in the build (yet!) but having these types defined helped me write this code.
+
+/**
+ * @typedef {object} FrontMatter
+ * @property {string?} canonicalUrl
+ *
+ * @typedef {object} Doc
+ * @property {FrontMatter} frontMatter
+ */
+
+// sjh - this is how to import the original type, if you need it later
+// *  * @param {import("@docusaurus/theme-common/lib/internal").DocContextValue} doc
+
+/**
+ *  * @param {Doc} doc
+ */
+function determineCanonical(doc) {
+  if (doc.frontMatter.canonicalUrl) {
+    return doc.frontMatter.canonicalUrl;
+  }
+
+  return "x";
+}
+
+module.exports = determineCanonical;
--- a/src/theme/DocItem/Metadata/determineCanonical.spec.js
+++ b/src/theme/DocItem/Metadata/determineCanonical.spec.js
@@ -20,5 +20,47 @@ describe("determineCanonical", () => {
 
       expect(result).toEqual("/docs/welcome/");
     });
+
+    describe("when that URL exists in a newer version", () => {
+      // it("returns the value of the canonicalUrl");
+    });
+
+    describe("when that URL does not exist in a newer version", () => {
+      // it("throws an exception");
+    });
+  });
+
+  describe("when the current doc has a canonicalId in its frontmatter", () => {
+    describe("when the current version has a doc with that ID", () => {
+      // it("returns the URL of the matching current version doc");
+    });
+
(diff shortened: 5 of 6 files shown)


pepopowitz:
@akeller @christinaausley I'm still working on building this functionality out, but I wanted to tag you early to get your feedback on the logic & mechanics I'm proposing in this PR. Let me know if you have questions! I'm sure I'm leaving some things out, and will likely fill in more context tomorrow morning when I think of it.

package.json
```diff
@@ -51,6 +51,7 @@
   "devDependencies": {
     "@playwright/test": "^1.32.2",
     "@swc/core": "^1.3.49",
+    "@types/jest": "^29.5.4",
```
pepopowitz, line 54: This package is needed so that I can type-check a test file. (It adds jest functions like `describe` and `it` to the global context, so that TypeScript doesn't complain about them.)

akeller:
I don't think you missed anything here. I did have to remind myself "moving a doc" includes renaming, moving it in the structure, renaming a parent directory, or any other impacts to the URL structure. 

I really like your test-first approach, too. It would make a great spotlight at our team meeting!

docusaurus.config.js
```diff
@@ -7,6 +7,9 @@ module.exports = {
   url: "https://docs.camunda.io",
   // baseUrl: "/camunda-cloud-documentation/",
   baseUrl: "/",
+  customFields: {
+    canonicalUrlRoot: "https://docs.camunda.io",
```
pepopowitz, line 11: Thinking ahead to unsupported sites -- they will not want to use their own host (unsupported.docs....) as the canonical. We'll want to point them over to the main site.

akeller:
The commits 🤣

Dockerfile.build
pepopowitz, line 1: I kept crashing my system locally by running `npm run build` in a console, so I stuck it in docker. It felt useful to share.

README.md
```diff
@@ -15,7 +15,7 @@ Did you know Camunda Engineers are the primary documentation authors?
 
 - [Documentation guidelines](./howtos/documentation-guidelines.md)
 - [Setup](./howtos/setup.md)
-- [Redirect rules](./howtos/redirect-rules.md)
+- [Moving content](howtos/moving-content.md)
```
pepopowitz, line 18: Renamed this doc because it is not just redirect rules anymore.

pepopowitz:
@akeller @christinaausley I've upgraded this from a draft to ready for review!

howtos/moving-content.md
```diff
@@ -158,3 +168,77 @@ If you wish to test `.htaccess` rules, you have a couple options:
...
+
+Specify a `canonicalUrl` property in the frontmatter of an older document, pointed at the relative URL of a document that exists within our documentation. Example:
+
+```
+---
+canonicalUrl: "/docs/apis-tools/operate-api/overview"
+---
+```
+
+The `canonicalUrl` is must match the path of a document exactly. Follow these rules to ensure that a `canonicalUrl` matches an existing document:
```
akeller, line 190:
```suggestion
The `canonicalUrl` must match the path of a document exactly. Follow these rules to ensure that a `canonicalUrl` matches an existing document:
```

howtos/moving-content.md
```diff
@@ -1,14 +1,24 @@
...
-Since the docs are always changing, it's possible for users to bookmark or share links to pages that move. To ensure that these users can still find their content, we use redirect rules to map old URLs to new URLs.
+Our docs are always changing, and sometimes we move content to be better organized. When we move content, we must intentionally preserve the connection between a document's old location and its new location. We do this in two ways:
+
+1. Specify [redirect rules](#redirect-rules) to map old URLs to new URLs. This ensures that users who have bookmarked or shared a link to the old page can still find their content.
+
+2. Specify [canonical references](#canonical-references) from previous versions to the moved location. This ensures that search engine crawlers can identify the canonical version of a document, and serve the most accurate results to users' search queries.
+
+This guide provides help on both techniques.
+
+## Redirect Rules
```
akeller, line 11: Sentence case

```suggestion
## Redirect rules
```

akeller (commented):
Couple of nit picks as suggestions.

Piece I'm missing or need to clarify/verify -

When I add a new page (author a new page), do I need to do anything? Declare a canonical at that point if it's in a supported version?

If it's in `/next/` and the page doesn't exist yet in a supported version, I do nothing. Correct?

pepopowitz:
@akeller these were really great questions. They prompted me to write up some scenarios, and now I'm not sure about one of them!

> When I add a new page (author a new page), do I need to do anything? Declare a canonical at that point if it's in a supported version?
> If it's in /next/ and the page doesn't exist yet in a supported version, I do nothing. Correct?

Correct on both accounts! You wouldn't need to declare any canonical frontmatter until a page had been moved. 

Here's how things would play out in a variety of scenarios: 

1. I add a new document in vNext, example `/docs/next/what-is-recombobulate/`. 
   
   `determineCanonical` would return the URL of the page with the version removed, `/docs/what-is-recombobulate/`. 

   As I say this out loud, this is the only case I'm not so certain about. It's covered under [this scenario of the unit tests](https://github.com/camunda/camunda-platform-docs/blob/3fc2ba2b077d7c3e36366708a31bc62d5a14a914/src/theme/DocItem/Metadata/determineCanonical.spec.js#L285-L295) and it originally made sense to me that we should remove versions from canonicals when in doubt. But I wonder if it would be better to self-reference for docs that are in next but nowhere else? 

2. That page graduates from vNext to vCurrent when we do our next release.

   * For the page in vNext, `/docs/next/what-is-recombobulate/`, `determineCanonical` would find the document in vCurrent with the same ID, and return `/docs/what-is-recombobulate/`. This is covered by [this test scenario](https://github.com/camunda/camunda-platform-docs/blob/3fc2ba2b077d7c3e36366708a31bc62d5a14a914/src/theme/DocItem/Metadata/determineCanonical.spec.js#L239-L263).
   * For the page in vCurrent, `/docs/what-is-recombobulate/`, `determineCanonical` would find the page itself, and return `/docs/what-is-recombobulate/`. This is covered by [this test scenario](https://github.com/camunda/camunda-platform-docs/blob/3fc2ba2b077d7c3e36366708a31bc62d5a14a914/src/theme/DocItem/Metadata/determineCanonical.spec.js#L239-L263).

3. I add another new document in the latest version, example `/docs/what-is-abominate/`.
   
   `determineCanonical` would match the page itself, and return `/docs/what-is-abominate/`. This is covered by [this test](https://github.com/camunda/camunda-platform-docs/blob/3fc2ba2b077d7c3e36366708a31bc62d5a14a914/src/theme/DocItem/Metadata/determineCanonical.spec.js#L181-L201).

4. I add another new document in an older version, example `/docs/8.0/what-is-exacerbate/`.

   `determineCanonical` would match the page itself, and return `/docs/8.0/what-is-exacerbate/`. This is covered by [this test](https://github.com/camunda/camunda-platform-docs/blob/3fc2ba2b077d7c3e36366708a31bc62d5a14a914/src/theme/DocItem/Metadata/determineCanonical.spec.js#L181-L201).

Let me know if it's useful to describe more scenarios in this way. Also, let me know if you disagree with how I'm currently handling scenario #1 😅

akeller:
@pepopowitz I would be very surprised if scenario 4 happened. This feels like a very rare exception or a backporting mistake. That said, your logic makes sense.

For scenario 1, having a self-referential canonical for `/next/` feels like the safest approach. The page exists but isn't indexed. The non-versioned URL doesn't exist yet. These pages graduate and qualify for an update in scenario 2, the indexable page with the non-versioned URL. 

I would update the docs to say what you have here:

> You wouldn't need to declare any canonical frontmatter until a page had been moved.

As a docs author or content maintainer, I know when action is required and when action is not required, so I know how to address build issues if they come up quickly.

I would also mention this in Slack when you merge this change and it's live. This will (hopefully) help manage expectations.

pepopowitz:
> For scenario 1, having a self-referential canonical for `/next/` feels like the safest approach. The page exists but isn't indexed. The non-versioned URL doesn't exist yet. These pages graduate and qualify for an update in scenario 2, the indexable page with the non-versioned URL.
> ...
> I would update the docs to say what you have here:
> 
> > You wouldn't need to declare any canonical frontmatter until a page had been moved.

Both these are addressed in new commits!

PR #2654, opened 2023-09-26
fix: remove unnecessary complexity from canonical regular expression

pepopowitz:
## Description

Fixes the docs for Safari versions <= 16.3, which [do not yet support lookbehind in regular expressions](https://caniuse.com/js-regexp-lookbehind).

The regular expression used here was overly complex anyway. We were previously doing something with the resulting `matches`. Now, the expression is only used for a test, and we don't use the match. 

Initially raised in [Slack](https://camunda.slack.com/archives/C026U8GBNSW/p1695712165348429) by @remcowesterhoud (cc @christinaausley).

## Proof that the tests still pass (because we don't run them in CI)

image

## PR Checklist

### DIFF
--- a/src/theme/DocItem/Metadata/determineCanonical.js
+++ b/src/theme/DocItem/Metadata/determineCanonical.js
@@ -107,7 +107,7 @@ function determineCanonicalFromDoc(currentDoc, currentPlugin) {
     .find((doc) => doc.id === unversionedId);
 
   if (match) {
-    if (match.path.match(/(optimize|docs)((next|[0-9\.]*)\/)/)) {
+    if (/(optimize|docs)((next|[0-9\.]*)\/)/.test(match.path)) {
       // This finds docs whose matches are non-latest versions.
       //  These docs would probably benefit from adding canonical frontmatter.
       console.log(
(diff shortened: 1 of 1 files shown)


src/theme/DocItem/Metadata/determineCanonical.js
```diff
@@ -107,7 +107,7 @@ function determineCanonicalFromDoc(currentDoc, currentPlugin) {
     .find((doc) => doc.id === unversionedId);
 
   if (match) {
-    if (match.path.match(/(?<=(optimize|docs)\/)((next|[0-9\.]*)\/)(?=.+)/)) {
+    if (/(optimize|docs)((next|[0-9\.]*)\/)/.test(match.path)) {
```
pepopowitz, line 110: This test is _also_ only used to emit a warning message to the console, which is more useful during the server build than when viewing the page in a browser. 

I'd imagine docusaurus exposes a context that tells us if we are operating on the server or browser. If you'd prefer, I can investigate this and add more guarding against emitting this message in the browser.

akeller: I don't have strong feelings about the additional investigation.

akeller (approved):

Issue #1904, opened 2023-04-05
Lint docs for specific words


akeller:
> 💡 We could use https://vale.sh/ to lint our docs for such words and instead of failing the CI we could use it to warn the author and reviewers of the usage with a suggestion about when to use which form.

_Originally posted by @korthout in https://github.com/camunda/camunda-platform-docs/pull/1900#discussion_r1158669767_

akeller:
Something to think about @christinaausley and @pepopowitz! Could be a quick win to implement and hugely valuable to reduce some of the manual work Christina does for reviews, specifically reviews that include certain terms (upgrade vs. update or casing on 3rd party terms).

korthout:
Another case where this could be applied is between American and British English styles. No need to do it all at once but I'm just highlighting the potential.

pepopowitz:
Yes, this would be great! 

I was originally thinking I'd add this as an item to https://github.com/camunda/camunda-platform-docs/issues/1742, but I stopped, because that one tracks reducing friction for _contributors_, not maintainers. I will peruse the list of open issues soon to see if there are enough similar issues to this, in which case I'll create another epic to track reducing friction for maintainers.

pepopowitz:
> Yes, this would be great!
> 
> I was originally thinking I'd add this as an item to #1742, but I stopped, because that one tracks reducing friction for _contributors_, not maintainers. I will peruse the list of open issues soon to see if there are enough similar issues to this, in which case I'll create another epic to track reducing friction for maintainers.

Nope, I lied. I _am_ adding this to #1742, because automating this would also tighten the contributor's feedback loop -- they can go back and fix these things before @christinaausley even gets to her review.

pepopowitz:
Note that vale has been added to this project in a `check-format` workflow, as part of #2675. That makes the work for this issue a little easier, as we need only configure to the rules we want.

akeller:
@christinaausley the foundation is ready for any additional Vale work. This would be an excellent tool for you to have in your tech writer toolbox and make your reviews easier (plus strengthen your partnership with the robot overlords). And so timely! https://camunda.slack.com/archives/C026U8GBNSW/p1707935866675739

christinaausley:
@pepopowitz wanna pair on this after our authentication/tutorial work? Assigning this to myself for now, but if you're interested feel free to assign yourself as well.

PR #2664, opened 2023-09-27
[do not merge] prototype: detect cross-instance/cross-version links (and report them as comments; and report prettier issues as comments too.)

### LINKED ISSUE #1104: Auto-format commits that come through GitHub UI
For contributors who commit through a local environment, we run prettier to auto-format changes via a pre-commit hook. If a contributor makes a commit through the GitHub UI, however, that pre-commit hook does not run. Recent examples: #1089 and 1094. 

This often results in unformatted code being committed, and CI failing because it runs prettier to _validate_ the format. This is unfriendly friction for contributors who want to make small, light-weight changes without setting up a local environment. 

### Implementation

There are a few ways I can think of to address this, but when implementing we are definitely not limited to these options: 

1. ~Add a GitHub action that formats & commits on all PRs/pushes.~

    **Update:** we are unable to configure our repo such that our workflows have write access on forks. [See this PR/comment](https://github.com/camunda/camunda-platform-docs/pull/1140#issuecomment-1213307313) for more details on what we do/do not have permissions to do against a forked repo. 

    _Out of date comments:_ When we initially configured prettier, [we intentionally avoided doing this](https://github.com/camunda/camunda-platform-docs/issues/1101) because the contributor will not have verified the auto-formatting changes in their local environment. I think my opinion has changed on this, now that we have a pre-commit hook to auto-format for contributors with local environments, and the gap right now is only for people who aren't checking things in a local environment _anyway_. 

2. Add a GitHub action that formats & commits when a PR is merged. 
  
    This _does_ have the potential to introduce changes that no one looked at. This isn't a baseless fear -- [here's an example of some auto-formatting that completely broke a code-snippet](https://camunda.slack.com/archives/C026U8GBNSW/p1660037553619989) (the linked Slack comment makes reference to [this page](https://docs.camunda.io/docs/components/modeler/feel/language-guide/feel-list-expressions/)). 

3. ~Add a GitHub action that listens for a specific command, and on-demand auto-formats & commits.~
  
    **Update:** we are unable to configure our repo such that our workflows have write access on forks. [See this PR/comment](https://github.com/camunda/camunda-platform-docs/pull/1140#issuecomment-1213307313) for more details on what we do/do not have permissions to do against a forked repo. 

    _Out of date comments:_ This is nice because it only happens when we ask it to....but it seems like it might be a bit more work than the first option.

4. Add a PR review with suggestions for each of the format violations. 

    The PR author can then accept the changes through the GitHub UI. We can use [this action](https://github.com/EPMatt/reviewdog-action-prettier#with-github-pr-review-reporter). 

### Also in scope

If we find that the solution covers all the right bases, and there is support for removing the pre-commit hook, we should do that as part of this work too.

### LINKED ISSUE #1876: Detect improperly formatted links

pepopowitz:
DO NOT MERGE! This PR is experimental, to explore ideas and gather feedback.

## Description

Initially intended to resolve #1876, but through a small additional effort I was able to extend it to resolve #1104. It also opens up the door for a variety of other quality checks.

This PR introduces automated PR review comments for two types of formatting -- Prettier, and [Vale](https://vale.sh/). It also intentionally includes two formatting violations, to demonstrate what happens when format violations are included in a pull request.

There are many details described below. I'm looking for feedback of any kind, for any or all of it.

## Detecting cross-instance/cross-version links

This is done with custom Vale rules. The rules are inspired by [our guidance on preferred link format](https://github.com/camunda/camunda-platform-docs/blob/main/howtos/documentation-guidelines.md#internal-links). There are four of them: 

1. Within the main docs instance, any link that starts with `docs/` but doesn't end with `.md` or `.mdx` triggers a warning (navigating by source file path is preferred).
2. Within the main docs instance, any link that starts with `/optimize/` triggers a warning (using `$optimize$` is preferred).
3. Within the optimize docs instance, any link that starts with `optimize/` but doesn't end with `.md` or `.mdx` triggers a warning (navigating by source file path is preferred).
4. Within the optimize docs instance, any link that starts with `/docs/` triggers a warning (using `$docs$` is preferred).

This doesn't cover _all_ possible links that could cause cross-version linking, but it covers the large majority of issues we currently have.

## Reporting as comments

There are multiple options for reporting from the Vale GitHub action, but I chose reporting as comments because it's the most visible. I also set these rules to **not** fail the check. This is because I don't have 100% confidence that these rules will always prevent us from merging a PR.

## Vale

Vale was mentioned in #1904 as a possible tool for linting our docs for specific words/phrases. I looked at multiple tools for finding cross-instance/cross-version links, but this suggestion inspired me to use Vale. Configuring it here unlocks a massive opportunity to automate certain things that @christinaausley is watching for, enabling her to focus on less tedious work. One obvious current example -- we could add a rule to check for use of `Camunda Platform`. 

There are also a ton of language style rules (e.g. passive voice), spelling, ...so many options... that we can easily enable. See [Vale's built-in rules](https://vale.sh/explorer/), the [write-good rules](https://github.com/errata-ai/write-good), and [Microsoft's rules](https://github.com/errata-ai/Microsoft) -- all of these are easily enabled, with a line or two of configuration; we can also use any of them as inspiration to write our own.

## Format-checking as separate from the `build-docs` workflow

There is a technical reason that I chose to extract the format-checking steps from the `build-docs` workflow. The `build-docs` workflow can be triggered on `push` for Camunda-org members, which prevents the reporting aspects in this PR from identifying newly introduced changes. There are nearly 500 violations of the cross-linking rules already in the codebase -- so showing _all_ errors is not an option.

Having said that, I actually _prefer_ having these formatting checks separate from the build itself. It felt like build-docs was doing a lot of things, and having these things run in parallel should speed up the build-docs workflow a bit.

## Prettier reporting

This is completely separate from the Vale checking, and didn't need to be part of this experiment. But once I'd extracted a check-format workflow, it made sense to also move prettier in here....and once I'd done that, I really wanted to see what it'd be like to have comments on the PR, so that fly-by changes can be fixed when they include formatting issues. This is a long-standing pain point, described more in #1104. 

It shouldn't. I'm going to break it into smaller pieces and open new PRs, after gathering feedback.

### DIFF
--- a/.github/workflows/check-format.yaml
+++ b/.github/workflows/check-format.yaml
@@ -1,0 +1,12 @@
+name: check-format
+on: [pull_request]
+
+jobs:
+  check-format:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Verify format
+        run: npm run format:check
+      - uses: errata-ai/vale-action@reviewdog
+        with:
+          filter_mode: added
--- a/.github/workflows/check-format.yaml
+++ b/.github/workflows/check-format.yaml
@@ -16,6 +16,7 @@ jobs:
         with:
           github_token: ${{ secrets.github_token }}
           reporter: github-pr-review
+          fail_on_error: true
       - name: Verify format with Vale
         uses: errata-ai/vale-action@reviewdog
         with:
--- a/.github/workflows/check-format.yaml
+++ b/.github/workflows/check-format.yaml
@@ -16,3 +16,4 @@ jobs:
       - uses: errata-ai/vale-action@reviewdog
         with:
           filter_mode: added
+          reporter: github-pr-review
--- a/.github/workflows/check-format.yaml
+++ b/.github/workflows/check-format.yaml
@@ -11,14 +11,14 @@ jobs:
           node-version: 16
       - name: Install Dependencies
         run: npm ci
+      - name: Verify format with Vale
+        uses: errata-ai/vale-action@reviewdog
+        with:
+          filter_mode: added
+          reporter: github-pr-review
       - name: Verify format with Prettier
         uses: EPMatt/reviewdog-action-prettier@v1
         with:
           github_token: ${{ secrets.github_token }}
           reporter: github-pr-review
           fail_on_error: true
-      - name: Verify format with Vale
-        uses: errata-ai/vale-action@reviewdog
-        with:
-          filter_mode: added
-          reporter: github-pr-review
--- a/.github/workflows/check-format.yaml
+++ b/.github/workflows/check-format.yaml
@@ -9,6 +9,8 @@ jobs:
       - uses: actions/setup-node@v1
         with:
           node-version: 16
+      - name: Install Dependencies
+        run: npm ci
       - name: Verify format
         run: npm run format:check
       - uses: errata-ai/vale-action@reviewdog
--- a/.github/workflows/check-format.yaml
+++ b/.github/workflows/check-format.yaml
@@ -5,6 +5,10 @@ jobs:
   check-format:
     runs-on: ubuntu-latest
     steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 16
       - name: Verify format
         run: npm run format:check
       - uses: errata-ai/vale-action@reviewdog
--- a/.github/workflows/check-format.yaml
+++ b/.github/workflows/check-format.yaml
@@ -11,9 +11,13 @@ jobs:
           node-version: 16
       - name: Install Dependencies
         run: npm ci
-      - name: Verify format
-        run: npm run format:check
-      - uses: errata-ai/vale-action@reviewdog
+      - name: Verify format with Prettier
+        uses: EPMatt/reviewdog-action-prettier@v1
+        with:
+          github_token: ${{ secrets.github_token }}
+          reporter: github-pr-review
+      - name: Verify format with Vale
+        uses: errata-ai/vale-action@reviewdog
         with:
           filter_mode: added
           reporter: github-pr-review
--- a/.vale.ini
+++ b/.vale.ini
@@ -1,0 +1,7 @@
+StylesPath = "styles/camunda"
+
+[{docs,versioned_docs}/**/*.{md,mdx}]
+BasedOnStyles = docsInstance
+
+[{optimize,optimize_versioned_docs}/**/*.{md,mdx}]
+BasedOnStyles = optimizeInstance
--- a/styles/camunda/docsInstance/hrefDocsToDocs.yml
+++ b/styles/camunda/docsInstance/hrefDocsToDocs.yml
@@ -1,0 +1,7 @@
+extends: existence
+message: "Improper link format: '%s'. Please specify the file extension."
+level: warning
+nonword: true
+scope: raw
+tokens:
+  - "\\[[^\\]]*\\]\\((\\.)?(\\/)?docs(?!.*(\\.md|\\.mdx))[^\\)]*\\)"
--- a/.github/workflows/build-docs.yaml
+++ b/.github/workflows/build-docs.yaml
@@ -24,11 +24,6 @@ jobs:
           node-version: 16
       - name: Install Dependencies
         run: npm ci
-      - name: Verify format
-        run: npm run format:check
-      - uses: errata-ai/vale-action@reviewdog
-        with:
-          filter_mode: added
       - name: Build
         run: npm run build
         env:
--- a/.github/workflows/build-docs.yaml
+++ b/.github/workflows/build-docs.yaml
@@ -26,6 +26,9 @@ jobs:
         run: npm ci
       - name: Verify format
         run: npm run format:check
+      - uses: errata-ai/vale-action@reviewdog
+        with:
+          filter_mode: added
       - name: Build
         run: npm run build
         env:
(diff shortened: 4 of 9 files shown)


.github/workflows/check-format.yaml
```diff
@@ -0,0 +1,24 @@
...
+        with:
+          node-version: 16
+      - name: Install Dependencies
+        run: npm ci
+      - name: Verify format with Prettier
+        uses: EPMatt/reviewdog-action-prettier@v1
+        with:
+          github_token: ${{ secrets.github_token }}
+          reporter: github-pr-review
+          fail_on_error: true
```
pepopowitz, line 19: This step no longer calls our `npm run format:check` script, but it _does_ still use our installed version of prettier. [The significance and importance of this is described by the action's author](https://github.com/EPMatt/reviewdog-action-prettier#why-does-the-action-require-a-packagejson-with-prettier-declared-as-a-dependency).

.vale.ini
```diff
@@ -0,0 +1,7 @@
+StylesPath = "styles/camunda"
+
+[{docs,versioned_docs}/**/*.{md,mdx}]
+BasedOnStyles = docsInstance
+
+[{optimize,optimize_versioned_docs}/**/*.{md,mdx}]
+BasedOnStyles = optimizeInstance
```
pepopowitz, line 7: We can easily configure the ability to share rules across the docs instances, but for this prototype I only included unique rules for each instance.

styles/camunda/docsInstance/hrefDocsToDocs.yml
```diff
@@ -0,0 +1,7 @@
+extends: existence
+message: "Improper link format: '%s'. Please specify the file extension."
+level: warning
+nonword: true
+scope: raw
+tokens:
+  - "\\[[^\\]]*\\]\\((\\.)?(\\/)?docs(?!.*(\\.md|\\.mdx))[^\\)]*\\)"
```
pepopowitz, line 7: This regex (and the others below) were tested by me with a variety of values. Trust me..????

styles/camunda/docsInstance/hrefDocsToDocs.yml
```diff
@@ -0,0 +1,7 @@
+extends: existence
+message: "Improper link format: '%s'. Please specify the file extension."
+level: warning
```
pepopowitz, line 3: I'm setting these as warnings, not errors, because I am again not 100% confident that they will _always_ be violations.

.vale.ini
akeller: I love the flexibility here!

styles/camunda/docsInstance/hrefDocsToDocs.yml
akeller: image

I ran this through several regex testing tools, and they all barfed on the same forward slash. When I add another backslash, it goes away. Two should be enough to escape the forward slash, and it works... so maybe these tools are using the wrong engine?

akeller:
This super slick and I love the amount of details you added, plus the amount of flexibility it looks like we have moving forward.

I did notice [reviewdog-action-prettier](https://github.com/EPMatt/reviewdog-action-prettier) appears to be a solo maintainer. Not an issue, but it's something to keep in mind if the maintainer walks away.

I am _not good_ at regex, so see my review comment. I was hoping to test it with a browser-based tool similar to redirects. Given that building these regex are really isolated to just our team and not authors, I don't see this as blocking.

akeller:
Oop, can you add a check for fully qualified URLs? - `https://docs.camunda.io/docs/blahblahblah`

Also fine to add an issue for for v2. These are super jarring for the experience in any version or instance.

styles/camunda/docsInstance/hrefDocsToDocs.yml
pepopowitz: It's actually a reverse problem -- vale/go needs the backslashes double-escaped, so the testable regex would have only one backslash everywhere there are 2.

Super annoying, I agree. 

Would a note above each regex that it can be tested in a browser by removing one `\` in every pair of `\` be helpful?

I also considered adding unit tests of the regexes to the codebase -- but thought against it, because the configured rules wouldn't actually use the expression in the tests. (In two ways -- (a) the tested regex would have half the `\` characters as the rule, and (b) the rule wouldn't import the tested regex directly.)

akeller: 🤯 please add a note. I agree with your thoughts on the unit tests. I wouldn't add them. Seems ripe for confusion and chaos.

pepopowitz:
> Oop, can you add a check for fully qualified URLs? - `https://docs.camunda.io/docs/blahblahblah`
> 
> Also fine to add an issue for for v2. These are super jarring for the experience in any version or instance.

I'll add this in my implementation PR! (For what it's worth, it finds 20 violations in the current codebase 😅 )

pepopowitz:
Implemented in #2675 & #2680

PR #2696, opened 2023-10-05
Document multi-tenancy for Zeebe gRPC

### LINKED ISSUE #2686: Zeebe gRPC documentation is expanded with multi-tenancy
```[tasklist]
### Zeebe gRPC API
- [x] DeployResource RPC documentation is expanded with multi-tenancy
- [x] CreateProcessInstance* RPCs documentation is expanded with multi-tenancy
- [x] Job-related RPC documentation is expanded to support multi-tenancy
- [x] Incident-related RPC documentation is expanded to support multi-tenancy
- [x] ThrowError-related RPC documentation is expanded to support multi-tenancy
- [x] PublishMessage RPC documentation is expanded with multi-tenancy
- [x] EvaluateDecision RPC documentation is expanded with multi-tenancy
- [x] (Deprecated) DeployProcess RPC (response-only) documentation is expanded with multi-tenancy
- [x] BroadcastSignal RPC documentation is expanded with multi-tenancy - currently not supported for non-`<default>` tenants
- [x] DeleteResource RPC documentation is expanded with multi-tenancy - currently not supported for non-`<default>` tenants
```

koevskinikola:
## Description

Document multi-tenancy for Zeebe gRPC.

:information_source: This PR is merged into the https://github.com/camunda/camunda-docs/tree/2561-multitenancy-sm branch where all multi-tenancy-related docs should be agregated before they are merged with `main`.

Closes #2686 

## PR Checklist

### DIFF
--- a/docs/apis-tools/grpc.md
+++ b/docs/apis-tools/grpc.md
@@ -95,6 +95,16 @@ Returned if:
 - Worker is blank (empty string, null)
 - Timeout less than 1 (ms)
 - maxJobsToActivate is less than 1
+- If multi-tenancy is enabled, and `tenantIds` is empty (empty list)
+- If multi-tenancy is enabled, and an invalid tenant ID is provided. A tenant ID is considered invalid if:
+  - The tenant ID is blank (empty string, null)
+  - The tenant ID is longer than 31 characters
+  - The tenant ID contains anything other than alphanumeric characters, dot (.), dash (-), or underscore (\_)
+- If multi-tenancy is disabled, and `tenantIds` is not empty (empty list), or has an ID other than `<default>`
+
+##### GRPC_STATUS_PERMISSION_DENIED
+
+- If multi-tenancy is enabled, and an unauthorized tenant ID is provided
 
 ### `BroadcastSignal` RPC
 
@@ -150,6 +160,7 @@ message CancelProcessInstanceResponse {
 Returned if:
 
 - No process instance exists with the given key. Note that since process instances are removed once they are finished, it could mean the instance did exist at some point.
+- No process instance exists with the given key for the tenants the user is authorized to work with.
 
 ### `CompleteJob` RPC
 
@@ -180,6 +191,7 @@ message CompleteJobResponse {
 Returned if:
 
 - No job exists with the given job key. Note that since jobs are removed once completed, it could be that this job did exist at some point.
+- No job exists with the given job key for the tenants the user is authorized to work with.
 
 ##### GRPC_STATUS_FAILED_PRECONDITION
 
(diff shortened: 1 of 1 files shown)


docs/apis-tools/grpc.md
```diff
@@ -150,6 +160,7 @@ message CancelProcessInstanceResponse {
 Returned if:
 
 - No process instance exists with the given key. Note that since process instances are removed once they are finished, it could mean the instance did exist at some point.
+- No process instance exists with the given key for the tenants the user is authorized to work with.
```
remcowesterhoud, line 167: 💭 Just thinking out loud. I wonder if we even need to specify this case. User makes a request, we can't find anything for the given tenant, so we just say it doesn't exist. If it's because the tenant doesn't match or not is up to the user.

Either way, it can't hurt having it here.

docs/apis-tools/grpc.md
```diff
@@ -1055,6 +1059,11 @@ Returned if:
 
 ### `DeleteResource` RPC
 
+:::warning
+This RPC is not tenant-aware. When multi-tenancy is enabled, resources can be
+deleted without considering the tenant ownership of the resource.
+:::
```
remcowesterhoud, line 1065: ❌ This isn't true is it? If multi-tenancy is enabled users can still only delete resources of the default tenant.

The way the processor works is it tries to find any process from de process state with the given key and tenant. If it finds any it deletes it.
If it doesn't find any it searches for a drg by tenant and key. If it finds one it deletes it.
If it doesn't find any we return a NOT_FOUND.

Obviously it'll take the tenant from the command. Since users can't set the tenant in the command I'd expect it to always get the value `<default>`.

remcowesterhoud (changes requested):
Thanks @koevskinikola 🚀 I believe there's 1 blocker that needs to be addressed. You also need to sign the CLA 😄

christinaausley:
@koevskinikola Once you've resolved comments from @remcowesterhoud I'll have a final technical look 👍

docs/apis-tools/grpc.md
koevskinikola: :+1: You have a point @remcowesterhoud. I'll adjust the docs.

docs/apis-tools/grpc.md
koevskinikola: @remcowesterhoud my reasoning for adding this line was to avoid confusion/support cases with users who may have had their authorization to a tenant revoked, or maybe decided to use a different token that is missing the appropriate tenant authorization.

In the scenarios above, this line will (hopefully) trigger a check for the authorizations, instead of users thinking it's a bug because a PI that they know was started can't be found.

WDYT?

remcowesterhoud: Yeah I think it's fair. Like I said it can't hurt having it here and it's already written so no point in taking it out. Worst that can happen is that it helps someone 😄

koevskinikola:
@remcowesterhoud I adjusted the docs according to your feedback. Please have a look when you have the time.

remcowesterhoud (approved):
LGTM!

koevskinikola:
:wave: @christinaausley you can start your review now.

PR #2704, opened 2023-10-05
chore: synchronize sidebars for vNext

pepopowitz:
## Description

Pre-work for #2432.

This PR synchronizes sidebars for the "next" version of the docs. The sidebars tend to drift significantly during new doc development, as it is non-intuitive to update two different sidebars files. This explains the rather large size of this PR.

## Not included

- Synchronization for other versions. That will come in a separate PR.

- [x] This change or page is part of a marketing blog, conference talk, or something else on a schedule. Required prior to version release next week.

## PR Checklist

### DIFF
--- a/hacks/generateOptimizeSidebars.md
+++ b/hacks/generateOptimizeSidebars.md
@@ -21,16 +21,24 @@ The script might require small tweaks depending on which instance you are genera
 
    The sidebar generation script operates on the DOM. Any levels of the nav that aren't in the DOM yet won't be included in the generated items.
 
-   I do this manually, because it doesn't take long, but you could script this step if you wanted to.
-
-   Tip: you can identify if there are any un-expanded menu categories by querying for them in your browser dev tools console:
+   The following script can be run in your browser dev tools, when viewing the documentation, to expand all categories in the left navigation:
 
    ```javascript
-   > document.querySelectorAll(".menu__link--sublist-caret[aria-expanded=false]").length
-   ⋖ 0
+   const count = () =>
+     document.querySelectorAll(".menu__link--sublist[aria-expanded=false]")
+       .length;
+   const expand = () =>
+     document
+       .querySelectorAll(".menu__link--sublist[aria-expanded=false]")
+       .forEach((x) => x.click());
+   async function expandAll() {
+     while (count() > 0) {
+       expand();
+       await new Promise((resolve) => setTimeout(resolve, 500));
+     }
+   }
+   await expandAll();
    ```
-
-   If that statement returns any number other than 0, there are unexpanded categories remaining.
 
 3. Paste the contents of generateOptimizeSidebars.js into your browser dev tools console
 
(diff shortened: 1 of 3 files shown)


hacks/generateOptimizeSidebars.md
```diff
@@ -21,17 +21,25 @@ The script might require small tweaks depending on which instance you are genera
...
+     document
+       .querySelectorAll(".menu__link--sublist[aria-expanded=false]")
+       .forEach((x) => x.click());
+   async function expandAll() {
+     while (count() > 0) {
+       expand();
+       await new Promise((resolve) => setTimeout(resolve, 500));
+     }
+   }
+   await expandAll();
```
pepopowitz, line 40: I'd been doing this by hand in the past. Automating it is much more tolerable.

optimize_sidebars.js
```diff
@@ -133,19 +167,13 @@ module.exports = {
...
-              "components/console/console-troubleshooting/feedback-and-support/"
-            ),
           ],
         },
       ],
     },
 
     {
       Modeler: [
-        docsLink("About Modeler", "components/modeler/about-modeler/"),
```
pepopowitz, line 147: This is an unfortunate change, but not something we can resolve before the next release. 

We have several pages like this, which we converted to "category" pages. They include a bit of content, and also list the pages inside them. Within the left nav, they appear as an expandable category, but when you click on them it takes you to the category page instead of only expanding.

Docusaurus does not support "category" types that don't point at a document within the current instance. In this example, this is the sidebars file for the Optimize instance....and this category page lives within the main docs instance. We can't put them in the sidebars as a simple link, or they'll make the sidebars look out of sync. 

I think the long-term answer is to work with Docusaurus to support category pages that point at URLs instead of document IDs. But that's not happening as part of this release.

christinaausley (approved):
🚀

optimize_sidebars.js
akeller: I'm not understanding how this will function. Should we avoid these category pages?

pepopowitz: I don't think we need to avoid them. Engineers seem to be happy about them. There is a little difference in how each of the instances (main vs optimize) handle them, but that's no different than the large amounts of sidebar drift we get in between releases. Clicking across instances in the left nav is something that I do pretty regularly to test the cross-instance functionality, but I don't think any actual users do it very often.

How this functions right now: 

In the original instance, the category page shows up as a navigable category. It not only expands, but you also go to a page, which usually enumerates the pages within that category. Example: https://docs.camunda.io/docs/components/modeler/about-modeler/ -- if you view another page next to it, then click back on Modeler, you'll see that it takes you to the category page.

In the opposite instance, the category page will show up as a category, but it will **not** be navigable. Meaning you can expand it in the tree, but you can't actually visit that page. Example: https://docs.camunda.io/optimize/components/what-is-optimize/ -- expand the Modeler option in the left nav, and you'll see that it functions differently than it did when you were in the Modeler section.

Long-term, if we work with Docusaurus to see if they're interested in supporting non-instance URLs in category sidebar items, we could make the latter work just like the former.

PR #2764, opened 2023-10-18
chore(connectors): updated references and run instructions

igpetrov:
## Description

chore(connectors): updated references and run instructions

⚠️ we decided to hold it off for now.

- [x] This change is not yet live and should not be merged until `hold` label is removed

## PR Checklist

### DIFF
--- a/docs/components/connectors/out-of-the-box-connectors/slack.md
+++ b/docs/components/connectors/out-of-the-box-connectors/slack.md
@@ -211,7 +211,16 @@ When using the **Slack inbound Connector** with an **Intermediate Catch Event**,
 For example, given that your correlation key is defined with `myCorrelationKey` process variable, and the request body contains `"event": {"text": "12345"}`, your correlation key settings will look like this:
 
 - **Correlation key (process)**: `=myCorrelationKey`
-- **Correlation key (payload)**: `=request.body.event.text`
+- **Correlation key (payload)**: `=if request.body.event = null then "" else request.body.event.text`
+
+:::note
+Please, keep in mind that when you subscribe to the Slack Events API, Slack will send an URL verification message first.
+Since a verification message does not contain `event` object, you have to specify the correlation key as following:
+`=if request.body.event = null then "" else <your payload extraction>` otherwise Zeebe will reject Slack verification message
+with HTTP code 422 `Correlation key not resolved: =request.body.event.text`.
+
+See an example of an URL verification message in [appendix](#slack-url-verification-request-example)
+:::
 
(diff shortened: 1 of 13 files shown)


docs/components/connectors/out-of-the-box-connectors/slack.md
```diff
@@ -211,7 +211,16 @@ When using the **Slack inbound Connector** with an **Intermediate Catch Event**,
 For example, given that your correlation key is defined with `myCorrelationKey` process variable, and the request body contains `"event": {"text": "12345"}`, your correlation key settings will look like this:
 
 - **Correlation key (process)**: `=myCorrelationKey`
-- **Correlation key (payload)**: `=request.body.event.text`
+- **Correlation key (payload)**: `=if request.body.event = null then "" else request.body.event.text`
```
sbuettner, line 214: @igpetrov As FEEL supports nullable properties, cant we just use `=request.body.event.text` here?

sbuettner: Or we make it more obvious like 
```
= if requests.body.type = "url_verification" then "" else request.body.event.text
```

igpetrov: > As FEEL supports nullable properties, cant we just use =request.body.event.text here? We could implemented a fallback value in the runtime.

We cannot, so I even added a not below what happens if we just set `=request.body.event.text`:

```
:::note
Please, keep in mind that when you subscribe to the Slack Events API, Slack will send an URL verification message first.
Since a verification message does not contain `event` object, you have to specify the correlation key as following:
`=if request.body.event = null then "" else <your payload extraction>` otherwise Zeebe will reject Slack verification message
with HTTP code 422 `Correlation key not resolved: =request.body.event.text`.

See an example of an URL verification message in [appendix](#slack-url-verification-request-example)
:::
```

I guess, this is how correlation resolved on `null` values.

The `= if requests.body.type = "url_verification" then "" else request.body.event.text` could be an option but this sounds sort of confusing for me. User might not know what is `url_verification`, what is `type`, etc. So in my example provided it says `if there is no event, return empty string` - simple as that.

@sbuettner WDYT?

igpetrov:
Converting to draft as per discussion with the team.

Reason: we would like to address Slack in more CX friendly way

akeller:
> Converting to draft as per discussion with the team.
> 
> Reason: we would like to address Slack in more CX friendly way

Come visit [ask-developer-experience](https://camunda.slack.com/archives/C0484UFR2VA) if you want input from the broader DevEx team. We do more than just docs 😉 and have a lot of opinions 😄

PR #2749, opened 2023-10-13
docs: changing ES supported versions

marcosgvieira:
## Description

This PR is updating supported versions of ElasticSearch as we found a bug in ElasticSearch that may impact importers on version 8.7.1. For this reason changing from version 8.7 to 8.8. 

**References:**
https://www.elastic.co/guide/en/elasticsearch/reference/current/release-notes-8.7.1.html
https://github.com/elastic/elasticsearch/issues/96078

## When should this change go live?

- [ ] This change is not yet live and should not be merged until {ADD_DATE} (apply `hold` label or convert to draft PR)?
- [ ] There is no urgency with this change.
- [ ] This change or page is part of a marketing blog, conference talk, or something else on a schedule.
- [ ] This functionality is already available but undocumented.
- [ ] This is a bug fix or security concern.

## PR Checklist

- [ ] I have added changes to the relevant `/versioned_docs` directory, or they are not for an **already released version**.
- [ ] I have added changes to the main `/docs` directory (aka `/next/`), or they are not for **future versions**.
- [ ] My changes require an [Engineering review](https://github.com/camunda/camunda-docs/blob/main/howtos/documentation-guidelines.md#review-process), and I've assigned an engineering manager or tech lead as a reviewer, or my changes do not require an Engineering review.
- [ ] My changes require a [technical writer review](https://github.com/camunda/camunda-docs/blob/main/howtos/documentation-guidelines.md#review-process), and I've assigned @christinaausley as a reviewer, or my changes do not require a technical writer review.

PR #2803, opened 2023-10-27
update elasticsearch support

christinaausley:
## Description

See https://camunda.slack.com/archives/C026U8GBNSW/p1698345189792509 and https://camunda.slack.com/archives/C026U8GBNSW/p1698393211411889.

@zszakolci We have the following in the upgrade guide, but is there a way you would like to see this more visible?

Screenshot 2023-10-27 at 8 20 44 AM

## When should this change go live?

- [ ] This change is not yet live and should not be merged until {ADD_DATE} (apply `hold` label or convert to draft PR)?
- [ ] There is no urgency with this change.
- [ ] This change or page is part of a marketing blog, conference talk, or something else on a schedule.
- [x] This functionality is already available but undocumented.
- [x] This is a bug fix or security concern.

## PR Checklist

- [x] I have added changes to the relevant `/versioned_docs` directory, or they are not for an **already released version**.
- [x] I have added changes to the main `/docs` directory (aka `/next/`), or they are not for **future versions**.
- [x] My changes require an [Engineering review](https://github.com/camunda/camunda-docs/blob/main/howtos/documentation-guidelines.md#review-process), and I've assigned an engineering manager or tech lead as a reviewer, or my changes do not require an Engineering review.
- [x] My changes require a [technical writer review](https://github.com/camunda/camunda-docs/blob/main/howtos/documentation-guidelines.md#review-process), and I've assigned @christinaausley as a reviewer, or my changes do not require a technical writer review.

PR #2670, opened 2023-09-29
Update supported-environments.md

toco-cam:
This table seems redundant to https://docs.camunda.org/enterprise/download/#production-distribution-without-elasticsearch

## When should this change go live?

- [ ] This change is not yet live and should not be merged until {ADD_DATE} (apply `hold` label or convert to draft PR)?
- [ ] There is no urgency with this change.
- [ ] This change or page is part of a marketing blog, conference talk, or something else on a schedule.
- [ ] This functionality is already available but undocumented.
- [ ] This is a bug fix or security concern.

## PR Checklist

- [ ] I have added changes to the relevant `/versioned_docs` directory, or they are not for an **already released version**.
- [ ] I have added changes to the main `/docs` directory (aka `/next/`), or they are not for **future versions**.
- [ ] My changes require an [Engineering review](https://github.com/camunda/camunda-platform-docs/blob/main/howtos/documentation-guidelines.md#review-process), and I've assigned an engineering manager or tech lead as a reviewer, or my changes do not require an Engineering review.
- [ ] My changes require a [technical writer review](https://github.com/camunda/camunda-platform-docs/blob/main/howtos/documentation-guidelines.md#review-process), and I've assigned @christinaausley as a reviewer, or my changes do not require a technical writer review.

PR #2841, opened 2023-11-01
fix(supported_envs): support a given minor and beyond

MaxTru:
* In accordance with our support strategy https://confluence.camunda.com/display/HAN/Camunda+8+Supported+Environments
  * Reason: OS and ES are full semver compliant, so this means they will not introduce breaking changes as part of minors. So customers can safely patch forward.
* Brought up in this conversation: https://camunda.slack.com/archives/C04QG866E3Z/p1698846971127109
* Initial PR: #2749 

## PR Checklist


MaxTru:
FYI for @megglos @abbasadel @johanwelgemoed @jwulf @marcosgvieira 

According to what we discussed and decided via https://confluence.camunda.com/display/HAN/Camunda+8+Supported+Environments:
> We communicate to customers that we support a target minor version of OS or ES and following minors
>   * Example: for Camunda 8.3 we support AWS OS 2.5+ and ES 8.7+.
>   * Reason: OS and ES are full semver compliant, so this means they will not introduce breaking changes as part of minors. So customers can safely patch forward.

**Please raise concerns if any :-) **

akeller:
☝️ edited the above comment to include @RomanJRW instead of jwulf

akeller (approved):
This change is straightforward to me, but prior to merging, I'd like to see responses from at least one of the component EMs.

akeller:
I'm still working through a backlog of notifications, so I apologize for the noise here.

We've made a few ES support changes over the past week, which expands what we need to change within the scope of this PR. 

- https://github.com/camunda/camunda-docs/pull/2830
- https://github.com/camunda/camunda-docs/pull/2803

Running this locally, I see a few other areas we need to change. Rather than create a new PR, I will add a commit shortly. 

I know we want to make it easier for Support to help customers, but I'd also like to acknowledge that changes like this explode the scope of otherwise "quick" PRs. I will open an issue in the backlog to see if we can leverage variables to only update this in a single place in the future.

akeller (dismissed):

versioned_docs/version-8.3/reference/supported-environments.md
```diff
@@ -30,11 +30,11 @@ Requirements for the components can be seen below:
...
+| Zeebe       | OpenJDK 17+  | Elasticsearch 8.8+                                                                                                                                  |
+| Operate     | OpenJDK 17+  | Elasticsearch 8.8+                                                                                                                                  |
+| Tasklist    | OpenJDK 17+  | Elasticsearch 8.8+                                                                                                                                  |
 | Identity    | OpenJDK 17+  | Keycloak 21.x, 22.x<br/>PostgreSQL 14.x, 15.x                                                                                                       |
-| Optimize    | OpenJDK 17+  | Elasticsearch 8.8.x+                                                                                                                                |
+| Optimize    | OpenJDK 17+  | Elasticsearch 8.8+                                                                                                                                  |
```
RomanJRW, line 37: @akeller - there is a separate matrix below for Optimize, do you think this should also be made consistent? The difference there is that for _some_ older versions, we didn't support all patches (see `7.16.2+` for example)

RomanJRW (commented):
In the past, we have had specific patch versions which we can't support and we have had to be specific at patch level. For example, with the `7.16` release of ES, there were security issues (from memory, I haven't gone back to check the specific reasons) that only got patched with `7.16.3`. Specifying the patch version in the docs allowed us to be explicit in saying that that 7.16.1-2 weren't supported but that `7.16.3+` were, because the older versions contained a vulnerability that could potentially affect our applications.

By dropping the patch version from this matrix, we lose this flexibility. What about including the patch but the specific version, like `8.8.0+`?

versioned_docs/version-8.3/reference/supported-environments.md
akeller: @toco-cam removed this table from future docs - https://github.com/camunda/camunda-docs/pull/2670.

akeller: This is complicated. Let's frame it like this - We are in the 8.3 docs, what version of Elasticsearch works will 8.3 + Optimize? Or more specifically, since Optimize can't operate in a vacuum, what ES version is required for 8.3? 

I'd love to get our supported environments to a point where it looks more like a unified platform experience, and the best way to get there is to be specific about what ES version works with the components together as a platform.

Can we follow up separately on the older Optimize versions? I think moving this Optimize table may help us here.

RomanJRW: Sure, makes sense. I would also argue that the old Optimize table is partly redundant. With versioned docs, we don't really need to list compatibility of older versions for a given docs page. You could also argue that we don't need to list compatibility for versions of Optimize no longer supported

akeller:
Between feedback on this PR and Slack: 

* https://camunda.slack.com/archives/C04QG866E3Z/p1698846971127109
* https://camunda.slack.com/archives/CHAC0L80M/p1698942501840449

We have a few things we need to clarify in the docs based on internal and customer feedback:

* What is the minimum supported version?
* Is it supported if we don't list it, but it's a higher number?
* What does it mean to write `x.y` and not `x.y.z`?

I want to add the following to the top of the page, regardless of where we land on `x.x+` or `x.x.x+`: 

> Unless otherwise specified, the versions listed on this page are the minimum version required. 
> 
> Most of our dependencies use [semantic versioning](https://semver.org/). If a `+` is added to a version number, you can read this as the minimum version _and_ anything higher or more recent will be compatible with Camunda. This includes more recent majors, minors, or patches unless otherwise specified.

MaxTru:
> I want to add the following to the top of the page, regardless of where we land on `x.x+` or `x.x.x+`:
> 
> > Unless otherwise specified, the versions listed on this page are the minimum version required.
> > Most of our dependencies use [semantic versioning](https://semver.org/). If a `+` is added to a version number, you can read this as the minimum version _and_ anything higher or more recent will be compatible with Camunda. This includes more recent majors, minors, or patches unless otherwise specified.


Thanks Amara. 

I struggle with:
> Unless otherwise specified, the versions listed on this page are the minimum version required.

=> because only if the `+` is there, this is true. 

Also this is not correct:
> This includes more recent majors, minors, or patches unless otherwise specified.

* 8.8+ would mean we support more recent minors (and patches)
* 8.8.0+ would mean we support more recent patches
* 8+ would mean we support more recent majors (and minors and patches)


What about something like this:

> A `+` added to a version number means that this is the minimum version. Any higher / more recent version will be compatible with Camunda. For example, 8.7+ means we support the minor version 7, and any higher minors and patches (but not majors).

akeller:
@MaxTru ok, I will make another attempt at clarifying this.

We use `x.x.0+` and `x.x.x+` on this page as well. In your opinion, do we need to standardize this or offer guidance on why one over the other?

MaxTru:
> @MaxTru ok, I will make another attempt at clarifying this.
> 
> We use `x.x.0+` and `x.x.x+` on this page as well. In your opinion, do we need to standardize this or offer guidance on why one over the other?

Let's standardize please, potentially linking to this one footnote explanation, which you are working on.

christinaausley:
@akeller Among all of the info in this issue, is there anything I can assist with? I'll note the ongoing work in my WCoE release today, but let me know if I can open any additional issues with this for my `Ready` column.

akeller:
Removed the Optimize table from 8.3 & 8.2, essentially a backport of https://github.com/camunda/camunda-docs/pull/2670

akeller:
@christinaausley please get some internal guidance together on when to use `x.x` and `x.0`, basically when to use `x` and when to use `0`, particularly when appending a `+`. IMO, `x` here is clearer and implies more recent versions are supported, but this might be something to verify with other technical writing teams and docs. 

I also see Helm charts used `>=`, which is totally different from the rest of the info on the page and potentially confusing.

It should be clear to teams and users that if we specify a version number, does that mean exactly that version or something else.

Create a new issue for this, as it's out of the scope of this particular issue. This PR offers improvements, but you will provide standards 🏆

akeller:
@MaxTru & reviewers have a look at the admonition I added here:

> The versions listed on this page are the minimum version required if appended with a `+`.
>
> Pay attention to where the `+` falls, as most of our dependencies follow [semantic versioning](https://semver.org/) (semver), where `x.y.z` correspond to MAJOR.MINOR.PATCH. Higher or more recent versions will be compatible with Camunda, with respect to semver.
>
> For example, 1.2+ means support for the minor version 2, and any higher minors (1.3, 1.4, etc.) and patches (1.2.1, 1.2.2, etc.), but not majors, like 2.x.

It's right on the line of over-documenting, but it clears up current confusion while tying this back to an industry standard. 

@christinaausley if reviewers give this an ok, can you do any final wordsmithing?

docs/reference/supported-environments.md
```diff
@@ -4,6 +4,15 @@ title: "Supported environments"
 description: "Find out where to run Camunda 8 components for SaaS and Self-Managed, including Optimize for both Camunda 8 and Camunda 7."
 ---
 
+:::note
+The versions listed on this page are the minimum version required if appended with a `+`.
+
+Pay attention to where the `+` falls, as most of our dependencies follow [semantic versioning](https://semver.org/) (semver), where `x.y.z` correspond to MAJOR.MINOR.PATCH. Higher or more recent versions will be compatible with Camunda, with respect to semver.
+
+For example, 1.2+ means support for the minor version 2, and any higher minors (1.3, 1.4, etc.) and patches (1.2.1, 1.2.2, etc.), but not majors, like 2.x.
```
RomanJRW, line 12:
```suggestion
For example, 1.2+ means support for the minor version 1.2, and any higher minors (1.3, 1.4, etc.) and patches (1.2.1, 1.2.2, etc.), but not majors, like 2.x.
```
Shouldn't it be this?

docs/reference/supported-environments.md
```diff
@@ -4,6 +4,15 @@ title: "Supported environments"
 description: "Find out where to run Camunda 8 components for SaaS and Self-Managed, including Optimize for both Camunda 8 and Camunda 7."
 ---
 
+:::note
```
RomanJRW, line 7: https://github.com/camunda/camunda-docs/pull/2841#pullrequestreview-1711162341

Still not against these changes, but I feel like it's quite possible that it doesn't give us enough flexibility as per my above comment. Will this note also need extended if we have to drop support for _some_ of the patch releases of a supported minor version?

docs/reference/supported-environments.md
akeller: 1 is the major version, 2 is the minor, but maybe this sentence could use some work.

docs/reference/supported-environments.md
akeller: @RomanJRW does this occur outside of Optimize? Regardless, instead of extending the note, we would explicitly mention what is not supported, patch or otherwise, in the table. 

Would you be more comfortable if it said the following instead (changing "will be compatible" to "should be compatible"):

> Higher or more recent versions should be compatible with Camunda, with respect to semver, unless otherwise noted.

RomanJRW: This issue can occur globally I would say. If there is a security issue that impacts one component, it will typically impact all of them. It happened with ES versions 7.10.x and also 7.17.x, so is not super common but also something we should consider. Happy to proceed as is and we can consider this scenario when it next happens

I think "will be" is better than "should be". We should document things like this with confidence imo 👍

akeller:
Does anyone have further feedback here? I want to make a small edit and then get this merged. Based on @RomanJRW's feedback, we need to continue the discussion, but I'd like to bring it to Slack rather than keep it in a PR.

docs/reference/supported-environments.md
toco-cam: > there were security issues (from memory, I haven't gone back to check the specific reasons) that only got patched with 7.16.3. Specifying the patch version in the docs allowed us to be explicit in saying that that 7.16.1-2 weren't supported but that 7.16.3+ were, because the older versions contained a vulnerability that could potentially affect our applications.

@RomanJRW How have we been impacted by this? Was it a support case? Shouldn't we ask, the customer to run the latest patch before we start supporting them? As there any reason for a customer not to install the latest patch?

RomanJRW: https://jira.camunda.com/browse/OPT-5824
We handled it for 7.16.2, and it related to the two support cases linked there

It's not about the latest patch, it's about the minimum patch. The scenario I am trying to highlight is that it would be wrong for us to say we support 8.9+, for example, when not _all_ patches of that version are actually compatible

christinaausley (approved):
I am happy with this. Can continue to make adjustments as @akeller noted in Slack.

christinaausley (approved):

christinaausley:
@akeller I'll let you make your edit and then can approve for merge 👍

akeller:
I've prompted on the private ELT channel to get a bit more feedback (or at least acknowledgment) from other components.

abbasadel (approved):

akeller:
Merging as-is, with the idea that we will likely need to iterate or modify this in the future.

PR #2875, opened 2023-11-07
docs(general): add configuring Entra ID page

Ben-Sheppard:
Closes https://github.com/camunda-cloud/identity/issues/2181
## Description

This PR introduces the early documentation relating to the Microsoft Entra ID (used to be Azure AD) authentication support arriving for Operate and Identity in 8.4.0-alpha1. 

Still to do:
- [x] Add in configuration options for Operate docs that use the new Identity Spring boot starter
- [ ] Alter docs to reflect a Helm approach vs env variables

- [x] This change is not yet live and should not be merged until {ADD_DATE} (apply `hold` label or convert to draft PR)?

## PR Checklist


docs/self-managed/platform-deployment/helm-kubernetes/guides/connecting-to-entra-id.md
```diff
@@ -0,0 +1,45 @@
+---
+id: connecting-to-entra-id
+title: "Connecting to Entra ID (Azure AD)"
+description: "Learn how to connect your Camunda Self-Managed deployment to Entra ID (Azure AD)."
+---
+
+:::note
+The following components support Entra ID authentication: Operate
```
christinaausley, line 8: After reading through this, I get the assumption that additional components will eventually support this? If not, may need to reword a few notes.

docs/self-managed/platform-deployment/helm-kubernetes/guides/connecting-to-entra-id.md
```diff
@@ -0,0 +1,45 @@
...
+1. Access the Entra ID admin area
+   and [register an application](https://learn.microsoft.com/en-us/entra/identity-platform/quickstart-register-app).
+   After registering the app, the **Overview** page will contain a Client ID; make a note of this value as it will be
+   required later on.
+
+2. Within the app registered in step
+   1, [configure a platform](https://learn.microsoft.com/en-gb/entra/identity-platform/quickstart-register-app#configure-platform-settings)
+   of type `web`. The expected redirect URI of the component you are configuring an app for can be found in [component-specific configuration](#component-specific-configuration).
+
+3. Once you have registered a platform for your app a [client secret should be added](https://learn.microsoft.com/en-us/entra/identity-platform/quickstart-register-app#add-a-client-secret). Make a note of the value of the client secret as it will be required later on.
```
christinaausley, line 33: client secret **should** be added or **will** be added?

christinaausley (commented):
Just a few comments for clarity 👍

docs/self-managed/platform-deployment/helm-kubernetes/guides/connecting-to-entra-id.md
Ben-Sheppard: Ah I've confused the language here I think! What needs to happen is the customer needs to go and create a secret for the app, so maybe this should be reworded to something like:

```
Once you have registered a platform for your app a client secret needs to be created, to do this see [adding a client secret](https://learn.microsoft.com/en-us/entra/identity-platform/quickstart-register-app#add-a-client-secret)
```

wdyt?

christinaausley: Just needs a period, then good to go!

docs/self-managed/platform-deployment/helm-kubernetes/guides/connecting-to-entra-id.md
Ben-Sheppard: Thats correct, in the next iterations more components will be added :)

Ben-Sheppard:
@christinaausley I made a slight tweak to where the azure doc is, at the moment in the helm/kubernetes directory seemed awkward

christinaausley (approved):

christinaausley:
Nice 👍 Feel free to merge when you are ready.

sdorokhova (approved):
LGTM

docs/self-managed/identity/user-guide/configuration/connecting-to-entra-id.md
```diff
@@ -0,0 +1,45 @@
+---
+id: connecting-to-entra-id
+title: "Connecting to Entra ID (Azure AD)"
+description: "Learn how to connect your Camunda Self-Managed deployment to Entra ID (Azure AD)."
```
dlavrenuek, line 4: I would expect to find this documentation in the "Installing on Microsoft AKS platform" section: https://docs.camunda.io/docs/next/self-managed/platform-deployment/helm-kubernetes/platforms/microsoft-aks/

Identity as a component is currently not involved with Azure at all, so we should also not hide this information there

dlavrenuek (changes requested):

docs/self-managed/operate-deployment/operate-authentication.md
```diff
@@ -137,29 +142,43 @@ export SPRING_PROFILES_ACTIVE=identity-auth
...
+| ----------------------------------- | ----------------------------------------- | --------------------------------------------------- |
+| CAMUNDA_IDENTITY_TYPE               | Type of Identity provider.                | AZUREAD (defaults to KEYCLOAK)                      |
+| CAMUNDA_IDENTITY_BASE_URL           | URL of the Identity component             | http://localhost:8080                               |
+| CAMUNDA_IDENTITY_ISSUER             | URL of issuer (Identity)                  | http://localhost:18080/auth/realms/camunda-platform |
+| CAMUNDA_IDENTITY_ISSUER_BACKEND_URL | Backend URL of issuer (Identity)          | http://localhost:18080/auth/realms/camunda-platform |
+| CAMUNDA_IDENTITY_CLIENT_ID          | Similar to a username for the application | operate                                             |
+| CAMUNDA_IDENTITY_CLIENT_SECRET      | Similar to a password for the application | XALaRPl...s7dL7                                     |
+| CAMUNDA_IDENTITY_AUDIENCE           | Audience for Operate                      | operate-api                                         |
+
```
dlavrenuek, line 159: I would actually also include this part in the other document. In the end all of the configuration variables will be the same for most components (probably except Web-Modeler and Connectors?), so Having it in a central place would reduce duplication. Here we could link to the `connecting-to-entra-id.md` page

dlavrenuek (changes requested):

docs/self-managed/identity/user-guide/configuration/connecting-to-entra-id.md
Ben-Sheppard: I have concerns about including this documentation in the helm section of our docs right now because these configuration variables are not natively supported in our helm charts currently. 

Although I agree regarding Identities hidden role here, if we look at the variable names "CAMUNDA_IDENTITY_ABC"  I think the Identity docs will be a natural place for readers to look. 

If we feel strongly about this though we could move it to the helm section but make a note that setting these variables requires the use of the direct env variables approach and not a specific param.

docs/self-managed/operate-deployment/operate-authentication.md
Ben-Sheppard: I like this suggestion, I'll refactor

docs/self-managed/identity/user-guide/configuration/connecting-to-entra-id.md
dlavrenuek: The section is actually `Helm/Kubernetes`, so it is also about k8s installation and not just helm. I would not directly map `_IDENTITY_` to the C8 Identity as a product, it can also be part of the `Identity and Access Management` configuration on component level. I am open for other suggestions where this page could be located but I have a strong opinion that Identity user docs is definitely the wrong place, since it is not related to Identity as a product and it's users at all.

If I think about experience for anyone installing Camunda on an Azure cluster, then that person would most likely follow the installation described on the page I've linked.

Ben-Sheppard: Ok, I've moved the page to this section now and also moved configuration to be more generalised here

dlavrenuek (approved):
Thank you for the changes, looks good from my perspective 👍

christinaausley (approved):

PR #2914, opened 2023-11-14
chore(migration): add upgrade notice for issue OPT-7419

PHWaechtler:
related to OPT-7421

## Description

We encountered a potential migration issue for 3.11 and recommend C7 customers to skip straight to 3.11.3 for a safer migration path.

These docs refer to the upcoming 3.11.3 patch which we are releasing asap, imo we can merge the docs even beforehand.

## PR Checklist

### DIFF
--- a/optimize/self-managed/optimize-deployment/migration-update/3.10-to-3.11_8.3.md
+++ b/optimize/self-managed/optimize-deployment/migration-update/3.10-to-3.11_8.3.md
@@ -17,6 +17,12 @@ Here you will find information about:
 - Changes in behavior (for example, due to a new feature)
 - Changes in translation resources
 
+## Known Issues
+
+## Migration of Camunda Platform 7 data for 3.11.0, 3.11.1 and 3.11.2
+
+Under some circumstances, the migration of Camunda 7 data for versions 3.11.0, 3.11.1 and 3.11.2 can cause issues with the tenant selection of report definitions and collection scopes. This only occurs if data is present in Elasticsearch with the value "zeebe" in the datasource field, which can happen for example if the "onboarding" dataset set is used in Optimize. To avoid this issue we recommend that Camunda 7 users skip 3.11.0, 3.11.1 and 3.11.2 and instead migrate straight to Optimize version 3.11.3.
+
 ## Changes in supported environments
 
 ### Elasticsearch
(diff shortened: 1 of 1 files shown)


christinaausley (approved):
Technical review in latest commit, but do you feel this should be noted anywhere outside of the update guide? If it could cause breaking changes in specific docs, we can add an admonition.

optimize/self-managed/optimize-deployment/migration-update/3.10-to-3.11_8.3.md
```diff
@@ -17,6 +17,12 @@ Here you will find information about:
 - Changes in behavior (for example, due to a new feature)
 - Changes in translation resources
 
+## Known issues
+
+## Migration of Camunda 7 data for 3.11.0, 3.11.1, and 3.11.2
+
+Under some circumstances, the migration of Camunda 7 data for versions 3.11.0, 3.11.1, and 3.11.2 can cause issues with the tenant selection of report definitions and collection scopes. This only occurs if data is present in Elasticsearch with the value `zeebe` in the datasource field, which can happen if the `onboarding` dataset set is used in Optimize, for example. To avoid this issue, we recommend Camunda 7 users skip 3.11.0, 3.11.1, and 3.11.2, and instead migrate straight to Optimize version 3.11.3.
```
RomanJRW, line 24: Do you think we should also include here that the customer onboarding dataset should not be used in C7 environments, and maybe a link to some config for how to disable it?

RomanJRW (approved):
No issues with the wording, but I would also like to see this in the docs.org announcements I think 🤔 

The fact that we only hear about this now makes me think that the impact is possibly smaller than we first thought and that most C7 users never had the problematic dataset. Or equally it could just be that adoption is slow 👀

optimize/self-managed/optimize-deployment/migration-update/3.10-to-3.11_8.3.md
PHWaechtler: Hm not sure, if they already have it then disabling the config isn't gonna remove it right? I would tend towards not complicating things, though I would like to follow up with checking if this was a normal use case to have the onboarding data in C7 and if so, whether we can adjust the set to not include the "zeebe" datasource so we dont have confusing data in the future.

RomanJRW: sure 👍 
I think this would be a good post-mortem topic. Re the zeebe datasource, this is needed for the dataset to work in SaaS, so I am not sure how changeable this would be for C7 customers

RomanJRW:
@christinaausley - Given the impact assessment in https://camunda.slack.com/archives/C065B30V03Y, I think it is ok to merge as is and not list this anywhere else. On the downloads page, we are only keeping the latest patch available for download, so this is probably enough. We can revisit this decision if any other customers run into this issue. With this in mind, I will go ahead and merge it 🙂

Issue #1778, opened 2023-03-06
Detect missing versions in PR changes


pepopowitz:
Part of #1742 

Currently it is left up to humans to notice when a PR's changes affect one version of the docs but should likely be synchronized to another version. This results in PRs merging prematurely, and requiring follow-up PRs when someone notices that it should be synchronized to another version.

We should write a GitHub action that looks at the files changed, and based on the rules below, comments on the PR if changes are likely to need synchronization.

Note that the comment should be worded as a *suggestion* -- there will always be cases where a change is intended in only one version, and we should still allow for those changes to happen. 

## Version synchronization suggestion rules

Iteration 1: 

- if vCurrent is affected, but vNext is not, suggest that the changes should also be applied to vNext
- if vNext is affected, but vCurrent is not, suggest that the changes should also be applied to vCurrent
- these rules apply to both the "docs" instance and the "optimize" instance

Iteration 2:

Per https://github.com/camunda/camunda-docs/pull/2922#issuecomment-1820839017, the following rules would also be meaningful:

- if vPrevious (or older) is affected, but vPrevious+1 through vNext are not, suggest that the changes should also be applied to vPrevious+1 though vNext.

## Implementation

```[tasklist]
### Implementation
- [ ] #2922
- [ ] #3091
- [ ] https://github.com/camunda/camunda-docs/pull/3116
- [ ] https://github.com/camunda/camunda-docs/pull/3402
```

## Tasks extracted from this one

- [ ] https://github.com/camunda/camunda-docs/issues/4765

pepopowitz:
Status: Currently collecting feedback of a prototype at https://github.com/camunda/camunda-docs/pull/2922.

pepopowitz:
Status: First implementation is up at #3091. I'll let it sit over the holidays, and hope to merge it prior to our release in January.

pepopowitz:
I've extracted the remaining work on this one, so that I can mark this done.

PR #2922, opened 2023-11-15
[DO NOT MERGE] prototype: Warn about changes in `current` but not `next`, and vice-versa.

pepopowitz:
This PR should probably not be merged in its current state. There is a bit of cleanup to do, and I'm not 100% sure we want it implemented.

It is a prototype of #1778. 

## Description

This prototype is a job thief! It seeks to steal a non-trivial portion of Amara's PR review responsibilities. It adds a workflow that is triggered on `pull_request`, which automatically comments on a PR when there are changes in the `next` version but not `current`, or the `current` version but not `next`.

The comment looks like this (screenshotted from this PR itself, see below):

image

## Summary of the logic to trigger a message

* If all the changed files in `next` are also changed in `current`, and vice-versa, no comment is added to the PR.
* If changed files exist in `next` that aren't also changed in `current`, a comment is added, and the message will list the files.
* If changed files exist in `current` that aren't also changed in `next`, a comment is added, and the message will list the files.
* The message, if it's presented, also includes some helpful context about why this message is appearing, and how to learn more about versioning in our docs.

## Some technical choices I made 

* It does **not** spam the PR with comments as commits are added. 
   If the missed version edits change throughout the life of the PR, the same comment will be updated with new information. We can change this behavior if we'd like, in two ways: (1) spam the PR every time a commit is pushed (🙅), or (2) delete & recreate the message each time a commit is pushed (🤷).
* The message is not "dismissable". It appears as an unresolvable comment on the PR.
   This is less about me "choosing" to do it this way, and more about me not knowing how to do it any other way. But we probably _could_....I'd just have to dig a lot more.
* The only version scenarios I implemented are: a file is edited in `next` but not `current`, or vice versa. 
   I don't think we should warn about older version changes, in either direction, because those are much more infrequent. I think the majority of times we make a change in `next` but not `current`, or `current` but not `next`, it is an accident, and we end up duplicating them. I might be wrong about this.

## Yet to do, but I will wait until I'm confident we want this

- [ ] Extend the check to include Optimize docs
- [ ] Finalize the messaging in the comments
- [ ] Probably trigger on `pull_request_target` instead of `pull_request`, to accommodate forked PRs. Regardless of whether it requires a change, that scenario should be tested.
- [ ] Update the message after all issues have been cleared up

## How this prototype moves forward

- If people think this is a good idea, I will re-implement it, including the "yet to do" list above. I will also add multiple "test" PRs _into_ the new PR, to confirm a variety of scenarios are covered as planned.
- If people think this is not a good idea, I will close this prototype and go back to the drawing board on #1778.

### DIFF
--- a/.github/workflows/check-versions.yaml
+++ b/.github/workflows/check-versions.yaml
@@ -7,28 +7,34 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout last two commits of the PR
+        # The last *two* commits are (1) everything in this branch, and (2) the parent of this branch.
         uses: actions/checkout@v3
         with:
           fetch-depth: 2
-      - name: Identify files changed in next but not current, or vice versa
+
+      - name: Identify files changed in next but not current, or vice versa. Build a message for them.
         run: |
+          # find all files changed in this PR, search for only within a folder, and remove that folder from the matches.
           next_files=($( git diff --name-only HEAD^ | grep ^docs/ | sed 's/^docs\///')) 
           current_files=($( git diff --name-only HEAD^ | grep ^versioned_docs/version-8.3/ | sed 's/^versioned_docs\/version-8.3\///' )) 
 
+          # remove file names duplicated in the two sets, leaving only files that are unique to one set.
           missing_files=(`echo ${next_files[@]} ${current_files[@]} | tr ' ' '\n' | sort | uniq -u`)
 
-          echo $missing_files
-
+          # exit if there are no offenders.
           if [ ${#missing_files} -eq 0 ]; then
             echo "No obvious missing version updates." 
             exit 0
           fi
 
+          # identify which set each of the unique files came from, by finding the intersection of the offenders and each set.
           only_in_next=(`echo ${next_files[@]} ${missing_files[@]} | tr ' ' '\n' | sort | uniq -D | uniq`)
           only_in_current=(`echo ${current_files[@]} ${missing_files[@]} | tr ' ' '\n' | sort | uniq -D | uniq`)
 
           message=""
 
+          # Append a message for the files only changed in next.
+          #  I'm unsure why some of the markdown here (the emphasis) gets formatted by GitHub, but some doesn't (the unordered list).
           if [ ${#only_in_next[@]} -gt 0 ]; then
             message+='These files were changed only in the *next* version. You might want to duplicate these changes in the *current* version.<br/><ul>'
             for file in "${only_in_next[@]}"; do
@@ -37,6 +43,8 @@ jobs:
             message+="</ul>"
           fi
 
+          # Append a message for the files only changed in current.
+          #  I'm unsure why some of the markdown here (the emphasis) gets formatted by GitHub, but some doesn't (the unordered list).
           if [ ${#only_in_current[@]} -gt 0 ]; then
             message+='These files were changed only in the *current* version. You might want to duplicate these changes in the *next* version.<br/><ul>'
             for file in "${only_in_current[@]}"; do
@@ -45,8 +53,10 @@ jobs:
             message+="</ul>"
           fi
 
+          # For debug purposes.
           echo $message
 
+          # Pass the message to the next step.
           echo POTENTIAL_MISSING_VERSIONS_MESSAGE="$message" >> $GITHUB_ENV
       - name: Comment on PR if version updates are potentially missing
         uses: thollander/actions-comment-pull-request@v2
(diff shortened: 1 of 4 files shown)


.github/workflows/check-versions.yaml
```diff
@@ -0,0 +1,61 @@
+name: Check versions
+
+on: pull_request
```
pepopowitz, line 3: Per [this action's docs](https://github.com/marketplace/actions/comment-pull-request#permissions), I anticipate that this will change to `pull_request_target` so that forked PRs have the necessary permissions for GitHub to add a comment.

.github/workflows/check-versions.yaml
```diff
@@ -0,0 +1,71 @@
...
+      - name: Comment on PR if version updates are potentially missing
+        uses: thollander/actions-comment-pull-request@v2
+        if: ${{ env.POTENTIAL_MISSING_VERSIONS_MESSAGE != '' }}
+        with:
+          message: |
+            :wave: :robot: Hello! Did you make your changes in all the right places? 
+
+            ${{ env.POTENTIAL_MISSING_VERSIONS_MESSAGE }}
+
+            You may have done this intentionally, but we wanted to point it out in case you didn't. You can read more about the versioning within our docs in our [documentation guidelines](https://github.com/camunda/camunda-docs/blob/c666b7e5e35bfe31468ce4ac679ea7b62ab86d73/howtos/documentation-guidelines.md#versions).
```
pepopowitz, line 70: Irritating that this is all on one line, but any line-breaks will appear in the GitHub comment.

.github/workflows/check-versions.yaml
```diff
@@ -0,0 +1,71 @@
...
+          comment_tag: execution
```
pepopowitz, line 71: I think we can call [this `comment_tag`](https://github.com/marketplace/actions/comment-pull-request#update-a-comment) anything, it's just a way for it to identify it so that it doesn't spam the PR. I started with it as `execution` because I was following an example, and didn't want to change it along the way to end up with a second comment. I'd change this in the final implementation.

christinaausley:
@pepopowitz This is epic.

So, this applies only to changes in `next` and `current`, yes? What if someone makes changes in `8.2` or `8.1` but forgets to make changes in the current version?

The argument here could be that we rarely see a change in `8.2` that isn't also in `8.3` and `next`, but I was just curious 👍

christinaausley:
> Is there a line in the sand where older version changes rarely make it to the newer versions? Are 8.1 changes also usually applied to everything newer?

Regardless, I think I might want to handle this iteratively -- enable this functionality for next & current versions, and introduce it for 8.2 and 8.1 after we're sure we feel good about it.

@pepopowitz Agreed here. Regardless, I really like it. Could always incorporate the following in future iterations:

- "These files were changed only in an _older version_ of docs. You might want to duplicate these in the `current` and `next` versions of docs." or something of the like

akeller (commented):
I really like this, and not just because it takes work off of my plate.

But I'm not the target user! @jonathanlukas would be a good user to ask, since he's often opening PRs ad-hoc and not in sync with the release cadence.

jonathanlukas:
@akeller thanks for pulling me in.

I LOVE the bot. One idea about the persistence of the message:

I use a tool to [display test results on PRs](https://github.com/camunda-community-hub/camunda-7-to-8-migration/pull/562#issuecomment-1827164787) and they somehow manage to update it.

I would love to have something like a dynamic status message that is updated (or is this the re-created thing mentioned?).

pepopowitz:
> I LOVE the bot. One idea about the persistence of the message:
> 
> I use a tool to [display test results on PRs](https://github.com/camunda-community-hub/camunda-7-to-8-migration/pull/562#issuecomment-1827164787) and they somehow manage to update it.
> 
> I would love to have something like a dynamic status message that is updated (or is this the re-created thing mentioned?).

@jonathanlukas from the body of this PR: 

> * It does not spam the PR with comments as commits are added.
>    If the missed version edits change throughout the life of the PR, the same comment will be updated with new information. We can change this behavior if we'd like, in two ways: (1) spam the PR every time a commit is pushed (🙅), or (2) delete & recreate the message each time a commit is pushed (🤷).

I _think_ the implemented behavior is what you're requesting. The initial message will be updated as the PR is changed. Let me know if I misunderstand!

One thing that I don't think I accounted for is that once all the issues are cleared up, no workflow will clear the last message. (Because the workflow that is updating the message is skipped if there are no missed versions.) I could see handling this iteratively in a follow-up PR, but it also might cause so much confusion that it needs to be included in this PR before merging. I welcome opinions 😅

jonathanlukas:
I think we are aligned on the message update! Thank you for double checking @pepopowitz 

Hm, the last message should be something like "You are good to go"? Because a closed PR could potentially be inspected.

pepopowitz:
Feedback on this prototype is positive, so I'm going to pursue a cleaner version in a new PR!

PR #2998, opened 2023-12-06
Clarify header values are config names in Zeebe docs

### LINKED ISSUE #2991: Clarify "field" in Zeebe config docs
It's not clear the "field" in the table isn't the name of the config option, but rather a field under the config option. The config option name is the section title or header.

At the [top of the page](https://docs.camunda.io/docs/self-managed/zeebe-deployment/configuration/gateway-config/#configuration), we mention the expected or anticipated format of the page, but we don't define "field." 

_Feedback via DM from @pierre-yves-monnet. Thanks!_

akeller:
## Description

Closes #2991.

I also added some info so the pages "matched" in how the configuration is described. 

NOTE - 8.1 is out of scope for backporting because the files don't exist in this version.

## PR Checklist


akeller:
@christinaausley I was having trouble coming up with a nice crisp sentence for this change. Can you have a look?

christinaausley:
> Field values represent the properties of that configuration.

Can you explain this for me?

akeller:
@christinaausley if you look at [this section](https://docs.camunda.io/docs/self-managed/zeebe-deployment/configuration/gateway-config/#zeebegatewaynetwork) of the Zeebe gateway config, you'll see the section title and table provide similar information to the YAML snippet. 

`network` is the config (config name? - there might be a better word for this, given the whole YAML file is config), and under or within the network config you have the fields. These fields are values that set the config within the network. 

This was inspired by the [multi-tenancy docs](https://docs.camunda.io/docs/self-managed/zeebe-deployment/configuration/gateway-config/#zeebegatewaymultitenancy). Some users may not understand that the config (or config name) isn't `enabled`, that's a field.

akeller:
😅 Whoops, I should have mentioned I was using `network` as an example here. With that, I would suggest we modify this slightly: 

> Configuration names are noted as the **header** of each documented section, while the **field** values represent properties to set the configuration.

megglos (approved):
Thanks for making this clearer!

PR #3035, opened 2023-12-12
Soften Amazon S3 references in broker config docs

### LINKED ISSUE #2984: Clarify `ZEEBE_BROKER_DATA_BACKUP_S3_ENDPOINT` can be used for MinIO
We support any S3 compatible storage system, but customers are unsure how to work with MinIO (vs. AWS S3).

More context (internal): https://camunda.slack.com/archives/CSQ2E3BT4/p1701439714479199

Quick win ideas
* add an example value to `ZEEBE_BROKER_DATA_BACKUP_S3_ENDPOINT`
* add an admonition (or something similar) to make it clear in the config docs that the S3 values are not specific to AWS S3

Outside the scope of docs
* modifications to the [broker.yaml.template](https://github.com/camunda/zeebe/blob/main/dist/src/main/config/broker.standalone.yaml.template) to remove AWS docs specific links

akeller:
## Description

Closes #2984. 

I made a few changes/choices in this PR, and I'm looking for feedback before backporting. 

* Added MinIO endpoint as an example value
* Added an admonition, specifically identifying Amazon S3 as an option, but not the only one.
* Added more generic references to soften the emphasis on Amazon S3, including the URLs. However, I found myself asking if we should include the Amazon S3 docs links if this is applicable config for _any_ S3 compatible storage and customers should know where to find their provider's docs. I'm also willing to delete any Amazon S3 docs link entirely to keep this generic. 

## PR Checklist


docs/self-managed/zeebe-deployment/configuration/broker.md
```diff
@@ -219,17 +219,23 @@ backup:
...
+| region               | Configure region. If no region is provided it will be determined as documented by your S3 compatible storage provider. If you use Amazon S3, it will be determined as [documented](https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/region-selection.html#automatically-determine-the-aws-region-from-the-environment). This setting can also be overridden using the environment variable `ZEEBE_BROKER_DATA_BACKUP_S3_REGION`                                                                                                        |                                               |
```
akeller, line 232: Ugh, the preview in GitHub is not so great. Here is an example where I keep the Amazon S3 link.

docs/self-managed/zeebe-deployment/configuration/broker.md
```diff
@@ -219,17 +219,23 @@ backup:
...
+| forcePathStyleAccess | When enabled, forces the s3 client to use path-style access. By default, the client will automatically choose between path-style and virtual-hosted-style. Should only be enabled if the s3 compatible storage cannot support virtual-hosted-style. Refer to your S3 compatible storage provider or the [Amazon S3 docs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/access-bucket-intro.html) for more information. This setting can also be overridden using the environment variable `ZEEBE_BROKER_DATA_BACKUP_S3_FORCEPATHSTYLEACCESS`. | False                                         |
```
akeller, line 236: Here is an example where I keep this as a single sentence, with the Amazon S3 link.

docs/self-managed/zeebe-deployment/configuration/broker.md
```diff
@@ -219,17 +219,23 @@ backup:
...
+| endpoint             | Configure URL endpoint for the store. If no endpoint is provided, it will be determined based on the configured region. This setting can also be overridden using the environment variable `ZEEBE_BROKER_DATA_BACKUP_S3_ENDPOINT`.                                                                                                                                                                                                                                                                                                                    | https://minio.default.svc.cluster.local:9000/ |
```
akeller, line 231: I added the MinIO example as https.

oleschoenburg: @akeller thanks! Maybe worth noting: The https://minio.default.svc.cluster.local:9000/ url is not very generic, it's the result of one particular way to deploy MinIO. The endpoint can be any arbitrary URL, as long as there's an API behind that that speaks S3.

@pierre-yves-monnet Let's keep this focused on S3, the other providers are a separate discussion.
> The idea is not to give the documentation of the provider but to explain what Zeebe expects for each parameter and link to the provider documentation

As far as I know, there is no such documentation for either MinIO or AWS S3. If you think otherwise, please link to it.

> For example, it's necessary to know that Zeebe waits for an access/Secret Key when he needs to call MinIo, because this is part of the Zeebe Code

This is not correct, Zeebe makes no distinction between MinIO and AWS S3. The auto-discovery features of the AWS S3 client that we use are documented here: https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/credentials-chain.html and can be used to configure the client for MinIO too.

docs/self-managed/zeebe-deployment/configuration/broker.md
```diff
@@ -238,7 +244,7 @@ backup:
   store: s3
   s3:
     bucketName: null
-    endpoint: null
+    endpoint: http://minio.default.svc.cluster.local:9000/
```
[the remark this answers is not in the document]
oleschoenburg: The format is correct :+1: 
Whether it's https or http, or rather what the _entire_ URL looks like, depends on how MinIO is deployed. There's no canonical URL, any URL we give as an example here needs to be adjusted to each deployment.

docs/self-managed/zeebe-deployment/configuration/broker.md
```diff
@@ -204,9 +204,9 @@ Use the same configuration on all brokers of this cluster.
...
+| ----- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------- |
+| store | Set the backup store type. Supported values are [NONE, S3, GCS]. Default value is NONE. When NONE, no backup store is configured and no backup will be taken. Use S3 to use any S3 compatible storage, including, but not limited to, Amazon S3. Use GCS to use [Google Cloud Storage](https://cloud.google.com/storage/). This setting can also be overridden using the environment variable `ZEEBE_BROKER_DATA_BACKUP_STORE`. | NONE          |
```
oleschoenburg, line 209: :bulb: At first I wanted to suggest calling it AWS S3 because this is an AWS product but it looks like their marketing agrees with you and calls it Amazon S3.

oleschoenburg (approved):
Thanks for taking care of it @akeller, changes look good for the most part.

docs/self-managed/zeebe-deployment/configuration/broker.md
akeller: I have friends at Amazon that also have to look up whether it is Amazon or AWS for each product 😅

docs/self-managed/zeebe-deployment/configuration/broker.md
```diff
@@ -219,17 +219,23 @@ backup:
...
+| accessKey            | Configure access credentials. If either accessKey or secretKey is not provided, the credentials will be determined as documented by your S3 compatible storage provider. If you use Amazon S3, it will be determined as [documented](https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/credentials.html#credentials-chain). This setting can also be overridden using the environment variable `ZEEBE_BROKER_DATA_BACKUP_S3_ACCESSKEY`                                                                                                  |                                               |
```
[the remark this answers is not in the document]
akeller: Changed it to:

> If either accessKey or secretKey is not provided, it will be determined as [documented](https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/credentials.html#credentials-chain), not based on your S3 compatible storage provider.

I debated adding something specific to Zeebe, but I don't think that's necessary here. We can iterate if we get more pushback or customer confusion.

akeller:
I removed the MinIO URL after @oleschoenburg suggested it was one possible way the URL could look based on how MinIO is configured. I don't want customers to look at it and get confused that their MinIO URL doesn't look like the example for reasons we may not have control of on the C8 side.

@pierre-yves-monnet: Instead of having different providers in the official documentation, Consulting should own supplementary docs or templates as they encounter customers' provider requirements. I worry that being too explicit about 3rd party vendors and platforms in the docs will imply we regularly test against these. 

Based on the feedback, I will move this PR out of draft.

akeller:
I'll revisit this PR when I'm back online in Jan. Adding a `hold` label so I can backport and re-request reviews :v:

akeller:
@christinaausley can I get your 👍 here to merge?

christinaausley (approved):

PR #3037, opened 2023-12-12
add known issue for keycloak

christinaausley:
## Description

Closes https://github.com/camunda/developer-experience/issues/200.

## PR Checklist

### DIFF
--- a/docs/reference/announcements.md
+++ b/docs/reference/announcements.md
@@ -19,27 +19,6 @@ were deprecated in `8.4`. Please use the dedicated Camunda Identity properties o
 
 As of the 8.4 release, Camunda is compatible with Elasticsearch 8.9+ and no longer supports older Elasticsearch versions. See [supported environments](/docs/reference/supported-environments.md).
 
-### Known issue in Keycloak
-
-There currently exists a [known issue in Keycloak](https://github.com/keycloak/keycloak/issues/12484) where Identity experiences a crashloop. You may see something similar to the following:
-
-```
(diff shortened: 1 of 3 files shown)


christinaausley:
I am largely struggling with **where** this should go, as it spans across all clients. Any insight is appreciated.

akeller:
I wouldn't add this to Announcements, but I recommend adding it to the Troubleshooting section (Self Managed > Identity > Troubleshooting). I would follow the same pattern with `Problem: / Solution: `

Now, is this a topic for the [Overview](https://docs.camunda.io/docs/self-managed/identity/troubleshooting/troubleshoot-identity/) or [Common problems](https://docs.camunda.io/docs/self-managed/identity/troubleshooting/common-problems/)? Or could we consolidate these pages? I'm not sure I see the difference in what content should go where.

dlavrenuek:
Thank you for adding the changes @christinaausley 🥇 I agree with @akeller that this should go to the common problems section, but also only in the 8.2 and 8.1 docs. A [fix for this behavior](https://github.com/camunda-cloud/identity/issues/2120) was released with 8.3 on our side, so 8.3+ should not be affected.

christinaausley:
> Now, is this a topic for the [Overview](https://docs.camunda.io/docs/self-managed/identity/troubleshooting/troubleshoot-identity/) or [Common problems](https://docs.camunda.io/docs/self-managed/identity/troubleshooting/common-problems/)? Or could we consolidate these pages? I'm not sure I see the difference in what content should go where.

Agreed here. I can open a separate issue to consolidate these, unless @dlavrenuek or @Ben-Sheppard have different ideas.

dlavrenuek (approved):
Thank you for the changes, looks good from my perspective 💯

akeller (approved):

PR #3050, opened 2023-12-14
Update local-kubernetes-cluster.md, wrong command for switching the c…

grafvonb:
Documentation shows the wrong command for switching the context using `kubectl`: 
`kubectl cluster-info --context kind-camunda-platform-local` 
It should be:
`kubectl config use-context kind-camunda-platform-local`

However, if you create the cluster using `kind` just shortly before, the context will be switched automatically to the new cluster.


maxdanilov (approved):
Thank you for the PR, great catch!

Issue #2569, opened 2023-09-10
Bug: API-tools cards have different sizes


MaxTru:
### Description

The cards on the api-tools page have different sizes, the orientation looks strange.

### How to reproduce

Visit https://docs.camunda.io/docs/apis-tools/working-with-apis-tools/.

See 
![Bildschirmfoto 2023-09-10 um 21 31 38](https://github.com/camunda/camunda-platform-docs/assets/42800119/46321aca-dd5d-4aaf-ad2c-49fa85582e99)

### Suggestion

Ensure same width of all cards.
(Maybe consult with a Designer what to do here.)

christinaausley:
Paired work with Steve on this today and tomorrow 👍

PR #3059, opened 2023-12-14
adjust doc card length and descriptions

christinaausley:
## Description

Closes https://github.com/camunda/camunda-docs/issues/2569 and https://github.com/camunda/developer-experience/issues/144.

🍐 **Paired work with @pepopowitz**.

## PR Checklist


christinaausley:
Tuesday update:

@pepopowitz and I met today, and successfully added an Optimize description!

As far as card sizing goes:

`display: -webkit-box;
    -webkit-box-orient: vertical;
    -webkit-line-clamp: 2;
    overflow: hidden;`
    
We applied those properties to the `p` element within a card.

1. We need to investigate further which browsers this will/won’t work on (specifically worried about this -webkit-box-orient property, and the value -webkit-box because we only looked at webkit-line-clamp.
2. Apply it via a CSS selector, on the CardLayout component of src/theme/DocCard/index.

Planning to look into this as time allows this week 👍

christinaausley:
> We need to investigate further which browsers this will/won’t work on (specifically worried about this -webkit-box-orient property, and the value -webkit-box because we only looked at webkit-line-clamp.

It looks like less than 1% of global users would run into an issue, and it's only for old _old_ versions and IE browser. I think this should be just fine.

christinaausley:
@pepopowitz I almost cried when I got this (at least, locally) on my first build.

I just added the properties to `src/theme/DocCard/styles.module.css` -- please have a look and let me know if you have any critiques.

I may just update the length of the Console description, since it looks weird as the only overflow fellow.

MaxTru (approved):

pepopowitz (approved):
This looks great! I confirmed the expected behavior in Chrome, Firefox, and Safari locally. 💪

PR #2944, opened 2023-11-21
feat: Operate OpenAPI explorer (part 2)

pepopowitz:
Supersedes #2867.

## Description

Another attempt at https://github.com/camunda/developer-experience/issues/158.

This PR implements a prototype of embedded-in-the-docs API interactivity for the Operate API. It is based on #2867, and includes multiple changes based on feedback from the first implementation.

```[tasklist]
### Progress
- [x] Duplicate content changes to Operate API Overview from `next` into `8.3`. (e.g. extract Authentication page, remove anything already covered by the interactive explorer.)
- [x] Docs on how to generate the OpenAPI docs
- [x] Make buttons more visible in the API explorer. They are white-bordered, and hard to find.
```

### Post-launch
- [ ] Talk to Operate team about how this affects them moving forward

## Preview site

A preview of the prototype is available [on Netlify](https://loquacious-pika-a05ac2.netlify.app/). [This URL](https://loquacious-pika-a05ac2.netlify.app/api/operate/docs/operate-public-api/) will take you directly to the new Operate API experience.

Note that the preview site uses the docusaurus setting `noIndex: true` to prevent crawlers from indexing it (and potentially introducing duplicate-content issues).

## What it looks like

(But really, the preview site is worth checking out.)

Note that this does not match exactly the preview site. I've added changes to make the buttons stand out a little more, as shown in this screenshot (see caveat below about the buttons).

image

 
## What's been changed in this version

This PR is based on #2867, with several bits of feedback addressed, and some new ideas: 

### Added an "internal link" icon

This is using the existing brand icon that I thought was closest to indicating an internal link. I don't think it's quite right for the job, but maybe others like it. 

image

I suspect that if we want to keep a link here, I will need to requisition an appropriate branded icon.

### Link to API explorer is removed from top nav

Per suggestion, the original top nav is retained. The only remaining way to find the Operate API Explorer is the sidebar: 

image

This will allow us to iterate without confusing users about which APIs are available or considered "first class".

### Operate API overview docs exist in only one place

I've removed the duplication of the Operate API overview page. Some details:

#### The remaining overview exists in the main docs instance, not the API docs instance.

This means [the overview](https://loquacious-pika-a05ac2.netlify.app/docs/next/apis-tools/operate-api/overview/) can be versioned along with the product, which might come in handy if they reference product features only available for specific versions.

[...]

PR #3068, opened 2023-12-15
fix: enable indexing in production (by swapping our default indexing strategy)

pepopowitz:
In #2944, I mistakenly left a configuration change that set `noIndex: true`. I did this so that when I deployed that PR as a preview site, it didn't get picked up by crawlers. Unfortunately, I overlooked it before merging, and now the production site is _also_ not indexed. 😱 

This PR: 

1. Resolves the fact that production is currently not indexed.
2. Swaps our default `noIndex` strategy, so that _everything_ is not indexed by default, and only the production build toggles `noIndex` to false. 

- [x] This is a bug fix or security concern (and should be released ASAP!).

## PR Checklist

### DIFF
--- a/.github/workflows/publish-prod.yaml
+++ b/.github/workflows/publish-prod.yaml
@@ -25,6 +25,8 @@ jobs:
           node-version: 16
       - name: Install Dependencies
         run: npm ci
+      - name: Enable Indexing
+        run: "sed -i 's/noIndex: true/noIndex: false/g' docusaurus.config.js"
       - name: Build
         run: npm run build
         env:
--- a/.github/workflows/publish-stage.yaml
+++ b/.github/workflows/publish-stage.yaml
@@ -20,8 +20,6 @@ jobs:
           node-version: 16
       - name: Install Dependencies
         run: npm ci
-      - name: Disable Indexing
-        run: "sed -i 's/noIndex: false/noIndex: true/g' docusaurus.config.js"
       - name: Update URL
         run: 'sed -i ''s!url: "https://docs.camunda.io"!url: "https://stage.docs.camunda.io"!g'' docusaurus.config.js'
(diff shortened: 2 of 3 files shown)


docusaurus.config.js
```diff
@@ -17,7 +17,7 @@ module.exports = {
   organizationName: "camunda", // Usually your GitHub org/user name.
   projectName: "camunda-docs", // Usually your repo name.
   trailingSlash: true,
-  // do not delete the following 'noIndex' line as it is modified for staging
+  // do not delete the following 'noIndex' line as it is modified for production
   noIndex: true,
```
pepopowitz, line 21: This _used_ to be set to false. It got changed to true in #2944.

.github/workflows/publish-prod.yaml
```diff
@@ -25,6 +25,8 @@ jobs:
           node-version: 16
       - name: Install Dependencies
         run: npm ci
+      - name: Enable Indexing
+        run: "sed -i 's/noIndex: true/noIndex: false/g' docusaurus.config.js"
```
pepopowitz, line 29: Important to note that the left/right sides of the `sed` command are swapped from what they used to be. Now, in production, we change that line from `noIndex: true` to `noIndex: false`....

.github/workflows/publish-stage.yaml
```diff
@@ -20,8 +20,6 @@ jobs:
           node-version: 16
       - name: Install Dependencies
         run: npm ci
-      - name: Disable Indexing
-        run: "sed -i 's/noIndex: false/noIndex: true/g' docusaurus.config.js"
```
pepopowitz, line 24: ....while previously, we were changing _in staging_ from `noIndex: false` to `noIndex: true`.

akeller (approved):

PR #3091, opened 2023-12-19
feat: Suggest missing changes across current/next versions 

pepopowitz:
## Description

Partially addresses #1778. A proper implementation of #2922. 

This is the first iteration, with a subset of version recommendation rules.

Adds a workflow that comments on a PR when changes were made in next but not current, or vice versa.

### What the message looks like

When there are suggested changes:

image

When suggested changes have been satisfied:

image

### Test cases

Test PRs were created based on this branch to demonstrate behavior in a variety of scenarios:

- [x] No suggestions (no changes in current or next) https://github.com/camunda/camunda-docs/pull/3092
- [x] No suggestions (changes in current are also in next) https://github.com/camunda/camunda-docs/pull/3104
- [x] Changes made in main current but not main next https://github.com/camunda/camunda-docs/pull/3093
- [x] Changes made in main next but not main current https://github.com/camunda/camunda-docs/pull/3105
- [x] Changes made in optimize current but not optimize next https://github.com/camunda/camunda-docs/pull/3106
- [x] Changes made in optimize next but not optimize current https://github.com/camunda/camunda-docs/pull/3107
- [x] Multiple changes suggested in one version https://github.com/camunda/camunda-docs/pull/3108
- [x] Multiple versions suggesting changes https://github.com/camunda/camunda-docs/pull/3109
- [x] Changes suggested, then satisfied https://github.com/camunda/camunda-docs/pull/3112

### Not included

Forked PRs. I ran into issues with checking out the correct SHA in #3111, due to changing the workflow to run on `pull_request_trigger`. I reverted this workflow to run on `pull_request`.

I'll try to make this run for forked PRs in the future. Running it on Camunda/internal PRs will likely help 98% of PRs, and we look at forked PRs with a more scrutinous eye anyway.

### Implementation notes

* The workflow gathers a list of all changed files in the PR, and checks for equal presence across the versions. It **does not** validate that the changes are duplicated exactly across versions; only that **there are changes** in the same files across versions.
* The workflow does not fail if suggested changes are identified. Because there's not 100% certainty that changes were forgotten in another version, I don't want to make it look like something is broken in the PR.
* If a comment is added to a PR for suggested changes, that comment is updated as the PR changes. When all suggested changes are implemented, the comment is updated to basically say "LGTM!".
* If there are no changes missed across versions in a PR from the start, no comment is ever added.
[...]

PR #3093, opened 2023-12-19
[DO NOT MERGE] test PR: changes in current but not next

pepopowitz:
## Description

**Scenario**: changes are made in a file in vCurrent, but not vNext.

**Expected result**: a comment suggesting that the changes in vCurrent also be made in vNext.

Never!


pepopowitz:

PR #3104, opened 2023-12-20
[DO NOT MERGE] test PR: changes in current & next

pepopowitz:
## Description

**Scenario**: changes are made in the same file in both vCurrent and vNext.

**Expected result**: there should be no message suggesting changes in another version.

Never!


pepopowitz:

PR #3105, opened 2023-12-20
[DO NOT MERGE] test PR: changes in next, but not current

pepopowitz:
## Description

**Scenario**: changes are made in a file in vNext, but not vCurrent.

Never!


pepopowitz:

PR #3108, opened 2023-12-20
[DO NOT MERGE] test PR: multiple changes missing in one version

pepopowitz:
## Description

**Scenario**: multiple files are changed in vNext, but none in vCurrent.

**Expected result**: a comment suggesting all the files also be changed in vCurrent.

Never!


pepopowitz:

PR #3110, opened 2023-12-20
[DO NOT MERGE] test PR: changes suggested from a forked PR

pepopowitz:
## Description

**Scenario**: changes are made in vNext but not vCurrent, in a forked PR.

Never!


pepopowitz:
I am not sure why this went wrong after rebasing on the target branch. The git diff command is only returning the workflow file, even though I think it should be returning only the content file. Gonna close and reopen to see if it goes better without a rebase.

PR #3111, opened 2023-12-20
[DO NOT MERGE] test PR: changes in vNext but not vCurrent, in a forked PR

pepopowitz:
## Description

**Scenario**: changes are made in vNext but not vCurrent, in a forked PR.

Never!

### DIFF
--- a/docs/apis-tools/working-with-apis-tools.md
+++ b/docs/apis-tools/working-with-apis-tools.md
@@ -7,6 +7,8 @@ description: "Interact programmatically with Camunda 8 using official Zeebe clie
 
 import DocCardList from '@theme/DocCardList';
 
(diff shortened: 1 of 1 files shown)


pepopowitz:
Ok, this just doesn't work in a forked PR.

The checkout action, because it's running on pull_request_target, is checking out the host's SHA, which is what this is merging into. So it's not running the action with the changes in this PR.

See [this conversation](https://github.com/actions/checkout/issues/518) for a long discussion of what _might_ be the right thing to check out on pull_request_target.

For now I'm going to revert #3091 to triggering on pull_request instead of pull_request_target, and exclude forked PRs from my acceptance criteria. I'll try to fix forked PRs in a follow-up.

Issue #3143, opened 2024-01-03
Temporarily remove section related to events for process instance migration


korthout:
The process instance migration concept documentation currently contains a section related to event subscriptions.

However, this functionality has been removed from version 8.4. This was documented as limitations with:
- https://github.com/camunda/camunda-docs/pull/3094

However, the event subscriptions section may confuse users as it suggests migrating process instances with boundary events and event sub-processes is possible.

I suggest we temporarily remove this section until that functionality is supported.

PR #3144, opened 2024-01-03
Remove reverted event subscriptions section from process instance migration concept

korthout:
## Description

Process instance migration can no longer deal with event subscriptions. While this is documented in the limitations with some specific scenarios, it should also be clear in the rest of the concept documentation. So, this removes the section discussing event subscriptions because it implies behavior that is not available.

I've also clarified the case where an active service task receives a boundary event in the target process definition which is also not supported at this time. It could be implied from the rule above it, but I'd rather have the limitations be explicit.

This pull request can be reverted when we eventually add this functionality.

closes https://github.com/camunda/camunda-docs/issues/3143

- [x] This change is not yet live and should not be merged until 8.4 release (apply `hold` label or convert to draft PR)?

## PR Checklist


berkaycanbc (approved):
LGTM @korthout ! 🚀

christinaausley (approved):

PR #3169, opened 2024-01-05
separate aurora instructions

christinaausley:
## Description

Closes https://github.com/camunda/developer-experience/issues/171.

## PR Checklist

### DIFF
--- a/sidebars.js
+++ b/sidebars.js
@@ -839,6 +839,7 @@ module.exports = {
             "self-managed/platform-deployment/troubleshooting",
           ],
         },
+        "self-managed/platform-deployment/amazon-aurora-postgresql",
         "self-managed/platform-deployment/docker",
         "self-managed/platform-deployment/manual",
       ],
--- a/docs/self-managed/identity/deployment/configuration-variables.md
+++ b/docs/self-managed/identity/deployment/configuration-variables.md
@@ -66,24 +66,11 @@ There are no default values for the variables above. See
 supported databases.
 :::
 
-#### Running Identity on Amazon Aurora PostgreSQL
+:::note
 
-Identity supports running on Amazon Aurora PostgreSQL.
-To connect Identity with your Amazon Aurora PostgreSQL instance, make the following configuration adjustments:
+See additional details for [running Identity on Amazon Aurora PostgreSQL](/self-managed/platform-deployment/amazon-aurora.md#running-identity-on-amazon-aurora-postgresql).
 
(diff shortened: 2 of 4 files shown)


christinaausley:
I can backport this to `8.3` if we like these changes.

sidebars.js
```diff
@@ -839,6 +839,7 @@ module.exports = {
             "self-managed/platform-deployment/troubleshooting",
           ],
         },
+        "self-managed/platform-deployment/amazon-aurora-postgresql",
```
akeller, line 842: This is at the wrong level in the hierarchy. However, we don't have a great space for it with existing docs. 

Do we have other database config in the works? Would it make sense to have a dedicated section for configuring different databases cross components?

christinaausley: I do think this would make sense, especially given the cross-component duplication we've been seeing.

I think this requires a separate issue though, to go through the existing documentation and outline what content could live in such a section (would it be too lean? should we leave as-is until we see more? do users like this new layout, or does it leave them more confused in terms of navigation across the docs?)

akeller: I would be ok with this as a separate issue, but I'm still not happy with the hierarchy as-is. @Ben-Sheppard is the Identity team authoring any additional database config? 

Would it make sense to move this into Helm/Kubernetes > Guides or Operational guides? I want the 3 deployment options very clear in the installation section.

Ben-Sheppard: We aren't at the moment (I'm not sure if Web Modeler are though?)- The only thoughts I have regarding the location is something that I've felt with other docs I've written/reviewed which is, the database config is really independent of the installation method IMO so I could use AWS Aurora in all Helm/K8s, Docker, and Manual (referenced from our current docs structure).

wollefitz: > (I'm not sure if Web Modeler are though?)

Neither are we.

> the database config is really independent of the installation method IMO so I could use AWS Aurora in all Helm/K8s, Docker, and Manual (referenced from our current docs structure).

Agree - I'd not put the AWS configuration below "Helm/Kubernetes" because of this reason, there should be a more generic location for cross-component configuration

akeller: @christinaausley let's discuss in our 1:1 this week to see if we can make a quick change here to the content structure.

christinaausley:
Hi folks -- as we don't have a high demand or influx of this semi-cross component content yet, I'm going to close this out.

Can always revisit as/if more related content rolls in.

PR #3021, opened 2023-12-11
fix(self-managed/zeebe): correct MT feature flag

nikku:
## Description

Using the updated flag I got [multi-tenancy working](https://github.com/camunda/camunda-platform/commit/9fc85a1c93c69144d8527c01d73321178de6e55d#diff-3fde9d1a396e140fefc7676e1bd237d67b6864552b6f45af1ebcc27bcd0bb6e9R25).

## When should this change go live?

- [ ] This change is not yet live and should not be merged until {ADD_DATE} (apply `hold` label or convert to draft PR)?
- [ ] There is no urgency with this change.
- [ ] This change or page is part of a marketing blog, conference talk, or something else on a schedule.
- [ ] This functionality is already available but undocumented.
- [ ] This is a bug fix or security concern.

## PR Checklist

- [ ] I have added changes to the relevant `/versioned_docs` directory, or they are not for an **already released version**.
- [ ] I have added changes to the main `/docs` directory (aka `/next/`), or they are not for **future versions**.
- [ ] My changes require an [Engineering review](https://github.com/camunda/camunda-docs/blob/main/howtos/documentation-guidelines.md#review-process), and I've assigned an engineering manager or tech lead as a reviewer, or my changes do not require an Engineering review.
- [ ] My changes require a [technical writer review](https://github.com/camunda/camunda-docs/blob/main/howtos/documentation-guidelines.md#review-process), and I've assigned @christinaausley as a reviewer, or my changes do not require a technical writer review.

PR #3171, opened 2024-01-05
fix(broker): add multitenancy configuration

nloding:
## Description

Add multitenancy variable configuration to the Zeebe Broker. It was previously undocumented, but if you are using an embedded gateway configuration, you need to use the broker settings rather than the gateway settings.

## PR Checklist


nloding:
After pushing this, I realized that the broker configuration guide doesn't have any of the `ZEEBE_BROKER_GATEWAY_SECURITY_*` variables documented either, which are required to set the `identity.baseUrl` property required by multi-tenancy ...

akeller:
@koevskinikola @Ben-Sheppard can we get your input here? This closes the loop on an open community feedback topic @nloding has been working. 

See these threads/PRs for more context:
* https://camunda.slack.com/archives/C0555930H7F/p1702658366563479
* https://github.com/camunda/camunda-docs/pull/3021

@Ben-Sheppard see @nloding's comment above this one re: `identity.baseUrl`.

akeller:
Adding the hold label so we can collaborate and not view this as blocking for the 8.4 release.

nloding:
I copied/pasted/updated the `security.authentication` variables from the gateway config to the broker config. There was one value in the gateway config (`ZEEBE_GATEWAY_SECURITY_AUTHENTICATION_IDENTITY_TYPE`) that I am not familiar with and isn't needed for MT. My outstanding questions, which may or may not be within the scope of this PR:

* With the authentication/Identity values in Zeebe being deprecated in 8.4, are my updates to the broker page valid?
* What other variables are missing? :)
* The variables on the gateway page are listed a different order than the broker page; I think they should follow the same order. The broker page is alphabetical, which I think makes sense for long term maintenance. Should we standardize this?

akeller:
> The variables on the gateway page are listed a different order than the broker page; I think they should follow the same order. The broker page is alphabetical, which I think makes sense for long term maintenance. Should we standardize this?

I'm not sure I see an order on either page. The order should match the template on the Zeebe repo.

koevskinikola:
Hi @akeller @nloding,

Unfortunately, I haven't had a lot of time in the last two weeks to review this PR. I will probably be able to review it properly after 22.01.2024.

If a review is more urgent, then please ask another Zeebe team member to have a look. The changes are more related to the structure of the docs related to the Zeebe config properties rather than multi-tenancy.

From just a quick look at the PR, I can note that the changes make the Broker configuration page look a bit inconsistent since we're only mirroring the Gateway properties related to multi-tenancy. That might be even more confusing to users.

docs/self-managed/zeebe-deployment/configuration/gateway.md
```diff
@@ -340,6 +340,10 @@ To use multi-tenancy, you must set [`authentication.mode`](#zeebegatewayclusters
 as well.
 :::
 
+:::note
+If you are using an embedded gateway, refer to the [broker configuration guide](./broker.md/#zeebebrokergatewaymultitenancy).
+:::
```
koevskinikola, line 345: :+1: This is pretty nice! It makes users aware that this setup is also possible.

docs/self-managed/zeebe-deployment/configuration/broker.md
```diff
@@ -51,6 +51,75 @@ broker:
     enable: false
 ```
 
+### zeebe.broker.gateway.multitenancy
```
koevskinikola, line 54: :x: I'm not sure adding only the multi-tenancy-related Gateway properties to the Broker config docs is a good idea.

For an embedded gateway setup, any Gateway property can be passed along to the `StandaloneBroker` by prefixing `zeebe.broker`. It might be confusing to users if they see some Gateway-related properties in this docs section, but not all.

This is a general problem that we're facing with Gateway-related properties in the Broker config docs.

:thinking: Maybe we can create a separate section for "Multi-tenancy configuration" (only a name suggestion) similar to the "Experimental configuration" at the end. In the "Multi-tenancy configuration" section we can link to the Gateway config docs related to MT, or move the new text that was added.

koevskinikola (dismissed):
@nloding @akeller I finally managed to review this PR.

In general the text looks good, I'm just worried that it will cause some confusion if we structure it like this. I left some suggestions on how we can avoid that.

docs/self-managed/zeebe-deployment/configuration/broker.md
akeller: I like your suggestion. I will add that to this PR and see if that gives us a temporary solution to move forward.

akeller:
Made a few adjustments to this PR.

nloding:
This looks great! 🚀

christinaausley (approved):

PR #3208, opened 2024-01-12
Update app matrix for 8.4

akeller:

## PR Checklist


akeller:
@RomanJRW I feel like I ask you this question so often 🙈, but we should still follow the 3.x numbering for Optimize when used like this, right?

RomanJRW:
@akeller - it kind of depends on context:
- If we are talking about C7 components, Optimize should be referred to as version 3.x. 
- If we are talking about C8 components, Optimize should be referred to as version 8.x

It's not ideal, but for now Optimize is a single artefact and this superficial difference in version is the status quo that allows C7 customers to not be confused by using a 8.x product, and for C8 customers to have consistent versioning across their stack.

For what it's worth, we would eventually like to have two completely different artefacts for Optimize (7 and 8) and some of this versioning confusion would go away, but this is a future bigger topic

RomanJRW (approved):

christinaausley (approved):

PR #3314, opened 2024-02-13
docs(ootbc): deprecate power Automate Connector

Oleksiivanov:
## Description

deprecate power Automate Connector

https://github.com/camunda/team-connectors/issues/617

TBD

## PR Checklist

- [x] I have added changes to the relevant `/versioned_docs` directory
- [x] I have added changes to the main `/docs` directory (aka `/next/`), and I've assigned an engineering manager or tech lead as a reviewer, or my changes do not require an Engineering review.
- [x] My changes require a [technical writer review](https://github.com/camunda/camunda-docs/blob/main/howtos/documentation-guidelines.md#review-process), and I've assigned @christinaausley as a reviewer,


christinaausley:
@pepopowitz @akeller In a case like this, how would you like me to handle redirects? I ask because we haven't gotten around to https://github.com/camunda/developer-experience/issues/151 yet. I can add a redirect to the Connectors overview page, but should I make this change in `next` as well, or only in our versioned docs and track this via a method @akeller outlined in this issue?

akeller:
@christinaausley make the change in `/next/` as well. 100% agree with redirecting to the overview page.

pepopowitz:
Re: redirects -- if the redirect is also made in vCurrent/older versions, then a future version release will already have the move accounted for and I won't need to create new redirect rules. From that perspective, it doesn't matter what you do in vNext.

Based on this I will add some clarity to https://github.com/camunda/developer-experience/issues/151, because it doesn't call out that the only moves that are hard to track are the ones that _only_ happen in vNext.

igpetrov (approved):

PR #3320, opened 2024-02-14
docs(web-modeler): update supported envs for pg-16

JacobiCamunda:
## Description

This updates the supported environments page to reflect the support of Postgres 16.

## PR Checklist


akeller:
@JacobiCamunda, as this list grows, would it be fair to say 13+ in the future? We don't appear to be deprecating support for older versions, so I'd like to know if we can clean this up or if we need to be specific about the exact major version for any reason.

JacobiCamunda:
> @JacobiCamunda, as this list grows, would it be fair to say 13+ in the future? We don't appear to be deprecating support for older versions, so I'd like to know if we can clean this up or if we need to be specific about the exact major version for any reason.

That's a good idea. But this would imply that we would always support all versions above 13. We do, but when a new version comes out, we have to check that and a new aurora version environment has to be set up for us to run tests again to be sure we really do. Meanwhile, customers could think we support the newest version, when in reality we don't know yet until tested. @jfriedenstab what do you think?

jfriedenstab:
@JacobiCamunda I agree, let's continue to list the specific versions we have tested.

@akeller: We will also remove support for older Postgres versions when they have reached their end-of-life date (according to https://www.postgresql.org/support/versioning/).

jfriedenstab (approved):

akeller:
Thanks @JacobiCamunda @jfriedenstab, fair enough. If you ever move to support all supported versions (including the most recently released), we should change this, but for now, this makes sense.

PR #3394, opened 2024-02-28
reorg bpmn references

christinaausley:
## Description

Closes https://github.com/camunda/developer-experience/issues/162.

A few inline comments, and a few BPMN models removed (if not referenced across our documentation) or moved (if referenced) into the BPMN folder.

This raises a few questions for me to investigate:

- In some BPMN instances we use screenshots rather than BPMN models? When do we use each case and when do we not?
- How would this impact screenshot automation?
- How does BPMN render from an SEO-perspective? Is it better to use screenshots? Is it better to use direct BPMN models? What are the pros and cons of each?
- Can a BPMN model ever be too large to render appropriately?

## PR Checklist


docs/apis-tools/go-client/go-get-started.md
```diff
@@ -193,6 +193,8 @@ Add the following to `main.go` at the bottom of `func main()`.
 	fmt.Println(response.String())
 ```
 
+<!--- This is a bit of an anomaly. Thoughts? --->
```
akeller, line 196: I'm not sure what this is referring to.

akeller:
I don't see anything blocking in the PR and I like that the files immediately download when you click the link. In other contexts I might not like that, but if I'm walking through a tutorial I don't want to leave the page if I don't have to.

My thoughts on your questions in the description 👇 

> In some BPMN instances we use screenshots rather than BPMN models? When do we use each case and when do we not?

My gut says we want to use BPMN models because they are consumable (either via file or raw text). Either way, we should include them with well-described text. 

> How would this impact screenshot automation?

BPMN models shouldn't be changing that often, but are arguably easier to maintain than screenshots.

> How does BPMN render from an SEO-perspective? Is it better to use screenshots? Is it better to use direct BPMN models? What are the pros and cons of each?

I'm super curious if you can find any info on SEO or even accessibility here.

> Can a BPMN model ever be too large to render appropriately?

We shouldn't have highly complex diagrams in the docs for this to be an issue.

docs/apis-tools/go-client/go-get-started.md
christinaausley: We reference `order-process.bpmn` in the code snippet, but I think this is fine and shouldn't have any impact on the migration of this file.

akeller: Won't be a problem.

There are bigger issues with consistency in this guide. I will create a separate issue.

akeller (dismissed):
Non-blocking for this PR - Do you think this is clear enough for our authors to know where to add BPMN files in the future? Or should we include something in the `howtos`?

christinaausley:
> Non-blocking for this PR - Do you think this is clear enough for our authors to know where to add BPMN files in the future? Or should we include something in the `howtos`?

Added guidance @akeller. Good thought.  🤠

akeller (approved):

PR #3405, opened 2024-03-04
Removing possible discriminatory language

itsmestefanjay:
## Description

I was informed during authoring certification questions, that indeed "white-listing" can be very discriminatory. Since I just copied a part from the docs, I wanted to follow -up here and fix it here as well.

Whenever you feel comfortable

## PR Checklist


christinaausley (approved):
Thank you for this catch!

Issue #3438, opened 2024-03-07
More robust table features


akeller:
Markdown tables can only do so much.

I want to sort or filter some tables of information, but Docusaurus doesn't have something OOTB. We should look into providing a react tables project option that allows for more full featured tables in our docs.

akeller:
React tables was pretty easy to add for https://github.com/camunda/camunda-docs/pull/4111, but I haven't heard any feedback or need for this, so I'm going to close it out for now.

PR #3439, opened 2024-03-07
[NEEDS FEEDBACK] - Quick win API feature table

akeller:
## Description

DO NOT MERGE!

This partially addresses https://github.com/camunda/doc-hospital/issues/13. 

Screenshot 2024-03-07 at 4 12 35 PM

Screenshot 2024-03-07 at 4 12 56 PM

There are a few things not ideal or incomplete about this PR:

* Markdown tables 
  * Table striping is not ideal when multiple lines per feature
  * No ability to filter or search the table (yes you can ctrl + f)
* Doesn't include product features list (yet)
* Lacks context in migration guide landing page
  * Should users look at this before concepts? 

I spent about 20 minutes on this PR. I copied the spreadsheet into a markdown table generator and cleaned up the `<br>` to make Docusaurus happy. I also added #3438 to the backlog because I would _love_ to filter or sort the columns (maybe by status?) but we can't do that today. 

Love it? Hate it? Iterate it? 😆


jonathanlukas:
Hi @akeller , I like that you want to get this information on the road. This is very useful for a migration.

Before publishing, I would like to:

* rethink the presentation data structure (for a document, a table might work, but for our docs, we could think of something "better", I like the idea with the ordering and sorting, maybe an interactive canvas?)
* align with existing migration guides: There will be overlaps for sure and possible cross-references that will make readers' lifes easier
* add explanations and links: features from the platform are referenced, so there should be links pointing to their documentation pages

akeller:
@jonathanlukas 100% agree - this was really just a low time investment option since we would likely have to build a custom implementation. I'm very open to how we want to present this information because a table didn't feel quite right, but neither did straight text. 

If you see a good example of how to present similar data, can you link me? I thought about splitting the table by status, and giving a status to each table, but I wasn't happy with that either. I'll also do some more research too.

akeller:
Closing this, will follow up via the [doc-hospital issue](https://github.com/camunda/doc-hospital/issues/13) when I'm back from my upcoming OOO.

Issue #3277, opened 2024-01-31
Document the Zeebe OpenAPI spec


koevskinikola:
### Description

With https://github.com/camunda/product-hub/issues/1926, Zeebe will provide its very own REST API.

The REST API endpoints should be documented in the Camunda 8 documentation. Preferably, the REST API endpoint docs are generated from the [Zeebe OpenAPI specification document](https://github.com/camunda/zeebe/blob/main/zeebe/gateway-protocol/src/main/proto/rest-api.yaml).

### Additional context

- Some clarifications on this task, and some ideas on how to implement it can be found [here](https://docs.google.com/document/d/1IxHvRBNa6VtQN53QjbHxfckD8mJqYxEuh5Uvza4pq80/edit#heading=h.4o91eke4a088)  (internal Google Doc)

pepopowitz:
@koevskinikola I've created a separate issue for myself, https://github.com/camunda/camunda-docs/issues/3421, to track the implementation of the API Explorer. Do you want to keep this issue for your team, and have it represent the related non-generated documentation you'll be creating?

koevskinikola:
Hey @pepopowitz, yes I'll keep this one for any manual docs required for the OpenAPI. If it turns out we don't need it, I'll close it later. Thanks!

megglos:
will be closed by https://github.com/camunda/camunda-docs/pull/3487

npepinpe:
Assigning myself to it to keep track of closing it.

npepinpe:
Authentication will look pretty much like Tasklist, with some minor changes (e.g. example API calls, no cookie support, etc.). But otherwise very similar.

pepopowitz:
@npepinpe I've added an Authentication page in #3529. Please review there to make sure my changes are accurate. (The page can be seen [here in the preview site](https://loquacious-pika-a05ac2.netlify.app/docs/next/apis-tools/zeebe-api-rest/zeebe-api-rest-authentication/), or [here for the commit itself](https://github.com/camunda/camunda-docs/pull/3529/commits/3239fa9790a827716f8422856aa0424c592f2b95)).

npepinpe:
Done thanks to @pepopowitz I've opened some PRs related, but it's been split into multiple things now.

Issue #3250, opened 2024-01-24
Release versions 8.5/3.13.0


pepopowitz:
Release new versions of the documentation: 8.5 for the main docs, and 3.13.0 for the Optimize docs.

**Due Date**: Tuesday, April 9, 2024, US morning.

```[tasklist]
### Tasks (in order)
- [ ] #3600
- [ ] #3601
- [ ] #3616 
- [ ] #3617 
- [ ] #3618 
- [ ] #3619 
- [ ] Cut versions 8.5/3.13.0
```

```[tasklist]
### Follow-up Tasks (blocking to this issue, usually a week or so after release)
- [x] submit sitemap.xml
```

```[tasklist]
### Follow-up Tasks (non-blocking to this issue, begin this work shortly after release)
- [ ] re-synchronize sidebars (1.3, 8.1, 8.2, 8.3, 8.4, 8.5, vNext)
- [ ] #3607 
- [ ] Archive 8.1/3.9.0
```

## Intentionally and notably not included

- Archive 1.3/3.7.0 -- we can't archive this version yet because Optimize 3.13 will not support OpenSearch. (See https://github.com/camunda/product-hub/issues/1908)

## Acceptance criteria

- The version dropdown reads "8.5" for the main docs instance.
- The version dropdown reads "8.5 / 3.13.0" for the Optimize docs instance.
- ~The version number on the Zeebe REST API Explorer reads "1.0" (see https://github.com/camunda/camunda-docs/pull/3487/files#r1529297935)~ -- no longer true, with #3529. It should match the main docs version that you're currently viewing..
- ~The Zeebe REST API Explorer does not emit an "unreleased" banner at the top (see https://github.com/camunda/camunda-docs/pull/3487/files#r1529295658).~ -- no longer true, with #3529. It should match the behavior of the main docs instance.
- An "unmaintained" banner displays for versions 8.1/3.9, because they have exceeded support.
- No "unmaintained" banner should display for versions 8.2/3.10, 8.3/3.11, 8.4/3.12, or 8.5/3.13. 
- An "unreleased" banner should display for versions Next.

## Previous examples

- synchronizing sidebars: #3163, #3174  
- releasing new versions: #3183, #3184
- archiving a previous version: #2190

pepopowitz:
fyi @christinaausley this is where I have tracked that we want to re-sync sidebars after the 8.5 release (under the "blocking follow-up issues list)

PR #3487, opened 2024-03-14
feat: Zeebe REST API Explorer

### LINKED ISSUE #3421: Zeebe REST API Explorer
Implement an API Explorer for the Zeebe REST API.

Deadline: April release

Part of, or related to, https://github.com/camunda/camunda-docs/issues/3277.

## Implementation notes

* The OpenAPI spec is available at https://github.com/camunda/zeebe/blob/main/zeebe/gateway-protocol/src/main/proto/rest-api.yaml
* Any updates to the spec can be PR'ed directly there ☝️ 
* Documentation outside of the API Explorer (e.g. authentication) will be handled by the Zeebe REST team. 

## Acceptance criteria 

* The API itself is currently in alpha. The API Explorer, and navigation to/from it within the docs, should make this clear. When we release 8.5 in April, the API will move to general availability, and the API Explorer should then reflect that.

pepopowitz:
## Description

Resolves #3421.

Introduces a Zeebe REST API Explorer.

What it looks like (but you should really check out the preview site, linked below): 

image

## Preview site 

https://loquacious-pika-a05ac2.netlify.app/api/zeebe/docs/zeebe-rest-api/ will take you directly into the API Explorer.

## What's included

* A very bare Overview page for the Zeebe REST API. This will be filled in more as part of #3277. This doc/section is intentionally only implemented in the "Next" version.
* A custom banner at the top of the Zeebe REST API Explorer that describes the unreleased state of the documentation: 
   image
   This is a custom banner, because the out-of-the-box banners available from Docusaurus emit confusing text for these API Explorers.
* A few small changes to the Zeebe REST OpenAPI spec, to bring language styles in line with our other docs and API Explorers. [See the exact changes I made in this implementation](https://github.com/camunda/camunda-docs/compare/cc4a1390ed40fe52abd86ff9c5d213b8cdca1c2b..10b826e#diff-c1086374f10b77b576f9dbb18ec7e6e0c04dfab9352adadd7a3cb699a12d31b0). I will follow up with a PR for the Zeebe REST team to integrate these changes at the source.
* A customization to the 3rd-party OpenAPI Docusaurus plugin we're using, to avoid a confusing duplication of schema descriptions (described in https://github.com/camunda/camunda-docs/pull/3487#discussion_r1529242015)
 
## What's not included but does not block this PR from merging

* Manual docs, like a more complete Overview and an Authentication page (or section), are covered by #3277.
* Follow-up with the Zeebe REST team to integrate my OpenAPI spec changes. I will handle this in a separate PR. 

- [x] This change or page is part of a marketing blog, conference talk, or something else on a schedule. -- required prior to 8.5 release.

## PR Checklist


package.json
```diff
@@ -19,12 +19,14 @@
     "serve": "docusaurus serve",
     "format": "prettier --write .",
     "format:check": "prettier --check .",
+    "prettier": "prettier",
```
akeller, line 22: ❓ Did you mean to add this?

api/zeebe/zeebe-openapi.yaml
```diff
@@ -0,0 +1,386 @@
...
+    variables:
+      host:
+        default: localhost
+        description: The hostname of a Zeebe Gateway.
+      port:
+        default: "8080"
+        description: The port of the Zeebe REST API server.
+      schema:
+        default: http
+        description: The schema of the Zeebe REST API server.
```
akeller, line 24: When this is released, will it be updated to match the other Explorers with `SERVER-URL`?

package.json
pepopowitz: I did! I don't really understand what's going on but for some reason running prettier to format the API pages right after they're generated ([example](https://github.com/camunda/camunda-docs/blob/10b826ea71aad1bd8c2fac68227b14d4b3d6b579/package.json#L29)) doesn't actually work, and I keep having to format things manually myself afterward! Adding this line allows me to do something like `npm run prettier -- --write api/zeebe/docs/` in my terminal, thus preventing me from accidentally committing a hardcoded path on line 20 above it, like I did in #3435.

api/zeebe/zeebe-openapi.yaml
pepopowitz: Actually I think the opposite -- I think I'm going to change all the other ones to use something like this, because I really like the way it plays out in the explorer.

From https://loquacious-pika-a05ac2.netlify.app/api/zeebe/docs/complete-a-user-task/, on the right side you have a "Base URL". If you hover over it, you get an "Edit" button. Clicking it expands the individual components of the URL, and you can then each component and click "Hide" to save the changes. Once you've done that, all examples with the explorer now use your new base URL.

It's not totally intuitive, but I think it's much better than the current experience with `SERVER-URL`, because you can't change that at all in the other explorers. (You can _try_, by clicking the "Edit" URL, but you can't actually change anything.)

If there's a concern about using `localhost` as the default, maybe there's some intermediate option to use instead, but I like the customizability this brings.

api/zeebe/zeebe-openapi.yaml
koevskinikola, line 1: @pepopowitz will we need to copy over the OpenAPI spec every time we change something there, or will this be automated somehow?

api/zeebe/zeebe-openapi.yaml
akeller: Ok great, I'm just looking for consistency either way at this point.

SERVER-URL isn't my favorite, but it was difficult to communicate what the SaaS and SM URLs could look like. If we can do that (or make the fields editable) and make this explorer actually send requests, that would be very, very cool. 

Sounds like you have this handled, though; carry on!

package.json
akeller: You saw where I was coming from here 😄

api/zeebe/zeebe-openapi.yaml
pepopowitz: We don't currently have plans to automate the updates. @christinaausley and I have plans to update this spec manually every quarter (and especially in time for minor releases). 

We're also trying to figure out what to do re: alpha releases -- do we want the spec updated here? Do we need to create two versions of the API explorer -- one for alphas and one for generally available? It's confusing because our APIs generally only have one version, not aligned with our main product versions.

In your team's case, how frequently do you anticipate the spec changing? And are API changes in between minor releases only applied to the alpha releases, or do they make it to the general audience immediately?

api/zeebe/zeebe-openapi.yaml
pepopowitz: Yeah that SERVER-URL nonsense was me wanting to scrub the generated URL from when I downloaded the spec, but not really grasping how good the experience could be with proper spec definition.

api/zeebe/zeebe-openapi.yaml
pepopowitz: I should rephrase -- _for now_ we plan to handle the updates, because I haven't figured out yet how to make the experience of updating a spec smooth enough that it doesn't require me clipping some thorns along the way. If I can get things to a state where updating the spec is as simple as running a single npm script, then this is probably something we would push back on the product teams to update when they want to.

pepopowitz: And actually, this alpha/v1 question raises another issue for me -- if the v1 API is updated outside of the cadence of our 6-month product releases, does that mean this API is already available for people to use @koevskinikola? And if that's the case, should I bother with adding a banner to the top of this API explorer that says "this API is in alpha"?

api/zeebe/zeebe-openapi.yaml
```diff
@@ -0,0 +1,386 @@
...
+          description: >
+            The user task with the given key is in the wrong state currently.
+            More details are provided in the response body.
+          $ref: "#/components/responses/ProblemResponse"
+        "400":
+          description: >
+            The assignment of the user task with the given key cannot be completed.
+            More details are provided in the response body.
+          $ref: "#/components/responses/ProblemResponse"
+  /user-tasks/{userTaskKey}/update:
```
tmetzke, line 107: ❌ As mentioned in the [original spec issue](https://github.com/camunda/zeebe/issues/16929#issuecomment-1999930146), this needs to be adjusted.
```suggestion
  /user-tasks/{userTaskKey}:
```

api/zeebe/zeebe-openapi.yaml
```diff
@@ -0,0 +1,386 @@
...
+          description: The list of candidate users of the task. Reset by providing an empty list.
+          items:
+            type: string
+          nullable: true
+        candidateGroups:
+          type: array
+          description: The list of candidate groups of the task. Reset by providing an empty list.
+          items:
+            type: string
+          nullable: true
```
tmetzke, line 354: ❓ @pepopowitz, any ideas why the additional properties here have the description from the parent object `UserTaskUpdateRequest` again?
image

tmetzke: 💭 The same happens for the `variables` in the user task completion endpoint
image

pepopowitz: It looks like the `additionalProperties: true` property on the schema object not only renders the `property/name*` property, but uses the same description as the schema object for the additional properties item. 

I'll see if I can figure out a way to avoid that, whether it's something I can override in the plugin or something I can change in the schema.

pepopowitz: I pushed up [a fix for this](https://github.com/camunda/camunda-docs/pull/3487/commits/1a27d42025a5ee20b36c173e947d2785cffd6771). I haven't yet deployed it to the preview site. 

## The problem

The plugin we're using to generate these docs handles `additionalProperties: true` in a funny way. It _does_ emit the `property *name` map, but it also duplicates the description from the parent schema.

I looked into multiple ways to work around this with alternative `additionalProperties` formats -- each of which displays in a way that makes sense using https://editor.swagger.io/, so I know it would be a valid schema -- but the plugin didn't emit any of these in a way that made sense. It seems to have limited support for the `additionalProperties` property.

## The workaround

Instead, I chose to swizzle (override) a rendering component from the OpenAPI plugin, to scrub the duplicate description in a context where it looks like the `additionalProperties` map is being rendered. Luckily for us, the `name` of the item being rendered is rather unique in these conditions. 

Here's what the two schemas in question now look like: 

image

image

And here's what https://editor.swagger.io/ displays for these properties, to demonstrate alignment: 

image

image


## Does the message `Additional properties allowed." capture everything?

If there's a message that would make more sense in both situations, let me know! I chose this message because it aligns with how Swagger displays it. I'd rather not override the messages on a per-field basis.

api/zeebe/zeebe-openapi.yaml
pepopowitz: Updated in https://github.com/camunda/camunda-docs/pull/3487/commits/216c0f1368040f5d9d97391e6e268121dc3b4cf0.

src/theme/DocVersionBanner/index.js
```diff
@@ -0,0 +1,39 @@
...
+  return (
+    <div
+      className={clsx(
+        className,
+        ThemeClassNames.docs.docVersionBanner,
+        "alert alert--warning margin-bottom--md"
+      )}
+      role="alert"
+    >
+      <div>This is unreleased documentation for the Zeebe REST API.</div>
```
pepopowitz, line 36: @christinaausley @akeller this is where I'm emitting a message at the top of the Zeebe API Explorer to clarify that it is not yet released. It looks like this: 

image

The language is I think slightly confusing, because it's not just that the docs are unreleased, the entire API is (I think)....but it aligns with the messaging that shows up on vNext docs on the rest of the site: 

image

Feel free to propose alternative text here!

cc @koevskinikola because the conversation in [this thread](https://github.com/camunda/camunda-docs/pull/3487#discussion_r1526492335) probably impacts this discussion.

src/theme/DocVersionBanner/index.js
```diff
@@ -0,0 +1,39 @@
+// Why is this swizzled?
+//   To emit a custom banner for the Zeebe REST API Explorer, which has not been released yet.
+// Swizzled from version 2.4.1.
+
+// TODO: Delete this swizzled component as part of 8.5 release!
```
pepopowitz, line 5: I'll add some acceptance criteria to #3250 so that I don't lose track of this.

docusaurus.config.js
```diff
@@ -148,6 +148,44 @@ module.exports = {
...
+      {
+        id: "api-zeebe-docs",
+        path: "api/zeebe",
+        routeBasePath: "api/zeebe",
+        sidebarPath: require.resolve("./api/zeebe/zeebe-sidebars.js"),
+        editUrl: "https://github.com/camunda/camunda-docs/edit/main/",
+        lastVersion: "current",
+        versions: {
+          current: {
+            label: "0.1",
```
pepopowitz, line 163: I think that when the 8.5 release happens, this should update to 1.0. I'll add acceptance criteria to #3250 so that I don't lose track of this.

pepopowitz:
FYI to all that I've re-published the preview site, and it is up to date with all changes as of this comment.

api/zeebe/zeebe-openapi.yaml
tmetzke: Great fix, @pepopowitz, I like the alignment with what Swagger displays 👍

api/zeebe/zeebe-openapi.yaml
koevskinikola: > We're also trying to figure out what to do re: alpha releases -- do we want the spec updated here?
@pepopowitz doesn't the `next` version of the docs represent alphas, i.e. the `SNAPSHOT` version of the docs? I think it would be good if we update the spec with every alpha release. Why?
* As we add new endpoints, they will need to be tested by the QA team, and they should be able to read docs on the feature (as a user would) in order to QA test it properly.
* Our users like to try out the features they are waiting on while we are developing them, so they can provide quick feedback before the feature is completely released. Providing docs earlier would help them understand how to use the new endpoints.

> In your team's case, how frequently do you anticipate the spec changing?
For the next 2 minor releases, I expect the spec to change significantly with every alpha release. With 8.5, we will be migrating all of the gRPC endpoints to REST, so there will be a lot of new endpoints every month. In 8.7, we might add additional endpoints (if we haven't managed to add them by 8.6). After that, it depends on what type of features we add. From experience working on the gRPC spec, there will be some changes to the existing endpoints with each minor release (ex. adding new properties, etc).

> And are API changes in between minor releases only applied to the alpha releases, or do they make it to the general audience immediately?
In general, API changes will only be applied to alpha releases, the API spec of past minor releases should remain unchanged to maintain backwards compatibility. However, if we discover that a bug was created due to how the API spec was defined in a past minor release, we may have to backport a change. But this shouldn't happend very often, and we should be careful how we do it.

> if the v1 API is updated outside of the cadence of our 6-month product releases, does that mean this API is already available for people to use @koevskinikola? And if that's the case, should I bother with adding a banner to the top of this API explorer that says "this API is in alpha"?
I'm not sure I understood this question. Do you mean if we update the API for our alpha/SNAPSHOTs? If yes, I think the same rules apply here, as with how we document the gRPC spec. From what I can see, we have a not there stating `This is unreleased documentation for Camunda 8 Docs Next version.`

src/theme/DocVersionBanner/index.js
koevskinikola: I wrote something similar in [the comment above](https://github.com/camunda/camunda-docs/pull/3487#discussion_r1531887349). I think it makes sense.

api/zeebe/zeebe-openapi.yaml
pepopowitz: Oof, this conversation really helped me realize how much I was misunderstanding the versioning of our APIs. I think that's where the source of confusion comes from in your responses -- the API Explorer I've added here _only exists for one version_, v1. In reality, I think we want the API Explorer embedded in the main versioned docs, so that the 8.5 Zeebe REST API endpoints can be different from 8.4, and 8.3, and all alphas can be properly documented in vNext.

I have a longer explanation [in Slack](https://camunda.slack.com/archives/C01H4NG9XDY/p1711405079397489?thread_ts=1710797951.637499&cid=C01H4NG9XDY). 

To reduce the noise, I plan to close this PR, and re-open it with the API Explorer properly embedded within the main versioned documentation. Recognizing that the minor release is coming soon, I intend on having this new PR ready to review tomorrow, and ideally merged by the end of the week.

pepopowitz:
Closing this PR. It will be re-opened with a different approach -- embedding the API Explorer into the main versioned docs, so that users can see the difference between the API in v8.3 vs v8.4 vs vNext. 

See [this thread](https://github.com/camunda/camunda-docs/pull/3487#discussion_r1538315827) and [this Slack conversation](https://camunda.slack.com/archives/C01H4NG9XDY/p1711405079397489?thread_ts=1710797951.637499&cid=C01H4NG9XDY) for more details.

PR #3510, opened 2024-03-22
draft out rate limit details

christinaausley:
## Description

Closes https://github.com/camunda/developer-experience/issues/274.

## PR Checklist

### DIFF
--- a/docs/apis-tools/administration-api/authentication.md
+++ b/docs/apis-tools/administration-api/authentication.md
@@ -61,14 +61,16 @@ Access tokens have a validity period found in the access token. After this time,
 
 ## Built-in rate limit
 
-The OAuth service has a built-in rate limit of about one request per second. This applies to clients with the same source IP address.
+The OAuth service has a built-in rate limit of about one request per second for all clients with the same source IP address.
 
-The officially offered [client libraries](/docs/apis-tools/working-with-apis-tools.md) (as well as the Node.js client) have already integrated with the auth routine, handle obtaining and refreshing an access token, and make use of a local cache.
+The officially offered [client libraries](/docs/apis-tools/working-with-apis-tools.md) (as well as the Node.js and Spring clients) have already integrated with the auth routine, handle obtaining and refreshing an access token, and make use of a local cache.
 
-If too many token requests are executed in a short time, the client is blocked for a certain time. Since the access tokens have a certain validity period, they must be cached on the client side.
+If too many token requests are executed from the same source IP address in a short time, all token requests from that source IP address are blocked for a certain time. Since the access tokens have a 24-hour validity period, they must be cached on the client side, reused while still valid, and refreshed via a new token request once the validity period has expired.
 
-When the rate limit triggers, the client will receive an HTTP 429 response. Note the following workarounds:
+When the rate limit is triggered, the client will receive an HTTP 429 response. Note the following workarounds:
 
-- Catch the token, as it is still valid for 24 hours. The official SDK already does this by default.
+- Cache the token as it is still valid for 24 hours. The official SDKs already do this by default.
 - Keep the SDK up to date. We have noted issues in older versions of the Java SDK which did not correctly cache the token.
-- Given the rate limit applies to clients with the same source IP address, be mindful of unexpected clients running within your infrastructure.
+- Given the rate limit applies to clients with the same source IP address, be mindful of:
+  - Unexpected clients running within your infrastructure.
+  - Updating all clients to use a current API key if you delete an API key and create a new one.
--- a/docs/apis-tools/administration-api/authentication.md
+++ b/docs/apis-tools/administration-api/authentication.md
@@ -63,6 +63,10 @@ Access tokens have a validity period found in the access token. After this time,
 
 The OAuth service rate limits about one request per second for all clients with the same source IP address.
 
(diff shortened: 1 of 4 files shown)


daniel-ewing (changes requested):
I made some changes.  You might not like these changes.  If you don't, then feel free to ignore me.   For the last two lines, is there any way to indent the bullets (make a sub-bullet list):

The OAuth service has a built-in token request rate limit of about one request per second for all clients with the same source IP address.
The officially offered [client libraries](/docs/apis-tools/working-with-apis-tools.md) (as well as the NodeJS and Spring clients) have already integrated with the auth routine, handle obtaining and refreshing an access token, and make use of a local cache.
If too many token requests are executed from the same source IP address in a short time, all token request from that source IP address are blocked for a certain time. Since an access token has a 24 hour validity period, it must be: cached on the client side, reused while still valid, refreshed via a new token request once its validity period has expired.
When the rate limit gets triggered, the client will receive an HTTP 429 response. Note the following workarounds:
- Cache the token as it is valid for 24 hours. The official SDKs already do this by default.
- Keep the SDK up to date. We have noted issues in older versions of the Java SDK which did not correctly cache the token.
- Given the rate limit applies to clients with the same source IP address, be mindful of:
-- Unexpected clients running within your infrastructure.
-- Updating all clients to use a current API key if you delete an API key and create a new one.

docs/components/console/manage-clusters/manage-api-clients.md
```diff
@@ -18,11 +18,19 @@ For the `Client Id` and `Client Secret`, a client application can request an acc
...
+The OAuth service has a built-in rate limit of about one request per second. This applies to clients with the same source IP address.
 
 The officially offered [client libraries](../../../apis-tools/working-with-apis-tools.md) (as well as the Node.js client) have already integrated with the auth routine, handle obtaining and refreshing an access token, and make use of a local cache.
-:::
+
+If too many token requests are executed in a short time, the client is blocked for a certain time. Since the access tokens have a certain validity period, they must be cached on the client side.
+
+When the rate limit triggers, the client will receive an HTTP 429 response. Note the following workarounds:
+
+- Catch the token, as it is still valid for 24 hours. The official SDK already does this by default.
```
christinaausley, line 31: Oops I think I meant cache here.

thomas-polnik (approved):

christinaausley:
@daniel-ewing Adjusted based on your comment 👍

docs/apis-tools/administration-api/authentication.md
```diff
@@ -59,6 +59,18 @@ curl --header "Content-Type: application/json" \
 Access tokens have a validity period found in the access token. After this time, a new access token must be requested.
 :::
 
-:::note
-The auth service has built-in rate limiting. If too many token requests are executed in a short time, the client is blocked for a certain time. Since the access tokens have a certain validity period, they must be cached on the client side.
-:::
+## Built-in rate limit
+
+The OAuth service has a built-in rate limit of about one request per second for all clients with the same source IP address.
```
multani, line 64: @christinaausley it's a minor change, but I think I would suggest to completely remove the wording `built-in`: I'm not sure why it was there in the first place and it doesn't indicate anything really useful:

```suggestion
## Rate limit

The OAuth service rates limit all clients with the same source IP address to about one request per second.
```

multani (changes requested):
I think it's also worth mentioning that **all the requests** are rate-limited, whether the request was successful or not.

It's especially important wrt. the last point you mentioned ("updating clients to use a new API key"): if a client can't authenticate because the credentials are expired and it "loops" trying to fetch a new token, its requests will count towards the rate limit.

I'm not sure what would be the best place to write this: maybe in the 3rd paragraph "if too many token requests are executed from the same source IP address in a short time"?

docs/apis-tools/administration-api/authentication.md
christinaausley: Done 👍

multani:
> I think it's also worth mentioning that **all the requests** are rate-limited, whether the request was successful or not.
> 
> It's especially important wrt. the last point you mentioned ("updating clients to use a new API key"): if a client can't authenticate because the credentials are expired and it "loops" trying to fetch a new token, its requests will count towards the rate limit.

@christinaausley do you have an idea also how to phrase this? This could also help support by pointing out to customers that problems on their side can affect themselves.

daniel-ewing:
Hi @christinaausley, following up on @multani's comment, does anyone like this:

The OAuth service has a built-in token request rate limit of about one request per second for all clients with the same source IP address.  
> I don't know the markdown syntax for a warning box - make this part a warning:
> All token requests count toward the rate limit, whether they are successful or not.  If any client is running with an expired or invalid API key, that client will continually make token requests.  That client will cause the rate limit to be exceeded for that IP address and it can block valid token requests from completing.

The officially offered ...

christinaausley:
@daniel-ewing @multani thank you for your help here! Added a note 👍

multani (approved):
This looks great to me, thanks @christinaausley and @daniel-ewing !
@daniel-ewing do you think this will help Support team to redirect relevant support tickets to in the future?

daniel-ewing:
@multani, I hope so! But if not, we tried...

PR #3566, opened 2024-04-02
refactor: generalize the strategy for generating API Explorers

pepopowitz:
## Description

* Refactors some developer tooling for generating API explorers, so that it makes sense for all known API explorers. 
   Steps that were previously chained sequentially in the package.json are now formalized in a JavaScript file, with the ability for each API Explorer to specify its custom needs. 
* Converts the Zeebe REST API Explorer to use this new generation workflow as an example.
* Documents this updated generation workflow.

## Not included in this PR

* Conversion of Tasklist and Operate APIs to use this new generation workflow. Those will be coming in two separate follow-up PRs.

- [x] This change or page is part of a marketing blog, conference talk, or something else on a schedule. I'd like to have this merged in time for next week's production release!

## PR Checklist

### DIFF
--- a/api/generate-api-docs.js
+++ b/api/generate-api-docs.js
@@ -1,0 +1,54 @@
+const { execSync } = require("child_process");
+
+// More strategies to come, for other APIs.
+const zeebe = require("./zeebe/generation-strategy");
+const apiStrategies = {
+  zeebe,
+};
+
+// Execute a command as if we were in the terminal
+function runCommand(command) {
+  const result = execSync(command, { stdio: "inherit" });
+  return result;
+}
+
+// API name must be passed in as an arg.
+const api = process.argv[2];
+if (api === undefined) {
+  const validAPIs = string.join(apiStrategies.join, ", ");
+  console.log(`Please specify an API name. Valid names: ${validAPIs}`);
+  process.exit();
+}
+
+// The API name must be recognized.
+const strategy = apiStrategies[api];
+if (strategy === undefined) {
+  const validAPIs = string.join(apiStrategies.join, ", ");
+  console.error(`Invalid API name ${api}. Valid names: ${validAPIs}`);
+  process.exit();
+}
+
+// All APIs will execute these same steps, with custom-per-API steps
+//   implemented by each API's generation-strategy.js.
+const steps = [
+  // Remove old docs
+  () => runCommand(`docusaurus clean-api-docs ${api} -p api-${api}-openapi`),
+
+  // Run any custom steps before generation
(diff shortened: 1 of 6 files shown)


api/zeebe/generation-strategy.js
```diff
@@ -0,0 +1,28 @@
...
+function removeDuplicateVersionBadge() {
+  // The generator adds a version badge to the Introduction file, but
+  //   we already have a version badge from the main docs layout.
+  console.log("removing duplicate version badge...");
+  replace.sync({
+    files:
+      "docs/apis-tools/zeebe-api-rest/specifications/zeebe-rest-api.info.mdx",
+    from: /^.*Version: .*$/m,
+    to: "",
+  });
```
pepopowitz, line 27: One thing I'm excited about with this approach is that this step should work cross-platform. The previous implementation of this was using `sed` to remove this line, which works great, but I think it would not have worked on most Windows machines. (I think.)

api/generate-api-docs.js
pepopowitz, line 1: Comments that I'd typically add inline for the PR are covered by the code comments.

api/zeebe/generation-strategy.js
```diff
@@ -0,0 +1,28 @@
+const replace = require("replace-in-file");
+
+const outputDir = "docs/apis-tools/zeebe-api-rest/specifications";
+function preGenerateDocs() {
+  // noop
```
pepopowitz, line 5: The `preGenerateDocs()` hook is obviously not yet needed in this PR, because I've defined it as a noop, but other APIs will need this hook. 

I could make the logic in `generate-api-docs` smart enough to not explode if there's no `preGenerateDocs()` function defined by the strategy, but I like the deliberate and explicit implementation of a specific interface here.

api/generate-api-docs.js
```diff
@@ -0,0 +1,54 @@
...
+  () => runCommand(`docusaurus gen-api-docs ${api} -p api-${api}-openapi`),
+
+  // Run any custom steps after generation
+  strategy.postGenerateDocs,
+
+  // Run prettier against the generated docs. Twice. Yes, twice.
+  //   I don't know why, but the first run always leaves an extra blank line,
+  //   which the second execution removes.
+  () => runCommand(`prettier --write ${strategy.outputDir}`),
+  () => runCommand(`prettier --write ${strategy.outputDir}`),
```
tmetzke, line 50: Really weird, but as long as it works 🤷

christinaausley (approved):
🥳

PR #3567, opened 2024-04-03
chore: update zeebe rest api docs

pepopowitz:
## Description

Updates the Zeebe REST API Explorer docs based on https://github.com/camunda/zeebe/issues/16759. 

## How did I generate these docs?

I used this update as an opportunity to test my refactoring of the generation process from #3566. (And then I rebased the re-generated docs on main, to avoid including 3566's changes in the diff of this PR.)

1. I replaced the api/zeebe/zeebe-openapi.yaml with the contents from https://github.com/camunda/zeebe/blob/main/zeebe/gateway-protocol/src/main/proto/rest-api.yaml.
2. I ran `npm run api:generate zeebe` in my terminal. (Note that that flavor of scripts is only available in #3566, not in the version of `main` that this branch is based on.)
 
## The generated docs are not quite right

**Update:** After the most recent commit, the UI shows the correct description, by brutally hacking the OpenAPI spec.

It looks like the tooling we're using to generate API docs doesn't handle the `description` fields properly in this updated OpenAPI spec. When a property is described by the `allOf` attribute to merge multiple schemas, the generator chooses the `description` from the first merged schema, rather than the `description` of the property itself. 

Below is a screenshot of how this appears in the UI. The screen on the left shows the docs from this PR; the screen on the right shows what it _should_ look like. As discussed with @tmetzke, the description that _does_ appear in this PR is adequate for now, and I will look into fixing this after the upcoming release. I've opened #3568 to track this.

image

- [x] This change or page is part of a marketing blog, conference talk, or something else on a schedule. Should be merged prior to cutting the next version of docs!

## PR Checklist


api/zeebe/zeebe-openapi.yaml
```diff
@@ -298,6 +322,8 @@ components:
       type: object
       properties:
         changeset:
+          allOf:
+            - $ref: "#/components/schemas/Changeset"
```
pepopowitz, line 326: this is the change that prompts the generator to select the wrong `description`.

tmetzke: Can we, for the time being, simply use the expected description in the `Changeset` schema instead of `A map of changes`? That way, the result is correct for now. When the issue in #3568 is fixed, we can use the OpenAPI YAML as is. Wdyt?

tmetzke: If we do this, we should probably add a note to the OpenAPI file in the docs as well so we're aware the next time we adjust it (given we do that before the issue is fixed).

pepopowitz: This is a good idea! I'll add a step to the pre-generation script to update the description of `Changeset`, and mark the step clearly as "temporary until the generator picks the correct description." Commit coming soon, on this PR.

pepopowitz: @tmetzke take a look at https://github.com/camunda/camunda-docs/pull/3567/commits/6d3d3725e6b26b1a6cf3bf934f66e195af126279. It hacks the description of the Changeset schema object, using a hardcoded value.

I started down a rabbit hole of (1) trying to capture the description from the UserTaskUpdateRequest.changeset property, (2) assigning it to the description of the Changeset, then (3) correcting the formatting, but it got pretty deep and I didn't think it was worth it. The risk of this current solution is that if the source description changes, we'll need to update this code.

Since it's a hack anyway, I thought that was appropriate, instead of spending the next day trying to figure out how to either customize the output of https://github.com/nodeca/js-yaml so that it doesn't create 300 unimportant formatting differences, or capturing _just_ one property from output re-generated by js-yaml and putting it in the correct place.

tmetzke: I love that you tried to make this as generic as possible, I was also wondering if that is easily achievable 👍
Since it looks like it's not, let's go with a pragmatic approach and fix it manually, plus, document it there as well.

api/zeebe/generation-strategy.js
```diff
@@ -15,6 +17,36 @@ module.exports = {
...
+        * \`followUpDate\` - reset by providing an empty String
+
+        Providing any of those attributes with a \`null\` value or omitting it preserves
+        the persisted attribute's value.
+
+        The assignee cannot be adjusted with this endpoint, use the Assign task endpoint.
+        This ensures correct event emission for assignee changes.`,
+  });
+}
+
```
tmetzke, line 49: ❓ Do we still need this when we also adjusted the YAML manually?
💭 We could add a comment to the YAML here in the docs project that this is manually adjusted currently. In the future, when we replace the YAML, we can see this change popping up. We could remove this hacky workaround then.

pepopowitz: My goal is to make the generation process as mistake-proof as possible. We _can_ rely on humans to catch the diff when they re-generate the docs and revert it, but it's nearly as likely that the diff would be overlooked, no one catches it in review, and the description is reverted in production.

The workflow I'm envisioning is that an alpha ships, someone grabs the updated OpenAPI spec, pastes it into `zeebe-openapi.yaml` here, and then runs `npm run api:generate zeebe`.... _then_ looks at the diff to see what's changed, or just opens a PR without looking at the diff. 
 
This step is written in a way that it will only replace the description if it still has the "incorrect" value. If the person re-generating the docs notices the diff _before_ re-generating, and reverts the description change, this action is effectively a noop. 

I don't feel good about hardcoding the contents of the description in this hack, but I do feel confident that this hack will prevent round-trips during re-generation, where the reverted description is not noticed until the PR is open. If you look at [the tasklist pre-generation script](https://github.com/camunda/camunda-docs/blob/4e82bbfeeaea7174bb1fec0b636fd7267a8d1ebc/api/tasklist/generation-strategy.js#L9-L10), or [the operate pre-generation script](https://github.com/camunda/camunda-docs/blob/main/api/operate/generation-strategy.js#L8), you can see other types of changes that can also be caught in this way. Eventually I hope to get all of those issues cleaned up at the source, but right now it's a nice way for me to automate cleanup that I'm doing upon generation, and I see this description as similar.

tmetzke: OK, I see, good arguments. Let's roll with that 👍

tmetzke (approved):
Thanks, @pepopowitz, for taking this on so quickly 👍

PR #3583, opened 2024-04-04
Update local-kubernetes-cluster.md

grafvonb:
Instead of `kubectl cluster-info --context kind-camunda-platform-local` should be `kubectl config use-context kind-camunda-platform-local`


aabouzaid (approved):

akeller:
@aabouzaid does this change need to be backported? We can backport in a follow up PR.

aabouzaid:
@akeller yes, the same for the other version.

akeller:
Since I cannot make changes to this PR I will merge as-is and backport in a follow up. Thank you @grafvonb!

PR #3641, opened 2024-04-09
fix?: maybe fix these two api explorer redirects?

pepopowitz:
## Description

For some reason, the old Operate & Tasklist API explorer URLs are not redirecting to their new homes. I'm not sure why this is. And it's actually even weirder, because you can _see_ the explorer on the screen for a second before it changes to a "Not found" screen. [Example](https://docs.camunda.io/api/operate/docs/operate-public-api/).

When investigating, I noticed the redirect rules for these explorers were missing the `[R=301,L]` options. I don't _think_ this would cause the URL to not redirect, because it still matches, and nothing matches after it....but it's the only thing I can see that looks funny about these two rules, so maybe?????

## PR Checklist


christinaausley (approved):
🤞

akeller:
LMK if this continues to be an issue and I'll do some digging too.

pepopowitz:
The results of this PR are positive in staging -- the old URLs now redirect to the new ones. 🤞

pepopowitz:
The redirects were fixed in production, too. 🙌

PR #3691, opened 2024-04-19
docs(self-managed): tidy up sidebar names and add openshift version

aabouzaid:
## Description

Keep the names in the sidebar consistent and add the OpenShift version (already supported for 8.5).

ASAP.

## PR Checklist

- [x] I have added changes to the relevant `/versioned_docs` directory.
- [x] I have added changes to the main `/docs` directory (aka `/next/`).
- [x] My changes do not require an Engineering review.
- [x] My changes do not require a technical writer review.

### DIFF
--- a/docs/self-managed/setup/deploy/openshift/redhat-openshift.md
+++ b/docs/self-managed/setup/deploy/openshift/redhat-openshift.md
@@ -12,9 +12,9 @@ We test against the following OpenShift versions and guarantee compatibility wit
 
 | OpenShift version | Supported          |
 | ----------------- | ------------------ |
-| 4.11.x            | :white_check_mark: |
 | 4.12.x            | :white_check_mark: |
 | 4.13.x            | :white_check_mark: |
+| 4.14.x            | :white_check_mark: |
 
 Any version not explicitly marked in the table above is not tested, and we cannot guarantee compatibility.
 
--- a/versioned_sidebars/version-8.5-sidebars.json
+++ b/versioned_sidebars/version-8.5-sidebars.json
@@ -1397,13 +1397,11 @@
                 },
                 "self-managed/setup/deploy/amazon/aws-marketplace"
               ],
-              "Microsoft Azure": [
+              "Microsoft (Azure)": [
                 "self-managed/setup/deploy/azure/microsoft-aks"
               ],
-              "Google Cloud Platform": [
-                "self-managed/setup/deploy/gcp/google-gke"
-              ],
-              "OpenShift": [
+              "Google (GCP)": ["self-managed/setup/deploy/gcp/google-gke"],
+              "Red Hat (OpenShift)": [
                 "self-managed/setup/deploy/openshift/redhat-openshift"
               ],
               "Other": [
--- a/optimize_sidebars.js
+++ b/optimize_sidebars.js
@@ -2139,7 +2139,7 @@ module.exports = {
             },
 
             {
-              "Microsoft Azure": [
+              "Microsoft (Azure)": [
                 docsLink(
                   "Microsoft AKS",
                   "self-managed/setup/deploy/azure/microsoft-aks/"
@@ -2148,7 +2148,7 @@ module.exports = {
             },
 
             {
-              "Google Cloud Platform": [
+              "Google (GCP)": [
                 docsLink(
                   "Google GKE",
                   "self-managed/setup/deploy/gcp/google-gke/"
@@ -2157,7 +2157,7 @@ module.exports = {
             },
 
             {
-              OpenShift: [
+              "Red Hat (OpenShift)": [
                 docsLink(
                   "Red Hat OpenShift",
                   "self-managed/setup/deploy/openshift/redhat-openshift/"
--- a/optimize_versioned_sidebars/version-3.13.0-sidebars.json
+++ b/optimize_versioned_sidebars/version-3.13.0-sidebars.json
@@ -2660,7 +2660,7 @@
               ]
             },
             {
-              "Microsoft Azure": [
+              "Microsoft (Azure)": [
                 {
                   "type": "link",
                   "label": "Microsoft AKS",
@@ -2669,7 +2669,7 @@
               ]
             },
             {
-              "Google Cloud Platform": [
+              "Google (GCP)": [
                 {
                   "type": "link",
                   "label": "Google GKE",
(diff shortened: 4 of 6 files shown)


versioned_sidebars/version-8.5-sidebars.json
akeller, line 1: @aabouzaid can you help me understand these changes? Since these are brand names I'm not sure why you are adding the parenthesis. Will we have other, for example, Microsoft technologies in this section? 

Amazon (AWS) is a bit of a special case because they flip flop on their brand names - some are AWS and some are Amazon.

docs/self-managed/setup/deploy/openshift/redhat-openshift.md
```diff
@@ -12,9 +12,9 @@ We test against the following OpenShift versions and guarantee compatibility wit
 
 | OpenShift version | Supported          |
 | ----------------- | ------------------ |
-| 4.11.x            | :white_check_mark: |
```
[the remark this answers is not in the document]
aabouzaid: Not sure if that's needed since OpenShift 4.11 officially [reached end-of-life](https://docs.openshift.com/dedicated/osd_architecture/osd_policy/osd-life-cycle.html#sd-life-cycle-dates_osd-life-cycle) on Dec 10, 2023 (by Red Hat).

versioned_sidebars/version-8.5-sidebars.json
aabouzaid: Amazon is the company name, the same as Microsoft and Google.
AWS, GCP, and Azure are the cloud services provided by those companies.

Hence we have some inconsistency in the naming ... so if we used AWS for Amazon, we should use GCP for Google.

akeller: Please follow up with @christinaausley on this topic. We use AWS for Amazon per their product brand name, not the company name. See https://confluence.camunda.com/display/HAN/Product+glossary.

If we want to capture the company name for SEO purposes, we can do that with metadata on individual pages via keywords.

aabouzaid: The point here is, if we use `Amazon (AWS)` why we don't use `Google (GCP)` :thinking:

The cloud platform is different from the product names. For example, the product is called [Amazon Simple Notification Service](https://aws.amazon.com/sns/) which is part of AWS ([Amazon Web Services](https://aws.amazon.com/)) so it makes sense to call it `Amazon SNS` but it's still part of AWS.

So IMO either we have it like this where we focus on the platform name (I'd go with this):
```
Amazon Web Services
Microsoft Azure
Google Cloud Platform
```

Or have the company name and cloud service:
```
Amazon (AWS)
Microsoft (Azure)
Google (GCP)
```

Or have the have the full and short names:
```
Amazon Web Services (AWS)
Microsoft Azure (Azure)
Google Cloud Platform (GCP)
```

christinaausley: ## AWS

We title things according to how Amazon titles their products. So, [Amazon DynamoDB](https://aws.amazon.com/dynamodb/) versus [AWS Lambda](https://aws.amazon.com/lambda/). I'm not sure why they do it this way, but we do have to write the product naming out exactly as they do.

In essence, _their company name is part of the product name_ -- for example, we would have someone call our product Camunda 7 or Camunda 8, but not just 7 or 8. On the flip side, we don't label our components consistently as "Camunda Tasklist" -- in our docs, we just say "Tasklist".

If someone outside of Camunda was writing an article, however, they might say "Camunda Tasklist" if it isn't already clear they are talking about a Camunda product.

I know this can be super confusing, but there are several unique use cases. In this case, Amazon constantly refers to "Amazon DynamoDB" and "AWS Lambda" rather than just "DynamoDB" and "Lambda".

## Microsoft

Microsoft only refers to this as Azure, with no company title proceeding it, so we name it as such, without the required company name preceding it.

## Google

Google has their own product naming conventions -- Google Cloud, Google Sheets, Gmail, etc. If they refer to a product directly as "Google Cloud" or "Google Sheets", rather than "GCP Sheets" or just "Cloud" then we need to follow this.

aabouzaid: @christinaausley That's the point here, the intention of that section is the cloud service (AWS, GCP, Azure), not any product (the products are in the sub-pages).

Hence, the Amazon section should be `Amazon Web Services` or `Amazon Web Services (AWS)`

christinaausley: See [this PR](https://github.com/camunda/camunda-docs/pull/2589) for maybe a better description as to what I was outlining above. I think we're just misunderstanding each other here?

I think for me to better understand your perspective, are these changes coming from user feedback or an issue I haven't seen? This might help me have a better grasp on the changes and make sure I don't need to adjust anything else if users are confused in the docs.

Regardless, this isn't anything super drastic and we can move forward with this, but I'm trying to understand so I can add appropriate guidance to the WCoE in the case we come across this in the future.

aabouzaid: @christinaausley The issue here I created those sections and now they don't match what I created for and also don't match the reality.

The sections are meant to represent the cloud services (`Amazon Web Services`, `Microsoft Azure`, `Google Cloud Platform`), and inside each section could have products like `Amazon Elastic Kubernetes Service (Amazon EKS)`, `Azure Kubernetes Service (AKS)`, `Google Kubernetes Engine (GKE)`.

When I see links like this, I see them as an inconsistency.
- https://docs.camunda.io/docs/self-managed/setup/deploy/amazon/amazon-eks/ (amazon/amazon-eks => it should be `aws/amazon-eks` because as mentioned the cloud service is called `Amazon Web Services`)
- https://docs.camunda.io/docs/self-managed/setup/deploy/azure/microsoft-aks/ (azure/microsoft-aks)
- https://docs.camunda.io/docs/self-managed/setup/deploy/gcp/google-gke/ (gcp/google-gke)

If you believe it's fine, I will remove that change from my PR.

christinaausley: Your changes are fine @aabouzaid -- I'll just make a note of this PR, in the case we have similar conversations across the docs in the future.

Do let me know if you receive any direct user feedback in the months ahead, and I'm happy to help clarify and clean up the docs anytime 👍

christinaausley (approved):

PR #3711, opened 2024-04-25
chore(action): check dead links with lychee

leiicamundi:
## Description

This pull request implements a comprehensive check of all external links within our documentation. This verification ensures that our clients always have access to the resources referenced in the documentation.

It leverages Lychee within an action (https://github.com/lycheeverse/lychee-action). A report will be generated every day at 3 AM.

For each pull request, the report will be available within the action summary (for example: https://github.com/camunda/camunda-docs/actions/runs/8832095632/attempts/1#summary-24248726137)

There are some false positives to address, but we have slightly less than 7% dead links in our documentation.

## PR Checklist

- [ ] The check links action is documented

### DIFF
--- a/.github/workflows/links.yml
+++ b/.github/workflows/links.yml
@@ -1,7 +1,6 @@
 name: Check external links
 
 on:
-  push:
   workflow_dispatch:
   schedule:
     - cron: "0 3 1 * *"


pepopowitz:
See also a lot of history with link-checking in #1941, #1973, and probably many other issues/PRs referenced by both of them. 

Also, it looks like we're trying to run lychee against the source files, and I think that's likely to fail, due to links that cross documentation instances with shortcodes (e.g. `$optimize$/....` to go from main docs to optimize docs).

.github/workflows/links.yml
```diff
@@ -0,0 +1,40 @@
+name: Links (Fail Fast)
```
pepopowitz, line 1:
```suggestion
name: Check external links
```

blocking -- this is a clearer description of this workflow.

.github/workflows/links.yml
```diff
@@ -0,0 +1,40 @@
...
+
+on:
+  push:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 3 * * *"
```
pepopowitz, line 7:
```suggestion
    - cron: "0 3 1 * *"
```

@christinaausley won't be looking at this more than once a month.

pepopowitz (changes requested):
Thanks for your work here @leiicamundi. I have some general thoughts below, in addition to a couple inline suggestions. I would merge if the inline suggestions are resolved, and item 2 below can be explained/understood.

1. @christinaausley will be the person who takes advantage of the report generated by this workflow. She will plan to schedule about an hour once per month to fix external links, using this report as a guide. As a result, we do not need to run this workflow every day, or we'll be generating a lot of noise that we'll be mostly ignoring. I added a suggestion to update the cron schedule to the first day of each month.
2. One of the first links I clicked on in the report took me to a page that existed. https://github.com/camunda/connector-sdk/releases/tag/0.2.0 appears 10 times in the report as a 404, but that URL resolves for me. Initially I thought it might be redirecting, and maybe lychee wasn't handling redirects properly...but it seems that _that exact URL_ exists, it's not even redirected. Do you have any thoughts about why this might be flagged as a 404? Do you think this is an edge-case false-positive, or do you think it reflects on the overall accuracy of the results?
3. I don't think we should put any more effort into correcting broken links as part of this PR. Broken external links are almost never trivial to resolve. They require time to dig into whether a particular link was moved or deleted, and where it was moved to; and we have a lot less context about these moves than we have about the content moves we make within our own documentation.
   * A good example of this is the link you corrected from Bernd's blog. It appears he moved his blog from medium to a personal domain. It might have been obvious to you from past experience that this move occurred, but for me it wouldn't have been. A google search might have turned up only one result that looked like it was the obvious new location...but it also might not have been obvious, and could have required a lot of spelunking to figure out where he'd moved his blog, which article was intended by the original path, etc.
   * As I mentioned in the previous item, @christinaausley has decided that she'll use this output to guide some regular time-boxed link-correcting time. That would be better than attempting to correct more links as part of this PR.
4. In one of your comments, you mention the idea of checking only content that has changed. This implies that the workflow would run as part of PR checks. In my opinion this is absolutely not what we want, for two reasons: 
   1. PRs being blocked for dead external links would be disruptive during our busiest and most overloaded times, e.g. trying to release a new minor version or handle a CVE. I imagine we'd turn the check off the first time we ran into that situation.
   2. External links are unlikely to go dead during the span of a PR. Human testing while a PR is open is likely to catch any typos in external links. The external links that need validation the most are the oldest ones, not the newest. 
5. Detecting broken links is something we've talked about and thought about a lot on the DX team. The high cost of fixing broken external links, and the fact that they happen completely out of our control, had previously convinced me not to automate detection of these issues. I'm satisfied with the loop of (a) a person finding a broken external link, then (b) opening a PR to correct that broken link.

   In the end it turns out that Christina will plan to use the output of this workflow, to dig away at them a little bit at a time....however it wasn't an urgent priority for us, and it solves a problem that I don't feel needed solving. Please reach out before spending time on work like this in the future, or open an issue for discussion. It would have been good for us to discuss this work ahead of time, so that we could share context, and you could use your time most effectively.

christinaausley:
I've got it on my calendar to start reviewing these links for correction at the beginning of each month for an hour or so, beginning next week! Thanks for flagging this @leiicamundi.

PR #3743, opened 2024-04-30
elasticsearch-exporter: Add link to index templates

falko:
It took me quite a while to dig the index templates out of the source based on the given hint. So, let's add the actual link and also point out that "this module" actually means the exporter source code: https://github.com/camunda/zeebe/tree/main/zeebe/exporters/elasticsearch-exporter/src/main/resources

## PR Checklist

- [ ] My changes require an [Engineering review](https://github.com/camunda/camunda-docs/blob/main/howtos/documentation-guidelines.md#review-process), and I've assigned the Engineerring DRI or delegate.


falko:
@christinaausley I don't have more time to work on this. Primarily, I wanted to report the documentation issue. Could your team please finish the versioned docs? This may also need special attention, when new releases come out.

akeller (changes requested):
This is one of those cases where we need to link to the version-specific URL. @npepinpe do you have any recommendations? I was having trouble finding a stable branch for 8.4 or the right tagged version. 

Similar to my conversation on another PR - https://github.com/camunda/camunda-docs/pull/3795#discussion_r1598660096

falko:
@npepinpe  The stable branches should work, right?
https://github.com/camunda/zeebe/tree/stable/8.5/zeebe/exporters/elasticsearch-exporter/src/main/resources

npepinpe:
Yes, every supported version has a stable branch (e.g. `stable/8.4`, `stable/8.5`). However, once the version is not supported anymore, we remove this. For the docs, we seem to keep a copy of the unsupported versions, so the links would become broken eventually. As an alternative, using the tag would work better, e.g.

- 8.5.0: https://github.com/camunda/zeebe/blob/8.5.0/zeebe/exporters/opensearch-exporter/src/main/resources/zeebe-record-job-template.json
- 8.4.0: https://github.com/camunda/zeebe/blob/8.4.0/exporters/opensearch-exporter/src/main/resources/zeebe-record-job-template.json
- 8.3.0: https://github.com/camunda/zeebe/blob/8.3.0/exporters/opensearch-exporter/src/main/resources/zeebe-record-job-template.json

And so on.

The downside here is:

1. `main` always points to latest, so on every release we need to update those links to point to the new tag.
2. The tags point to specific patch versions. Though, I don't expect us to change these things much between patches, it's not impossible if we introduced something incorrect.

falko:
Then a combination of linking to stable branches for maintained versions and to the last release tag for older unmaintained versions would work.

christinaausley:
> #3795 (comment)

@akeller Unless you have immediate thoughts, let's discuss this (as in both PRs) and the solution in our 1:1 this week? Would be worth a style guide addition.

christinaausley:
I'll make a note to myself to adjust the latest link per each release 👍

akeller (approved):

PR #3756, opened 2024-05-02
chore(dual-region): Improve phrasing around supported envrionments

maxdanilov:
## Description

Improve phrasing around supported environments as per this discussion: https://camunda.slack.com/archives/C0486FPV076/p1713528935569049?thread_ts=1713370090.038159&cid=C0486FPV076

Mainly around removing the point that k8s is the only possible option for dual-region and putting k8s as the recommended installation method only.

## PR Checklist

- [x] My changes require an [Engineering review](https://github.com/camunda/camunda-docs/blob/main/howtos/documentation-guidelines.md#review-process), and I've assigned the Engineerring DRI or delegate.
- [x] My changes require a [technical writer review](https://github.com/camunda/camunda-docs/blob/main/howtos/documentation-guidelines.md#review-process), and I've assigned @christinaausley as a reviewer.

### DIFF
--- a/docs/self-managed/concepts/multi-region/dual-region.md
+++ b/docs/self-managed/concepts/multi-region/dual-region.md
@@ -29,7 +29,7 @@ By contrast, an **active-passive** setup designates one region as the main or ac
 
 :::danger
 
-Running dual-region setups requires developing, testing, and executing of custom [operational procedures](./../../operational-guides/multi-region/dual-region-ops.md) matching your environments. This page outlines key points to consider.
+Running dual-region setups requires developing, testing, and executing custom [operational procedures](./../../operational-guides/multi-region/dual-region-ops.md) matching your environments. This page outlines key points to consider.
 
 :::
 
@@ -37,7 +37,7 @@ Running dual-region setups requires developing, testing, and executing of custom
 
 <DualRegion />
 
-The depicted architecture consists of two regions. For illustrative purposes, we're showing Kubernetes-based installation. Each region houses a Kubernetes cluster with Camunda 8 deployment. Those two Camunda 8 setups are capable to communicate with each other.
+The depicted architecture consists of two regions. For illustrative purposes, we're showing a Kubernetes-based installation. Each region houses a Kubernetes cluster with Camunda 8 deployment. Those two Camunda 8 setups are able to communicate with each other.
 
 One of the regions will be considered **active** and the other **passive**. User traffic must only reach the **active** region. We consider **Region 0** (underlined in green) the active region and **Region 1** the passive region. In this case, user traffic would only go to **Region 0**. **Region 1** would be considered passive and used in case of the loss of the active region. Due to Zeebe's data replication, you can recover from an active region loss by utilizing the passive region without much downtime.
 
@@ -99,10 +99,10 @@ In the event of a total active region loss, the following data will be lost:
     - Tasklist **8.5+**
     - Zeebe **8.5+**
     - Zeebe Gateway **8.5+**
-- For Helm chart installation method, two Kubernetes clusters are required
+- For the Helm chart installation method, two Kubernetes clusters are required
   - OpenShift is not supported
 - Network
-  - The regions (e.g. two Kubernetes clusters) need to be able to connect to each other (for example, via VPC peering)
+  - The regions (for example, two Kubernetes clusters) need to be able to connect to each other (for example, via VPC peering)
     - See an [example implementation](/self-managed/setup/deploy/amazon/amazon-eks/dual-region.md) of two VPC peered Kubernetes clusters based on AWS EKS.
   - Maximum network round trip time (**RTT**) between the regions should not exceed **100 ms**
   - Open ports between the two regions:
@@ -119,8 +119,8 @@ In the event of a total active region loss, the following data will be lost:
 
 ## Limitations
 
-- We are recommending using Kubernetes dual-region setup, with [Camunda Helm chart](/self-managed/setup/install.md) installed into two Kubernetes clusters
-  - Using alternative installation methods (e.g. with docker-compose) is way more complex and is not covered in our documentation
+- We recommend using a Kubernetes dual-region setup, with [Camunda Helm chart](/self-managed/setup/install.md) installed in two Kubernetes clusters.
+  - Using alternative installation methods (for example, with docker-compose) is not covered in our documentation.
 - Looking at the whole Camunda platform, it's **active-passive**, while some key components are active-active.
   - There's always one active and one passive region for serving active user traffic.
   - Serving traffic to both regions will result in a detachment of the components and users potentially observing different data in Operate and Tasklist.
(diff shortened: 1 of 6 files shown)


maxdanilov:
(not asking for Christina's review yet as this PR yet to be reviewed internally)

docs/self-managed/concepts/multi-region/dual-region.md
```diff
@@ -120,8 +121,8 @@ In the event of a total active region loss, the following data will be lost:
 
 ## Limitations
 
-- Camunda 8 must be installed with the [Camunda Helm chart](/self-managed/setup/install.md).
-  - Alternative installation methods (with docker-compose installation, for example) are not supported.
+- We are strongly recommending using Kubernetes dual-region setup, with [Camunda Helm chart](/self-managed/setup/install.md) installed into two Kubernetes clusters
```
MaxTru, line 124: What is the meaning of `strongly recommended` vs `recommended`?

MaxTru (changes requested):
Thanks Maxim! LGTM. 

Can we please look at one more thing?

The page still has a very prominent disclaimer which reduces the readers confidence in Camundas ability to work in dual-region environments.

```
:::danger
- Customers must develop and test [operational procedures](./../../operational-guides/multi-region/dual-region-ops.md) in non-production environments based on the framework steps outlined by Camunda **before applying them in production setups**.
- Before advancing to production go-live, validating these procedures with Camunda is strongly recommended.
- Customers are solely responsible for detecting any regional failures and implementing the necessary [operational procedures](./../../operational-guides/multi-region/dual-region-ops.md).
:::
```

What do we want to say here? Currently what stands out to me is:
* `customers must develop and test` -> what does `must` imply here? What happens if a customer does not test their setups and why are we the ones dictating this?
* `validating these procedures with Camunda is strongly recommended.`  -> what is the intention behind `strongly recommended`? Our architecture is ready for multi-region setups, and experienced customers may just go ahead. If they want, they can validate with us, but why "strongly recommend"?
* ...

I feel that we can bring the message along without being overly cautious. Why not just state something like:

```
:::danger
- Running dual-region setups requires developing, testing, and execution of custom operational procedures matching your environment so to detect and act on regional failures. This page outlines key dimensions to consider.
:::
```

docs/self-managed/concepts/multi-region/dual-region.md
```diff
@@ -101,12 +101,13 @@ In the event of a total active region loss, the following data will be lost:
...
+- Network
+  - The regions (e.g. two Kubernetes clusters) need to be able to connect to each other (for example, via VPC peering)
     - See an [example implementation](/self-managed/setup/deploy/amazon/amazon-eks/dual-region.md) of two VPC peered Kubernetes clusters based on AWS EKS.
-  - Maximum round trip time (RTT) of 100ms between the two Kubernetes clusters
-- Open ports between the two Kubernetes clusters
+  - Maximum network round trip time (**RTT**) between the regions should not exceed **100 ms**
+  - Open ports between the two regions:
   - **9200** for Elasticsearch for Zeebe to push data cross-region
   - **26500** for communication to the Zeebe Gateway from client/workers
   - **26501** for the Zeebe brokers and Zeebe Gateway communication
```
Langleu, line 113: should those be indented one further?

docs/self-managed/concepts/multi-region/dual-region.md
```diff
@@ -138,7 +139,7 @@ In the event of a total active region loss, the following data will be lost:
 - Zeebe cluster scaling is not supported.
 - Web-Modeler is a standalone component and is not covered in this guide.
   - Modeling applications can operate independently outside of the automation clusters.
-- Kubernetes service meshes are currently unsupported, and we advise against their use for the setup.
+- We advise against using Kubernetes service meshes for the dual-region setup.
```
Langleu, line 142: at this point, what's the difference to just omitting it?

A customer can disregard it anyway and we haven't tested it to give a factual reasoning on why we advise against it. It results in more questions and potentially more support requests / forum posts on the topic than just not mentioning it at all.

docs/self-managed/concepts/multi-region/dual-region.md
```diff
@@ -39,7 +39,7 @@ By contrast, an **active-passive** setup designates one region as the main or ac
 
 <DualRegion />
 
-The illustrated architecture consists of two regions. Each region houses a Kubernetes cluster in which Camunda 8 is deployed. Those two Kubernetes clusters are capable of communicating with each other.
+The illustrated architecture consists of two regions. For illustrative purposes, we're showing Kubernetes-based installation. Each region houses a Kubernetes cluster with Camunda 8 deployment. Those two Camunda 8 setups are capable to communicate with each other.
```
Langleu, line 42: Maybe `depicted` instead of having `illustrated` followed by `illustrative`

```
The depicted architecture consists of two regions. For illustrative ...
```

Langleu (changes requested):
small remarks, otherwise looks good.

docs/self-managed/concepts/multi-region/dual-region.md
maxdanilov: Good catch, fixing.

docs/self-managed/concepts/multi-region/dual-region.md
maxdanilov: Good point, fixing it.

docs/self-managed/concepts/multi-region/dual-region.md
maxdanilov: Yep, fixing.

docs/self-managed/concepts/multi-region/dual-region.md
maxdanilov: This is to highlight that complexity of other installation methods is much higher. But fine by me to only leave `recommended`.

I've also searched the docs and we're having multiple cases of the `strongly recommended`, so it's not something totally new to use in this place.

maxdanilov:
@MaxTru thank you for the feedback.

> customers must develop and test -> what does must imply here? What happens if a customer does not test their setups and why are we the ones dictating this?

This is a reminder for the customers to always test their setups in the pre-prod environments before going to prod. We don't want them to go to prod not confident in their own procedures and then claim Camunda is legally responsible for issues inflicted by an incorrect setup untested from their side. It's especially important in multi-region setups due to the high complexity of the operational procedures.
I agree, though, that it also should be applied in general production setups anyway, not only in multi-region.

> validating these procedures with Camunda is strongly recommended. -> what is the intention behind strongly recommended? Our architecture is ready for multi-region setups, and experienced customers may just go ahead. If they want, they can validate with us, but why "strongly recommend"?

We can mark it as optional (`customers can validate their procedures with Camunda`), wdyt? Or removing it altogether.

> I feel that we can bring the message along without being overly cautious. Why not just state something like:

That's a good approach, I'll iterate on it.

MaxTru (approved):

docs/self-managed/setup/deploy/amazon/amazon-eks/dual-region.md
```diff
@@ -9,10 +9,10 @@ description: "Deploy two Amazon Kubernetes (EKS) clusters with Terraform for a p
 import CoreDNSKubeDNS from "./assets/core-dns-kube-dns.svg"
 
 :::warning
-Review our [dual-region concept documentation](./../../../../concepts/multi-region/dual-region.md) before continuing to understand the current limitations and restrictions of this setup, as well as the disclaimer concerning support from Camunda.
+Review our [dual-region concept documentation](./../../../../concepts/multi-region/dual-region.md) before continuing to understand the current limitations and restrictions of this blueprint setup, as well as the disclaimer concerning support from Camunda.
```
MaxTru, line 12: >  as well as the disclaimer concerning support from Camunda.
Which disclaimer would this refer to now?

MaxTru (changes requested):

docs/self-managed/setup/deploy/amazon/amazon-eks/dual-region.md
maxdanilov: removed it, was a leftover, thanks for noticing 👍

MaxTru (dismissed):

maxdanilov:
@christinaausley we're ready for the review 🙇

christinaausley (approved):

PR #3843, opened 2024-05-23
clean up existing links

christinaausley:
## Description

Closes https://github.com/camunda/developer-experience/issues/57.

@akeller -- I'll note this in the upcoming WCoE release. Would you also like a batch of fixes for `8.5`, or every supported version? As far as links that referenced `/docs/next/components/link.md` specifically, I fixed this in every version, but only `next` for `/docs/components/link.md`.

## PR Checklist


akeller (dismissed):
> @akeller -- I'll note this in the upcoming WCoE release. Would you also like a batch of fixes for 8.5, or every supported version? As far as links that referenced /docs/next/components/link.md specifically, I fixed this in every version, but only next for /docs/components/link.md.

Yes please another batch for 8.5 fixes, but no need to spend extra cycles on every supported version at this point.

This did remind me, we should check in the ask-product-management if they can share metrics around the distribution of users across versions. Can I have you follow up on this? This may impact my thoughts around backporting (in general and for these link fixes).

akeller (approved):

PR #3855, opened 2024-05-28
Add Go client repo rename changes to 8.6

akeller:
## Description

Following up to the Go client `camunda/zeebe` -> `camunda/camunda` repo rename, specifically for 8.6+ projects.

## PR Checklist

### DIFF
--- a/docs/apis-tools/go-client/go-get-started.md
+++ b/docs/apis-tools/go-client/go-get-started.md
@@ -37,7 +37,7 @@ go mod init zb-user/zb-example
 2. To use the Zeebe Go client library, run the following:
 
 ```bash
-go get github.com/camunda/zeebe/clients/go/v8@v8.5.1
+go get github.com/camunda/camunda/clients/go/v8@v8.5.1
 ```
 
 This adds the following dependency to your `go.mod`, it should look similar to this:
@@ -47,7 +47,7 @@ module github.com/zb-user/zb-example
 
 go 1.21
 
-require github.com/camunda/zeebe/clients/go/v8@v8.5.1
+require github.com/camunda/camunda/clients/go/v8@v8.5.1
 ```
 
 3. Set the connection settings and client credentials as environment variables in your terminal:
@@ -71,8 +71,8 @@ package main
 import (
 	"context"
 	"fmt"
-	"github.com/camunda/zeebe/clients/go/v8/pkg/zbc"
-	"github.com/camunda/zeebe/clients/go/v8/pkg/pb"
+	"github.com/camunda/camunda/clients/go/v8/pkg/zbc"
+	"github.com/camunda/camunda/clients/go/v8/pkg/pb"
 	"os"
 )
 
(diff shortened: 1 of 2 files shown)


akeller:
@Zelldon, I softened the Announcement a bit, but please make sure I captured the right info.

docs/apis-tools/go-client/go-get-started.md
```diff
@@ -37,7 +37,7 @@ go mod init zb-user/zb-example
 2. To use the Zeebe Go client library, run the following:
 
 ```bash
-go get github.com/camunda/zeebe/clients/go/v8@v8.5.1
+go get github.com/camunda/camunda/clients/go/v8@v8.5.1
```
Zelldon, line 40: ❌ I guess this is for the upcoming version right? So we need to adjust the version as well.

Zelldon, line 40:
```suggestion
go get github.com/camunda/camunda/clients/go/v8@v8.6.0
```

docs/apis-tools/go-client/go-get-started.md
```diff
@@ -47,7 +47,7 @@ module github.com/zb-user/zb-example
 
 go 1.21
 
-require github.com/camunda/zeebe/clients/go/v8@v8.5.1
+require github.com/camunda/camunda/clients/go/v8@v8.5.1
```
Zelldon, line 50:
```suggestion
require github.com/camunda/camunda/clients/go/v8@v8.6.0
```

docs/reference/announcements.md
```diff
@@ -10,6 +10,25 @@ Scheduled release date: 8th of Oct 2024
 
 Scheduled end of maintenance: 14th of April 2026
 
+### Zeebe repo rename impacts Go client
+
+The Camunda 8 Github repository was renamed from `http://github.com/camunda/zeebe` to `http://github.com/camunda/camunda` on May 23, 2024.
+
+Applications that are already compiled, deployed, and running are **not affected**.
+
+Applications importing the Zeebe Go client under development will fail to compile if referencing the old repository after May 23, 2024. The Zeebe Go Client path should reflect the renamed repo as follows:
```
Zelldon, line 19: That is not 100% true. 

Maybe we could explain that due to our repo name we renamed also the go module for consistency reasons. 

This means that new versions are provided are under the new name only. In consequence if they want to use a newer version of the client they have to migrate to the new module name.

Does this make sense?

Zelldon (commented):
Thanks @akeller for preparing this 👍 I think we need to adjust some parts see below.

docs/reference/announcements.md
akeller: Let me see if I'm understanding this correctly. Instead of a date, the driver is the version number? Would this be starting in 8.6.0?

akeller: Adjusted in the latest commit.

I did not include "Applications that are already compiled, deployed, and running are **not affected**." since the rest of the message assume active development.

Zelldon (approved):
Thanks! 

:shipit:

Issue #3749, opened 2024-05-01
Clarify which Camunda versions are supported, unsupported, no longer maintained


ev-codes:
## Problem

As a user, I want to know what versions of Camunda are currently supported so I know how long I can expect to receive support for that version.

It's unclear to me:
- what versions are currently supported ([Announcements](https://docs.camunda.io/docs/next/reference/announcements/) seems to be the best place, but I found this difficult to find, and I'm not sure if this is kept up-to-date)
- what versions are unsupported
- what versions are "no longer actively maintained" - there seems to be some inconsistency between the version selector (screenshot 1) and the callout at the top of the 8.1 home page (screenshot 2)

<img width="1035" alt="Version selector has 'Unsupported versions' section" src="https://github.com/camunda/camunda-docs/assets/7648845/0aaabae8-dae9-4319-8727-6a4f45e8fa68">

<img width="571" alt="Callout at top of 8.1 home page" src="https://github.com/camunda/camunda-docs/assets/7648845/4e11d8a8-6962-48d3-a1dc-d56d037e6438">

## Potential solution(s):

- Update the version selector (screenshot 1) so that 8.1 is considered an unsupported version
- If "no longer actively maintained" and "unsupported" mean the same thing, consider aligning this version
- Adjust page content to make "Supported Versions" more clear to SEO and human browsers. Perhaps a new Supported Versions page would be helpful here?

I'm happy to make small code changes, but it's unclear what the best path forward would be here 😄

akeller:
Thanks for the feedback @ev-codes! There are a few things going on here that are not super obvious.

@pepopowitz is actively archiving the latest unsupported versions of Camunda 8 and Optimize. You can follow the progress [here](https://github.com/camunda/camunda-docs/issues/3731). He may chime in with more info as our DRI of docs infra.

We move unsupported versions to separate static doc sites and then adjust the dropdown. We only index the current version (non-version URL), and we take great pride in carefully monitoring our SEO. 

But it sounds like you are looking for something more - clearer indications about what versions are supported. Oddly enough, this isn't something we did on docs.camunda.org (the dropdown... err drop...up? 😆 is just a list of all versions ever), and I haven't heard much feedback on this experience being confusing.

ev-codes:
@akeller you're welcome, thanks for your quick reply!

It sounds like most of my points are being addressed or are addressed by your extra context, thank you 😄 

[...]

PR #3862, opened 2024-05-30
Add comprehensive table to release notes page

akeller:
## Description

* Adds table to release notes page (index page)
* Lists all supported versions, including scheduled end of maintenance dates 

👇 Oops, typo'd the date. But it's fixed in the PR itself.
image

## PR Checklist

- [x] My changes require a [technical writer review](https://github.com/camunda/camunda-docs/blob/main/howtos/documentation-guidelines.md#review-process), and I've assigned @christinaausley as a reviewer.


akeller:
@ev-codes, this PR wouldn't change the info in the drop down, but would it help address https://github.com/camunda/camunda-docs/issues/3749? 

I could add more context to why we still support 1.3 to this page as well.

akeller:
@MaxTru for 8.2-8.4 I linked to the changelog in camunda-platform, thus I call the link "Camunda Platform." Should I call this "Camunda 8 Core" as well, or would this be misleading because not all core components would be available in the camunda-platform repo? 

I kept the "scheduled" terminology in the table, but WDYT? Should I add "Scheduled release notes" to the column header?

ev-codes:
@akeller yes, this would help address https://github.com/camunda/camunda-docs/issues/3749 !

Thank you for considering and addressing the feedback.

> I could add more context to why we still support 1.3 to this page as well.

I think providing more context about 1.3 support is a good idea!

MaxTru:
>@MaxTru for 8.2-8.4 I linked to the changelog in camunda-platform, thus I call the link "Camunda Platform." Should I call this "Camunda 8 Core" as well, or would this be misleading because not all core components would be available in the camunda-platform repo?

Hi @akeller ,
I think your suggestion makes sense since it is just the reality that we used camunda platform for these releases, which is different to the integrated repo we have now.
Thanks

akeller:
@camunda/tech-writers, during review, please note that not all supported versions were in scope for this so we won't backport beyond the current version at this time. 

Also, I added a brief disclaimer for CC 1.3/Optimize 3.7, which I was going to link to https://github.com/camunda/issues/issues/182, but I need to check internally if this is the most up-to-date issue on this topic. Given that this is only part of the PR (and not the focus) I would like to move forward and follow up with another PR if this one is ready to merge.

conceptualshark (approved):

PR #3874, opened 2024-06-03
Add the tech-writers team to the add-to-project workflow

conceptualshark:

## PR Checklist


.github/workflows/add-to-project.yaml
```diff
@@ -8,7 +8,7 @@ on:
 
 jobs:
   sync-board:
-    if: github.event.requested_reviewer.login == 'akeller' || github.event.requested_reviewer.login == 'christinaausley'
+    if: github.event.requested_reviewer.login == 'akeller' || github.event.requested_reviewer.login == 'christinaausley' || github.event.requested_reviewer.login == 'conceptualshark' || github.event.requested_team.name = 'tech-writers'
```
pepopowitz, line 11:
```suggestion
    if: github.event.requested_reviewer.login == 'akeller' || github.event.requested_reviewer.login == 'christinaausley' || github.event.requested_reviewer.login == 'conceptualshark' || github.event.requested_team.name == 'tech-writers'
```

[This workflow execution](https://github.com/camunda/camunda-docs/actions/runs/9355620374) caught a syntax error here.

pepopowitz (changes requested):
Curious to see how it would handle group-assignments, I opened #3875 to test the changes in this PR. It caught a syntax error for us! 🙌

.github/workflows/add-to-project.yaml
conceptualshark: Should be updated!

.github/workflows/add-to-project.yaml
```diff
@@ -8,7 +8,7 @@ on:
 
 jobs:
   sync-board:
-    if: github.event.requested_reviewer.login == 'akeller' || github.event.requested_reviewer.login == 'christinaausley'
+    if: github.event.requested_reviewer.login == 'akeller' || github.event.requested_reviewer.login == 'christinaausley' || github.event.requested_reviewer.login == 'conceptualshark' || github.event.requested_team.name == 'tech-writers'
```
pepopowitz, line 11: I don't think `github.event.requested_team.name == 'tech-writers'` is doing what we expect it to. I updated my [test PR](https://github.com/camunda/camunda-docs/pull/3875) on this branch, and here is the execution of this workflow when I added `tech-writers` as a reviewer: https://github.com/camunda/camunda-docs/actions/runs/9357583360. The workflow is skipped, because all conditions on this line were false. 

Is there somewhere in the GitHub docs that you saw `github.event.requested_team.name` should be populated? When I look at [these docs for the payload on a `review_requested` event](https://docs.github.com/en/rest/using-the-rest-api/issue-event-types?apiVersion=2022-11-28#review_requested), I see the `requested_reviewer` property but no `requested_team`.

christinaausley: > I think that when a member of a group adds a review, GitHub removes the group as a reviewer and adds the individual instead

This is what I've found while reviewing PRs today 👍

pepopowitz: It looks like GitHub added Christina as a reviewer after https://github.com/camunda/camunda-docs/pull/3884#pullrequestreview-2096981836, but didn't trigger a "review_requested" workflow for it. That's not what I expected, but it seems fine for this workflow! 👍

pepopowitz (approved):
💪 thanks for fighting through this one!

Issue #3857, opened 2024-05-29
Document the upgrade version flow for self-managed


hamza-m-masood:
**Describe the use case:**

It should be clear in our documentation on what version the customer should upgrade to next. We recommend to upgrade the latest patch release of the next minor version. 

For example, if the customer is on `9.2.0` then the customer should upgrade to `9.3.5` assuming that 5 is the latest patch release of the minor version 3.

**Desired outcome and acceptance:**
The above should be clear to our customers in our self-managed documentation.

akeller:
Is this something the Distro team will take on? If you need assistance or collaboration, pull in @conceptualshark and we'll get it added on the DevEx board.

We have the following: 
> When updating from one minor version to the next, you do not need to update to each patch version along the way. However, do not skip minor versions when updating.

But it looks like you want more specifics about patches when, historically, we've focused on minors.

hamza-m-masood:
@akeller  Just a small paragraph in [this page](https://docs.camunda.io/docs/self-managed/setup/upgrade/) about what I mentioned should suffice. I'm happy for @conceptualshark to take this issue. You can add me as reviewer.

PR #3889, opened 2024-06-05
document the upgrade version flow for self-managed

conceptualshark:
## Description

Clarify the recommended upgrade path for Camunda Helm charts, and clean up the Helm page introduction to match this information. 

## PR Checklist


christinaausley (dismissed):

docs/self-managed/setup/upgrade.md
```diff
@@ -5,12 +5,10 @@ sidebar_label: "Upgrade"
...
 
-To upgrade to a more recent version of the Camunda Helm charts, there are certain things you need to keep in mind.
-
-:::caution
-
-Ensure to review the [instructions for a specific version](#version-update-instructions) before starting the actual upgrade.
+To upgrade to a more recent version of the Camunda Helm charts, configuration adjustments may be required. Review the [instructions for specific versions](#version-update-instructions) prior to starting an upgrade.
 
+:::note
+The recommended upgrade path for any given installation is to the **latest patch** release of the **next minor version**.
```
akeller, line 11: I'd like to see a review from the Distribution team, specifically for this note. Specifically because in the [update instructions](https://docs.camunda.io/docs/self-managed/operational-guides/update-guide/introduction/) we mention the following:

> When updating from one minor version to the next, you do not need to update to each patch version along the way. However, do not skip minor versions when updating.

Should we also mention (again), in this PR, that you should not skip minor versions? Or do we think this is clear enough as is?

akeller: @hamza-m-masood, can we get your review here?

jessesimpson36:
I'm reviewing this now

docs/self-managed/setup/upgrade.md
jessesimpson36: Sorry @akeller , you'll just have to put up with a review from me :stuck_out_tongue_winking_eye: . 

I do not find the two statements:

(statement A)
> When updating from one minor version to the next, you do not need to update to each patch version along the way. However, do not skip minor versions when updating.

and

(statement B)
> The recommended upgrade path for any given installation is to the latest patch release of the next minor version.

to be contradictory.  I think the difference between them is that one states what not to do, and the other describes what to do. And I think both are clear that one minor upgrade at a time is the acceptable approach.

However, one thing that I'm realizing is that the "meaning" of a minor version is different now that we have the detached version numbers.  We should re-think whether this statement is intended for people thinking about the camunda version numbers or the helm chart version numbers.  if it's the helm chart version numbers, we need to say 

```suggestion
:::note
The recommended upgrade path for any given installation is to the **latest patch** release of the **next major version**.
```

jessesimpson36:
I'm good with this change. I feel my comment left some points that were out of the scope of this PR, so I will approve now, and if you want to continue discussion on those points before merge, we can at your discretion.

jessesimpson36 (approved):

docs/self-managed/setup/upgrade.md
akeller: @jessesimpson36 appreciate you 🙂 

Great point about the detached version numbers. The only additional thing I would add to your suggestion is to qualify that this is for Helm charts.

@conceptualshark can you have a look at this feedback and adjust the PR?

conceptualshark: @jessesimpson36 @akeller The original request is [here](https://github.com/camunda/camunda-docs/issues/3857) - I was making the assumption it meant camunda versions, but documented on the Helm chart page. Would it be appropriate to update this note to include a specific recommendation for each?

akeller: I'm good with saying this page refers to major versions because we are within the context of Helm. So, we recommend the latest patch release of the next major version (of the Helm chart). I would rather not go into more detail about each. If we get more feedback on this page, we can always adjust it.

akeller (approved):

PR #3949, opened 2024-06-14
docs(sm) - update secrets extraction for version 8.5

upgradingdave:
## Description

There is no longer a `camunda-postgresql` secret as of 8.5. Here's a suggestion to improve this page: 

https://docs.camunda.io/docs/self-managed/setup/upgrade/#secrets-extraction

I also updated the `Note`. Many of the components are customizable, and so this secrets extraction script will be different for each environment. So, rather than adding comments such as ("uncomment if using console"), I suggest we give a script including everything with a Note to remind folks to customize it as needed. 

## PR Checklist

### DIFF
--- a/docs/self-managed/setup/upgrade.md
+++ b/docs/self-managed/setup/upgrade.md
@@ -54,7 +54,7 @@ For a successful upgrade, you first need to extract all secrets that were previo
 
 :::note
 
-You also need to extract all secrets that were generated for Keycloak, since Keycloak is a dependency of Identity.
+The following is meant as a sample script to build upon. Be sure to adapt the following script to your specific configuration. For example, when using an Identity Provider other than Keycloak, the keycloak related secrets listed are unnecessary. It will likely be necessary to change this script to account for the secrets found in your specific environment.
 
 :::
 
--- a/docs/self-managed/setup/upgrade.md
+++ b/docs/self-managed/setup/upgrade.md
@@ -54,7 +54,7 @@ For a successful upgrade, you first need to extract all secrets that were previo
 
 :::note
 
-The following is meant as a sample script to build upon. Be sure to adapt the following script to your specific configuration. For example, when using an Identity Provider other than Keycloak, the keycloak related secrets listed are unnecessary. It will likely be necessary to change this script to account for the secrets found in your specific environment.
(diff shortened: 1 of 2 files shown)


aabouzaid:
I assign this to @jessesimpson36 as Jesse is working on the upgrade topic.

christinaausley:
@jessesimpson36 I'll give this a final approval following your review 👍

jessesimpson36:
I think this is a bit more nuanced. There is a secret named `camunda-postgresql` and a secret named `camunda-identity-postgresql`.  As dave pointed out, these two secrets are used for the same purpose; i.e. to provide credentials for postgresql. However, depending on your kind of installation, you might have one secret or the other. 

The reason for there being two secrets for the same purpose is that one is provided by the bitnami postgresql subchart. However, we also want customers to be able to access an external database (a postgresql server that exists outside of the cluster). For those situations, the subchart is disabled, but we still need a secret to provide credentials to the third party postgresql server. and that secret exists in the camunda chart  and  we disable it whenever the postgresql subchart is enabled.

Perfect world scenario is that these two secrets would be named the same thing and nobody would know the difference, however, I don't think it can be changed now, because if we changed the secret names it'd affect existing users installs.

For 8.7 and newer, we will be mandating that existingSecrets be created and managed outside the helm chart to remove this confusion, and also remove the secrets extraction from the upgrade process.  For 8.6, we will display a warning encouraging users to use existingSecrets.  

We  could favor  whichever secret name is provided for non-external postgresql instances for the time being, which should be `camunda-postgresql`.  

do you want to get on a call with me to discuss/confirm @upgradingdave ?

upgradingdave:
Thanks for the explanation @jessesimpson36. Yeah, maybe we could jump on a quick call and I can update this PR to explain this more clearly. I'll catch up with you on slack, thanks!

christinaausley:
@upgradingdave Also happy to hop on a call and help with any docs changes, just let me know 😄

akeller:
Checking in on open PRs - we'll want to merge this well before our minor release. @upgradingdave what's the status? How can we get this over the finish line 🏁

upgradingdave:
Hi @akeller - I simplified this down to just a small note reminding folks that the secrets extraction script is meant as a sample and will likely need to be customized for each environment. I think it should be good to go now.

jessesimpson36 (approved):

christinaausley (approved):

PR #4018, opened 2024-07-04
docs: document reverse proxy config for job streams

### LINKED ISSUE #3903: Document recommended configuration for worker job streaming with reverse proxies
## Description

We had two incidents now due to long living job streams being closed unexpectedly with 504 errors. This is most prominent with nginx if users configure the parameter `grpc_read_timeout`, as idle workers will have no "response" possibly for a long time. It seems the nginx project has no plans to forward the HTTP/2 pings, or consider those to keep the connection alive.

So the recommendation for now is the following:

- On nginx, set a high `grpc_read_timeout`, e.g. `12h`
- On the client side, set the job worker `streamTimeout` to be a little less than the nginx timeout, e.g. `10h` . It's a good idea anyway to occasionally load balance your long living streams.
- Upstream failures will likely be detected as 502s due to the client sending keep alives (default interval is `45s` , but it's configurable per client), and if upstream is gone then the ping will not be forwarded and the client will receive a 502, and the connection will be closed by nginx.

While this is nginx specific, it can be useful for other reverse proxies (e.g. Traeffik) which may have similar issues.

My only question here is where this kind of documentation should lived - in the self-managed operation guides (but where?), or with the job worker documentation. For this, please consult someone from the DevEx team for their opinions during kickoff.

### Context

- This affects us in SaaS, where the default `grpc_read_timeout` is 10 minutes and 1 second. Idle workers are immediately closed after this with a 504.
- This was related to two support issues:
  - https://jira.camunda.com/browse/SUPPORT-22069
  - https://jira.camunda.com/browse/SUPPORT-22216

npepinpe:
## Description

This PR documents the proper configuration for long living job streams if there is a reverse proxy or load balancer between the user's clients and the Camunda gateway.

closes #3903

## PR Checklist


docs/apis-tools/go-client/job-worker.md
```diff
@@ -106,6 +106,12 @@ To avoid your workers being overloaded with too many jobs, e.g. running out of m
 
 **If streaming is enabled, back pressure applies to both pushing and polling**. You can then use `MaxJobsActive` and `Concurrency` as a way to soft-bound the memory usage of your worker. For example, given a maximum variable payload for a job of 1MB, `MaxJobsActive = 32`, and `Concurrency = 10`, then a single worker could use up to 42MB of memory. You can estimate a worst case scenario using the configured maximum message size, as no job payload will ever exceed this.
 
+#### Proxying
+
+If you're using a reverse proxy or a load balancer between your worker and your gateway, you may need to configure additional parameters to ensure the worker is not killed unexpectedly. If you observe regular 504 timeouts, consider reading [this guide](../../../self-managed/zeebe-deployment/zeebe-gateway/job-streaming).
```
lenaschoenburg, line 111: :question: Is "killed" not misleading? The symptom is more likely that workers get disconnected, right?

docs/self-managed/zeebe-deployment/zeebe-gateway/job-streaming.md
```diff
@@ -0,0 +1,22 @@
...
+
+:::note
+Note that this configuration is _only_ required for reverse proxies which do not support forwarding HTTP/2 keepalive (on either side). See, for example, [this nginx ticket](https://trac.nginx.org/nginx/ticket/1887).
+
+If your proxy supports it, then you don't need to do anything.
+:::
+
+The general recommendation would be to apply the following configuration:
+
+- On your client, set an explicit stream timeout, say, 1h.
```
lenaschoenburg, line 17: :wrench: Mention or link to the config for that

lenaschoenburg (dismissed):
:wrench: I think this needs to added to the versioned docs too, to 8.4 at least I think.

docs/apis-tools/go-client/job-worker.md
npepinpe: True, it's more that the request is failed unexpectedly :+1:

docs/self-managed/zeebe-deployment/zeebe-gateway/job-streaming.md
npepinpe: It's really client specific though, isn't it? :thinking: I could link to each official client I suppose.

lenaschoenburg: I'd link to the java client just to have an example.

docs/self-managed/zeebe-deployment/zeebe-gateway/job-streaming.md
```diff
@@ -0,0 +1,23 @@
+---
+id: job-streaming
+title: "Job streaming"
+sidebar_label: "Job streaming"
+description: "Streaming job workers is expected to be long-lived to cut down on the latency overhead involved with re-creating a stream and propagating this throughout the cluster."
+---
+
+Streaming job workers is expected to be long-lived to cut down on the latency overhead involved with re-creating a stream and propagating this throughout the cluster. This may require special configuration, especially if you're using a reverse proxy in front of your gateway. Typically, this will affect you in the form of HTTP 504 (gateway timeout) being returned to your job streaming worker at regular intervals.
```
christinaausley, line 8: What do we mean by "Streaming job workers is expected to be long-lived" -- as in, we expect to support this longterm?

lenaschoenburg: No, it's that job workers are expected to _run_ for a long time, i.e. not just connect once for a single job and then disappear again. It's better when they are long-lived because the initial setup, i.e. establishing connections, registering it with the gateways and brokers etc. is relatively expensive and shouldn't be done very often.

conceptualshark (commented):
@npepinpe I pushed a few updates, but I think this is in good shape otherwise. I assume it should be backported to previous versions as well?

npepinpe:
Looks good :+1: Hm, there's no job streaming before 8.4, so we could backport to 8.5 and 8.4.

conceptualshark (approved):
Went ahead and backported to the requested versions - all looks good on my end, so approving here unless there's any other engineering reviews waiting.

PR #4025, opened 2024-07-10
chore(cache): Implement a cache for docusaurus build

andromaqui:
## Description

This pr adds a cache for the preview environment build. The deploy-preview-env action will be:
- Caching every subsequent run after the first and reducing the build down to 6 minutes 
To achieve this, an action which [caches docusaurus builds](https://github.com/docuactions/cache/blob/main/action.yml) has been used. 

####  A cache gets created and stored for subsequent runs

- [commit](https://github.com/camunda/camunda-docs/pull/4025/commits/d2223b385423eca9f910c60bf5e0b160ff63c439)
- [logs from the build](https://github.com/camunda/camunda-docs/actions/runs/9990388237/job/27610977359#step:31:6) ( I had to change sth so the cache got actually  created in the next run)
- [next run](https://github.com/camunda/camunda-docs/actions/runs/9990705051/job/27611949419) got reduced to 6 minutes & commit for the run can be found [here](https://github.com/camunda/camunda-docs/pull/4025/commits/8951e2026e8f3db02b7e8a5ffd0c271017accc3e)

## PR Checklist


.github/workflows/preview-env-deploy.yml
```diff
@@ -2,6 +2,9 @@ name: preview-env-deploy
 on:
   pull_request:
     types: [labeled, synchronize]
+  push:
+    branches:
+      - master
```
cmur2, line 7: I'd consider moving this logic to a dedicated workflow with a clearly distinctive name like `preview-env-build-cache` to avoid mixing concerns (this workflow doesn't deploy `main`) as that could lead to mistakes later:

All other `preview-env-deploy` workflows that the Infra team maintains are not ran on `main` so the logic is surprising for the reader and someone editing this file might e.g. forget the `if: github.ref != 'refs/heads/main'` condition on a newly added step.

.github/workflows/preview-env-deploy.yml
```diff
@@ -28,6 +32,18 @@ jobs:
...
 
+      - name: Restore docusaurus cache
+        if: github.ref != 'refs/heads/main'
+        uses: actions/cache/restore@v4
+        with:
+          path: |
+            ${{ github.workspace }}/.docusaurus
+            ${{ github.workspace }}/**/.cache
+          key: |
+            ${{ runner.os }}-docusaurus-preview-env-${{ hashFiles('**/package-lock.json', '**/npm-shrinkwrap.json', '**/yarn.lock', '**/pnpm-lock.yaml') }}
```
cmur2, line 43: I'm not super familiar with this docusaurus project, how did you come up with the part `${{ hashFiles('**/package-lock.json', '**/npm-shrinkwrap.json', '**/yarn.lock', '**/pnpm-lock.yaml') }}` and are all of those needed?

.github/workflows/preview-env-deploy.yml
```diff
@@ -28,6 +32,18 @@ jobs:
             secret/data/products/camunda-docs/ci/preview-environment PREVIEW_ENV_BUCKET_NAME;
             secret/data/products/camunda-docs/ci/preview-environment PREVIEW_ENV_GCLOUD_SA_KEY;
 
+      - name: Restore docusaurus cache
+        if: github.ref != 'refs/heads/main'
+        uses: actions/cache/restore@v4
+        with:
+          path: |
+            ${{ github.workspace }}/.docusaurus
```
cmur2, line 40: Are files in this cache dependent on the base URL? If the base URL on master is going to be different that those for individual PRs, will there still be a speed-up?

.github/workflows/preview-env-deploy.yml
andromaqui: This is actually a very good point.

.github/workflows/preview-env-deploy.yml
andromaqui: yeah you are right actually I didnt think of that, let me verify this

.github/workflows/preview-env-deploy.yml
andromaqui: The inspiration this mechanism came from the already existent action which [caches docusaurus builds](https://github.com/docuactions/cache/blob/main/action.yml). I am under the impression that  these files dont get cached but are used in order to generate a unique key for the hash

cmur2 (approved):
> I think the only thing we could implement is an improvement on the build time in every subsequent push after the first one.

Very correct conclusion. Thanks for the lean implementation 👍🏼

PR #4031, opened 2024-07-15
Add ES exporter to application.yaml

akeller:
## Description

Open discussion - https://camunda.slack.com/archives/C074T8BTTM2/p1721078565460779?thread_ts=1720034668.231519&cid=C074T8BTTM2

Takes the following instructions, packaged with the release artifacts, and makes them more consumable:

```
How to run
==========

Prerequisites:
1. Download Elasticsearch 8.9.2 (from https://www.elastic.co/downloads/elasticsearch)
2. For non-production cases, disable Elasticsearch's security packages by setting the xpack.security.* configuration options to false in ELASTICSEARCH_HOME/config/elasticsearch.yml
3. Start Elasticsearch by running ELASTICSEARCH_HOME/bin/elasticsearch (or ELASTICSEARCH_HOME\bin\elasticsearch.bat on Windows)

Start Camunda:
1. Run bin/camunda (or bin\camunda.bat on Windows)
```

## PR Checklist


akeller:
For reviewers (@christinaausley @conceptualshark @pepopowitz)- I'm looking for the following, feedback-wise:

* Is it overkill to include the entire `application.yaml` because of spacing/nesting purposes? Do you have a suggestion for how to make these easy to explain & maintain if the monorepo (source of truth) doesn't include the exporter by default?
  * If this is too challenging to explain & maintain, I'd like to push back on the idea that this _should_ be a default and users who want to run Zeebe by itself should remove the exporter.  
* Does breaking up Step 1 (installation step) by technology make sense? ES needs to be running for Camunda to start successfully anyway. 
  * Implementing Step 1 as a reusable component seems to cause the headers to not show up in the right-hand sidebar, but the header links appear to work... was this just weird on my computer, or is this the case for you too? Will this be confusing? 
  
  
Please set aside some time to run this on your local machine or work together to see how it runs. Let's set a deadline to do this by next Friday (July 26). This is a high priority (part of the getting started experience), but not urgent. If the urgency increases, I'll move forward as-is and open a new PR based on your feedback.

docs/guides/react-components/install-plain-java.md
```diff
@@ -8,8 +8,169 @@ For this installation, you must have:
...
+      elasticsearch:
+        className: io.camunda.zeebe.exporter.ElasticsearchExporter
+        args:
+          url: http://localhost:9200
+          index:
+            prefix: zeebe-record
+```
+
+:::note
+Spacing is important! Ensure four spaces and proper nesting of configuration.
```
pepopowitz, line 33:
```suggestion
Spacing is important! Indent the `exporters` element four spaces to properly nest the configuration.
```

"Ensure four spaces" threw me off because the example uses two-space indentation....but I think you meant that the `exporters` element itself is indented four spaces.

docs/guides/react-components/install-plain-java.md
```diff
@@ -8,8 +8,169 @@ For this installation, you must have:
...
+
+<details>
+<summary>Still need help?</summary>
```
pepopowitz, line 36: I do like the use of details/summary to hide the full application.yaml.

pepopowitz (changes requested):
Most of my feedback is non-blocking, but one in particular (marked "blocking") is preventing me from moving forward to test this out. I am not clear on which specific settings I need to change for elasticsearch.

docs/guides/react-components/install-plain-java.md
```diff
@@ -8,8 +8,169 @@ For this installation, you must have:
 - OpenJDK 21+ locally installed
 - Camunda `8.6.0-alpha2` or later
 
+### Download and configure Elasticsearch
+
+1. Download [Elasticsearch 8.9.2](https://www.elastic.co/downloads/past-releases/elasticsearch-8-9-2).
+2. Disable Elasticsearch's security packages by setting the `xpack.security.*` configuration options to `false` in `ELASTICSEARCH_HOME/config/elasticsearch.yml`.
```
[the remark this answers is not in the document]
akeller: I added this to the bottom of the file that was entirely commented out. No additional config needed, just the single line at the bottom of the file, which for me was line 84:

`xpack.security.*: false`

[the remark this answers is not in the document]
conceptualshark: The wildcard security config didn't work for me; it always seems to fail with `java.lang.IllegalArgumentException: unknown setting [xpack.security.*] did you mean [xpack.security.user]?`. 

To reproduce: 
I wasn't sure where to put `xpack.security.*: false` initially, and also ended up with entering what you have on like 84. Attempting to run this fresh install of elasticsearch then auto-generates a series of security config options which don't appear to care that `xpack.security.*: false` was set. It'd always fail on the wildcard line with or without the generated config.

Ultimately, I started with a clean install again, and ran elasticsearch for the first time to autogenerate the config. I then set everything generated manually to false, and removed the extra lines, and ended up with this:
```yaml
xpack.security.enabled: false

xpack.security.enrollment.enabled: false

# Enable encryption for HTTP API client connections, such as Kibana, Logstash, and Agents
xpack.security.http.ssl:
  enabled: false

# Enable encryption and mutual authentication between cluster nodes
xpack.security.transport.ssl:
  enabled: false 
  ```
  It appears to be running after that point. I tested a few of these lines individually to see the minimum config needed, and I think all of them might need to be set.

File: docs/guides/react-components/install-plain-java.md (hunk 2 above)
Comment (conceptualshark) [reply docs/guides/react-components/install-plain-java.md:36]: Agreed - if it seems like the spacing might be an issue, stating that they can replace the entire contents with the provided file could help.

File: docs/guides/react-components/install-plain-java.md (hunk 1 above)
Comment (pepopowitz) [reply docs/guides/react-components/install-plain-java.md:33]: If there is a consistent sibling next to it, then I think that makes sense! But I don't have a strong opinion either way.

File: docs/guides/react-components/install-plain-java.md [hunk 4]
```diff
@@ -8,8 +8,169 @@ For this installation, you must have:
...
+3. Start Elasticsearch by running `ELASTICSEARCH_HOME/bin/elasticsearch` (or `ELASTICSEARCH_HOME\bin\elasticsearch.bat` on Windows).
+
+### Download and configure Camunda
+
 1. Download the [latest release artifact](https://github.com/camunda/camunda/releases), starting with [8.6.0-alpha2](https://github.com/camunda/camunda/releases/tag/8.6.0-alpha2).
```
conceptualshark, line 19: Running alpha2 seemed to result in a version mismatch for me, but running alpha3 went fine:

`io.camunda.operate.schema.SchemaStartup - Schema startup failed: Error(s) in index schema: Newer version(s) for operation (8.4.0) already exists: [8.4.1]`

docs/guides/react-components/install-plain-java.md
```diff
@@ -8,8 +8,169 @@ For this installation, you must have:
...
-2. Download [Elasticsearch 8.9.2](https://www.elastic.co/downloads/past-releases/elasticsearch-8-9-2).
-3. For non-production cases, disable Elasticsearch's security packages by setting the `xpack.security.*` configuration options to `false` in `ELASTICSEARCH_HOME/config/elasticsearch.yml`.
-4. Start Elasticsearch by running `ELASTICSEARCH_HOME/bin/elasticsearch` (or `ELASTICSEARCH_HOME\bin\elasticsearch.bat` on Windows).
-5. To start Camunda, run `bin/camunda` (or `bin\camunda.bat` on Windows).
+2. Modify the `application.yaml` to include the Elasticsearch exporter by adding the following to line 91:
```
conceptualshark, line 20: We should format this the same way we do for finding and editing the elasticsearch config, and specify that this is in the `config` folder.

docs/guides/react-components/install-plain-java.md
```diff
@@ -8,8 +8,169 @@ For this installation, you must have:
 - OpenJDK 21+ locally installed
 - Camunda `8.6.0-alpha2` or later
```
conceptualshark, line 9: I'm not sure if we need Camunda as a prereq if we're going to direct them to install it below - this might be fine with just OpenJDK as the requirement.

docs/guides/react-components/install-plain-java.md
conceptualshark, line 14: Where ever we land with the install steps, a version of this line that reads a little more clearly could look like the following: 
```suggestion
2. Navigate to the directory where you installed Elasticsearch, and open `/config/elasticsearch.yaml`. Add the line `xpack.security.*: false` to the bottom of the configuration to disable Elasticsearch's security packages.
```

docs/guides/react-components/install-plain-java.md
akeller: I prefer to be extra explicit until this is GA'd (and not only available in specific alphas).

docs/guides/react-components/install-plain-java.md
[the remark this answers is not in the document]
akeller: @conceptualshark just to verify, you needed to add extra security config was with alpha2, alpha3, or both?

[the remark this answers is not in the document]
conceptualshark: @akeller This was prior to any Camunda install, so only when attempting the "Download and configure elasticsearch" section.

1. Download elasticsearch
2. Edit `config/elasticsearch.yml` to include `xpack.security.*: false`
3. Running `bin/elasticsearch` results in `java.lang.IllegalArgumentException: unknown setting [xpack.security.*] did you mean [xpack.security.user]?`.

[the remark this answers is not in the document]
akeller: Oh! So just to stand up Elasticsearch! I misunderstood. I didn't run into this.

docs/guides/react-components/install-plain-java.md
akeller: @romansmirnov thoughts on this?

romansmirnov: Assumption: The same ES instance has been used, but first, alpha3 was started, creating an index in version 8.4.1, and then alpha2 was started, expecting the index to be in version 8.4.0.

Can you please confirm my understanding? If so, I don't recommend using the ES instance and switching from a new C8 version to an old C8 version.

docs/guides/react-components/install-plain-java.md
romansmirnov: In my opinion, it should still contain the statement that it is only meant for non-production use cases. If they go in production, users should consider to configure security in ES (whatever it looks like).

docs/guides/react-components/install-plain-java.md
```diff
@@ -8,8 +8,169 @@ For this installation, you must have:
...
 
+### Download and configure Elasticsearch
+
+1. Download [Elasticsearch 8.9.2](https://www.elastic.co/downloads/past-releases/elasticsearch-8-9-2).
+2. Disable Elasticsearch's security packages by setting the `xpack.security.*` configuration options to `false` in `ELASTICSEARCH_HOME/config/elasticsearch.yml`.
+3. Start Elasticsearch by running `ELASTICSEARCH_HOME/bin/elasticsearch` (or `ELASTICSEARCH_HOME\bin\elasticsearch.bat` on Windows).
```
pepopowitz, line 15:
```suggestion
3. Start Elasticsearch by running `ELASTICSEARCH_HOME/bin/elasticsearch` (or `ELASTICSEARCH_HOME\bin\elasticsearch.bat` on Windows).

Confirm Elasticsearch is running by visiting `http://localhost:9200` in a browser. If the response doesn't include version information formatted as JSON, you will need to troubleshoot your installation.
```

This should help us catch failed Elasticsearch installations proactively. 

And if we want to add an example, this is what mine looks like: 

```json
{
    "name": "Stevens-MBP.attlocal.net",
    "cluster_name": "elasticsearch",
    "cluster_uuid": "kl8KQQnLQlqkoS1gvbai_A",
    "version": {
        "number": "8.9.2",
        "build_flavor": "default",
        "build_type": "tar",
        "build_hash": "e8179018838f55b8820685f92e245abef3bddc0f",
        "build_date": "2023-08-31T02:43:14.210479707Z",
        "build_snapshot": false,
        "lucene_version": "9.7.0",
        "minimum_wire_compatibility_version": "7.17.0",
        "minimum_index_compatibility_version": "7.0.0"
    },
    "tagline": "You Know, for Search"
}
```

docs/guides/react-components/install-plain-java.md
```diff
@@ -8,8 +8,169 @@ For this installation, you must have:
...
+
+```
+    exporters:
+      elasticsearch:
+        className: io.camunda.zeebe.exporter.ElasticsearchExporter
+        args:
+          url: http://localhost:9200
+          index:
+            prefix: zeebe-record
+```
```
pepopowitz, line 30:
~~~suggestion
2. Modify the `application.yaml` to include the Elasticsearch exporter by adding the following as a child of the `zeebe`/`broker` configuration element:

```
zeebe:
  broker:
    ...
    exporters:
      elasticsearch:
        className: io.camunda.zeebe.exporter.ElasticsearchExporter
        args:
          url: http://localhost:9200
          index:
            prefix: zeebe-record
```
~~~


My alpha2 download had different line numbers than yours (not sure why). Specifying the _logical_ placement of this configuration element should avoid problems with that drift, and help avoid mis-indentation and mis-placement.

docs/guides/react-components/install-plain-java.md
pepopowitz: After further review, I think this can be aided by my comment on the previous block -- specifying the _logical_ location of the configuration block, instead of physical.

docs/guides/react-components/install-plain-java.md
```diff
@@ -8,8 +8,169 @@ For this installation, you must have:
...
+
+</details>
+
+:::
+
+Save the file. Without performing this step, no data will be visible in Operate or Tasklist.
+
+3. To start Camunda, run `bin/camunda` (or `bin\camunda.bat` on Windows).
+
```
pepopowitz, line 173:
```suggestion

It may take a few minutes for startup to complete. When the message `Started StandaloneCamunda in ___ seconds` is displayed, the application is ready to use.

```

Non-blocking, but I was personally not sure when I should expect it to start responding. 

This is based on these messages being the last output before I was able to finally hit mine: 

```
2024-07-19 15:16:56.786 [] [main] [] INFO 
      io.camunda.application.StandaloneCamunda - Started StandaloneCamunda in 196.726 seconds (process running for 198.442)
2024-07-19 15:16:56.797 [] [main] [] INFO 
      io.camunda.operate.schema.migration.SchemaMigration - SchemaMigration finished.
```

docs/guides/react-components/install-plain-java.md
```diff
@@ -8,8 +8,169 @@ For this installation, you must have:
...
+:::tip
+Operate can be found at `http://localhost:8080/` and Tasklist can be found at `http://localhost:8080/tasklist`.
+:::
```
pepopowitz, line 176: I feel like there is a default user/password I can use to log in at this point, but I can't remember what it is or where to find it! We should definitely mention it here.

pepopowitz (changes requested):
I took a second pass. As I was finishing, I noticed in Slack that I may have been asked to use alpha3, but I went through with alpha2. Oops? This might be a worthless review.

In the end, I can see a login screen for operate at localhost:8080 (though I don't know what credentials to log in with). I see localhost:8080/tasklist trying to serve _something_, but it's basically a blank gray page. I am wondering if it would serve an actual tasklist if I had logged into localhost:8080.

docs/guides/react-components/install-plain-java.md
akeller: demo/demo

docs/guides/react-components/install-plain-java.md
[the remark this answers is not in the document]
akeller: Find them is still super non-trivial IMO, but I agree this helps a bit.

docs/guides/react-components/install-plain-java.md
akeller: Since this is a get started guide, I hope no one uses this in production, so I was also hoping I could be brief here and remove it. We can add this back, but I would also want to add a link to our production setup guides.

docs/guides/react-components/install-plain-java.md
```diff
@@ -8,8 +8,174 @@ For this installation, you must have:
 - OpenJDK 21+ locally installed
 - Camunda `8.6.0-alpha2` or later
 
+### Download and configure Elasticsearch
+
+1. Download [Elasticsearch 8.9.2](https://www.elastic.co/downloads/past-releases/elasticsearch-8-9-2).
+2. Disable Elasticsearch's security packages by setting the `xpack.security.*` configuration options to `false` in `ELASTICSEARCH_HOME/config/elasticsearch.yml`.
+3. Start Elasticsearch by running `ELASTICSEARCH_HOME/bin/elasticsearch` (or `ELASTICSEARCH_HOME\bin\elasticsearch.bat` on Windows).
```
christinaausley, line 15: > 3. Start Elasticsearch by running `ELASTICSEARCH_HOME/bin/elasticsearch` (or `ELASTICSEARCH_HOME\bin\elasticsearch.bat` on Windows).

I assume `ELASTICSEARCH_HOME/bin/elasticsearch` works for Mac/Linux, and Windows is the only exception worth noting here?

conceptualshark:
I ran through this whole process in UTM to isolate it from the build errors I was having on my machine and: success!

I changed the ES security config `xpack.security.enabled: false` _prior_ to running ES for the first time, and it worked fine. I also no longer got schema errors running alpha2, so can confirm that worked on a fresh/isolated install as well. 

My assumption is testing the C8Run script caused some versioning/config issues there, which might just be good to have in mind if we end up with folks attempting multiple methods of installing/running.

christinaausley:
> Is it overkill to include the entire application.yaml because of spacing/nesting purposes?

I don't love this, because users will only scroll so far down a page in general, and it can make the page a bit messy. Are we able to link to it, or include it in something like a <details> element or collapsible/expandable code snippet? I've seen this a few times across our docs.

If not, is there a specific reason why users aren't able to find these details in elements like these already?

akeller:
Hi All - thanks for the reviews and feedback. I will do one more pass today to close this out.

Related, we _will_ swap in C8SM Run to be our single option for getting started with SM. If you have more input on that part of the conversation, please go to [#e2e-docs-experience-getting-started-guides](https://camunda.slack.com/archives/C077J92G8LX). Regardless, it's important if we do ever need to have both of these, we can easily pull these changes back in and offer parallel paths.

docs/guides/react-components/install-plain-java.md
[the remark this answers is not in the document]
akeller: Added to an incoming commit.

docs/guides/react-components/install-plain-java.md
akeller: Added a warning admonition in an incoming commit.

akeller: @conceptualshark I also modified your suggestion slightly, as I _think_ we only need the one line of config as you mentioned in a recent comment on this PR.

docs/guides/react-components/install-plain-java.md
akeller: I'll re-request a review and see where we are at with this, but resolving this convo for now.

docs/guides/react-components/install-plain-java.md
akeller: @conceptualshark I don't know if we should expect people using the same ES instance (purposefully or accidentally). Can I give you a follow up here to maybe add a small troubleshooting section or think more about this experience for new users?

conceptualshark: The only thing I can think of that would have impacted this is having started C8SM Run on this machine (it was not running at the time). When I tested this PR, I ran ES + alpha2, failed with the above error, stopped both, then reinstalled and ran ES + alpha3, so I am not sure why an index (or even ES instance) would have been maintained. 

Unfortunately, I can't add anything for resolving it, because it still doesn't work on my machine (I had to run it in a VM).

conceptualshark (commented):
I think this is in good shape! If we're eager to merge, my suggestion isn't blocking.

conceptualshark (approved):
🎉

Issue #4036, opened 2024-07-17
Add spring support policy for spring sdk


felix-mueller:
Add spring support policy for spring sdk here: https://docs.camunda.io/docs/next/apis-tools/spring-zeebe-sdk/getting-started/#version-compatibility

https://camunda.slack.com/archives/C037W9NMATG/p1721203971434809

"For new minor releases we usually automatically update to new minor releases of our dependencies when they are released. This general rule also applies to spring. Thus, on main (8.6) we are already on spring boot 3.3.1"

christinaausley:
This has been merged -- closing.

PR #4052, opened 2024-07-19
add spring support policy

christinaausley:
## Description

Related to https://github.com/camunda/camunda-docs/issues/4036.

## PR Checklist


christinaausley:
@felix-mueller @megglos I assume this should also be backported to `8.5`? Ideally I'd like a statement/example that we don't need to change for every version of the docs. Would it be sufficient to just say "For new minor releases, Camunda typically automatically updates to new minor releases of our dependencies when they are released. This general rule also applies to Spring."?

If not, I may revisit some of the phrasing here, unless we want to default to a specific version comparison with `8.5`, for example.

akeller (commented):
We reference [clients](https://docs.camunda.io/docs/next/reference/supported-environments/#clients) on the supported environment page, but don't include the Spring SDK. 

I don't love the phrasing here and have concerns similar to @christinaausley's comment about changing this for every new version. We can be more definitive here and update the table per minor version but also consider moving this content to the supported environment page, a page we already know changes for each minor.

felix-mueller:
@akeller @christinaausley - from my perspective we can be more generic, my message was just an example for 8.6.
we can also add spring sdk to supported environments and link back to the version matrix (similar to helm cli). 
let's keep the statement itself spring specific for now since it might have more implications and i dont see the need to go through it right now.

christinaausley:
@felix-mueller I think I'm following correctly here -- let me know based on my latest changes.

felix-mueller:
@christinaausley lgtm

felix-mueller (approved):

christinaausley:
@akeller Can I get one final review from you here? 😄

akeller (approved):

PR #4064, opened 2024-07-25
clarify possible instances of zeebe rollback

christinaausley:
## Description

Closes https://github.com/camunda/developer-experience/issues/345.

Clarifying unique instances of rollback.

## PR Checklist

### DIFF
--- a/docs/self-managed/zeebe-deployment/operations/update-zeebe.md
+++ b/docs/self-managed/zeebe-deployment/operations/update-zeebe.md
@@ -160,11 +160,11 @@ Because the update is no longer rolling and all brokers are shut down at the sam
 If updated brokers log the error message `Failed to install partition` and do not become healthy, look for more details to understand if this is caused by the rolling update.
 
 If the error is caused by `Cannot upgrade to or from a pre-release version`, Zeebe detected that either the version you started from or the version you updated to is a pre-release version.
+
 This is not permitted because pre-release versions such as alpha releases are considered unstable and do not guarantee compatibility with any other version.
-If you attempted to update from a normal version to a pre-release version, you can roll back to the previous normal version.
 
 :::note
-While in most instances version rollbacks are not supported, in this particular use case a version rollback is possible.
+If you attempted to update from a minor release to a pre-release or alpha version, it is possible to roll back to the previous version of Zeebe. Note that version rollbacks are not supported in most other instances.
 :::
 
 If the log message includes `Snapshot is not compatible with current version`, the rolling update failed and manual recovery is required.
@@ -196,11 +196,7 @@ In that case, it is caused by updated brokers sharing snapshots with not yet upd
 This should resolve automatically once the broker is updated.
 :::
 
-If this persists, you may want to [force the update](#rolling-update-is-not-completing).
-
-:::note
-In this specific case, a rollback is possible because users may restart Zeebe with the "skipped" minor version. [Backup and restore](/self-managed/operational-guides/backup-restore/backup-and-restore.md) is possible, but results in partial data loss of everything following the backup. Therefore, a rollback is preferable in this use case.
-:::
+If this persists, you can [force the update](#rolling-update-is-not-completing). Alternatively, it is possible to restart Zeebe with the "skipped" minor version. Note that version rollbacks are not supported in most other instances.
 
 ## Offline update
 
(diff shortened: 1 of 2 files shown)


docs/self-managed/zeebe-deployment/operations/update-zeebe.md
```diff
@@ -194,6 +198,10 @@ This should resolve automatically once the broker is updated.
 
 If this persists, you may want to [force the update](#rolling-update-is-not-completing).
 
+:::note
+In this specific case, a rollback is possible because users may restart Zeebe with the "skipped" minor version. [Backup and restore](/self-managed/operational-guides/backup-restore/backup-and-restore.md) is possible, but results in partial data loss of everything following the backup. Therefore, a rollback is preferable in this use case.
```
lenaschoenburg, line 202: :thought_balloon: I think including backup and restore here adds to the confusion, especially because _taking_ a backup is not possible in this state. Restoring from a backup is _always_ possible, no need to mention it here.

docs/self-managed/zeebe-deployment/operations/update-zeebe.md
```diff
@@ -163,6 +163,10 @@ If the error is caused by `Cannot upgrade to or from a pre-release version`, Zee
 This is not permitted because pre-release versions such as alpha releases are considered unstable and do not guarantee compatibility with any other version.
 If you attempted to update from a normal version to a pre-release version, you can roll back to the previous normal version.
 
+:::note
+While in most instances version rollbacks are not supported, in this particular use case a version rollback is possible.
```
lenaschoenburg, line 167:
```suggestion
While in most instances version rollbacks are not supported, in this particular use case a version rollback of Zeebe is possible because the "new" version never became active, thanks to the safety check.
```

conceptualshark, line 167: Could we condense these two lines? Either as a single note, or as one line in the `Cannot upgrade to or from a pre-release version` paragraph, to make it clear this is referring to that error only. 

`If you attempted to update from a minor release to a pre-release or alpha version, it is possible to roll back to the previous version of Zeebe. Note that version rollbacks are not supported in most other instances.`

conceptualshark: I'm not sure the extra explanation adds a lot here. We don't otherwise talk about what the safety checks or doing, and can possibly introduce new points of confusion re: "new" vs "old" instead of clarifying. If we really think our developers need this info, I'd suggest something like: 

`..a version rollback of Zeebe is possible, as the included safety checks prevented the pre-release version from becoming active.`

docs/self-managed/zeebe-deployment/operations/update-zeebe.md
conceptualshark, line 202: Similar to above, is it beneficial to combine this section?
```suggestion

If this persists, you can [force the update](#rolling-update-is-not-completing). Alternatively, it is possible to restart Zeebe with the "skipped" minor version.  Note that version rollbacks are not supported in most other instances.
```

docs/self-managed/zeebe-deployment/operations/update-zeebe.md
lenaschoenburg: I'd keep the addition of "Zeebe" then because rollback of the other applications might not be possible.

docs/self-managed/zeebe-deployment/operations/update-zeebe.md
christinaausley: Updated in latest commit 👍

docs/self-managed/zeebe-deployment/operations/update-zeebe.md
christinaausley: Updated in latest commit -- let me know what you think.

christinaausley: Updated in latest commit.

docs/self-managed/zeebe-deployment/operations/update-zeebe.md
christinaausley: Adjusted in latest commit.

conceptualshark (approved):
👍 This looks clear to me!

Issue #3047, opened 2023-12-13
Example values in config docs tables should match the YAML snippet casing


akeller:
In the Zeebe [broker](https://docs.camunda.io/docs/self-managed/zeebe-deployment/configuration/broker-config/) and [gateway](https://docs.camunda.io/docs/self-managed/zeebe-deployment/configuration/gateway-config/) docs, we provide a table and YAML snippet to describe the configuration. Particularly in cases with a boolean value, the example value is capitalized in the table and lowercase in the YAML snippet. 

In both the table and the YAML snippet, boolean values should be lowercase. 

This change should be backported or added to all versions where these config docs exist.

PR #4118, opened 2024-08-06
true/false in config docs

akeller:
## Description

Closes https://github.com/camunda/camunda-docs/issues/3047

## PR Checklist


mesellings (approved):
I've reviewed the changes, and all the default value changes to lower case look good to me.

Non-blocking: I know it's not part of this PR scope, but I did notice the following:

* For the descriptions, we could standardise and make them less passive. For example, 'When true, Importer module is enabled" > "Enable the Importer module".
* The use of "whether" could be changed to use "if" as per the Camunda writing style guide: https://camunda.com/brand/writing-style-guide/

akeller:
@mesellings, thanks for the review! Can you add an issue to the backlog for this? To be complete, we'll also need to change the config "docs" in the codebase.

mesellings:
> @mesellings, thanks for the review! Can you add an issue to the backlog for this? To be complete, we'll also need to change the config "docs" in the codebase.

Will do - forgive my ignorance, what do you mean by "config docs in the codebase"? Where would this be?

conceptualshark:
If we're going through the trouble of updating these, my preference is to use `inline code` when referring to specific boolean values (or any value), but I don't think we have a company style for this use.

akeller:
> > @mesellings, thanks for the review! Can you add an issue to the backlog for this? To be complete, we'll also need to change the config "docs" in the codebase.
> 
> Will do - forgive my ignorance, what do you mean by "config docs in the codebase"? Where would this be?

On a quick review of the templates, we may not need to change this at all! 

The templates for broker config really drive this documentation. We moved it into the docs to make it easier to discover. At the top of the [page](https://docs.camunda.io/docs/next/self-managed/zeebe-deployment/configuration/broker-config/) we link to the [template](https://github.com/camunda/camunda/blob/main/dist/src/main/config/broker.yaml.template).

> A complete broker configuration template is available in the [Zeebe repo](https://github.com/camunda/camunda/blob/main/dist/src/main/config/broker.yaml.template).

We do something similar for gateway config.

mesellings:
Issue created: https://github.com/camunda/camunda-docs/issues/4123

PR #4119, opened 2024-08-07
Add spring boot test annotation

ingorichtsmeier:
Make the template runnable

## Description
When you just copy the snippet, the Zeebe engine won't start.

You have to add the `@SpringBootTest` annotation as well.

## PR Checklist

### DIFF
--- a/docs/components/best-practices/development/testing-process-definitions.md
+++ b/docs/components/best-practices/development/testing-process-definitions.md
@@ -70,6 +70,7 @@ A test can now look like the following example. The complete source code is avai
 
 ```java
 @ZeebeSpringTest
+@SpringBootTest
 class TestTwitterProcess {
 
     @Autowired
--- a/versioned_docs/version-8.3/components/best-practices/development/testing-process-definitions.md
+++ b/versioned_docs/version-8.3/components/best-practices/development/testing-process-definitions.md
@@ -70,6 +70,7 @@ A test can now look like the following example. The complete source code is avai
 
 ```java
 @ZeebeSpringTest
+@SpringBootTest
 class TestTwitterProcess {
 
     @Autowired
--- a/versioned_docs/version-8.4/components/best-practices/development/testing-process-definitions.md
+++ b/versioned_docs/version-8.4/components/best-practices/development/testing-process-definitions.md
@@ -70,6 +70,7 @@ A test can now look like the following example. The complete source code is avai
 
 ```java
 @ZeebeSpringTest
+@SpringBootTest
 class TestTwitterProcess {
 
     @Autowired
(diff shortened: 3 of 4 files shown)


akeller:
Hi @ingorichtsmeier, we actually [removed this](https://github.com/camunda/camunda-docs/pull/3780) in April based on a request from Support. Is there something else going on here with either the docs or code?

ingorichtsmeier:
@akeller, unfortunately, Emma didn't give a reason for her change request. 

I tried to run the example that I copied (without `@SpringBootTest`) and the Zeebe engine didn't start. Hmm...

After looking at the complete Twitter example on GitHub, I found the missing annotation.

I think it depends on the developers' expectation. If you want a Spring integration test, starting the Zeebe engine in the background and executing the tests on it, you need this annotation.

I don't know, which test would run without it. Maybe @epollum has more background?

There isn't anything else than my confusion. I tried to reduce my mental load by adding the annotation to the copy-template.

pepopowitz (approved):
I confirmed locally that the tests require this attribute in order to pass. 

@ingorichtsmeier @akeller what do you think about waiting a day or so for @epollum to provide feedback on why we removed this attribute in #3780 and then merging? (We can always revert if we have a good reason to after merging.)

akeller:
Thanks @pepopowitz! Can you also ping Emma on Slack? I'm not sure how often she's on GitHub.

pepopowitz:
For what it's worth, I think we should merge this, and revisit if we get feedback about the original removal.

akeller:
> For what it's worth, I think we should merge this, and revisit if we get feedback about the original removal.

Good enough for me! 🟢

epollum:
@ingorichtsmeier ,

I just tried this again with the @SpringBootTest and it does not work, it doesn't either the embedded or test containers that [zeebe-process-test](https://github.com/camunda/zeebe-process-test) should use and instead tries to connect to an external zeebe cluster

Please see https://github.com/epollum/c8-kind/tree/main/zeebe-application for example.

ingorichtsmeier:
After taking a look at my example and the given twitter example, I will close my pull request by now.

Reasons:
* spring-boot-starter-camunda-sdk is not an option to replace the community driven spring-boot-starter-camunda right now. In version 8.5.6 the sdk has a dependency to the community library, which is a grey area for support.
* In version 8.5, no test support for the sdk is available. It's coming with 8.6-alpha4. Tests still require the community test library.
* The twitter example uses a lot of spring boot support (Instead of my example from the hackdays), for example the `@MockBean` for the twitter service. Rewriting it for `@ZeebeProcessTest` is an extensive change to the example.

My personal test class will work with `@ZeebeProcessTest` as well when I remove the `@Autowire` from the Zeebe client. 

It feels halfhearted to me to just remove `@SpringBootTest`annotation from the example for showing, that spring-zeebe is not officially supported. 

But it is easier to wait for the release of 8.6 when the spring SDK supports tests and removes the dependency to the community library completely. This would be a good point to improve the best practice (and the example) with full supported libraries.

PR #4141, opened 2024-08-12
docs(guides): use registry and repository parameter

ManuelDittmar:
## Description
Bitnami Charts do have docker.io hardcoded as registry, therefore the current guide does not work if you want to update the registry. Also camunda components are now making correct use of registry vs. repository.

## PR Checklist


aabouzaid (changes requested):
I'd suggest using `global.image.registry` given we can use it for all of our components instead of setting that for each component individually.

christinaausley:
@ManuelDittmar Are you just waiting on additional review here?

ManuelDittmar:
@christinaausley yes

christinaausley:
> @christinaausley yes

CC @aabouzaid 😄

Issue #4117, opened 2024-08-06
Unify API Auth documentation


pepopowitz:
Part of #3649.

Unify the content on our various API Authentication docs, so that each API Auth guide contains the following: 

* Steps to generate a token, in either a SaaS (if available) or self-managed (if available) environment
   * Clear description of environment variables to be downloaded when creating a client
* An example auth call, in either a SaaS (if available) or self-managed (if available) environment
* An example usage of a generated token, in either a SaaS (if available) or self-managed (if available) environment
* A description of token expiration
* Consistent authentication requests between SaaS and self-managed, such that plugging the sample SaaS auth request into an OOTB self-managed environment with the correct variables, will work. (See https://github.com/camunda/camunda-docs/issues/4092#issuecomment-2266054121)

```[tasklist]
### Tasks
- [ ] #4078 
- [ ] Console SM API
- [ ] Operate API
- [ ] Optimize API
- [ ] Tasklist API
```

pepopowitz:
Complete!

PR #4152, opened 2024-08-13
feat: rewrite Tasklist API auth guide

pepopowitz:
## Description

Part of #4117.

Rewrites the Tasklist API Auth guide to align with specs defined in #4117.

Once approved, I'll backport to the current version of the docs as part of this PR.

## SaaS

image

## Self-Managed

image

## PR Checklist


docs/apis-tools/tasklist-api-rest/tasklist-api-rest-authentication.md
```diff
@@ -5,63 +5,137 @@ sidebar_position: 2
...
-
-Refer to the following example with _curl_:
+import Tabs from "@theme/Tabs";
+import TabItem from "@theme/TabItem";
+
+All Tasklist API requests require authentication. To authenticate, generate a JWT token and pass it in each request.
+
+## Generating a token
+
+<Tabs groupId="authentication" defaultValue="saas" queryString values={
```
akeller, line 15: Bit of a nit pick, but could you [sync](https://docusaurus.io/docs/2.x/markdown-features/tabs#syncing-tab-choices) the tabs so they are both on SaaS or both on SM, not a combination? This requires the same `groupId` prop.

docs/apis-tools/tasklist-api-rest/tasklist-api-rest-authentication.md
```diff
@@ -5,63 +5,137 @@ sidebar_position: 2
...
+   ```
+5. A successful authentication response looks like the following:
+   ```json
+   {
+     "access_token": "<TOKEN>",
+     "expires_in": 300,
+     "refresh_expires_in": 0,
+     "token_type": "Bearer",
+     "not-before-policy": 0
+   }
```
akeller, line 53: Do we need to do anything about the difference in the responses? The expiration immediately caught my eye.

docs/apis-tools/tasklist-api-rest/tasklist-api-rest-authentication.md
```diff
@@ -5,63 +5,137 @@ sidebar_position: 2
...
+  {
+    "id": "12345",
+    "name": "Do something",
+    ...
+  }
+]
 ```
 
-## Authentication for Self-Managed cluster
+## Token expiration
```
akeller, line 139: Similar to my other comment - do we need to mention the change in default value? 86400 vs. 300?

akeller (dismissed):
I'm approving ✅ 

"But you left comments!" - I did, but nothing sufficiently blocking. Across these changes I see such huge improvements to the readability and presentation of content, I'm happy with it! 💪

alessandrocavalli (dismissed):
Looks good to me!

christinaausley (commented):
I think some of @mesellings feedback rolls into this one as well, but otherwise, I'm happy with this 👍

docs/apis-tools/tasklist-api-rest/tasklist-api-rest-authentication.md
pepopowitz: Definitely, this is a good idea, and I misunderstood the `groupId` property. I remember thinking "oh I bet this has to be unique on the page, there's no way docusaurus will sync these tabs across different sets, I'd better use different groupIds or it's going to cause some weird behavior/an error." I should have more faith 😅

docs/apis-tools/tasklist-api-rest/tasklist-api-rest-authentication.md
pepopowitz: This specific output comes from a self-managed environment. In a SaaS environment, the only notable difference is the absence of a `refresh_expires_in` property. The default `expires_in` differs across environments, but in a SaaS environment that value can be just about anything, because the tokens are cached on the server rather than re-generated on each auth request. 

So a SaaS user would see an extra field in this sample response compared to reality. Do you think that's worse than a SM user seeing an absence of a field in this sample response compared to reality? I don't personally think either is very confusing, so I'd avoid showing a different sample response based on environment, to avoid unnecessary complexity.

pepopowitz: Oh, and now I see from your next comment that you were referring to the _value_ of the `expires_in`, not the presence of the `refresh_expires_in`. 😅 😬

docs/apis-tools/tasklist-api-rest/tasklist-api-rest-authentication.md
pepopowitz: I don't think so. That strays into environment configuration more than "interacting via API," IMO. If you disagree, I'm open to suggested content here.

akeller: Can you raise this with the Tasklist team to see if this default needs to be communicated somewhere?

pepopowitz: it's an identity thing. Mentioned at https://camunda.slack.com/archives/CRQAL5A1M/p1724356189900659.

pepopowitz: > It’s configurable in Keycloak at client level, or at realm level. You need to access Keycloak as an admin user, and then you can configure it.

pepopowitz:
This PR is ready for a final review, with all required feedback addressed, and backported to v8.5.

akeller (approved):

Issue #4070, opened 2024-07-26
Missing information in connect to an oidc provider documentation


maryarm:
### Description
On the [docs page](https://docs.camunda.io/docs/next/self-managed/setup/guides/connect-to-an-oidc-provider/), there are some missing information regarding environment variables and considerations:

1. The `SPRING_PROFILES_ACTIVE=oidc` is missed from Environment variables tab.
2. The fact that `IDENTITY_INITIAL_CLAIM_NAME` and `IDENTITY_INITIAL_CLAIM_VALUE` env variables are only applied once in during startup of identity only when there is no mapping rule in mapping_rules table and the changes are not effective afterward, is missed from documentation and can confuse customers. 

### Possible solution
The documentation should be updated to correctly reflect these two item.

akeller:
Added this to the DevEx board - @conceptualshark is this something you can take a stab at and add Maryam as a reviewer?

conceptualshark:
👍 Can do - I have some open work on this page, and this should fit right in.

PR #4161, opened 2024-08-15
Correct Redirect URIs, initial claims usage in OIDC docs 

### LINKED ISSUE #4066: Microsoft Entra example helm config is not clear regarding redirect URLs
### Description
On the [docs page](https://docs.camunda.io/docs/next/self-managed/setup/guides/connect-to-an-oidc-provider/) regarding connecting to Microsoft Entra, we suggest in the helm example to `see table below` for the redirect URL, this is incorrect, the redirect URL in the table at the bottom of the page is the complete value that should be set in Microsoft Entra itself, the `redirectUrl` param in our helm chart example should actually be the root of the component. For example:

The redirect uri for Identity in entra could be `https://identity.example.com/auth/login-callback` but the `redirectUrl` value in our helm charts would be `https://identity.example.com`.

### Possible solution
The examples should be updated to correctly reflect the difference, and the table should perhaps be more descriptive about what the value is for.

conceptualshark:
## Description

Resolves #4066, #4070 

- add the difference between redirect URIs required for microsoft entra vs helm 
- add appropriate spring profile var to the environment variables
- warn that initial claims cannot be updated using this configuration, and must instead be changed manually 
- minor cleanup

## PR Checklist


docs/self-managed/setup/guides/connect-to-an-oidc-provider.md
```diff
@@ -108,6 +103,10 @@ global:
 </TabItem>
 </Tabs>
 
+:::warning
+Once set, your initial claim name and value cannot be updated using environment or Helm values, and must be changed directly in the database.
```
mesellings, line 107: Non-blocking suggestion (nbs):

Once set, you cannot update your initial claim name and value using environment or Helm values. You must change these values directly in the database.

docs/self-managed/setup/guides/connect-to-an-oidc-provider.md
```diff
@@ -236,13 +240,13 @@ It's also important to follow the [steps described here](https://learn.microsoft
...
+| Identity    | **Microsoft Entra ID:** `https://<IDENTITY_URL>/auth/login-callback` <br/><br/> **Helm:** `https://<IDENTITY_URL>`         |                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                |
```
mesellings, line 245: Nbs: On the web page, these values sometimes break onto the next line if longer, but sometimes don't - you could standardise this for all of them by adding a <br/> after before each value, so they are always on the next line?
e.g. **Microsoft Entra ID:**<br/>`https://<OPERATE_URL>/identity-callback`

docs/self-managed/setup/guides/connect-to-an-oidc-provider.md
```diff
@@ -236,13 +240,13 @@ It's also important to follow the [steps described here](https://learn.microsoft
...
+| Component   | Redirect URI                                                                                                               | Notes                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          |
```
mesellings, line 243: For this whole table, I am getting it cut off - this might be due to table within a tab + long code value names behaviour? 
image

mesellings (approved):
Lgtm - just a few suggestions, I don't  sure if the table cut-off is a blocker as it is like this in prod, so I've approved anyway

christinaausley (approved):

PR #4168, opened 2024-08-17
Automate rest api docs generation

nicpuppa:
This PR add a GH workflow to automate the creation of REST API specification in our documentation


.husky/pre-commit
```diff
@@ -1,7 +1,7 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
-
-export NVM_DIR="$HOME/.nvm"
-[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
+#
+#export NVM_DIR="$HOME/.nvm"
+#[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
```
nicpuppa, line 5: ❓ I'm not sure about this, can someone clarify why this is needed ? The workflow doesn't work with this lines

pepopowitz: From https://github.com/camunda/camunda-docs/commit/2b5c57c1e2bab6b74c2b678b10e67ae68379e387#r88648535: "let's see if anyone complains" -- it took almost two years, but it finally happened! 😅 

I think these lines were snuck in for a dev environment that was running nvm, and it's intended to _not_ cause problems for someone that isn't. @nicpuppa can you point me at a failed workflow, so I can try to identify if this script can be rewritten to support its intended use, and also work for this workflow? I won't spend a lot of time investigating, and if I can't get anywhere, then I support the removal of these lines.

nicpuppa: This is the [failed workflow](https://github.com/camunda/camunda-docs/actions/runs/10563015584/job/29262273110). 
This is the [husky issue](https://github.com/typicode/husky/issues/1176) that point me to this removal

pepopowitz: From that linked issue:

>  it was a minor version difference for node. Specifically he was running node 16.17 and the .nvmrc file for the project specified 16.13. 

Looking closer, this seems likely the culprit here, too. I didn't expect it, but [the image used by the workflow](https://github.com/actions/runner-images/blob/ubuntu22/20240818.1/images/ubuntu/Ubuntu2204-Readme.md) does include nvm! I would have expected it not to be there, in which case that line wouldn't have executed anything. 

But it does, and since [our .nvmrc](https://github.com/camunda/camunda-docs/blob/main/.nvmrc) hasn't been updated in a while, it's likely using a version that isn't installed on the workflow instance. 

The couple of solutions I can think of are: 
1. remove these lines. This might mean that some contributors need to update their shell profile so that nvm is loaded there, instead of in this pre-commit hook. (Unless I misunderstand why it was originally introduced, this is likely what should have been done all along.)
2. Update our .nvmrc to a more recent version (maybe even just the most recent 16, not sure which versions are installed on the workflow image). I don't really want to do this right now, because it feels like it could explode, and is definitely out of scope of this PR.

All this to say, I support removing these lines. @akeller @nicpuppa do you have a preference whether we remove them in this PR, or whether I open a separate PR since it is outside the scope of this PR?

.github/workflows/sync-rest-api-docs.yaml
```diff
@@ -0,0 +1,49 @@
...
+        uses: peter-evans/create-pull-request@v5
+        with:
+          token: "${{ secrets.GITHUB_TOKEN }}"
+          path: docs
+          title: Update camunda REST API doc
+          body: This PR updates the camunda REST API
+          commit-message: "docs: update REST API spec"
+          branch: update-rest-doc
+          delete-branch: true
+          reviewers: camunda/tech-writers
```
pepopowitz, line 48:
```suggestion
          reviewers: camunda/tech-writers,pepopowitz
```

Blocking: [The docs](https://github.com/peter-evans/create-pull-request?tab=readme-ov-file#action-inputs) say this should work; I haven't tested to confirm. I would also like to be assigned  about these PRs.

.github/workflows/sync-rest-api-docs.yaml
```diff
@@ -0,0 +1,49 @@
...
+            exit 0
+          fi
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v5
+        with:
+          token: "${{ secrets.GITHUB_TOKEN }}"
+          path: docs
+          title: Update camunda REST API doc
+          body: This PR updates the camunda REST API
```
pepopowitz, line 44: Blocking: We will want more details in this body, especially to clearly describe the expected action. We don't necessarily have to figure out everything in this PR, but I think we should definitely add more context, and iterate it on the future. 

@akeller what do you think we should include in this body? Some things I'm thinking of: 

* More thorough description of what these changes represent, and how they were generated.
* Link back to the workflow that generated it 
* Clearly state a default urgency ("not urgent")
* Clarify that it only affects the "next" version of the docs

.github/workflows/sync-rest-api-docs.yaml
```diff
@@ -0,0 +1,49 @@
...
+          commit-message: "docs: update REST API spec"
+          branch: update-rest-doc
+          delete-branch: true
+          reviewers: camunda/tech-writers
+          base: main
```
pepopowitz, line 49:
```suggestion
          base: main
          labels: deploy
```

non-blocking: Do we want to default to deploying a preview site? I have no strong opinion here, but if others do, I think adding this line would do it.

pepopowitz (changes requested):
I love this, and we (read: me, probably) should think about extending this to other APIs when possible. 

I have some blocking feedback, especially around details of the PR that gets opened. I also started some conversations in my feedback, and we should not hesitate in those conversations to qualify things as worthy of follow-up PRs instead of blocking this one.

.husky/pre-commit
akeller: You can remove them in this PR so we have the context all in the same PR.

.github/workflows/sync-rest-api-docs.yaml
akeller: +1 here, you should also be added as a reviewer

.github/workflows/sync-rest-api-docs.yaml
akeller: Ideally, we get everything in the [PR template](https://github.com/camunda/camunda-docs/blob/main/.github/pull_request_template.md), which you captured above. We don't need the template filled out, just the info.

.github/workflows/sync-rest-api-docs.yaml
akeller: Preview site will be the easiest way for us to validate the changes and do any final review. This would be my preference, but like you said, it's non-blocking.

nicpuppa: I like this 👍

.github/workflows/sync-rest-api-docs.yaml
nicpuppa: Got it 👍

nicpuppa:
Hey @pepopowitz @akeller I addressed all the suggestions, please have another look 😄

.github/workflows/sync-rest-api-docs.yaml
pepopowitz: > This is why I think that the PRs coming from this workflow should be approved and merged straight forward

I appreciate your optimism, but I don't believe that mistakes won't slip through, especially as this process is forming. Yes, the specs should arrive here having already been reviewed, but (a) technical writers haven't even agreed yet to review at that stage, (b) as far as I'm aware there's nothing enforcing reviews from the devex team on those upstream spec changes, and (c) we won't know about spec changes that we aren't tagged on. 

If we see these come through a bunch of times with no missed descriptions or technical writing edits, I might feel more confident merging them blindly. But especially at the beginning, I don't share that optimism. Please assign myself and the technical writers for now.

pepopowitz (approved):
This looks great, thanks for making changes!!

akeller (approved):
👍  No complaints!

Issue #4011, opened 2024-07-02
Consolidate Intro to Camunda 8 + Intro to Components


akeller:
As we move to position Camunda 8 as a unified platform, we should introduce Camunda 8 as a platform, not a set of components.

With that, we should consolidate [Introduction to Camunda 8](https://docs.camunda.io/docs/guides/) & [Introduction to Components](https://docs.camunda.io/docs/components/).

The final location is TBD - can be either the current location in /guides/ or /components/. 

--
This phase 2 item is part of the overarching project around #3988.

christinaausley:
This is ready for review with [Consolidate intro to camunda #4181](https://github.com/camunda/camunda-docs/pull/4181), though note a few comments in the PR description for discussion.

christinaausley:
@camunda/tech-writers @akeller I think the existing PR is a fine baseline and starting point for general ideas, but I should have started conversation on a Miro board before launching changes -- I'm going to convert this to a draft, and add a hold label.

See the [Miro board](https://miro.com/app/board/uXjVKm1QkNs=/?share_link_id=898204352738) for continued discussion, where I've outlined a second set of proposed changes. Will bring into the PR once finalized 👍

PR #4175, opened 2024-08-20
Docs site landing page redesign

mesellings:
## Description

Redesign of the landing page to include:

* New text and icons for the six landing cards.
* Faded hero bg image for improved readability.
* Edited hero text for conciseness and relevance. 
* Hero buttons follow the product dark mode pattern (blue).
* Additionally, fixed the layout of the three What's new buttons as these were not mobile responsive.

This PR relates to and closes the work in https://github.com/camunda/developer-experience/issues/340.

## PR Checklist

- [X] My changes require a [technical writer review](https://github.com/camunda/camunda-docs/blob/main/howtos/documentation-guidelines.md#review-process), and I've assigned @camunda/tech-writers as a reviewer.


docusaurus.config.js
```diff
@@ -6,7 +6,7 @@ const latestVersion = require("./src/versions").versionMappings[0].docsVersion;
 module.exports = {
   title: "Camunda 8 Docs",
   tagline:
-    "Orchestrate complex process flows, automate across people, systems, and devices, and transform your organization.",
+    "Get started with Camunda 8 SaaS or Self-Managed universal process orchestration.",
```
akeller, line 9: 🤔 I don't love this, but I can't figure out why. Is this copy you got from Marketing? 

> Get started with Camunda 8, the universal process orchestrator. 

> Get started with Camunda 8 SaaS or Self-Managed for universal process orchestration.

src/css/custom.css
akeller, line 1: I noticed you skipped really rounding out the buttons. Is that a future follow-up? I am ok with this since we didn't have the bandwidth to pick up matching the product styles.

src/pages/index.js
```diff
@@ -122,7 +124,7 @@ function Home() {
         )}
         <div className={clsx("hero hero--secondary", styles.heroBanner)}>
           <div className="container">
-            <h2 className="hero__title">Want to know what's new?</h2>
+            <h2 className="hero__title">What's new</h2>
```
akeller, line 127: 💭 We should have some guidance about how/when the content in this section gets updated.

akeller (commented):
I don't see anything highly blocking, but I did leave some comments about items I want to make sure we either add to the backlog or discuss as a team. 

The tagline is the only thing preventing me from really giving an outright approval. I can't put my finger on what's going on there, but it reads a bit awkward to me.

docusaurus.config.js
mesellings: This isn't from Marketing, but I tried to include the following important elements in it: Get Started (echoes the button), 2 main products/paths (SaaS and Self-Managed), and the "universal process orchestration" phrasing that is a focus currently. I think it also needs to be snappier, as the previous tagline was too long to scan quickly.

If we stick with the current style, I like your second option: "Get started with Camunda 8 Saas or Self-Managed for universal process orchestration."

We could also use alternative tagline styles such as:

* Standard documentation subtitle. e.g. "Explore Camunda 8 SaaS and Self-Managed documentation."
* Search focused (once we have a search bar). e.g. "Find answers to your technical questions and learn more about Camunda 8."

@christinaausley what do you think?

src/css/custom.css
mesellings: We had a lot of discussion about buttons/styles and currently there is quite a disparity between product/marketing regarding buttons and design systems. The upshot was that we are loosely okay to choose and even develop our own style for the docs, but that ideally we should follow product (rectangle) vs marketing (pill-shaped). 

For now, I felt this subtler rounding was a good compromise, as:
* The rectangular shaped buttons in the Carbon Design system would mean all styling in the docs site should probably align and be 90 degree shapes, which would make the site look dated. 
* The pill-shaped buttons similarly dictate a certain style, which moves too far away from the product.
* The docs site currently uses this subtle rounding across all other elements, such as feedback buttons, admonitions, badges etc.

Some of the designers talked about a refresh coming at some point on the marketing site anyway, so it would be interesting to see how that evolves and what they do about buttons.

I would like to look at the general design of the site and elements (cards, font, search box, landing pages, colours etc) in a future issue definitely - I think we can vastly improve the site look and feel with a few minor targeted changes.

src/pages/index.js
mesellings: Great idea 👍

docusaurus.config.js
christinaausley: How about "Get started with the Camunda 8 SaaS and Self-Managed documentation to orchestrate your process flows."?

src/pages/index.js
christinaausley: Perhaps by each alpha or minor release?

src/pages/index.js
```diff
@@ -8,61 +8,61 @@ import styles from "./styles.module.css";
...
   },
   {
-    title: "I'm looking for a particular component",
-    imageUrl: "img/homegrown-workflow.png",
+    title: "Components",
+    imageUrl: "img/home-components.png",
     url: "/docs/components/",
     description:
-      "Find product manual content for each component in Camunda 8. Together, these components comprise the Camunda 8 SaaS experience.",
+      "Explore components such as Modeler, Zeebe, Connectors, Operate, Tasklist, and Optimize.",
```
christinaausley, line 22: I would add Console before Modeler, and I'm having a hard time naming Connectors a component -- let's remove it here until I can define in my work on our [Camunda intro docs](https://github.com/camunda/camunda-docs/issues/4011).

docusaurus.config.js
pepopowitz: another alternative, but I have no stake in this:

src/pages/styles.module.css
```diff
@@ -32,13 +40,33 @@
...
+  grid-auto-rows: 250px;
+  gap: 2.5rem;
+}
+
+/* new styles */
+
+.padmd {
+  padding: 0 2rem;
+}
+
```
pepopowitz, line 59: This class is used in only one place, and follows the "utility CSS" approach of styling. The pre-existing styles in this stylesheet all follow a "semantic CSS" approach of styling. I don't think we should mix these two approaches in the same stylesheet -- it introduces uncertainty for someone coming in and making new changes, about which style to follow. 

If the utility class was used many times, I might feel differently, but especially since it's only used once, I'd prefer for us to apply this specific padding in the semantic class that needs it. 

The one catch with that, in this specific example, is that the element that requires this padding applies the `.features` style......which is reused on a second different element on the page. The best way to fix this, IMO, is to create separate semantic classes for the two call sites, and duplicate the shared styles across them. Then we can apply this unique style to the one class that needs it.

This was a little hard to explain, so I'm worried it doesn't make sense. Please let me know if it doesn't!

src/pages/styles.module.css
```diff
@@ -32,13 +40,33 @@
...
+}
+.featuresh3 {
+  line-height: 1.4;
+  font-size: 24px;
+}
+
+.rowflex {
+  gap: 30px;
+  display: flex;
+  justify-content: center;
```
pepopowitz, line 71: Similar to my previous comment about utility vs semantic, this utility class is only used once. We should instead apply the style through a semantically named class (e.g. "buttonWrapper" or something).

src/pages/styles.module.css
```diff
@@ -32,13 +40,33 @@
...
+  gap: 2.5rem;
+}
+
+/* new styles */
+
+.padmd {
+  padding: 0 2rem;
+}
+
+.featuresp {
```
pepopowitz, line 60:
```suggestion
.featuresDescription {
```

1. See my previous comments on semantic CSS vs utility -- this class is named after an HTML element (`p`), and I'd prefer to maintain the semantic-CSS approach already in place for this page, to avoid mixing metaphors.
2. We should also use a case that communicates word boundaries. The file is already using `camelCase` so I chose that for this suggestion; I don't actually care if we did `kebab-case` instead (e.g. `featuresDescription`.)

This would require corresponding changes in the `pages/index.js` file too.

src/pages/styles.module.css
```diff
@@ -32,13 +40,33 @@
...
+  line-height: 1.4;
+}
+.featuresh3 {
```
pepopowitz, line 63:
```suggestion
.featuresTitle {
```

See previous comments re: mixing styling approaches & word boundaries. Note also that this would require corresponding changes in the pages/index.js.

pepopowitz (changes requested):
This looks great, but I have some changes I'd like to see made in the CSS. Let me know if any of it is unclear!

src/pages/index.js
mesellings: Changed to: "Explore components such as Console, Modeler, Zeebe, Operate, Tasklist, and Optimize."

src/pages/styles.module.css
mesellings: @pepopowitz Great review - I've changed all the CSS to use semantic references + camelCase as you suggest, and it's much better thank you!

_(I was trying to prepare for if/when we moved styles out into the main CSS and that are applied more globally if/when we start using these components in other areas, but we can tackle that if/when we actually need to and it was a bit messy 👍)_

docusaurus.config.js
mesellings: I like your alternative @christinaausley, just shortened for verb + description pattern, so I've changed it to:

"Start orchestrating your processes with Camunda 8 SaaS or Self-Managed."

mesellings:
@akeller @christinaausley @pepopowitz I've made all the changes as suggested, please re-review 👍

christinaausley (approved):
Looks great -- thank you for polishing this up!

pepopowitz (approved):
Thanks for these changes!

akeller (approved):
🚀 :shipit: 🚢

pepopowitz:
oh! Commenting on another PR reminded me that I wanted to mention something here -- 

In a conversation with Amara yesterday, she mentioned that you all had expressed in a project Slack channel that you'd like to merge this by end of week. I don't have visibility into that channel (and I don't want it 😅), but I would have had no idea of your timeline without her telling me.

In our PR template, we have the "When should this change go live" section, with some checkboxes to call out a timeline. In future PRs, can we keep that updated with any timeline objectives? Even if it's not a hard target-driven deadline, and something like "I'd like to merge this by the end of the week because I'll be out next week", it is useful information for reviewers. In this specific example it took me a few days to finally review this; I might have prioritized it higher on my list if I'd known your goal date for merge.

mesellings:
> oh! Commenting on another PR reminded me that I wanted to mention something here --
> 
> In a conversation with Amara yesterday, she mentioned that you all had expressed in a project Slack channel that you'd like to merge this by end of week. I don't have visibility into that channel (and I don't want it 😅), but I would have had no idea of your timeline without her telling me.
> 
> In our PR template, we have the "When should this change go live" section, with some checkboxes to call out a timeline. In future PRs, can we keep that updated with any timeline objectives? Even if it's not a hard target-driven deadline, and something like "I'd like to merge this by the end of the week because I'll be out next week", it is useful information for reviewers. In this specific example it took me a few days to finally review this; I might have prioritized it higher on my list if I'd known your goal date for merge.

Apologies @pepopowitz yes you're correct, we should have kept you in the loop but it got missed out - my fault entirely 👍

PR #4180, opened 2024-08-21
fix: link to telemetry sample asset via absolute github URL

pepopowitz:
## Description

Discussed [in Slack](https://camunda.slack.com/archives/C026U8GBNSW/p1724234114734109).

Replaces a relative link to a `.json` asset with an absolute link to the GitHub source.

The relative link was emitted as a non-existent path, with a trailing slash after `.json`. This is a known issue in Docusaurus. 

## PR Checklist

### DIFF
--- a/docs/self-managed/console-deployment/telemetry.md
+++ b/docs/self-managed/console-deployment/telemetry.md
@@ -36,7 +36,7 @@ When the telemetry configuration is set to `online` (recommended), Console colle
 
 ### Example
 
-You can download an example [data file](./telemetry-sample.json). Below is an extract to highlight the types of data collected:
+You can download an example [data file](https://github.com/camunda/camunda-docs/blob/main/docs/self-managed/console-deployment/telemetry-sample.json). Below is an extract to highlight the types of data collected:
 
 ```json
 {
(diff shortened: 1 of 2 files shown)


docs/self-managed/console-deployment/telemetry.md
```diff
@@ -36,7 +36,7 @@ When the telemetry configuration is set to `online` (recommended), Console colle
 
 ### Example
 
-You can download an example [data file](./telemetry-sample.json). Below is an extract to highlight the types of data collected:
+You can download an example [data file](https://github.com/camunda/camunda-docs/blob/main/docs/self-managed/console-deployment/telemetry-sample.json). Below is an extract to highlight the types of data collected:
```
pepopowitz, line 39: Unfortunately, when we cut a new version, this URL will *not* be automatically adjusted to the new version number, i.e. all versions of this page moving forward will point at the sample file associated with `vNext`. We don't have a canonical source for this file outside the docs.

I think this is fine, and I don't really have any good ideas on how to address this without what feels like too much effort. I'm open to suggestions, though!

Although maybe it also suggests that moving this file to the `static` folder is _not_ as bad an approach as I think it is, since we aren't truly using the versioning of this asset anyway. 🤔

akeller: I don't know enough about how this file may change minor version to minor version. @theburi, do you anticipate many changes to this file? Or any changes to this file with new minors (or alphas)?

theburi: It may change in 8.7. but I would not expect it to change that much over time.

akeller (approved):

PR #4189, opened 2024-08-22
Standardize component config introductions/options 

conceptualshark:
## Description

An attempt to start the component configuration pages on the path of some sort of standardization. This was scoped down to only tackle the introduction and config options for each page, and is not meant to tackle bigger areas of restructuring, rewriting, etc.

## PR Checklist


christinaausley (approved):

christinaausley:
CC @camunda/distribution for review here 👍

docs/self-managed/console-deployment/configuration.md
```diff
@@ -9,14 +9,21 @@ description: "Read details on the configuration variables of Console Self-Manage
 Console Self-Managed is available only to [Enterprise customers](/reference/licenses.md#console).
 :::
 
-Console Self-Managed can be configured using environment variables and configuration parameters.
+Console can be configured using environment variables, configuration parameters, or a combination of both. When configuring your Console setup, keep in mind the following:
```
akeller, line 12: This is probably a PM question, but do we want to emphasize "Console Self-Managed" as the component name? Or are they similar enough that we can drop "Self-Managed" with the right context?

hamza-m-masood: I think it is a good idea to specify self-managed as well, since it is a different development stream. It is quite different from SaaS.
cc @theburi

docs/self-managed/console-deployment/configuration.md
```diff
@@ -71,7 +80,7 @@ console:
       value: online
 ```
 
-## Using a different OpenID Connect (OIDC) authentication provider than Keycloak
+## Configure an OpenID Connect (OIDC) provider
```
hamza-m-masood, line 83: I think it is still a good idea to mention keycloak somewhere in the title, just so it is clear that the OIDC provider and keycloak are mutually exclusive. 
Or maybe mention it in a paragraph under the title. 
What do you think @conceptualshark ?

docs/self-managed/identity/deployment/configuration-variables.md
```diff
@@ -5,10 +5,18 @@ sidebar_label: "Configuration variables"
...
-## Core configuration
+- If both configuration files and environment variables are present, environment variables overwrite settings in configuration files.
+- The existing configuration is applied at startup, and changes made to the configuration will not be applied at runtime.
+
+For more information on Self-Managed configuration options, see [configuring components](/self-managed/operational-guides/application-configs.md).
+
+## Environment variables
+
+:::note
+Underscores in environment variables correspond to configuration file key levels.
```
hamza-m-masood, line 18: Maybe this is true for all configuration options in identity, but I don't think it holds true for other components. I think it would be better to use `Generally,...` at the start

akeller:
Removing `deploy` since it's currently broken. 😢

conceptualshark:
Removed the less-certain pieces of this update (if underscores always represent key levels) and included feedback re: Keycloak specificity, specifying Console Self Managed. Also removed a reference to Console being enterprise-specific and will follow up on other instances of that in a separate issue.

christinaausley (approved):
I love this! 🥳

akeller (commented):
👍 No concerns.

I did notice some repeated content that could be added to a partial/reusable component, but would consider that a good item to follow up on.

theburi (approved):

hamza-m-masood (changes requested):
I have raised my concerns in this comment:
https://github.com/camunda/camunda-docs/pull/4189#discussion_r1877974620

PR #4251, opened 2024-09-04
TW edits: Note and formatting

jwulf:
## Description

Here is a suggested edit for the note on start messages and correlationKey values.


mesellings (approved):
Great work, thanks @jwulf - approved 👍

akeller:
Should this be backported or is this only relevant for `/next/`? (Or is there another PR for that?)

mesellings:
> Should this be backported or is this only relevant for `/next/`? (Or is there another PR for that?)

I think this will need to be backported, but I will do that in the original PR (this one was just an edit from Josh to the note I agreed with the devs)

PR #4254, opened 2024-09-05
JDK version for spring-zeebe-sdk should be >= 21

grafvonb:
Using JDK 17 causes the following error which indicates wrong version of the JDK: 

```
[ERROR]   bad class file: /home/user/.m2/repository/io/camunda/spring-boot-starter-camunda-sdk/8.5.3/spring-boot-starter-camunda-sdk-8.5.3.jar(/io/camunda/zeebe/spring/client/configuration/ZeebeClientProdAutoConfiguration.class)
[ERROR]     class file has wrong version 65.0, should be 61.0
```


akeller:
Marking this as a bug and following up internally.

akeller:
Hi @grafvonb there was a [bug with 8.5.3](https://github.com/camunda/camunda/issues/19658). Are you able to update to 8.5.4? This should resolve this issue for you.

grafvonb:
Hi @akeller, yup, it works with 8.5.4+ again. Thank you.

PR #4260, opened 2024-09-05
Update/audit external links 

conceptualshark:
## Description

Some easy wins in updating external links for September docs housekeeping.

## PR Checklist


docs/self-managed/console-deployment/overview.md
```diff
@@ -8,5 +8,3 @@ description: "Console Self-Managed provides key insights into automation cluster
 Camunda Console Self-Managed offers a centralized overview for Camunda 8 installations, designed to enhance operational efficiency and oversight within Enterprise environments. Console Self-Managed provides key insights into automation cluster deployments, process orchestration usage, and streamlining usage tracking.
 
 Camunda Console Self-Managed provided is available as a container image. Refer to the [installation guide](/self-managed/setup/overview.md) for details on how to install this component.
-
-Participate in development by sharing your feedback on our [ProductBoard page](https://portal.productboard.com/q4b4cm3z3fkx5zesq51o3mz3).
```
akeller, line 12: @javi-alzorriz-camunda do we want to provide an opportunity for people to provide feedback via ProductBoard?

akeller (approved):
I left a non-blocking comment, but I'm fine with this merging.

docs/self-managed/console-deployment/overview.md
conceptualshark: If we need to swap something new in here, I can follow-up with a separate PR to update it. 👍

akeller: @conceptualshark, We should not use ProductBoard in the public-facing docs; it's for internal use only. Instead, non-Camundi would follow the feature request process in JIRA. I don't think we have any more links in ProductBoard and this is covered on the [contact page](https://docs.camunda.io/contact/).

PR #4272, opened 2024-09-09
docs(dual-region): update base branch for 8.5 release scripts

Langleu:
## Description

Adjusts 8.5 docs for the dual-region topic to select the appropriate 8.5 stable branch for scripts and active references.

It also removes a temporary `legacy` mode that was required as we had everything on a single branch prior.

While I don't think it requires a tech-writer review, I've had instances in the past where I did minor changes and was asked afterwards to always double check, so here we are! 😁 

related to https://github.com/camunda/team-infrastructure-experience/issues/241 on the InfraEx side.

## PR Checklist

### DIFF
--- a/versioned_docs/version-8.5/self-managed/operational-guides/multi-region/dual-region-ops.md
+++ b/versioned_docs/version-8.5/self-managed/operational-guides/multi-region/dual-region-ops.md
@@ -82,7 +82,7 @@ For the **failback** procedure, your recreated region cannot contain any active 
 
 :::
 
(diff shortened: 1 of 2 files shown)


Langleu:
since the Helm chart isn't published yet etc. I would probably keep it as is as it's still 8.5 on the next branch.
With the Helm chart release (11) + closer to release with stable images everywhere, we can add a branch and change that in the 8.6 docs branch that will cut the new release.

So we stay `next` --> `main`
`8.5` --> `stable/8.5`
when the new docs PR is created for 8.6 changes, we can add the `stable/8.6` branch and change the docs solely for that version.

conceptualshark (approved):

PR #4301, opened 2024-09-13
Update the Spring developer guide

camundom:
### Description

Improved the guide to align with the upcoming release of Camunda 8 Desktop Modeler and C8 Run. This includes revising the Prerequisites section, removing redundant steps now covered by the new Self-managed setup, and improving the clarity and structure of the steps for getting started.

### What's changed?

- **Prerequisites**
    - Includes links to the OpenJDK distribution for easy access.
    - Includes links to download the Desktop Modeler & C8 Server
    - Helpful tip that informs the user if they have already taken the survey to look for the Download links in their email.

- **Discarded Step 1: Install Camunda 8 Self-managed**
    - Removed this step entirely as it’s now redundant with the introduction of C8 Run

- **Steps 2 & 3**
    - Updated to mention the latest Maven version - 3.3.3
    - Improved discoverability of the Properties panel and readability
    - Added link to ‘What is FEEL?’

- [x] This on a **specific schedule** and the assignee will coordinate a release with the DevEx team. (apply `hold` label or convert to draft PR)

## PR Checklist


docs/guides/getting-started-java-spring.md
```diff
@@ -36,18 +36,22 @@ For example, in this guide we will outline a BPMN model to receive a payment req
...
-## Step 1: Install Camunda 8 Self-Managed
+## Prerequisites
 
-<Install/>
+Before getting started, ensure you have the following in your local environment:
 
-## Step 2: Create a new Spring Boot project
+- Access to your preferred code editor or IDE
+- [OpenJDK 21+](https://jdk.java.net/)
+- [Camunda 8 Desktop Modeler](https://camunda.com/download/modeler/) and [Camunda 8 Server<br/>](https://camunda.com/download/modeler/)_If you've already signed up for a Self-managed account, you can find the download links in your welcome email._
```
pepopowitz, line 47: blocking, in one way or another: 

Previous to this, we've been creating a separate prerequisites sub-document for each different form of installation. I can get on-board with eliminating the sub-documents, especially if this is the final form of this guide. But if this isn't the final expected form of this guide, or the previous practice of keeping these bits in a sub-document resonates with you, we should stick with that approach. 

If we _do_ stick with this approach of pulling this section inline instead of importing from a sub-document, then there are a couple things we should clean up: 

1. The `[comment]: SmPrereqs/` line -- it's unclear to me why we would preserve that path as a "comment". 
2. The now-unused import lines at the top of this file: 
   ```
   import SmPrereqs from './react-components/sm-prerequisites.md'
   import Install from './react-components/install-plain-java.md'
   ```
3. (maybe) The unused sub-documents in [the `react-components` folder](https://github.com/camunda/camunda-docs/tree/4300-update-the-spring-developer-guide-to-include-the-new-installation-of-self-managed/docs/guides/react-components).

pepopowitz (dismissed):
Yay, C8RUN!!!!! Thanks for updating this!!!

I think my one comment will result in changes one way or another.

docs/guides/getting-started-java-spring.md
camundom: Thanks for taking a look and sharing your feedback @pepopowitz !

This would be the final form of this guide. Would it be good to add these steps to the separate prerequisites sub-document instead, replacing the older format of installing SM? AFAIK, there are two other guides that use prerequisites and the sub-document would help the other guides as well.

> The [comment]: SmPrereqs/ line -- it's unclear to me why we would preserve that path as a "comment".

This is temporary. If we go ahead with using the sub-document, then this line would get uncommented.

> The now-unused import lines at the top of this file:

These would be removed as well.

PS: This is my first contribution towards docs where I've edited/written using markdown and understanding the structure was new to me :)

akeller:
FYI @conceptualshark

docs/guides/getting-started-java-spring.md
conceptualshark: There is [a component](https://github.com/camunda/camunda-docs/blob/main/docs/guides/react-components/install-c8run.md) for reusing the C8Run installation content created already, but I haven't yet dropped it into existing pages like this. I would recommend using a separate component than the ones that already exist for the Java install, though, as I am not sure each page is ready to use with C8Run.

docs/guides/getting-started-java-spring.md
```diff
@@ -36,18 +36,22 @@ For example, in this guide we will outline a BPMN model to receive a payment req
...
 
-## Step 1: Install Camunda 8 Self-Managed
+## Prerequisites
 
-<Install/>
+Before getting started, ensure you have the following in your local environment:
 
-## Step 2: Create a new Spring Boot project
+- Access to your preferred code editor or IDE
+- [OpenJDK 21+](https://jdk.java.net/)
```
conceptualshark, line 46: Do we want to be opinionated about this link? I opted against including a specific link/recommendation in the C8Run documentation, as each user may have a different preference for install.

docs/guides/getting-started-java-spring.md
```diff
@@ -36,18 +36,22 @@ For example, in this guide we will outline a BPMN model to receive a payment req
...
+- [Camunda 8 Desktop Modeler](https://camunda.com/download/modeler/) and [Camunda 8 Server<br/>](https://camunda.com/download/modeler/)_If you've already signed up for a Self-Managed account, you can find the download links in your welcome email._
```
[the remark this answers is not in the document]
akeller: **The name will be Camunda 8 Run.** After our meeting today, @melanie-butcher will communicate this to the respective teams.

@camundom, is your timeline for this PR aligned with the 8.6? Or just the marketing pages/assets being ready too?

akeller:
After the prj-dev-gettingstarted meeting today, I made the following changes:

* Removed the note referring to the tutorial only being for non-prod
* Adjusted the 2nd note about greenfield - greenfield is not a term that resonates well with our users based on user testing, so we removed it

docs/guides/getting-started-java-spring.md
akeller: I want to be more opinionated, but my first concern is this is the wrong link to OpenJDK. We want something like https://openjdk.org/install/

docs/guides/getting-started-java-spring.md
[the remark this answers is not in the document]
camundom: @akeller, thanks for the update! Yes, the timeline for this PR is 8.6. The marketing pages are ready for preview and undergoing testing, while the download email is WIP.

docs/guides/getting-started-java-spring.md
camundom: The idea in providing a direct link was to make it easier for users to quickly access the right resource—the same link OpenJDK provides on their installation page. We also received user feedback requesting direct links for easier access. The link to https://openjdk.org/install/ works fine as well.

conceptualshark: I don't really love the OpenJDK install page - I think it's a little misleading/confusing, and OpenJDK offers no follow-up on setting JAVA_HOME when installed in this way. 

Using homebrew or apt-get for mac and linux installs is going to be a lot more straightforward, but this opens us up to needing to document multiple paths of entry because OpenJDK's documentation isn't terribly robust. I just want to make sure that being opinionated doesn't actually get us fielding more questions than it answers.

conceptualshark:
@akeller What led to removing the non-production notice? Is that something we'd need to repeat in other guides?

docs/guides/getting-started-java-spring.md
conceptualshark: Either way, a note to make sure JAVA_HOME is set correctly would probably help; that way, no matter the installation, that check happens before they continue.

akeller:
> @akeller What led to removing the non-production notice? Is that something we'd need to repeat in other guides?

User feedback/testing - @camundom, this would be outside the scope of this PR, but do you have a recommendation for our other guides?

conceptualshark:
@camundom Is this still waiting for marketing/email journey updates? Is there a version of this update we can deliver for 8.6?

docs/guides/getting-started-java-spring.md
```diff
@@ -36,18 +29,26 @@ For example, in this guide we will outline a BPMN model to receive a payment req
...
-<SmPrereqs/>
+## Prerequisites
 
-## Step 1: Install Camunda 8 Self-Managed
+Before getting started, ensure you have the following in your local environment:
 
-<Install/>
+- Access to your preferred code editor or IDE
+- [OpenJDK 21+](https://openjdk.org/install/)
+- [Camunda 8 Desktop Modeler](https://camunda.com/download/modeler/) and [Camunda 8 Run<br/>](https://camunda.com/download/modeler/)_If you've already signed up for a Self-Managed account, you can find the download links in your welcome email._
```
theburi, line 38: I noticed that C8Run is linked to the modeller link. 
Could we add a link to C8Run installation page? 
I would not expect users to wait for email or look for email when they are already committed to trying out this guide. 
Plus, we do not get the download link in our email. It is not as easy as downloading an app. it needs a few more steps.

akeller:
Removing `deploy` since it's currently broken. 😢

akeller:
I resolved the merge conflicts that may _not_ be in the spirit of this PR, but I wanted to get the `deploy` tag working again, which required merging in changes from main. @camundom please have a look.

docs/guides/getting-started-java-spring.md
camundom: It looks like the download emailer hasn't gone up yet. I'll add the link to the C8 Page.

akeller:
@camundom is this ready to merge? It will need to be backported to appear in the 8.6 docs, but it's something DevEx can help with. Just let us know.

camundom:
> @camundom is this ready to merge? It will need to be backported to appear in the 8.6 docs, but it's something DevEx can help with. Just let us know.

@akeller this is ready to merge

christinaausley:
I will backport and merge this now.

christinaausley (approved):

docs/guides/getting-started-java-spring.md
```diff
@@ -35,16 +31,26 @@ For example, in this guide we will outline a BPMN model to receive a payment req
...
 
-## Step 1: Install Camunda 8 Self-Managed
+## Prerequisites
+
+Before getting started, ensure you have the following in your local environment:
 
-<Install/>
+- Access to your preferred code editor or IDE
+- [OpenJDK 21+](https://openjdk.org/install/)
+- Camunda 8 Desktop Modeler and Camunda 8 Run
```
conceptualshark, line 40: Can we link to the docs for both here?

christinaausley: I will incorporate this in a separate PR @conceptualshark 👍

PR #4348, opened 2024-09-25
Document instance migration phases 3 and 4

korthout:
## Description

closes camunda/camunda#22680

WIP

## PR Checklist

WIP


docs/components/concepts/process-instance-migration.md
```diff
@@ -153,11 +153,18 @@ This section explains how to deal with catch events when migrating a process ins
 
 You decide what happens to the associated event subscription through the mapping instructions for the catch events:
 
-- If a catch event is mapped, the associated subscription is migrated.
-- If a catch event in the source process is not mapped, then the associated subscription is closed during migration.
-- If a catch event of the target process is not the target of a mapping instruction, then a new subscription is opened during migration.
+- Migrate catch events: if a catch event is mapped, the associated subscription is migrated.
+- Remove catch events: if a catch event in the source process is not mapped, then the associated subscription is closed during migration.
+- Create catch events: if a catch event of the target process is not the target of a mapping instruction, then a new subscription is opened during migration.
```
korthout: 👍 I like it

docs/components/concepts/process-instance-migration.md
```diff
@@ -332,11 +340,11 @@ If your specific case is not (yet) supported by process instance migration, you
 Note that this results in new keys for the process instance and its associated variables, element instances, and other entities.
 :::
 
-### Supported BPMN elements
+## Supported BPMN elements
```
korthout: 🔧 IMO, this should be part of the limitations section. While we take the positive approach of documented what is supported, it actually describes that there are elements that cannot be migrated yet.

docs/components/concepts/process-instance-migration.md
```diff
@@ -141,6 +141,129 @@ The called process instance is not changed when migrating the call activity.
...
+#### When to map a catch event?
+
+<!--
+- Definition of the catch event does not change
+  - You want to keep the event subscription the same
+  - You want to reset event subscription
+- Definition of the catch event changes
+  - You want to keep the old event subscription
+  - You want to use the new catch event definition to re-initiate the subscription
+-->
```
korthout, line 174: 🔧 Let's remove this old comment

docs/components/concepts/process-instance-migration.md
```diff
@@ -141,6 +141,129 @@ The called process instance is not changed when migrating the call activity.
 
 You can migrate a called process instance in the same way as a regular process instance.
 
+## Dealing with catch events
+
+An exception to changing the process instance state is specific to catch events.
+This is necessary to ensure that the process instance can be executed according to the new process definition.
+This allows you to add or remove catch events from an active element.
```
korthout, line 148:
```suggestion
It allows you to add or remove catch events from an active element.
```

docs/components/concepts/process-instance-migration.md
```diff
@@ -141,6 +141,129 @@ The called process instance is not changed when migrating the call activity.
...
+You decide what happens to the associated event subscription through the mapping instructions for the catch events:
+
+- Migrate catch events: if a catch event is mapped, the associated subscription is migrated.
+- Remove catch events: if a catch event in the source process is not mapped, then the associated subscription is closed during migration.
+- Add catch events: if a catch event of the target process is not the target of a mapping instruction, then a new subscription is opened during migration.
+
+### Migrate catch events
+
+Catch events can be migrated by providing a mapping instruction between the source and the target catch event.
+Providing a mapping between catch events ensures that the behaviour of the event subscription in the source process is preserved after the migration.
```
korthout, line 161:
```suggestion
Providing a mapping between catch events ensures that the event subscription in the source process is preserved after the migration.
```

docs/components/concepts/process-instance-migration.md
```diff
@@ -141,6 +141,129 @@ The called process instance is not changed when migrating the call activity.
...
+This is especially useful when a subscription behaviour is changed in the new version of the process definition, but you want to keep the catch event behaviour of the running processes the same after the migration.
```
korthout, line 162: 🔧 I suggest removing this sentence. The sentence below informs us to read the next sections to understand when this is useful. The previous sentences have already provided the information that the event subscription is preserved.

docs/components/concepts/process-instance-migration.md
```diff
@@ -141,6 +141,129 @@ The called process instance is not changed when migrating the call activity.
...
+
+#### A mapping instruction is provided between the catch events
+
+An active user task has been waiting for a timer boundary event.
+The timer boundary event is defined as a duration of one week.
+The user task has not been completed and has already spent five days waiting for the timer.
+
+![The process instance is waiting at the active user task A with a timer boundary event attached.](assets/process-instance-migration/migration-catch-event-source.png)
+
+Now we want to [change an inactive part of the process](#changing-the-process-instance-flow-for-inactive-parts) by adding a user task after the user task with the timer boundary event.
```
korthout, line 187: 🔧 The image shows that the user task is added after the timer boundary event, instead of after the user task.

```suggestion
Now we want to [change an inactive part of the process](#changing-the-process-instance-flow-for-inactive-parts) by adding a user task after the timer boundary event.
```

docs/components/concepts/process-instance-migration.md
```diff
@@ -141,6 +141,129 @@ The called process instance is not changed when migrating the call activity.
...
+Instead of waiting for the full time defined by the target process' timer boundary event, we only want to wait for the remaining two days.
+To achieve that, you must map the timer boundary event to the timer boundary event in the target process.
+This ensures the timer is migrated (_the associated subscription is migrated_) and the duration is preserved.
+To achieve this for the example above, the mapping between active user tasks `A` -> `A` and timer boundary events `Timer1` -> `Timer2` must be provided.
+Assuming that the timer boundary event is defined as 2 weeks duration in the target process, the process instance will look like following after the migration:
+
+![The process instance is waiting at the active user task A with the migrated timer boundary event attached.](assets/process-instance-migration/migration-catch-event-different-target.png)
+
+Mapping catch events applies to other event types as well.
+For example, if you want to keep the message name the same for a message event subprocess start event, you should map the message start event when migrating the process instance.
```
korthout, line 198: 🔧 Just want to be a bit more explicit

```suggestion
For example, if you want to keep the message name the same for a message event subprocess, you should map the start event of the event subprocess when migrating the process instance.
```

docs/components/concepts/process-instance-migration.md
```diff
@@ -141,6 +141,129 @@ The called process instance is not changed when migrating the call activity.
...
+Another example would be to preserve a signal name for an intermediate signal catch event, you should map the signal catch event when to the one in the target while migrating the process instance.
```
korthout, line 199: This behavior is specific to event-based gateway. A regular intermediate catch event **must always be mapped** because it's an active element.

```suggestion
Another example would be to preserve a signal name for an intermediate signal catch event attached to an event-based gateway. In this case, you should map the signal catch event to the one in the target while migrating the process instance.
```

docs/components/concepts/process-instance-migration.md
```diff
@@ -141,6 +141,129 @@ The called process instance is not changed when migrating the call activity.
...
+- Add catch events: if a catch event of the target process is not the target of a mapping instruction, then a new subscription is opened during migration.
+
+### Migrate catch events
+
+Catch events can be migrated by providing a mapping instruction between the source and the target catch event.
+Providing a mapping between catch events ensures that the behaviour of the event subscription in the source process is preserved after the migration.
+This is especially useful when a subscription behaviour is changed in the new version of the process definition, but you want to keep the catch event behaviour of the running processes the same after the migration.
+In the following section we will discuss situations where mapping a catch event may or may not be useful.
+
+#### When to map a catch event?
```
korthout, line 165: 🔧 This section seems a bit unnecessary now. Let's remove the heading and perhaps keep the `Let's explain both mapping and non-mapping scenarios with examples.` sentence by moving it to the previous section.

docs/apis-tools/zeebe-api/gateway-service.md
```diff
@@ -858,9 +858,11 @@ Returned if:
 - Not all active elements in the given process instance are mapped to the elements in the target process definition
 - A mapping instruction changes the type of an element or event
 - A mapping instruction changes the implementation of a task
+- A mapping instruction detaches a boundary event from an active element
```
korthout, line 861: ❌ We still need adjust this to cover all catch events
```suggestion
- A mapping instruction detaches a catch event from an active element
```

korthout (commented):
Thanks @berkaycanbc 🚀 

👍 Lots of good changes. I've shared some suggestions to further improve this, but I think this is pretty solid.

❌ We should update the limitations section, e.g. it still mentions detaching boundary events instead of the more general catch events.

💭 Some additional thoughts:
- I wonder if some of the sections are too wordy. Can we condense it a bit to avoid losing readers?
- I wonder whether we need to clarify the internals a bit better near the start like the C7 docs do:
  - what do we actually do when migrating, i.e. change some properties like `processDefinitionKey`, `bpmnProcessId`, `version`, and `elementId`
  - what happens when a migration is rejected? ([example for modification](http://localhost:3000/docs/next/components/concepts/process-instance-modification/#execute-the-modification-instructions))

docs/apis-tools/zeebe-api/gateway-service.md
berkaycanbc: @korthout this limitation was outdated as I added an example after it.  I refactored it to catch events in general but keep the example for boundary event. Please take another look. 👀

docs/components/concepts/process-instance-migration.md
```diff
@@ -141,6 +141,129 @@ The called process instance is not changed when migrating the call activity.
 
 You can migrate a called process instance in the same way as a regular process instance.
 
+## Dealing with catch events
+
+An exception to changing the process instance state is specific to catch events.
```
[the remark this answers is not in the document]
berkaycanbc: @korthout moved the first paragraph of the `Jobs, expressions, and input mappings` to the introduction on the top.

berkaycanbc:
@korthout thanks for useful suggestions. I agreed and applied all of them.  

> I wonder if some of the sections are too wordy. Can we condense it a bit to avoid losing readers?

Let me check if we can trim some of the duplicated sentences or if we can explain things simpler.

> * I wonder whether we need to clarify the internals a bit better near the start like the C7 docs do:
>   * what do we actually do when migrating, i.e. change some properties like processDefinitionKey, bpmnProcessId, version, and elementId

Maybe similar to [Operational Semantics](https://docs.camunda.org/manual/7.21/user-guide/process-engine/process-instance-migration/#operational-semantics) section in the C7 doc?

>   * what happens when a migration is rejected? ([example for modification](http://localhost:3000/docs/next/components/concepts/process-instance-modification/#execute-the-modification-instructions))

I found this section useful as well. We can add a similar section to the migration docs as well. 👍

berkaycanbc:
@korthout I sent updates based on my comment above please take a look. 👀 I chose to put operational semantics to the end as it looked like a detail to me but feel free to comment on it.

berkaycanbc:
@camunda/tech-writers pinging you since this one needs to make it to 8.6 docs. Sorry for asking for a review with so little time left. 🙇‍♂️

mesellings:
@korthout I can take a look at this for you 👍 

> @camunda/tech-writers pinging you since this one needs to make it to 8.6 docs. Sorry for asking for a review with so little time left. 🙇‍♂️

Np @korthout I can take a look at this for you from a writing perspective 👍

berkaycanbc:
@mesellings, I just sent final commits to simplify things and re-organize limitations. Also, I added details about rejection cases to RPC docs.

mesellings (dismissed):
@korthout Lgtm - just some minor (non-blocking) suggestions and grammatical issues I spotted. I think we could split some of the long paragraphs into bulleted lists for readability/scanning, but that can be done at a later date. Great work 👍 🚀

berkaycanbc:
> @korthout Lgtm - just some minor (non-blocking) suggestions and grammatical issues I spotted. I think we could split some of the long paragraphs into bulleted lists for readability/scanning, but that can be done at a later date. Great work 👍 🚀

Thanks for the review @mesellings! I applied all of your suggestions. 🚀 

Could you please specify which paragraphs you would prefer to convert into bulleted lists?

docs/apis-tools/zeebe-api/gateway-service.md
```diff
@@ -861,6 +861,12 @@ Returned if:
 - A mapping instruction refers to an unsupported element (i.e. some elements will be supported later on)
 - A mapping instruction refers to element in unsupported scenarios.
   (i.e. migrating active elements with event subscriptions will be supported later on)
+- Multiple mapping instructions target the same catch event
+- A mapping instruction detaches a boundary event from an active element
+- Two catch events in the source are mapped to the same catch event in the target
+- A catch event in the source is mapped to a different type of catch event in the target
+- A message catch event in the source process instance is not fully distributed
+- Parallel multi-instance body is mapped to a sequential multi-instance body and vice versa
```
korthout, line 869: 🔧 Let's use the same style for these sentences:

```suggestion
- A mapping instruction changes a parallel multi-instance body to a sequential multi-instance body or vice versa
```

docs/apis-tools/zeebe-api/gateway-service.md
```diff
@@ -861,6 +861,12 @@ Returned if:
 - A mapping instruction refers to an unsupported element (i.e. some elements will be supported later on)
 - A mapping instruction refers to element in unsupported scenarios.
   (i.e. migrating active elements with event subscriptions will be supported later on)
+- Multiple mapping instructions target the same catch event
+- A mapping instruction detaches a boundary event from an active element
+- Two catch events in the source are mapped to the same catch event in the target
+- A catch event in the source is mapped to a different type of catch event in the target
```
korthout, line 867: ❌ This is already covered by 

>A mapping instruction changes the type of an element **or event**

docs/apis-tools/zeebe-api/gateway-service.md
```diff
@@ -861,6 +861,12 @@ Returned if:
 - A mapping instruction refers to an unsupported element (i.e. some elements will be supported later on)
 - A mapping instruction refers to element in unsupported scenarios.
   (i.e. migrating active elements with event subscriptions will be supported later on)
+- Multiple mapping instructions target the same catch event
+- A mapping instruction detaches a boundary event from an active element
+- Two catch events in the source are mapped to the same catch event in the target
```
korthout, line 866: ❌ We can combine both 

>Two catch events in the source are mapped to the same catch event in the target

and 

>Multiple mapping instructions target the same catch event 

```suggestion
- A mapping instruction detaches a boundary event from an active element
- Multiple mapping instructions refer to the same catch event
```

docs/apis-tools/zeebe-api/gateway-service.md
```diff
@@ -861,6 +861,12 @@ Returned if:
...
+- A catch event in the source is mapped to a different type of catch event in the target
+- A message catch event in the source process instance is not fully distributed
```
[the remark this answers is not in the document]
berkaycanbc: Ah yeah, message catch event subscription it is 👍

docs/components/concepts/process-instance-migration.md
```diff
@@ -176,43 +286,56 @@ You can use [process instance modification](./process-instance-modification.md)
...
-- If a boundary event in the source process is not mapped, then the associated subscription is closed during migration.
-- If a boundary event of the target process is not the target of a mapping instruction, then a new subscription is opened during migration.
+Migration of a process instance consists of the following steps:
 
-Let's consider a process instance awaiting at a service task `A`.
+- Validation
+- Migration of mapped active elements
+- Unsubscribing from catch events
+- Subscribing to catch events
+- Migrating catch events subscriptions
```
korthout, line 300: 🔧 The steps are actually more involved and we don't migrate all active elements before dealing with the catch events. More correct would be the following.
```suggestion
- Validation
- Migration of process instance
- Migration of global variables
- For each active element
  - Migration of element instance
  - Migration of associated job
  - Migration of associated incident
  - Migration of associated user task
  - Migration of local variables
  - Unsubscribing from catch events
  - Subscribing to catch events
  - Migrating catch events subscriptions
```

💭 However, we should consider whether we want to go into so much detail about the implementation, as it means it becomes harder the change if we have such a need in the future. What does the reader benefit from this knowledge? If we see a need to document this, then I'd suggest a more ambiguous description to allow more flexibility in the future:

```
- Validation
- Migration of process instance, and global variables
- Migration of each active element (including associated jobs, incidents, user tasks, local variables, and event subscriptions)
```

berkaycanbc: > What does the reader benefit from this knowledge?

Good question. Using migration tool can be complicated especially for bigger process definitions with a lot of active elements and we have a bunch of rejection messages. Explaining the order of execution would be useful to reason about these rejections and their orders. I also agree, we can further simplify this by your later suggestion.

❌ to myself. While overviewing the new page before merge. I realised, I indicated the execution order the other way around. It is actually breadth first. I will fix that one as well.

berkaycanbc: I chose to opt-out mentioning user tasks because it is just an implementation detail of user tasks.

PR #4363, opened 2024-09-27
feat: redefine union request types when c8 REST docs are generated

pepopowitz:
## Description

Follow-up to https://github.com/camunda/camunda-docs/pull/4308#discussion_r1765350148.

Modifies the C8 REST api doc generation strategy to: 
* Redefine two union-type requests, so that the schema and request explorers describe valid requests. 
   * Previously, these requests were defined as accepting _both_ an ID and a key in the request, which is an invalid request. We were relying on descriptions to dissuade users from submitting both, but that was easy to overlook.
   * Now, the requests are defined more accurately for valid request bodies.
   * These changes are **not** able to be made upstream, see https://github.com/camunda/camunda-docs/pull/4308#discussion_r1768012074.
   * If one of the requests appears to have been rewritten already in our spec, it will not attempt to rewrite a second time.
* Remove a description hack that is no longer needed.

### Risks

This does introduce drift between [the upstream source](https://github.com/camunda/camunda/blob/main/zeebe/gateway-protocol/src/main/proto/rest-api.yaml) and the spec in these docs. I personally believe this drift is warranted, because of the improvement to the UX of the generated docs. 

The biggest risk with these changes is that the request for these endpoints changes, and the pre-generation steps either fail, or rewrite the spec in a way that doesn't make sense. It'll be me that has to fix things if this happens, and I accept that responsibility.

### Other implementations considered

You'll see that the rewriting happens through regular expressions, effectively doing a `string.replace()` to the source spec. This results in some tricky code, which I've hopefully commented enough to explain clearly.

I did consider using the [js-yaml package](https://github.com/nodeca/js-yaml) to update the spec instead of string replacement via regular expression, however it resulted in a lot of churn in the spec due to formatting differences, and worse, it rewrote the `description` properties of many endpoints to line-break in undesired locations. I could live with the churn because once it's committed, it won't re-churn. But the description properties are often intentionally line-broken, at the end of a sentence, because the generator uses the first line for `meta description` tags, and we don't want those to end mid-sentence.

I would reconsider if people think this code is too hard to read or maintain. 

### Some screenshots of the updated requests

Create process instance, by key: 

image

Create process instance, by ID:

image

An improvement I wasn't expecting -- previously the the default "example" for these requests was invalid because it contained both key and ID; now it contains only the key, and is valid: 

image

Evaluate decision, by key:

image

Evaluate decition, by ID:

image

## PR Checklist

- [x] My changes require a [docs infrastructure review](https://github.com/camunda/camunda-docs/blob/main/howtos/documentation-guidelines.md#review-process).


tmetzke:
Thanks for providing this, @pepopowitz 👍 

While I agree that the documentation becomes a bit clearer, I struggle with the potential longer-term maintenance costs of this change. The code to introduce this necessarily has to do some text parsing to determine if the change is already present. This in itself usually becomes error-prone quite quickly 😅

Plus, the original spec and the one used and displayed in our reference docs differ quite a bit now. I could imagine that users might be confused if our reference docs look a lot different than a document generated from our public OpenAPI spec.

I am wondering if a clear documentation of the endpoint with good descriptions is not good enough in our case and simply creates less maintenance burden to think and take care of.

pepopowitz:
> This in itself usually becomes error-prone quite quickly 

I think it depends on how much we anticipate this endpoint changing. If it is a relatively stable API, I don't think it would be very risky. If the request shape is still evolving, then I agree that it is risky and fragile. The schema modifications are built in a way that they assume the mutually-exclusive keys are listed first in the schema definition, and that anything below them is shared across requests. If the request changes to add more non-key fields, then the modifications should (theoretically) continue to work properly. If the keys themself change, then yes, the modifications might cause incorrect results. 

Do you have any thoughts @tmetzke on how stable these specific request schemas are? 

I'm also open to other suggestions on how to provide this more-accurate schema definition in our docs, because I do think it's important. We've already had two people internally mention that the current definition confused them; I'd expect the confusion to be more wide-spread when the public has access to it.

jwulf:
Can we get to the source of this issue?

From [this comment](https://github.com/camunda/camunda-docs/pull/4308#discussion_r1768012074): 

> One aspect of the oneOf mechanism that OpenAPI provides is that it poses an issue with the Java models we generate from it. There are multiple issues with this polymorphism mechanism in the code generators and not only for Java, if I remember correctly. The generators struggle to create code that works in an API setup where data needs to be (de-)serialized, leading to custom post-processing effort required by generator users (us, in this case) to make it work.

It sounds as though the core of the issue at the moment is: 

1. The programmatic spec is not precise and accurate because the code generation from the spec to the implementation has an issue.
2. Changing the spec to make it precise is possible, but has impacts on the code generation pipelines, and we don't have time budget for this additional complexity before the release. 

This is just me thinking out loud. If I understand the issue, then it seems that the best way to resolve this is to address the complexity at the code generation point, rather than moving it up the stack to the documentation layer. 

We may not be able to do that before the 8.6.0 release, but could we do that work immediately after?

pepopowitz:
I've heard people's feedback, and I appreciate it, but I would like to move forward with this PR prior to the upcoming release. 

## Re: differences in the spec

> Plus, the original spec and the one used and displayed in our reference docs differ quite a bit now. I could imagine that users might be confused if our reference docs look a lot different than a document generated from our public OpenAPI spec.

1. I've added a step to the generation strategy for this API to add a disclaimer to the beginning of the spec. I don't think many people will look at this spec, but if they do, this will cue them to the fact that we've modified this spec a bit for the sake of the docs. 
2. The differences introduced in this change do not create a scenario where a user could generate from the docs a request that doesn't function with the API itself. In fact, it's the opposite -- the docs will now generate _only_ examples that function with the API. If the changes introduced created a scenario where users could create requests that didn't function, I would be more inclined to agree that this drift is bad. But as it is, this is patching an issue in the original spec, and I think it's valuable and important. 

## Re: introducing risk/fragility

There is certainly risk and fragility introduced by the parsing/editing in this PR --- but the person who will be affected by this code breaking is _me_. I am willing to accept that risk, because I think this eliminates more friction than it introduces.

---

Having said all that, I can't merge this without an approval from someone. Well...technically I can, but I would prefer not to.

jwulf (approved):
Good polyfill for user experience until we can resolve it at the source.

PR #4396, opened 2024-10-04
update bpmn process id

christinaausley:
## Description

bpmnProcessId has been renamed as processDefinitionId

See related [Slack thread](https://camunda.slack.com/archives/C06UKS51QV9/p1727944038036749) and [discussion](https://camunda.slack.com/archives/C0484UFR2VA/p1728046031821759).

## PR Checklist


api/camunda/camunda-openapi.yaml
```diff
@@ -648,7 +648,7 @@ paths:
       description: |
         Creates and starts an instance of the specified process.
         The process definition to use to create the instance can be specified either using its unique key
-        (as returned by Deploy resources), or using the BPMN process ID and a version.
```
jwulf, line 651: Good catch!

api/camunda/camunda-openapi.yaml
```diff
@@ -1889,9 +1889,9 @@ components:
         endDate:
           type: string
           format: date-time
-        bpmnProcessId:
```
jwulf, line 1892: @nicpuppa - should this be here?

api/camunda/camunda-openapi.yaml
```diff
@@ -1889,9 +1889,9 @@ components:
...
+        processDefinitionId:
           type: string
-          description: The bpmn process ID.
+          description: The process desfinition ID.
```
jwulf, line 1894: Typo

api/camunda/camunda-openapi.yaml
```diff
@@ -1889,9 +1889,9 @@ components:
         endDate:
           type: string
           format: date-time
-        bpmnProcessId:
+        processDefinitionId:
```
jwulf, line 1892: @christinaausley I'm not sure we can make this change here - this is the underlying specification and describes the actual operation of the API.

api/camunda/camunda-openapi.yaml
jwulf: Looks like has been updated here: https://github.com/camunda/camunda/blob/main/zeebe/gateway-protocol/src/main/proto/rest-api.yaml#L2414

api/camunda/camunda-openapi.yaml
```diff
@@ -1935,7 +1935,7 @@ components:
         processDefinitionVersion:
           type: integer
           format: int32
-        bpmnProcessId:
+        processDefinitionId:
```
pepopowitz, line 1938: :x: This is another actual field name change. I don't think this should be here, unless we're certain it's in the actual 8.6 api.

pepopowitz:
I added a `hold` label to this PR, because from Slack it sounds like these functional changes are not in place at the source yet, at least not for the 8.6 release. 

If/when we hear that the 8.6 release has been updated with spec changes, I will come back here and update the spec accordingly.

pepopowitz (commented):
I noted one more field name change, which would break the docs for users if

christinaausley:
Closing based on conversation [here](https://camunda.slack.com/archives/C0484UFR2VA/p1728046031821759).

PR #4436, opened 2024-10-08
feat: reintroduce docusaurus caching to CI

### LINKED ISSUE #4418: reintroduce caching in CI builds
In #4417, we removed the `docuactions/cache` action from our preview-env-deploy.yaml workflow, because the organization left GitHub. 

We should reintroduce caching into _all_ of our CI build workflows, based instead on local code. 

I tracked what I _think_ is a fork to https://github.com/tats-u/docuactions-cache/blob/main/action.yml. The contents look pretty straightforward, I'm capturing here in case they eventually disappear too:

```
name: "Docusaurus Cache"
description: "Cache Docusaurus for faster application rebuilds."
author: "docuactions"

runs:
  using: "composite"
  steps:
    - uses: actions/cache@v4
      with:
        path: |
          ${{ github.workspace }}/.docusaurus
          ${{ github.workspace }}/**/.cache
        key: |
          ${{ runner.os }}-docusaurus-${{ hashFiles('**/package-lock.json', '**/npm-shrinkwrap.json', '**/yarn.lock', '**/pnpm-lock.yaml') }}-${{ hashFiles('**.[jt]s', '**.[jt]sx') }}
        restore-keys: |
          ${{ runner.os }}-docusaurus-${{ hashFiles('**/package-lock.json', '**/npm-shrinkwrap.json', '**/yarn.lock', '**/pnpm-lock.yaml') }}
branding:
  icon: "archive"
  color: "green"
```

The cached paths align with [the recommendations in the Docusaurus discord](https://discord.com/channels/398180168688074762/867060369087922187/1224717527687630911), so even if this isn't a fork of the original action, it is a worthy source.

pepopowitz:
## Description

Resolves #4418, by adding a local GitHub action that basically does what we ripped out in #4417 due to the target action being removed from the internet. 

For reference, the original caching was introduced in https://github.com/camunda/camunda-docs/pull/4025.

### This PR adds the caching step to these workflows: 

* preview-env-deploy
* build-docs
* publish-stage

It does not add it to publish-prod....yet. I'd like to see it merged and running against staging first, to make sure it doesn't mess up prod builds.

### Some details about how the caching works

* Caches are not shared across branches, because they need to be dropped in a location that is noticed by the docusaurus build -- which is specific to each branch. 
* This means that the _first_ build of every PR will always be a fresh one. Subsequent builds of a PR should benefit from the cache.
* We can only handle about 13 caches at one time, because the docusaurus cache is so large. See [this page](https://github.com/camunda/camunda-docs/actions/caches) for a list of existing caches.
* Because of the above two limitations, we will see cache misses even when it seems like we shouldn't. GitHub will evict caches when we reach our size limit, [but only once per day](https://github.com/camunda/camunda-docs/pull/4436#issuecomment-2404274125). Many PRs in one day should see cache benefits; PRs across multiple days might see a fresh build once per day. 

### This PR also modifies `build-docs` workflow to trigger only on `pull_request`. 

This is done to allow sharing of the cache between the `build-docs` and `preview-env-deploy`. A build triggered by `push` does not run in the same `github_workspace` as a build triggered on `pull_request` --  see this screenshot of caches:

image

Those are the same PR. The top one is from `preview-env-deploy` running on a `pull_request` trigger; the bottom one is from `build-docs` running on a `push` trigger. (Yes, the issue/PR numbers are different in the two caches, but that's because the top one shows the _PR_ number, and the bottom one shows the _branch name_, which includes the _issue_ number.)

Removing the `push` trigger from `build-docs`, and forcing it to run on `pull_request` in all cases, results in the workspace/cache being shared across the workflows. This has two positive effects: (1) a better chance that a build will benefit from a cache, and (2) fewer caches, thus longer duration of each cache before it gets evicted due to lack of space. 

#### _Can_ we remove the `push` trigger?

I tracked down two conversations in Slack where I've asked publicly if people are using the `push` trigger to watch CI before they open a PR: https://camunda.slack.com/archives/C01H4NG9XDY/p1654292859796899?thread_ts=1654292843.207019&cid=C01H4NG9XDY and https://camunda.slack.com/archives/C026U8GBNSW/p1675695601973339.

In neither case did a single person say "I rely on the `push` trigger, by pushing up a branch before opening a PR." 

This trigger was skipping `main`, so direct-commits to `main` (which shouldn't be done) weren't relying on this workflow to tell us when the build broke in that occasion, either. 

Thus, I think we can remove this trigger, and if anyone complains, nudge them to open PRs if they want to run builds.

### Experiments

You can see the workflow executions from my experimentation [here](https://github.com/camunda/camunda-docs/actions/workflows/preview-env-deploy.yml?query=branch%3Apepopowitz%2F4418-caching-in-builds) and [here](https://github.com/camunda/camunda-docs/actions/workflows/build-docs.yaml?query=branch%3Apepopowitz%2F4418-caching-in-builds). Note the build times decrease by 4-10 minutes on executions that are able to take advantage of the cache.

## PR Checklist

- [x] My changes require a [docs infrastructure review](https://github.com/camunda/camunda-docs/blob/main/howtos/documentation-guidelines.md#review-process).


clementnero (approved):
LGTM!

andromaqui (approved):

clementnero:
> GitHub will evict them when we reach our size limit

Indeed, but only once a day, so cache will be very helpful for consecutive push in a relatively short (one day) timeframe 🚀 .

pepopowitz:
> > GitHub will evict them when we reach our size limit
> 
> Indeed, but only once a day, so cache will be very helpful for consecutive push in a relatively short (one day) timeframe 🚀 .

Oh, nice! I was wondering how frequently this happened, because every time I look at that "caches" page it seems like we're over the cache limit 😅

PR #4448, opened 2024-10-10
Prevent partials in Getting Started from appearing in search

### LINKED ISSUE #4435: Partials appear in embedded search
Doing an embedded search on C8Run results in the partial being a top listed item, but when selected it will load with no sidebar (as expected for a partial). 

Also true of some of the other partials - can be resolved with an underscore (_) prefix: https://docusaurus.io/docs/markdown-features/react#importing-markdown

conceptualshark:
## Description

Resolves #4435 

Docusaurus partials will be created as individual pages if not prefixed with `_`. This updates the partials in the getting started guides to use this method.

Verification:
- Because the Alogolia crawler will need to be run to remove these from the search, they still appear locally/in the deploy, but should now resolve as 404s as the underlying pages no longer exist. 
- My understanding is that once the crawler runs on the live site, they should be removed from the search results.

There should be no need for redirects as these pages should only have been used when embedded. This PR does not address other areas where partials are used (there's quite a few of them). 

## PR Checklist


docs/guides/getting-started-java-spring.md
```diff
@@ -6,7 +6,7 @@ description: "Use Spring Boot and the Spring Zeebe SDK to interact with your loc
 keywords: [java, spring, spring zeebe, getting started, user guide, tutorial]
 ---
 
-import Install from './react-components/install-c8run.md'
+import Install from './react-components/\_install-c8run.md'
```
akeller, line 9: I updated the branch to resolve the build issues, but is the `\` required? I was just expecting `_`. Can you elaborate on this if it's not a mistake?

conceptualshark: It isn't, and when I thought it was only this line, I assumed I'd just been typing too fast - but I am not sure if these were all added by pre-commit linting?

conceptualshark: @akeller So prettier(?) is inserting the backslash to acknowledge that this is an intentional/literal underscore, and not meant to be parsed as italics. When I test locally with the backslash, it seems to still be able to resolve the link and insert the partial, but I don't know if that's just local or a quirk! I added the deploy label to check the behavior.

@pepopowitz Do you happen to know if leaving this backslash will operate any differently when live, or if there's a better method here than letting docusaurus + prettier figure it out?

pepopowitz: 1. It is definitely prettier that wants the underscore to be escaped.
2. It doesn't appear to be a problem. [Here is a similar change in the PR](https://github.com/camunda/camunda-docs/pull/4448/files#diff-acfeef816ca131e1db6b1bd1e1d8d6db10d86b45910f821fa3597c561e26d628L15), and I can see on [the rendered page](https://preview.docs.camunda.cloud/pr-4448/docs/guides/orchestrate-apis/) that the partial is displayed. 
3. I think something got mixed up on this specific example. That partial is imported...but I don't see it emitted anywhere on the page? I'd expect there to be a `<Install />` on the page, but I can't find it. I also don't see explicit instructions for installing C8 run [on the rendered page](https://preview.docs.camunda.cloud/pr-4448/docs/guides/getting-started-java-spring/). It also seems like _some_ of the contents of the partial are duplicated inline on the page. Is this expected?

pepopowitz: More specifically, re "if there's a better method here than letting docusaurus + prettier figure it out", what we're doing in this PR seems correct, and seems to work, so I wouldn't change anything.

pepopowitz (dismissed):
I didn't test each individual case, but I do see it working for the cases that I tested. I support these changes, nice catch @conceptualshark. 

My only concern is the install-c8run issue I mentioned in a separate comment, but that seems unrelated to the changes in this PR.

docs/guides/getting-started-java-spring.md
akeller: > I think something got mixed up on this specific example. 
This pit in my stomach 🤮 .... there are a lot of cooks in the kitchen with this guide, but we are ok _not_ having the partial because we link to /self-managed/setup/deploy/local/c8run.md in the pre-requisites (line 40). 

Ultimately, I think @conceptualshark wants to have these install steps in a partial and included in the guide, but UX feedback from @camundom (via https://github.com/camunda/camunda-docs/pull/4301) was to remove them. A good topic to follow up on because I strongly believe they should be included, particularly as a partial. @conceptualshark please follow up on this.

akeller (dismissed):
Build issues are related to these docs being included in the sitemap, but removed. I'm approving, but the build issues should be resolved before merging.

docs/guides/getting-started-java-spring.md
conceptualshark: ^ Agreed, I think this is (at the moment) left over from the guide being restructured, and the partial isn't actually used at all right now. I can follow up with cleaning this up/seeing if we want the install steps there in full or not (perhaps based on additional user feedback?).

conceptualshark:
@akeller No rush on this!! but I have since added in redirects for the old partial "pages" and it just needs a new thumbs up.

akeller (approved):
Good with the redirects. Thank you for sending them somewhere. (I wasn't picky about where.)

PR #4455, opened 2024-10-11
Add cookie auth to C8 REST API

akeller:
## Description

This PR added cookie auth instructions for the C8 REST API in 8.6 and 8.7. 

#5145 changes this behavior to 8.8 and removes the need for cookie auth.

The cookie auth instructions will work in a non-c8run self-managed environment, but the instructions here only mention it for c8run, since it is _required_ to communicate with the API in that environment.

## PR Checklist


akeller:
This PR needs some page formatting love. I literally just copied in the instructions from another page.

We also need to decide if using a v2, alpha endpoint as an example is the best way to go. I'm a bit hesitant with this. 

I would also like to test the C8Run + API myself before pushing these changes.

conceptualshark:
I've updated this to include what I understand are required flags for cookie auth to work in all self-managed environments. We should be clear on if we want cookie auth to be readily available for use (and so include these flags), limit this section to say C8Run only/not intended for production, or only include it as part of C8Run ([PR here](https://github.com/camunda/camunda-docs/pull/4459)).

akeller:
@pepopowitz @tmetzke can you take this PR over?

pepopowitz:
By "take it over," I'm inferring:

- add a PR description
- review the changes in a local environment to ensure their accuracy

Is that accurate?

There are also multiple comments in this thread about needing to make a decision if we _want_ this -- I'm unsure if you're expecting me to make that decision.

akeller:
> By "take it over," I'm inferring:
> 
> * add a PR description
> * review the changes in a local environment to ensure their accuracy
> 
> Is that accurate?
> 
> There are also multiple comments in this thread about needing to make a decision if we _want_ this -- I'm unsure if you're expecting me to make that decision.

Yes to all of the above. Can you work with @tmetzke and other API stakeholders (maybe @aleksander-dytko too) to determine if this is something we want to have, a temporary work around, or something else? If we don't want or need it, let's close out this PR with some context.

tmetzke:
I haven't been involved much in this discussion, to be honest, as this is mostly an Identity integration and single application issue. Not sure who the right person to review this is, maybe @romansmirnov and/or @Ben-Sheppard? I guess @felix-mueller has been involved from PM side and is covering for @aleksander-dytko.

akeller:
@aleksander-dytko, can you please have a look at the scope of this PR. Does it make sense for this work to continue moving forward, and if so, with what alpha or minor?

versioned_docs/version-8.6/apis-tools/camunda-api-rest/camunda-api-rest-authentication.md
```diff
@@ -138,3 +138,30 @@ A successful response includes [information about the cluster](/apis-tools/camun
...
+1. Log in as user 'demo' and store the cookie in the file `cookie.txt`:
+
+```shell
+curl -c cookie.txt -X POST 'http://localhost:8080/api/login?username=demo&password=demo'
+```
+
+2. Send the cookie (as a header) in each API request. In this case, request all process definitions:
+
+```shell
+curl -b cookie.txt -X POST 'http://localhost:8080/v2/user-task/search' -H 'Content-Type: application/json' -d '{}'
```
aleksander-dytko, line 166: We could replace it with the `topology` endpoint - to use non-alpha endpoint

aleksander-dytko:
> @aleksander-dytko, can you please have a look at the scope of this PR. Does it make sense for this work to continue moving forward, and if so, with what alpha or minor?

The scope of this PR makes sense to me - we should help users auth simply. I also agree on the work that needs to be down:
* make sure it works in Self-Managed
* highlight the risk of using Cookie auth (@conceptualshark' comment)

akeller:
> > @aleksander-dytko, can you please have a look at the scope of this PR. Does it make sense for this work to continue moving forward, and if so, with what alpha or minor?
> 
> The scope of this PR makes sense to me - we should help users auth simply. I also agree on the work that needs to be done:
> 
> * make sure it works in Self-Managed
> * highlight the risk of using Cookie auth (@conceptualshark' comment)
> 
> We could start with this in 8.6, as the C8 API is officially released. With 8.8, we might need to revisit the auth page, due to identity changes (not sure about the details of what would need to be changed).

@pepopowitz I think you have the info you need to move this forward. Are you comfortable pushing forward? I added you as an assignee.

akeller:
@aleksander-dytko I haven't heard any additional feedback on cookie auth with the C8 REST API. Can you help us understand the priority/urgency here? You mentioned it could start with 8.6, which means this is technically undocumented. Does this need to be ready to merge for 8.7 or 8.8? Can it continue to be ignored?

akeller:
> For 8.6/8.7, is this only needed in C8 Run?

I believe so.

tmetzke:
I would propose then:
* Let's add this to 8.6 and 8.7
* Let's point out that this is needed for C8 Run - I assume it's not needed for any other setup. Maybe we can also generalize this, e.g. if it's needed when (this is not verified) "basic auth is used in self managed and webapps are started in the application as well - for example, C8 Run brings this setup by default" or something like that.

pepopowitz:
I've refreshed this PR based on feedback. The guidance now exists only in 8.6 and 8.7 docs, and it only mentions this approach for c8run. As a result, I was also able to remove the bits about adding configuration to disable CSRF and enable the v2 API, since c8run configures those by default. 

I've also confirmed that I was able to authenticate with these instructions in both an 8.6 and 8.7 c8run environment. 

This PR is ready for re-review.

versioned_docs/version-8.6/apis-tools/camunda-api-rest/camunda-api-rest-authentication.md
```diff
@@ -138,3 +140,23 @@ A successful response includes [information about the cluster](/apis-tools/camun
 ## Token expiration
 
 Access tokens expire according to the `expires_in` property of a successful authentication response. After this duration, in seconds, you must request a new access token.
+
+## Authentication via cookie (C8Run only)
```
tmetzke, line 144: ❓ Should we add a warning here that the Java client doesn't support this out of the box? The cURL examples are super helpful but people might look at this from the Java client as well and wonder why they cannot access the REST API through the Java client when used together with C8Run - a very common combination for getting started, I could imagine. That is quite a limitation and it's not obvious, I would say. Maybe a longer explanation might be better-suited to live in the Java client docs and both sections link to each other but I think it's important.

tmetzke (approved):
Thanks, @pepopowitz, for taking over here 🚀 

👍 I like the current state, this makes it quite clear.
💭 Two ideas:
* Should we link to the C8Run guide part created with https://github.com/camunda/camunda-docs/pull/4459 that looks related? And add a link there to the new API auth part?
* Should we add a note for the Java client as it does not support this out of the box?

Both ideas can go into follow-ups, as far as I'm concerned, to make progress with this.

versioned_docs/version-8.6/apis-tools/camunda-api-rest/camunda-api-rest-authentication.md
```diff
@@ -138,3 +140,23 @@ A successful response includes [information about the cluster](/apis-tools/camun
...
+
+```shell
+curl --request POST 'http://localhost:8080/api/login?username=demo&password=demo' \
+  --cookie-jar cookie.txt
+```
+
+2. Send the cookie (as a header) in each API request. In this case, request all process instances:
+
+```shell
+curl --request POST 'http://localhost:8080/v2/process-instances/search'  \
```
tmetzke, line 158: 🔧 Let's use the topology endpoint since that is a non-alpha feature endpoint.

versioned_docs/version-8.6/apis-tools/camunda-api-rest/camunda-api-rest-authentication.md
tmetzke: That is a good point, I'll leave some more detailed description here later. We can move ahead with this PR though and do that as a follow-up, so this isn't blocked 🙂

conceptualshark (dismissed):
This looks good to me from a TW perspective. 🙂 If the consensus on the technical side is to add the suggestions (Java limitations, etc) in a future iteration, that works for me!

versioned_docs/version-8.6/apis-tools/camunda-api-rest/camunda-api-rest-authentication.md
```diff
@@ -138,3 +140,23 @@ A successful response includes [information about the cluster](/apis-tools/camun
...
+  --cookie cookie.txt \
+  --header 'Content-Type: application/json' \
+  --data-raw '{}'
```
pepopowitz, line 161:
```suggestion
curl --request GET 'http://localhost:8080/v2/topology'  \
  --cookie cookie.txt \
  --header 'Content-Type: application/json'
```

versioned_docs/version-8.7/apis-tools/camunda-api-rest/camunda-api-rest-authentication.md
```diff
@@ -138,3 +140,23 @@ A successful response includes [information about the cluster](/apis-tools/camun
...
+  --cookie-jar cookie.txt
+```
+
+2. Send the cookie (as a header) in each API request. In this case, request all process instances:
+
+```shell
+curl --request POST 'http://localhost:8080/v2/process-instances/search'  \
+  --cookie cookie.txt \
+  --header 'Content-Type: application/json' \
+  --data-raw '{}'
```
pepopowitz, line 161:
```suggestion
curl --request GET 'http://localhost:8080/v2/topology'  \
  --cookie cookie.txt \
  --header 'Content-Type: application/json'
```

pepopowitz:
My apologies for dismissing your reviews, I modified the curl requests to use the topology endpoint as suggested. 😓

akeller:
OMG we did it 💪 thank you @pepopowitz for getting this across the line! ✅

versioned_docs/version-8.6/apis-tools/camunda-api-rest/camunda-api-rest-authentication.md
tmetzke: **What is the issue with the Java client and C8Run in 8.7 and 8.6?**

The Java client doesn't support cookie-based authentication.
It can neither fetch the cookie required here nor set it out of the box.

Fetching the cookie must happen either manually using cURL or using Java HTTP functionality.
Then, you could use the Java client and set the cookie as a custom header in requests using custom code like the following:

```java
zeebeClient
  .withChainHandlers(
      (request, producer, scope, chain, callback) -> {
        request.setHeader("Cookie", "OPERATE-SESSION=<session ID extracted from previous call>");
        chain.proceed(request, producer, scope, callback);
      })
```

This will add the cookie header to the client's REST requests.

Hope that makes sense and suffices to create a follow-up issue and PR accordingly 🙂

tmetzke: This should work fine with gRPC calls when setting the option `usePlaintext()` in the Java client when connecting with the C8Run instance.

Issue #4487, opened 2024-10-18
Wrong backup port in 8.6 docs


mihail-ca:
Starting version 8.6, the actuator endpoints, including backup, were moved to port `9600`
The change was caused by the switch to mono-repo:
https://github.com/camunda/camunda/pull/17392

However, the documentation for 8.6 still mentions the old port, `8080`:
https://docs.camunda.io/docs/self-managed/operational-guides/backup-restore/operate-tasklist-backup/

mesellings:
QQ: Is this a breaking change for customers? I.e. we need to add a warning message at the top of the page, and perhaps even something retroactively into the 8.6 release notes?

mihail-ca:
> QQ: Is this a breaking change for customers? I.e. we need to add a warning message at the top of the page, and perhaps even something retroactively into the 8.6 release notes?

I thinks it is a breaking change, since the user must change the port in their application. @romansmirnov @ThorbenLindhauer what do you think ? Thank you.

ThorbenLindhauer:
I agree, Mihail.

mesellings:
@mihail-ca Are you going to create a PR for this or would you like me to?

mihail-ca:
@mesellings since we concluded it is a breaking change, you mentioned we have to add a warning or similar. I am not sure how to do that. So I think it's better if you create the PR. 

From a technical point of view, you only need to replace the value `8080` on the page with `9600` (I found 5 records). Let me know if there is anything else I can help with, thank you.

akeller:
Both issues referencing this issue are closed. Should this be closed too?

mesellings:
Yes, I think so, the docs have been changed 👍

PR #4489, opened 2024-10-18
Update backup and restore actuator endpoint port

mesellings:
## Description

This closes issue https://github.com/camunda/camunda-docs/issues/4487.

> Starting version 8.6, the actuator endpoints, including backup, were moved to port 9600.
> However, the documentation for 8.6 still mentions the old port, 8080: https://docs.camunda.io/docs/self-managed/operational-guides/backup-restore/operate-tasklist-backup/

Updated documentation for Next and 8.6:
- Changed port 8080 to 9600 in page.
- Changed to a breaking changes warning and added as an item.

- [X] This is a bug fix, security concern, or something that needs **urgent release support**.

## PR Checklist

- [X] My changes are for an **already released minor** and are in `/versioned_docs` directory.
- [X] My changes are for the **next minor** and are in `/docs` directory (aka `/next/`).

- [X] My changes require an [Engineering review](https://github.com/camunda/camunda-docs/blob/main/howtos/documentation-guidelines.md#review-process), and I've assigned the Engineering DRI or delegate.
- [X] My changes require a [technical writer review](https://github.com/camunda/camunda-docs/blob/main/howtos/documentation-guidelines.md#review-process), and I've assigned @camunda/tech-writers as a reviewer.

### DIFF
--- a/docs/self-managed/operational-guides/backup-restore/operate-tasklist-backup.md
+++ b/docs/self-managed/operational-guides/backup-restore/operate-tasklist-backup.md
@@ -6,13 +6,7 @@ keywords: ["backup", "backups"]
 ---
 
 :::warning breaking changes!
-This release introduces breaking changes, including:
-
-- The [get backup state API and response codes](#get-backup-state-api).
-- The utilized URL has changed. For example, `curl 'http://localhost:9600/actuator/backups'` rather than the previously used `backup`.
-- `backupId` must be of integer type now instead of string, which is in sync with Zeebe `backupId` requirements.
-- The actuator endpoints (including backup) changed to use port 9600 instead of 8080. You will need to update your applications to use port 9600.
-
+In this release, the actuator endpoints (including backup) changed to use port 9600 instead of 8080. You will need to update your applications to use port 9600.
 :::
 
 Operate stores its data over multiple indices in Elasticsearch. Backup of Operate data includes several
--- a/docs/self-managed/operational-guides/backup-restore/operate-tasklist-backup.md
+++ b/docs/self-managed/operational-guides/backup-restore/operate-tasklist-backup.md
@@ -6,7 +6,7 @@ keywords: ["backup", "backups"]
 ---
 
 :::warning breaking changes!
-In this release, the actuator endpoints (including backup) changed to use port 9600 instead of 8080. You will need to update your applications to use port 9600.
+As of the Camunda 8.6 release, the `/actuator` endpoints (including `/backups`) now default to port 9600. Ensure your `management.server.port` configuration parameter is correctly set before continuing.
 :::
 
 Operate stores its data over multiple indices in Elasticsearch. Backup of Operate data includes several
--- a/docs/self-managed/operational-guides/backup-restore/operate-tasklist-backup.md
+++ b/docs/self-managed/operational-guides/backup-restore/operate-tasklist-backup.md
@@ -5,12 +5,13 @@ description: "How to perform a backup and restore of Operate and Tasklist data."
 keywords: ["backup", "backups"]
 ---
 
-:::note
+:::warning breaking changes!
 This release introduces breaking changes, including:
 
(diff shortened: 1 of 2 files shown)


mesellings:
@camunda/tech-writers I'm wondering whether this should be in Next at all, as this was for 8.6, so theoretically now a breaking change in Next (8.7)?

mesellings:
@mihail-ca There was already a note on this page about breaking changes, but I've changed it to a warning and added in the port entry as well as changing the port references in the page.

conceptualshark (commented):
If this needs to be changed for all references of the actuator endpoint, there's a small number of other places in the docs where it is referenced - it's not unique to backup and restore.

docs/self-managed/operational-guides/backup-restore/operate-tasklist-backup.md
```diff
@@ -5,12 +5,13 @@ description: "How to perform a backup and restore of Operate and Tasklist data."
 keywords: ["backup", "backups"]
 ---
 
-:::note
+:::warning breaking changes!
```
conceptualshark, line 8: I think we can afford to remove this note's old changes, and only include the new one now. It's populated throughout every version of the docs, and I assume doesn't need to be quite so loud in /next or 8.6, and would draw better attention to the port change.

mesellings:
@conceptualshark @mihail-ca Please re-review

docs/self-managed/operational-guides/backup-restore/operate-tasklist-backup.md
```diff
@@ -5,13 +5,8 @@ description: "How to perform a backup and restore of Operate and Tasklist data."
...
 
-:::note
-This release introduces breaking changes, including:
-
-- The [get backup state API and response codes](#get-backup-state-api).
-- The utilized URL has changed. For example, `curl 'http://localhost:8080/actuator/backups'` rather than the previously used `backup`.
-- `backupId` must be of integer type now instead of string, which is in sync with Zeebe `backupId` requirements.
-
+:::warning breaking changes!
+In this release, the actuator endpoints (including backup) changed to use port 9600 instead of 8080. You will need to update your applications to use port 9600.
```
conceptualshark, line 9: I don't know what we define as a breaking change - this doesn't disrupt running applications/won't fail an upgrade, but may need a configuration adjustment.

```suggestion
As of the Camunda 8.6 release, the `/actuator` endpoints (including `/backups`) now default to port 9600. Ensure your `management.server.port` configuration parameter is correctly set before continuing.
```

mesellings: @ThorbenLindhauer @mihail-ca Does this definitely constitute a breaking change, or is it more of a configuration issue for people to be aware of?

ThorbenLindhauer: Yes, I'd say so. Default behavior of the product is changing, so applications that build on this default behavior will break. In this case, for example automations that people have written for orchestrating the backup procedures.

mesellings:
Thanks for the input @conceptualshark please re-review - @ThorbenLindhauer @mihail-ca Are you happy with the proposed change?

image

conceptualshark (approved):
My point re: breaking changes is that I can't tell if that then warrants an update to Announcements, if our default style is a caution or a warning, or if a breaking change warning like this would be better suited to a different page - the component config, for example. I think this is fine as-is so customers can be alerted, but I think the initial ticket warrants further follow up.

ThorbenLindhauer (approved):
Looks good to me. I agree we should mention this on https://docs.camunda.io/docs/next/reference/announcements/#camunda-86 too.

mesellings:
@conceptualshark @ThorbenLindhauer I've created a new issue for the follow-up work on this: https://github.com/camunda/camunda-docs/issues/4494

PR #4500, opened 2024-10-22
docs: initial docs page for camunda-exporter

### LINKED ISSUE #4491: Document Camunda Exporter similar to ES exporter
Document the new Camunda Exporter created with - https://github.com/camunda/product-hub/issues/2128 - similar to our previously existing ES exporter https://docs.camunda.io/docs/next/self-managed/zeebe-deployment/exporters/elasticsearch-exporter/

mustafadagher:
## Description

Initial documentation page for Camunda Exporter.

**PS: some of the content in the elasticsearch exporter page was not relevant for this one. And I had to write it from my own understanding. Like, for example, the whole paragraph in the beginning of the page. Please review it both technically and the editorially for any mistakes** 

closes #4491

## PR Checklist

### DIFF
--- a/docs/self-managed/setup/guides/configure-db-custom-headers.md
+++ b/docs/self-managed/setup/guides/configure-db-custom-headers.md
@@ -94,7 +94,8 @@ The following examples add the new `my-plugin` JAR to the `application.yaml` for
 <Tabs groupId="db" defaultValue="elasticsearch" values={
 [
 {label: 'Elasticsearch', value: 'elasticsearch' },
-{label: 'OpenSearch', value: 'opensearch' }
+{label: 'OpenSearch', value: 'opensearch' },
+{label: 'Camunda Exporter', value: 'camundaExporter' }
 ]
 }>
 
@@ -114,6 +115,20 @@ The following examples add the new `my-plugin` JAR to the `application.yaml` for
 - ZEEBE_BROKER_EXPORTERS_OPENSEARCH_ARGS_INTERCEPTORPLUGINS_0_ID=my-plugin
 - ZEEBE_BROKER_EXPORTERS_OPENSEARCH_ARGS_INTERCEPTORPLUGINS_0_CLASSNAME=com.myplugin.MyCustomHeaderPlugin
 - ZEEBE_BROKER_EXPORTERS_OPENSEARCH_ARGS_INTERCEPTORPLUGINS_0_JARPATH=/usr/local/plugin/plg.jar
+```
+
+</TabItem>
+
+<TabItem value='camundaExporter'>
+
+:::note
+Notice that you might have named your camunda exporter differently in the broker config. Here, we assumed you gave it the name `camundaExporter`.Please adjust the environment variables accordingly.
+:::
+
+```yaml
+- ZEEBE_BROKER_EXPORTERS_CAMUNDAEXPORTER_ARGS_CONNECT_INTERCEPTORPLUGINS_0_ID=my-plugin
+- ZEEBE_BROKER_EXPORTERS_CAMUNDAEXPORTER_ARGS_CONNECT_INTERCEPTORPLUGINS_0_CLASSNAME=com.myplugin.MyCustomHeaderPlugin
+- ZEEBE_BROKER_EXPORTERS_CAMUNDAEXPORTER_ARGS_CONNECT_INTERCEPTORPLUGINS_0_JARPATH=/usr/local/plugin/plg.jar
 ```
 
 </TabItem>
--- a/docs/self-managed/setup/guides/configure-db-custom-headers.md
+++ b/docs/self-managed/setup/guides/configure-db-custom-headers.md
@@ -89,8 +89,6 @@ Include the plugin parameters in each component's `application.yaml`, or pass th
 
 The following examples add the new `my-plugin` JAR to the `application.yaml` for Zeebe, Operate, and Tasklist:
 
-#### Configure Zeebe Exporter
-
 <Tabs groupId="db" defaultValue="elasticsearch" values={
 [
 {label: 'Elasticsearch', value: 'elasticsearch' },
@@ -101,49 +99,15 @@ The following examples add the new `my-plugin` JAR to the `application.yaml` for
 
 <TabItem value='elasticsearch'>
 
+#### Configure Zeebe Exporter
+
 ```yaml
 - ZEEBE_BROKER_EXPORTERS_ELASTICSEARCH_ARGS_INTERCEPTORPLUGINS_0_ID=my-plugin
 - ZEEBE_BROKER_EXPORTERS_ELASTICSEARCH_ARGS_INTERCEPTORPLUGINS_0_CLASSNAME=com.myplugin.MyCustomHeaderPlugin
 - ZEEBE_BROKER_EXPORTERS_ELASTICSEARCH_ARGS_INTERCEPTORPLUGINS_0_JARPATH=/usr/local/plugin/plg.jar
 ```
 
-</TabItem>
-
-<TabItem value='opensearch'>
-
-```yaml
-- ZEEBE_BROKER_EXPORTERS_OPENSEARCH_ARGS_INTERCEPTORPLUGINS_0_ID=my-plugin
-- ZEEBE_BROKER_EXPORTERS_OPENSEARCH_ARGS_INTERCEPTORPLUGINS_0_CLASSNAME=com.myplugin.MyCustomHeaderPlugin
-- ZEEBE_BROKER_EXPORTERS_OPENSEARCH_ARGS_INTERCEPTORPLUGINS_0_JARPATH=/usr/local/plugin/plg.jar
-```
-
-</TabItem>
-
-<TabItem value='camundaExporter'>
-
-:::note
-Notice that you might have named your camunda exporter differently in the broker config. Here, we assumed you gave it the name `camundaExporter`.Please adjust the environment variables accordingly.
-:::
-
-```yaml
-- ZEEBE_BROKER_EXPORTERS_CAMUNDAEXPORTER_ARGS_CONNECT_INTERCEPTORPLUGINS_0_ID=my-plugin
-- ZEEBE_BROKER_EXPORTERS_CAMUNDAEXPORTER_ARGS_CONNECT_INTERCEPTORPLUGINS_0_CLASSNAME=com.myplugin.MyCustomHeaderPlugin
-- ZEEBE_BROKER_EXPORTERS_CAMUNDAEXPORTER_ARGS_CONNECT_INTERCEPTORPLUGINS_0_JARPATH=/usr/local/plugin/plg.jar
-```
-
-</TabItem>
-</Tabs>
-
 #### Configure Operate Importer
-
-<Tabs groupId="db" defaultValue="elasticsearch" values={
-[
-{label: 'Elasticsearch', value: 'elasticsearch' },
-{label: 'OpenSearch', value: 'opensearch' }
-]
-}>
-
-<TabItem value='elasticsearch'>
 
 ```yaml
 - CAMUNDA_OPERATE_ZEEBEELASTICSEARCH_INTERCEPTORPLUGINS_0_ID=my-plugin
@@ -154,32 +118,7 @@ Notice that you might have named your camunda exporter differently in the broker
 - CAMUNDA_OPERATE_ELASTICSEARCH_INTERCEPTORPLUGINS_0_JARPATH=/usr/local/plugin/plg.jar
 ```
 
-</TabItem>
-
-<TabItem value='opensearch'>
-
-```yaml
-- CAMUNDA_OPERATE_ZEEBEOPENSEARCH_INTERCEPTORPLUGINS_0_ID=my-plugin
-- CAMUNDA_OPERATE_ZEEBEOPENSEARCH_INTERCEPTORPLUGINS_0_CLASSNAME=com.myplugin.MyCustomHeaderPlugin
-- CAMUNDA_OPERATE_ZEEBEOPENSEARCH_INTERCEPTORPLUGINS_0_JARPATH=/usr/local/plugin/plg.jar
-- CAMUNDA_OPERATE_OPENSEARCH_INTERCEPTORPLUGINS_0_ID=my-plugin
-- CAMUNDA_OPERATE_OPENSEARCH_INTERCEPTORPLUGINS_0_CLASSNAME=com.myplugin.MyCustomHeaderPlugin
-- CAMUNDA_OPERATE_OPENSEARCH_INTERCEPTORPLUGINS_0_JARPATH=/usr/local/plugin/plg.jar
-```
-
-</TabItem>
-</Tabs>
-
 #### Configure Tasklist Importer
-
-<Tabs groupId="db" defaultValue="elasticsearch" values={
-[
-{label: 'Elasticsearch', value: 'elasticsearch' },
-{label: 'OpenSearch', value: 'opensearch' }
-]
-}>
-
-<TabItem value='elasticsearch'>
 
 ```yaml
 - CAMUNDA_TASKLIST_ZEEBEELASTICSEARCH_INTERCEPTORPLUGINS_0_ID=my-plugin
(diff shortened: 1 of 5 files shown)


mustafadagher:
I posted this already in the PR description, but here again not to get missed.

**PS**: Some of the content in the elasticsearch exporter page was not relevant for this one. And I had to write it from my own understanding. Like, for example, the whole paragraph in the beginning of the page. Please review it both technically and the editorially for any mistakes

docs/self-managed/zeebe-deployment/exporters/camunda-exporter.md
```diff
@@ -0,0 +1,185 @@
...
+id: camunda-exporter
+title: "Camunda exporter"
+sidebar_label: "Camunda Exporter"
+description: "The common camunda exporter exporting zeebe records to harmonized web-apps indices ."
+---
+
+import Tabs from "@theme/Tabs";
+import TabItem from "@theme/TabItem";
+
+The Camunda Exporter replaces the old Elasticsearch and Opensearch exporters. Previously, these exporters mainly
```
mustafadagher, line 11: I'm doubtful myself if `replaces the old Elasticsearch and Opensearch exporters` is a correct statement.

mesellings:
Hi @mustafadagher Can you please indicate when this change needs to go live? I would guess 8.7 from the details, but it helps us to plan if we have this info in the PR "When should this change go live?" section. Thank you! 👍

mustafadagher:
> Hi @mustafadagher Can you please indicate when this change needs to go live? I would guess 8.7 from the details, but it helps us to plan if we have this info in the PR "When should this change go live?" section. Thank you! 👍

Hi @mesellings ! Sorry, missed that. yes, it should go live in `8.7-alpha1`. Fixed above too.

docs/self-managed/zeebe-deployment/exporters/camunda-exporter.md
```diff
@@ -0,0 +1,189 @@
...
+exported raw records from Zeebe streams into several indices. Camunda webapps (e.g., Operate, Tasklist) then imported
+and transformed these records into useful forms. The Camunda Exporter aims to write records directly into
+Elasticsearch/Opensearch in the form needed by the webapps, eliminating the need for additional importing or
+transformation. This increases process instance throughput and reduces the latency of changes appearing
+in Operate and Tasklist webapps by cutting off the importer from the data pipeline.
+
+:::note
+The indexes are created as required, and will not be created twice if they already exist. However, once disabled, they
+will not be deleted (that is up to the administrator.) Similarly, data is never deleted by the exporter, and must be
+deleted by the administrator when it is safe to do so.
```
Zelldon, line 21: This is actually not true. Identity will delete users.

docs/self-managed/zeebe-deployment/exporters/camunda-exporter.md
```diff
@@ -0,0 +1,189 @@
...
+A [retention](#retention) policy can be configured to automatically delete data after a certain number of days.
+:::
+
+## Configuration
+
+:::note
+As the exporter is packaged with Zeebe, it is not necessary to specify a `jarPath`.
+:::
```
Zelldon, line 29: 🙃 I feel like these notes are more confusing than helpful. If I read this guide the first time I don;t know what a jarPath is.

Zelldon: Maybe we move it downwards

docs/self-managed/zeebe-deployment/exporters/camunda-exporter.md
```diff
@@ -0,0 +1,189 @@
...
+| -------------- | ----------------------------------------------------------------------------------------------------------------------------- | --------------------------- |
+| type           | the type of the underlying search engine to export to. Accepted values are `elasticsearch` or `opensearch`.                   | elasticsearch               |
+| clusterName    | The name of the elasticsearch/opensearch cluster to export to.                                                                | elasticsearch               |
+| dateFormat     | Defines a custom date format that should be used for fetching date data from the engine (should be the same as in the engine) | yyyy-MM-dd'T'HH:mm:ss.SSSZZ |
+| socketTimeout  | Defines the socket timeout in milliseconds, which is the timeout for waiting for data.                                        |                             |
+| connectTimeout | Determines the timeout in milliseconds until a connection is established.                                                     |                             |
```
Zelldon, line 70: 🔧 I still find this confusing. I feel we need to change this somehow or make this simpler. It is not clear to me why we have these both.

Zelldon: What happens after the timeout?

docs/self-managed/zeebe-deployment/exporters/camunda-exporter.md
```diff
@@ -0,0 +1,189 @@
...
+
+</TabItem>
+
+<TabItem value="index">
+
+| Option                | Description                                                                                                                                                                                 | Default |
+| --------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------- |
+| prefix                | This prefix will be appended to every index created by the exporter; must not contain `_` (underscore).                                                                                     |         |
+| numberOfShards        | The number of [shards](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules.html#_static_index_settings) used for each new record index created.                   | 3       |
+| numberOfReplicas      | The number of shard [replicas](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules.html#dynamic-index-settings) used for each new record index created.           | 0       |
```
Zelldon, line 98:
```suggestion
| numberOfShards        | The number of [shards](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules.html#_static_index_settings) used for each created index.                   | 3       |
| numberOfReplicas      | The number of shard [replicas](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules.html#dynamic-index-settings) used for each created index.           | 0       |
```

docs/self-managed/zeebe-deployment/exporters/camunda-exporter.md
```diff
@@ -0,0 +1,189 @@
...
+| variableSizeThreshold | Defines a threshold for variable size. Variables exceeding this threshold are split into two properties: `FULL_VALUE` (full content, not indexed) and `VALUE` (truncated content, indexed). | 8191    |
+| shardsByIndexName     | A map where the key is the index name and the value is the number of shards, allowing you to override the default `numberOfShards` setting for specific indices.                            |         |
+| replicasByIndexName   | A map where the key is the index name and the value is the number of replicas, allowing you to override the default `numberOfReplicas` setting for specific indices.                        |         |
```
Zelldon, line 101: ❓ This means setting this would also win right?

docs/self-managed/zeebe-deployment/exporters/camunda-exporter.md
```diff
@@ -0,0 +1,189 @@
...
+size) can be controlled by configuration.
+
+| Option | Description                                                                                                                                                    | Default |
+| ------ | -------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------- |
+| delay  | Delay, in seconds, before force flush of the current batch. This ensures that even when we have low traffic of records, we still export every once in a while. | `5`     |
+| size   | The amount of records a batch should have before we flush the batch                                                                                            | `1000`  |
+
+With the default configuration, the exporter will aggregate records and flush them to Elasticsearch/Opensearch:
+
+1. When it has aggregated 1000 records. 2. Five seconds have elapsed since the last flush (regardless of how many records were aggregated).
```
Zelldon, line 118:
```suggestion
1. When it has aggregated 1000 records. 
2. Five seconds have elapsed since the last flush (regardless of how many records were aggregated).
```

docs/self-managed/zeebe-deployment/exporters/camunda-exporter.md
```diff
@@ -0,0 +1,189 @@
...
+A retention policy can be set up to delete old data.
+When enabled, this creates an Index Lifecycle Management (ILM) Policy that deletes the data after the specified
+`minimumAge`.
+All index templates created by this exporter apply the created ILM Policy.
+
+| Option     | Description                                                                  | Default |
+| ---------- | ---------------------------------------------------------------------------- | ------- |
+| enabled    | If `true` the ILM Policy is created and applied to the index templates       | `false` |
+| minimumAge | Specifies how old the data must be, before the data is deleted as a duration | `30d`   |
+| policyName | The name of the created and applied ILM policy                               |         |
```
Zelldon, line 133: ❓ There is a default value for the policy name right?

docs/self-managed/zeebe-deployment/exporters/camunda-exporter.md
```diff
@@ -0,0 +1,189 @@
...
+
+Here is an example configuration of the exporter:
+
+```yaml
+---
+exporters:
+  # Camunda Exporter ----------
+  # An example configuration for the camunda exporter:
+  #
+  # These setting can also be overridden using the environment variables "ZEEBE_BROKER_EXPORTERS_CAMUNDAEXPORTER_..."
```
Zelldon, line 153: 🔧 Worth to mention how the args look like, because this often lead to confusion, as this is not with `_`

docs/self-managed/zeebe-deployment/exporters/camunda-exporter.md
```diff
@@ -0,0 +1,189 @@
+---
+id: camunda-exporter
+title: "Camunda exporter"
+sidebar_label: "Camunda Exporter"
+description: "The common camunda exporter exporting zeebe records to harmonized web-apps indices ."
```
Zelldon, line 5:
```suggestion
description: "The Camunda exporter exporting aggregated data to harmonized web-apps indices."
```

Zelldon (commented):
Thanks @mustafadagher I had some comments, which would be good to take a look at :)

docs/self-managed/zeebe-deployment/exporters/camunda-exporter.md
mustafadagher: that's true for users indeed, not for any other records though. I will think of a way to re-write this or remove it.

docs/self-managed/zeebe-deployment/exporters/camunda-exporter.md
mustafadagher: we can configure both separately in the client. we had one configuration variable before but actually we used this same value to set both variables in the client.

Also for operate importer they had both variables configured separately.

I don't see where does your confusion come from, but maybe if you elaborate a bit more, it could help me better understand.

>What happens after the timeout?

From the es/os client perspective, if it fails to establishes a connection before the configured `connectTimeout` it will throw a `ConnectionTimeoutException`, and if it waits longer than the `socketTimeout` to get a response to one of the client request, it will throw `SocketTimeoutException`. And both are actually separate concepts and, I believe, deserve separate configurations.

docs/self-managed/zeebe-deployment/exporters/camunda-exporter.md
mustafadagher: yes, the ones in the map will override the default values set above

docs/self-managed/zeebe-deployment/exporters/camunda-exporter.md
mustafadagher: we used to have a default in zeebe, but here not.

docs/self-managed/zeebe-deployment/exporters/camunda-exporter.md
mustafadagher: I removed this statement and re-phrased the paragraph

docs/self-managed/zeebe-deployment/exporters/camunda-exporter.md
mustafadagher: I removed the statement completely for now

docs/self-managed/zeebe-deployment/exporters/camunda-exporter.md
Zelldon: Maybe you can explain me (me as a user using Camunda 8) why I should configure them both and why different values would be beneficial for me.

docs/self-managed/zeebe-deployment/exporters/camunda-exporter.md
Zelldon: Isn;t this an issue then? If I enable this and have no name it would fail no? I need a default name I think

docs/self-managed/setup/guides/configure-db-custom-headers.md
```diff
@@ -116,6 +117,20 @@ The following examples add the new `my-plugin` JAR to the `application.yaml` for
 - ZEEBE_BROKER_EXPORTERS_OPENSEARCH_ARGS_INTERCEPTORPLUGINS_0_JARPATH=/usr/local/plugin/plg.jar
 ```
 
+</TabItem>
+
+<TabItem value='camundaExporter'>
+
+:::note
+Notice that you might have named your camunda exporter differently in the broker config. Here, we assumed you gave it the name `camundaExporter`.Please adjust the environment variables accordingly.
```
Zelldon, line 125: Good point 👍🏼

docs/self-managed/zeebe-deployment/exporters/camunda-exporter.md
mustafadagher: Not really an issue, if you enable retention, you should give a name to the policy. it's not consistent with other exporters, but not an issue.

I can't remember if that was a decision, but I guess it was removed when there was a confusion whether we will have different policies or not.

Anyway, we can bring back a default value, but here I'm just documenting what there is right now.

docs/self-managed/zeebe-deployment/exporters/camunda-exporter.md
mustafadagher: can youn explain a bit more what do you mean?

docs/self-managed/zeebe-deployment/exporters/camunda-exporter.md
mustafadagher: >Maybe you can explain me (me as a user using Camunda 8) why I should configure them both and why different values would be beneficial for me.

If you don't see a value as a user, then don't. just use the same value for both. if you know what you are doing, and have a good reason to have different values for those, then you have an option.

 >Thanks. What happens then ? Is the exporter retrying or are we stuck? Why should I configure a higher or lower value.

We create the client in the `Exporter.open()`, if it fails to connect, it will retry until successful.

We make requests to the exporter in subsequent exports or flushes, etc. if those fail due to a socket-timeout, the exporter `last exported position` should not change, and the next export will retry flushing the failed batches.

Zelldon: Thanks Mustafa.

docs/self-managed/zeebe-deployment/exporters/camunda-exporter.md
Zelldon: Was thinking about an example of setting something in the ARGS section like:

`ZEEBE_BROKER_EXPORTERS_CAMUNDAEXPORTER_ARGS_INDEX_NUMBEROFSHARDS`

For example related support case. https://jira.camunda.com/browse/SUPPORT-23300

mustafadagher: I added a note on how to convert it and an example below

Zelldon: Double underscore?

mustafadagher: oh no, that's wrong. I guess I relied on co-pilot completing the sentence and didn't check for correctness 😄

mustafadagher: fixed

docs/self-managed/zeebe-deployment/exporters/camunda-exporter.md
```diff
@@ -0,0 +1,192 @@
...
+
+:::note
+The indexes are created as required, and will not be created twice if they already exist. However, once disabled, they
+will not be deleted (that is up to the administrator.)
+A [retention](#retention) policy can be configured to automatically delete data after a certain number of days.
+:::
+
+## Configuration
+
+The exporter can be enabled by configuring it with the `classpath` in the broker settings.
```
Zelldon, line 26:
```suggestion
The exporter can be enabled by configuring it with the `className` in the broker settings.
```

Zelldon (approved):
Great work thanks @mustafadagher 👏🏼

conceptualshark (commented):
I've gone through and updated a bit of this content, including making sure the links to nested tables work, and simplifying some of the introduction. I've also restructured the custom headers page to make the tabs work a little more cleanly.

Do the following pages also need to be updated as part of this work:
https://docs.camunda.io/docs/next/self-managed/zeebe-deployment/exporters/ - added to the list here?
https://docs.camunda.io/docs/next/self-managed/zeebe-deployment/configuration/broker-config/#zeebebrokerexporters - added to the options here?

Additionally, would it be useful to add a note to the Tasklist and Operate config that importers/exporters are not required when using the Camunda Exporter? For example, as a note [on this Tasklist page](https://docs.camunda.io/docs/next/self-managed/tasklist-deployment/tasklist-configuration/#zeebe-elasticsearch-or-opensearch-exporter), and/or the [Importer and archiver page](https://docs.camunda.io/docs/next/self-managed/tasklist-deployment/importer-and-archiver/).

Zelldon:
I wouldn;t do this as part of this PR. We have to write an extensive update guide, where we can add then also some notes about importers (like need to be run for a while until they can be shutdown).

Zelldon:
@conceptualshark can we get this merged for the alpha release, wdyt?

conceptualshark (approved):
👍 Thanks for adding the extra links. Would you like any final engineering review before merging? Once merged, it will go live on /next with the next docs release, unless you'd like for me to merge it closer to the alpha.

PR #4514, opened 2024-10-24
Remove the message correlation guide from 8.5 and above

conceptualshark:
## Description

Relates to https://github.com/camunda/developer-experience/issues/69

Remove the message correlation guide from versions 8.5 and above.

## PR Checklist


docs/components/concepts/messages.md
```diff
@@ -4,7 +4,7 @@ title: "Messages"
 description: "Learn how process instances can respond to incoming messages."
 ---
 
-Process instances can respond to incoming messages. Published messages must be mapped onto a process instance. This step is called **[message correlation](/guides/message-correlation.md)**.
+Process instances can respond to incoming messages. Published messages must be mapped onto a process instance. This step is called message correlation.
```
christinaausley, line 7: There are about 90 references of message correlation per version of docs -- perhaps we should create a backlog issue to add this term to the docs glossary and link to it if we are removing the doc on message correlation?

akeller: Good idea, I am ok either doing that in the scope of this PR or a separate one.

conceptualshark: Would it make sense to add to the glossary, or link to message correlation [here](https://docs.camunda.io/docs/next/components/modeler/bpmn/message-events/#message-correlation)?

christinaausley: I think linking to the existing documentation would be fine 👍

conceptualshark:
@christinaausley Updated this to replace the old link to the message correlation section in Message events - I think it should be good to go now.

christinaausley (approved):

akeller (approved):
Thank you. I know we don't enjoy removing guides, but it's the right thing to do in this case. Thanks for adjusting!

PR #4539, opened 2024-10-30
Remove suggestion to convert from YAML to env vars for optimize config

conceptualshark:
## Description

Remove the suggestion that Optimize YAML options can be converted in a standard way to environment variables.

## PR Checklist


RomanJRW (commented):
Not objecting the change, but please seem my reason for preferring removing the note entirely [here](https://camunda.slack.com/archives/C0484UFR2VA/p1730319216617809?thread_ts=1730299907.861359&cid=C0484UFR2VA)

conceptualshark:
@RomanJRW Given that it doesn't apply to all the properties and would be difficult to annotate individually, I've taken your suggestion to remove it for now. I think if there is an initiative to document all properties (including their matching env vars) in the future, that would be a better method of handling this, rather than introducing anything ambiguous here.

RomanJRW (approved):
Thanks for the quick change! 👍

PR #4540, opened 2024-10-30
add new web modeler deploy config

conceptualshark:
## Description

Relates to https://github.com/camunda/web-modeler/issues/10571 - no changes needed; the documents on deployment pipelines/stages look compatible with the expected self managed behavior, and may only need screenshot updates when the SaaS and SM visuals are aligned

Implements to https://github.com/camunda/web-modeler/issues/10570 - added the ID field to the SM web modeler config table, and updated references to the Docker distro to include the option in Helm as well.

## PR Checklist

### DIFF
--- a/docs/components/modeler/web-modeler/play-your-process.md
+++ b/docs/components/modeler/web-modeler/play-your-process.md
@@ -14,7 +14,7 @@ Play is a Zeebe-powered playground environment within Web Modeler for validating
 
 To use Play, open a BPMN diagram and click the **Play** tab. Read the [limitations and availability section](#limitations-and-availability) if this section is missing.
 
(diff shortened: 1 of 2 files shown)


conceptualshark:
@nithinssabu My instinct here was to mark the new ID field as Required in some way, as it sounds like you will be unable to deploy without it? If this is the case, it would be handy to list which other config options are required vs not, and I could add those in as well, or follow up.

mesellings (approved):
Lgtm, nothing useful to add - great work! 👍 🚀 🥇

nithinssabu:
> @nithinssabu My instinct here was to mark the new ID field as Required in some way, as it sounds like you will be unable to deploy without it? If this is the case, it would be handy to list which other config options are required vs not, and I could add those in as well, or follow up.

@conceptualshark If the ID field is not present or unique, then the application won't start up and the reason will be clear from the error message. In my opinion, it's fine to not have this in the documentation.

nithinssabu (approved):
Thank you 🙂

PR #4549, opened 2024-11-01
fix: Removing references to OpenSearch proxy

grlimacan:
## Description

In https://github.com/camunda/camunda-optimize/issues/10058 we have decided to de-scope the implementation for proxy for OpenSearch, so we are updating the documentation to reflect that

Same applies to the response buffer limit as described in https://github.com/camunda/camunda/issues/24083

## PR Checklist


conceptualshark (approved):
This looks good to me from my understanding of the linked issues. Just to be certain, it seems like the ability to use the proxy was not ever implemented, only documented, and so there is no need to announce the removal?

grlimacan:
> This looks good to me from my understanding of the linked issues, but I'm not certain if an engineering review would be prudent. Just to be certain, it seems like the ability to use the proxy was not ever implemented, only documented, and so there is no need to announce the removal?

That's correct, it was never implemented, so no announcement necessary

Issue #3014, opened 2023-12-08
Rework Connectors + MT docs


akeller:
Most component documentation for configuring MT includes a default value of "false" to signify MT is disabled. However, Connectors default value is `<default>`. If this is the true default value, we should make it clear and look into deviating from the table format so users understand or expect this isn't a boolean value as it appears in the other MT config docs.

Thread opened for Connectors - https://camunda.slack.com/archives/C0555930H7F/p1701875344024639

Docs - https://docs.camunda.io/docs/self-managed/connectors-deployment/connectors-configuration/#multi-tenancy

PR #4552, opened 2024-11-01
rework connectors mt doc

christinaausley:
## Description

Closes https://github.com/camunda/camunda-docs/issues/3014.

## Open questions

1. Why do Connectors take a tenant id?
2. Where would I find this id, [Identity](https://docs.camunda.io/docs/self-managed/identity/user-guide/tenants/managing-tenants/)?
3. We have two example values - t1 and tenant1. Should these be adjusted to myTenant as well?

## PR Checklist

### DIFF
--- a/docs/self-managed/connectors-deployment/connectors-configuration.md
+++ b/docs/self-managed/connectors-deployment/connectors-configuration.md
@@ -1,6 +1,7 @@
 ---
 id: connectors-configuration
 title: Configuration
+description: "Configure the Connector runtime environment based on the Zeebe instance to connect to, the Connector functions to run, and secrets available to the Connectors."
 ---
 
 import Tabs from "@theme/Tabs";
@@ -72,7 +73,7 @@ CAMUNDA_OPERATE_CLIENT_USERNAME=demo
 CAMUNDA_OPERATE_CLIENT_PASSWORD=demo
 ```
 
-When running against a self-managed environment you might also need to configure Identity properties instead of username and password:
+When running against a Self-Managed environment, you might also need to configure Identity properties instead of username and password:
 
 ```bash
 CAMUNDA_OPERATE_CLIENT_URL=http://localhost:8081
@@ -87,8 +88,7 @@ CAMUNDA_IDENTITY_CLIENT_SECRET=<YOUR_OPERATE_CLIENT_SECRET>
 
 <TabItem value='operate'>
 
-Disabling Operate polling will lead to inability to use inbound (e.g., webhook) capabilities.
-However, if you still wish to do so, you need to start your Connector runtime with the following environment variables:
+Disabling Operate polling will lead to inability to use inbound capabilities like webhooks. If you still wish to do so, start your Connector runtime with the following environment variables:
 
 ```bash
 CAMUNDA_CONNECTOR_POLLING_ENABLED=false
@@ -102,7 +102,7 @@ OPERATE_CLIENT_ENABLED=false
 ## Manual discovery of Connectors
 
 By default, the Connector runtime picks up outbound Connectors available on the classpath automatically.
-To disable this behavior, use the following environment variables to configure Connectors and their configuration explicitly:
+To disable this behavior, use the following environment variables to configure Connectors explicitly:
 
 | Environment variable                          | Purpose                                                       |
 | :-------------------------------------------- | :------------------------------------------------------------ |
@@ -111,9 +111,9 @@ To disable this behavior, use the following environment variables to configure C
 | `CONNECTOR_{NAME}_INPUT_VARIABLES` (optional) | Variables to fetch for worker with `NAME`                     |
 | `CONNECTOR_{NAME}_TIMEOUT` (optional)         | Timeout in milliseconds for worker with `NAME`                |
 
-Through that configuration, you define all job workers to run.
+Through this configuration, you define all job workers to run.
 
-Specifying optional values allow you to override `@OutboundConnector`-provided Connector configuration.
+Specifying optional values allows you to override `@OutboundConnector`-provided Connector configuration.
 
 ```bash
 CONNECTOR_HTTPJSON_FUNCTION=io.camunda.connector.http.rest.HttpJsonFunction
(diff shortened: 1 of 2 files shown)


conceptualshark (commented):
I think my lingering question here that doesn't seem to be covered in the PR summary is still if `<default>` is actually the default value. The summary only captures changing the example values, which makes sense for clarity, but I worry we might be over-correcting.

The [multi-tenancy docs](https://docs.camunda.io/docs/self-managed/concepts/multi-tenancy/#the-tenant-identifier) call out `<default>` as a specific reserved identifier so my assumption is this is the case. If this is true, maybe we can make a follow-up issue to update the instructions, since it seems a little unclear to all of us.

christinaausley:
> I think my lingering question here that doesn't seem to be covered in the PR summary is still if `<default>` is actually the default value. The summary only captures changing the example values, which makes sense for clarity, but I worry we might be over-correcting.
> 
> The [multi-tenancy docs](https://docs.camunda.io/docs/self-managed/concepts/multi-tenancy/#the-tenant-identifier) call out `<default>` as a specific reserved identifier so my assumption is this is the case. If this is true, maybe we can make a follow-up issue to update the instructions, since it seems a little unclear to all of us.

@sbuettner Do you have any thoughts here?

sbuettner:
Yes, `<default>` is the default value of the zeebe client configuration: https://github.com/camunda/camunda/blob/aa4b4e93b2110db2b1e75b7c9cdb5c92192fc16d/clients/java/src/main/java/io/camunda/zeebe/client/api/command/CommandWithTenantStep.java#L25

docs/self-managed/connectors-deployment/connectors-configuration.md
```diff
@@ -223,75 +223,71 @@ java -cp 'connector-runtime-application-VERSION-with-dependencies.jar:...:my-sec
 
 ## Multi-tenancy
 
-The Connector Runtime supports multiple tenants for inbound and outbound Connectors.
+The Connector Runtime supports multiple tenants for inbound and outbound Connectors. These are configurable in [Identity](/self-managed/identity/user-guide/tenants/managing-tenants.md).
+
 A single Connector Runtime can serve a single tenant or can be configured to serve
-multiple tenants. By default, the runtime uses the `<default>` tenant ID for all
-Zeebe related operations like handling Jobs and publishing Messages.
+multiple tenants. By default, the runtime uses the default tenant ID for all
```
conceptualshark, line 229:
```suggestion
multiple tenants. By default, the runtime uses the tenant ID `<default>` for all
```

I think this is one of the references where being explicit about the default being `<default>` is clearer.

docs/self-managed/connectors-deployment/connectors-configuration.md
```diff
@@ -223,75 +223,71 @@ java -cp 'connector-runtime-application-VERSION-with-dependencies.jar:...:my-sec
...
+The Connector Runtime uses the following environment variables to configure multi-tenancy:
 
-| Name                                       | Description                                                     | Default value |
-| ------------------------------------------ | --------------------------------------------------------------- | ------------- |
-| ZEEBE_CLIENT_DEFAULT-TENANT-ID             | The default tenant ID used to communicate with Zeebe            | `<default>`   |
-| ZEEBE_CLIENT_DEFAULT-JOB-WORKER-TENANT-IDS | The default tenants IDs (comma separated) used to activate jobs | `<default>`   |
+| Name                                       | Description                                                    | Default value |
+| ------------------------------------------ | -------------------------------------------------------------- | ------------- |
+| ZEEBE_CLIENT_DEFAULT-TENANT-ID             | The default tenant ID used to communicate with Zeebe           | `myTenant`    |
+| ZEEBE_CLIENT_DEFAULT-JOB-WORKER-TENANT-IDS | The default tenant IDs (comma separated) used to activate jobs | `myTenant`    |
```
conceptualshark, line 244: Same as above, we can be explicit about the default being `<default>` here, but this might benefit from updating the description if my understanding is correct? 

```suggestion
| ZEEBE_CLIENT_DEFAULT-TENANT-ID             | The default tenant ID used to communicate with Zeebe. Changing this value will set a new default tenant ID used for fetching jobs and publishing messages .          | `myTenant`    |
| ZEEBE_CLIENT_DEFAULT-JOB-WORKER-TENANT-IDS | The default tenant IDs (comma separated) used to activate jobs. To run the Connector Runtime in a setup where a single runtime serves multiple tenants, add each tenant ID to this list. | `myTenant`    |
```

conceptualshark (changes requested):
I left some suggestions for the two places I think using `<default>` still works. I think adding in an explanation to the table may help clarify this section, and otherwise the new myTenant examples make it a lot clearer when and how you're overriding the defaults. 👍

christinaausley:
I have applied your proposed changes -- thank you for having a look @conceptualshark!

docs/self-managed/connectors-deployment/connectors-configuration.md
```diff
@@ -223,75 +223,71 @@ java -cp 'connector-runtime-application-VERSION-with-dependencies.jar:...:my-sec
...
-| ZEEBE_CLIENT_DEFAULT-JOB-WORKER-TENANT-IDS | The default tenants IDs (comma separated) used to activate jobs | `<default>`   |
+| Name                                       | Description                                                                                                                                                                              | Default value |
+| ------------------------------------------ | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------- |
+| ZEEBE_CLIENT_DEFAULT-TENANT-ID             | The default tenant ID used to communicate with Zeebe. Changing this value will set a new default tenant ID used for fetching jobs and publishing messages .                              | `myTenant`    |
+| ZEEBE_CLIENT_DEFAULT-JOB-WORKER-TENANT-IDS | The default tenant IDs (comma separated) used to activate jobs. To run the Connector Runtime in a setup where a single runtime serves multiple tenants, add each tenant ID to this list. | `myTenant`    |
```
conceptualshark, line 244: I forgot to flag this in the earlier review, sorry! I think the default in the table should remain `<default>`. Other than that, this matches the way I understand things, but I'd feel most comfortable with a technical overview to be certain.

christinaausley:
@sbuettner @conceptualshark mind having a final look over this?

sbuettner (approved):

conceptualshark (approved):
🚀 !

Issue #2719, opened 2023-10-10
Prepare for docusaurus v3


pepopowitz:
> Docusaurus v3 is now [in beta](https://github.com/facebook/docusaurus/discussions/9312) and the official release is around the corner. This is the perfect time to start preparing your site for this new major version.

https://docusaurus.io/blog/preparing-your-site-for-docusaurus-v3

There is anecdotal data from the write-the-docs slack that performance is improved in v3.

akeller:
I saw they anticipate releasing V3 in the next few days unless something big comes up.

akeller:
I'm glad to see they acknowledged the [MDX upgrade and infrastructure changes](https://docusaurus.io/blog/releases/3.0#conclusion) were the bulk of this release. 

I did see two interesting features highlighted:

* [Unlisted content](https://docusaurus.io/blog/releases/3.0#unlisted-content)
* [Code blocks](https://docusaurus.io/blog/releases/3.0#code-blocks)

I don't know how much we'll be able to do with code blocks, but I'm a sucker for interactive experiences.

I didn't realize `draft:true` was already available for docs and pages (I thought this was a blog only feature). This and the `unlisted:true` option could help us with some of the feedback that came up in retros around proactive docs authoring or reviewing content without running it locally. CC @christinaausley 

☝ I'll create two issues for the draft and unlisted content, respectively. I'd like to see us incorporate a strategy for using the draft feature and then when we move to V3, a strategy for using unlisted.

akeller:
This came up during our [8.4 docs retro](https://github.com/camunda/developer-experience/issues/133) as a possible idea to explore to keep our build times down.

akeller:
Here is a _dump_ of the `npx docusaurus-mdx-checker` results - https://docs.google.com/spreadsheets/d/1IvTrYlxrqr1BgOqY8csPDyFH5Rc_2-hrtvajVJskvTY/edit?usp=sharing

PR #4561, opened 2024-11-01
Remove {{}} in Optimize update guide links

akeller:
## Description

Related to #2719, fixes build issues by getting docs.camunda.org syntax out. 

## PR Checklist

### DIFF
--- a/optimize/self-managed/optimize-deployment/migration-update/camunda-7/2.1-to-2.2.md
+++ b/optimize/self-managed/optimize-deployment/migration-update/camunda-7/2.1-to-2.2.md
@@ -21,7 +21,7 @@ Here you will find information about:
 When updating Optimize, certain features might not work out of the box for the old data. This is because old versions of Optimize
 do not fetch data that is necessary for the new feature to work. For this update, the following features do not work on the old data:
 
-- [Process instance parts](./components/userguide/process-analysis/report-analysis/process-instance-parts.md).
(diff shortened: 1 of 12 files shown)


akeller:
Oops, sorry for the misleading branch name. I also couldn't easily figure out where this content should point to, so I didn't include links.

akeller:
If this builds, this will take care of all the links in the docs.camunda.org syntax.

However, there will still be instances where `{{<` is used, particularly for referencing the version alias (not something we do on docs.camunda.io, but is an interesting idea!). Because these don't break the build, I will leave them for now.

conceptualshark (commented):
Should these links resolve to the versioned docs? Right now, they're sending me to /next, and I am not sure of the Optimize versioned docs can figure out what version it's in without backing out of the current directory.

akeller:
> Should these links resolve to the versioned docs? Right now, they're sending me to /next, and I am not sure of the Optimize versioned docs can figure out what version it's in without backing out of the current directory.

This does look like a bug. Given the links didn't resolve at all previously, but now go to a page that exists, I think it's a risk I'm willing to accept. Also these versions are technically too old to be supported and the C7 docs will move back to docs.camunda.org soon. 

Any concerns with my logic here @conceptualshark?

conceptualshark (approved):

conceptualshark:
👍 Works for me - happy to approve!

Issue #4137, opened 2024-08-09
C8SM Zeebe REST API is left out of ingress installation guide


jessesimpson36:
[SUPPORT-23061](https://jira.camunda.com/browse/SUPPORT-23061)

In  the above ticket, the customer navigated to
https://docs.camunda.io/docs/apis-tools/zeebe-api-rest/zeebe-api-rest-overview/

and was missing some key information regarding how to set up the zeebe rest api, and making requests against it.

While working on the ticket, the customer specifically asks:
1. Which service should be called ( kubernetes port-forward ) to access it?
2. Which URL should we use?

It should be noted somewhere that the service that provides the rest api is the zeebe gateway (in case customers have issues connecting to their ingress, they should be able to test via port-forwards).  And perhaps there could be a link from above link to the self-managed documentation where we set up the ingress.

I also noticed that the "[Combined and separated ingress setup](https://docs.camunda.io/docs/self-managed/setup/guides/ingress-setup/)" does not include the necessary options to enable the zeebe rest api.

a snippet like 
```diff
zeebeGateway:
  contextPath: "/zeebe"
  ingress:
    grpc:
      enabled: true
      className: nginx
      host: "zeebe.camunda.example.com"
+    rest:
+      enabled: true
+      className: nginx
+      host: "camunda.example.com"
+      path: "/zeebe"
```
should be added.

upgradingdave:
This is a great suggestion, +1 for this!

We might want to also highlight that ingress path must match `contextPath` for each app, for example:
-  `global.identity.auth.<app>. redirectUrl` should match `<app>.contextPath` for operate, tasklist, optimize, modeler, connectors, and console. 
-  `zeebeGateway.ingress.rest.path` should match `zeebe.contextPath`

akeller:
Assigning @conceptualshark and @pepopowitz to this since its an API + SM topic. @jessesimpson36 if Distribution Team plans to deliver these docs changes, you can remove them and assign them as reviewers.

jessesimpson36:
Sounds good @akeller . I was waiting for our backlog grooming to see if we would assign implementation of this to our next sprint, but we decided not to. I am willing to review a PR on this topic though.

PR #4602, opened 2024-11-13
Add the REST config to the Ingress guides

conceptualshark:
## Description

Relates to #4137 
This addresses:
- adding the Zeebe REST API configuration into the Ingress guides
- adding a note that the ingress path should match the context path to the (pre-8.7 combined) Ingress guides
- linking to the ingress config, clarify the context path is for Zeebe Gateway, and update organization of the context path section in the Zeebe REST API overview

I've added these changes to the available documentation, back through 8.5 for the Ingress docs - prior to that the formatting changes. 

I've added these options into /next assuming that if the /v1 APIs remain available, it will still be useful, and if they're removed, we can address removing it at the same time as any other updates, but let me know if this isn't the right call.

## PR Checklist


pepopowitz (approved):
makes sense to me, and clarifies. Thank you!

conceptualshark:
@jessesimpson36 Would you be able to double-check the update here? Thanks!

akeller:
> I've added these options into /next assuming that if the /v1 APIs remain available, it will still be useful, and if they're removed, we can address removing it at the same time as any other updates, but let me know if this isn't the right call.

I'd like to see a member of the Zeebe team pulled in to review this, given the Zeebe REST API is actively being removed. @conceptualshark can you reach out on #ask-zeebe?

jessesimpson36 (approved):
I read the content and verified the referenced options. This all looks good to me.

conceptualshark:
@akeller @jessesimpson36 With the zeebe REST API officially out of the docs for /next, I've updated this PR to remove the reference and remove the config from 8.7. Do you mind double-checking for a final review?

docs/self-managed/setup/guides/ingress-setup.md
```diff
@@ -28,6 +28,10 @@ In this setup, a single Ingress/domain is used to access Camunda 8 web applicati
 
 ![Camunda 8 Self-Managed Architecture Diagram - Combined Ingress](../../assets/camunda-platform-8-self-managed-architecture-diagram-combined-ingress.png)
 
+:::note
+**Operate, Tasklist, Optimize, Modeler, Connectors, Console:** The Ingress path value for each Component (`global.identity.auth.<component>.redirectUrl`) should match the `contextPath` for that Component.
+:::
```
akeller, line 33:
```suggestion
:::note
**Operate, Tasklist, Optimize, Modeler, Connectors, Console:** The Ingress path value for each component (`global.identity.auth.<component>.redirectUrl`) should match the `contextPath` for that component.
:::
```

akeller (commented):
❓ For the C8 REST API (v2), do we need to add a similar change? Basically, adding info on [this page](https://preview.docs.camunda.cloud/pr-4602/docs/next/apis-tools/camunda-api-rest/camunda-api-rest-overview/#context-paths) with a similar link to the ingress setup instructions? 

Otherwise, the changes look consistently applied across the relevant versions.

docs/self-managed/setup/guides/ingress-setup.md
akeller: Plus the other versions.

conceptualshark:
@akeller I've updated the Camunda REST API page as described! I think this should be the last place where this info was located/unclear. 🤞

akeller (approved):

PR #4648, opened 2024-11-22
Add orchestration cluster glossary terms

akeller:
## Description

~This is a DRAFT because a crisp definition is not found internally. This PR will need to be nurtured by an @camunda/tech-writers to work with @felix-mueller(?) on a definition.~

We should also mention the 1:many relationship between ~a management cluster~ Web Modeler + Console and an orchestration cluster in SM, but I wasn't sure if we wanted to do that here when the definition should be crisp with a link to more info. 

## PR Checklist


docs/reference/glossary.md
```diff
@@ -149,6 +149,10 @@ The log is comprised of an ordered sequence of records written to persistent sto
 
 - [Partitions](/components/zeebe/technical-concepts/partitions.md#partition-data-layout)
 
+### Management cluster
```
theburi, line 152: Lets call it "Web Modeler and Console clusters" we are not planning to use Management cluster for now.

docs/reference/glossary.md
```diff
@@ -163,6 +167,10 @@ A message contains information to be delivered to interested parties during exec
 
 - [Messages](/components/concepts/messages.md)
 
+### Orchestration cluster
+
+An orchestration cluster includes Zeebe, Operate, Tasklist, and Optimize, existing alongside the management cluster.
```
theburi, line 172: Orchestration cluster= Orchestration Cluster Core + Connectors+ Connectors Runtime
 Orchestration Cluster Core = Zeebe, Operate, Tasklist,  Optimize and identity

akeller:
@theburi I followed up on your [doc](https://docs.google.com/document/d/1Bb8Yx8a4BN7Q4byGKN33IB_eggS-Av9j-u4mhU27oeQ/edit?tab=t.0) since it sounds like some of this has changed while I was out. See my comments, particularly regarding SaaS and "core".

docs/reference/glossary.md
akeller: I removed this term

docs/reference/glossary.md
akeller: I added both terms, but referred to the "Orchestration Cluster Core" as follows:

> ### Orchestration core
> An orchestration core or orchestration cluster core includes Zeebe, Operate, Tasklist, Optimize, and Identity.

Would this work @theburi? Or do you want use to be specific and refer to this as "orchestration cluster core"?

akeller:
@theburi can you review these terms for good working definitions? We can refine them in another PR, but I want to make sure the tech writers know how to use these moving forward.

conceptualshark (commented):
This matches my understanding of orchestration cluster vs core, with two comments:

- As far as I understand, the Core will not change, but additional features might be added to the orchestration cluster - IDP and RPA? So the definition as written will need to be expanded when that occurs. I'm wondering if we have a more generic way to refer to "The orchestration core and any other optional components"? 
- From what I've seen in discussions and PRs, we're using "orchestration cluster" over "core" almost exclusively. Is "core" something we want to push externally? Does it matter to a customer if they know "core" is only X components, or can they just assume their orchestration cluster is _always_ X components, and also inclusive of anything else added to it? If we rarely use "core" in the docs it can make the communication a little more confusing. More of a question for @theburi.

conceptualshark (approved):
Based on further discussion with @theburi, going ahead and approving this, as it matches what was confirmed as approved in [this internal doc](https://docs.google.com/document/d/1Bb8Yx8a4BN7Q4byGKN33IB_eggS-Av9j-u4mhU27oeQ/edit?tab=t.0#heading=h.cwo8qxehm1w4).

We'll just need to remember that if/when additional features get added to the orchestration cluster, they will need to be added to this list.

PR #4762, opened 2024-12-12
docs: remove alpha feature nature from C8 Query API

### LINKED ISSUE #4761: Mark C8 REST Query API as stable
## Description

Remove the alpha feature state of the C8 Query API from the overview.

## Out of scope

Removing the alpha annotation from the specification happens in https://github.com/camunda/camunda/issues/26022

tmetzke:
## Description

Remove the alpha documentation from the C8 REST Query API.

closes #4761 

## PR Checklist

### DIFF
--- a/docs/apis-tools/camunda-api-rest/camunda-api-rest-overview.md
+++ b/docs/apis-tools/camunda-api-rest/camunda-api-rest-overview.md
@@ -39,17 +39,4 @@ spring.servlet.multipart.max-request-size=4MB
 
 For example, if you increase the `maxMessageSize` to 10MB, increase these property values to 10MB as well.
 
-### Query API
-
-:::warning
-Query API endpoints do not currently support [resource authorizations][], and can be used to expand user access to restricted resources. If you use resource permissions, allowing public access to those endpoints is not recommended.
-:::
-
-All Query API endpoints contain an `(alpha)` declaration. Those endpoints are not accessible by default in Camunda 8 clusters.
-
-You can enable the [alpha feature][] search endpoints by setting either the configuration property `camunda.rest.query.enabled` to `true`,
-or the environment variable `CAMUNDA_REST_QUERY_ENABLED` to `true`.
-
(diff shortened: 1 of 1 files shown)


christinaausley (approved):

pepopowitz (approved):
1. In terms of sequencing, we should probably wait to merge this PR until https://github.com/camunda/camunda/issues/26022 is complete and the API spec has been updated here in the docs.
2. @tmetzke I interpret this PR to mean that we do not intend to have alpha features in the API in the future. Is that accurate? If it's not, and it's more a case of we've graduated all _current_ alpha endpoints, then I think we should leave this explanation in the docs for when another alpha endpoint comes. 
3. If we are indeed done with alpha endpoints, then there is also some logic I can remove in the API generation process (which docusaurus-ifies the alpha notes in the spec). I will take care of this.

tmetzke:
> 1. In terms of sequencing, we should probably wait to merge this PR until [Remove alpha label from query API in OpenAPI spec camunda#26022](https://github.com/camunda/camunda/issues/26022) is complete and the API spec has been updated here in the docs.

Yep, that makes sense; I'll try to get that one in today.

> 2. @tmetzke I interpret this PR to mean that we do not intend to have alpha features in the API in the future. Is that accurate? If it's not, and it's more a case of we've graduated all _current_ alpha endpoints, then I think we should leave this explanation in the docs for when another alpha endpoint comes.
> 3. If we are indeed done with alpha endpoints, then there is also some logic I can remove in the API generation process (which docusaurus-ifies the alpha notes in the spec). I will take care of this.

No, this means we're removing the alpha status from the Query API. That is also the only thing that was documented here. All other (still existing) alpha-state endpoints remain in the API. They each describe why they are marked as an alpha feature. I am not 100% sure what's in our docusaurus routine for alpha endpoints. If it's just adding a visible `:::note:::` element that links to the official alpha feature documentation, we can keep that.

Issue #4782, opened 2024-12-16
Add Node SDK to SDK section


akeller:
The fully supported [Node SDK](https://docs.camunda.io/docs/apis-tools/node-js-sdk/) needs to be added to the [SDK section](https://docs.camunda.io/docs/apis-tools/working-with-apis-tools/#sdks). It's correctly present in the sidebar, but not on the main page.

akeller:
@christinaausley can you pick this up this week?

christinaausley:
I'll have this done tomorrow 👍

PR #4784, opened 2024-12-17
add nodejs sdk to overview page

christinaausley:
## Description

Closes https://github.com/camunda/camunda-docs/issues/4782.

## PR Checklist


mesellings (approved):
Lgtm 👍 🚀 

One non-blocking issue that perhaps is just something to look at when we move to docusaurus 3 is that the card height is dependent on the text content - here the node card is now taller than the one next to it:

image


I would expect card components to behave slightly differently ideally, but I guess our options are:

1) Adjust the meta description for the node.js page, i.e. remove ", available via npm.", except this then conflicts with SEO description length.
2) Leave as is, take a look at card components in Docusaurus 3.

I guess we shouldn't be trying to fit content to a responsive layout, so I think 2 is our best option, but up to you  - just something I noticed?

christinaausley:
@mesellings Good catch. This was bothering me so for the time being I adjusted the description length, as it isn't dire either way. Cheers to Doc3 when this perhaps won't be an issue?

mesellings (approved):
Lgtm 👍 🚀

PR #4805, opened 2024-12-20
d3: regen 8.6 API docs

pepopowitz:
## Description
Part of #4799.

- Rewrites the API generation logic to support non-next versions of docs
- Regenerates API docs for vCurrent
   - Operate
   - Tasklist
   - Camunda
   - Admin SM
   - Zeebe 

## When should this change go live?

- [ ] This is a bug fix, security concern, or something that needs **urgent release support**.
- [ ] This is already available but undocumented and should be released within a week.
- [ ] This on a **specific schedule** and the assignee will coordinate a release with the DevEx team. (apply `hold` label or convert to draft PR)
- [ ] This is part of a scheduled alpha or minor. (apply alpha or minor label)
- [ ] There is **no urgency** with this change and can be released at any time.

## PR Checklist

- [ ] My changes are for an **already released minor** and are in `/versioned_docs` directory.
- [ ] My changes are for the **next minor** and are in `/docs` directory (aka `/next/`).

- [ ] My changes require an [Engineering review](https://github.com/camunda/camunda-docs/blob/main/howtos/documentation-guidelines.md#review-process), and I've assigned the Engineering DRI or delegate.
- [ ] My changes require a [technical writer review](https://github.com/camunda/camunda-docs/blob/main/howtos/documentation-guidelines.md#review-process), and I've assigned @camunda/tech-writers as a reviewer.

PR #4801, opened 2024-12-19
d3: regenerate vNext API docs

pepopowitz:
Targets the branch for #4799.

## Description

Regenerates API docs for vNext, to resolve a bunch of markdown errors and accommodate the upgraded API docs plugin.

### Progress

- [x] camunda rest api, vNext
- [x] operate rest api, vNext
- [x] tasklist rest api, vNext
- [x] administration sm rest api
- [x] update duplicate-badge-removal regex to accommodate changes in API generator plugin 

## When should this change go live?

- [ ] This is a bug fix, security concern, or something that needs **urgent release support**.
- [ ] This is already available but undocumented and should be released within a week.
- [ ] This on a **specific schedule** and the assignee will coordinate a release with the DevEx team. (apply `hold` label or convert to draft PR)
- [ ] This is part of a scheduled alpha or minor. (apply alpha or minor label)
- [ ] There is **no urgency** with this change and can be released at any time.

## PR Checklist

- [ ] My changes are for an **already released minor** and are in `/versioned_docs` directory.
- [ ] My changes are for the **next minor** and are in `/docs` directory (aka `/next/`).

- [ ] My changes require an [Engineering review](https://github.com/camunda/camunda-docs/blob/main/howtos/documentation-guidelines.md#review-process), and I've assigned the Engineering DRI or delegate.
- [ ] My changes require a [technical writer review](https://github.com/camunda/camunda-docs/blob/main/howtos/documentation-guidelines.md#review-process), and I've assigned @camunda/tech-writers as a reviewer.

PR #4828, opened 2025-01-09
d3: fix alpha URLs for 8.6 c8 REST API docs

pepopowitz:
Targets the branch for #4799.

## Description

- Fixes the 8.6 C8 REST API docs, which were targeting an alpha document that moved in vNext.
- Adjusts the C8 REST API generation strategy to vary its alpha document path by version, so that I don't have to fix this same issue when I generate the older versions.

### There are still two broken links in the generated API docs

These are because the C8 REST docs reference an 8.5-specific document, which is not yet included in this branch. Eventually it will be, and those broken links should resolve.

image

 
## PR Checklist


akeller (approved):
Thank you 🙏 

Is there a better/smarter way for us to catch these link changes? Or is that part of this, include a set of linked docs as part of the generation strategy and add more moving forward?

pepopowitz:
> Thank you 🙏
> 
> Is there a better/smarter way for us to catch these link changes? Or is that part of this, include a set of linked docs as part of the generation strategy and add more moving forward?

This is maybe not exactly what you're thinking. The URLs this PR corrects were actually introduced by the API doc generation process, they weren't snuck into the spec upstream. 

So in this case, we already own these links, and the issue is that I didn't notice that I was using an invalid URL in #4805 and #4801.

Issue #4498, opened 2024-10-22
Document update guide for prefix migration


ChrisKujawa:
## Context

Operate and Tasklist support a feature to add a custom prefix to indices. As we harmonize indices, this makes it a bit more complicated. We want to use harmonized indices for different web-apps, and reuse indices from Operate for example for things in Tasklist, etc.

There is no good way to make this possible with customized indices, without migration/reindexing data on an update to 8.7. This was an accepted/agreed way to handle such.

## Need
We need to write a guide for users/customers that use this custom index feature, such that they can migrate the index structure to a supported format. 

It was [decided](https://docs.google.com/document/d/1w1VzwyvvBzu0vnjCUj4UcLxTCVae8iMXAMA5_m6xs1s/edit#bookmark=id.ycvvxxg2yt70) that this is something users need to do as preparation work in order to migrate to 8.7. If we would implement this, it would make migration logic quite complex and would be hard to cover all cases. 

## AT

As a user I can read the document and understand how to migrate to 8.7. 

1. If I use the prefixed indices I have to manually migrate the indices
    2. The guide needs to be verified with QA and manually during writing
    3. Code snippets or commands need to be easily copied, the guide needs to be clear and crisp.
4. If I don't use the feature, I'm not affected.

ChrisKujawa:
We should have the same warning as we have for optimize:

https://docs.camunda.io/optimize/next/self-managed/optimize-deployment/configuration/shared-elasticsearch-cluster/#:~:text=Changing%20the%20value%20of%20*.settings.index.prefix%20after%20an%20instance%20was%20already%20running%20results%20in%20new%20indexes%20being%20created%20with%20the%20new%20prefix%20value.%20There%20is%20no%20support%20in%20migrating%20data%20between%20indexes%20based%20on%20different%20prefixes.

everywhere

EuroLew:
# Prefix migration guide
This migration guide only applies for scenarios where operate and task-list both have prefixes and the new harmonised indices will also have a prefix.

## Making the Elasticsearch instance reachable
The migration relies on communicating directly with Elasticsearch so therefore the instance must be reachable on `http://localhost:9200` 

- Manual installations - should already be reachable
- Kubernetes - a port forward should be used to make the instance reachable
``` shell
kubectl port-forward svc/camunda-platform-elasticsearch 9200:9200
```
## Creating harmonised indices

The first step involves creating the harmonised indices indices for migration 

``` shell
[...]

PR #4844, opened 2025-01-13
Prefix migration guide

EuroLew:
## Description

Currently the image used for creating harmonised indices is `SNAPSHOT` however when https://github.com/camunda/camunda/pull/26972 makes it into a pre-release the image tag will be changed to that version.

PR for bash script is here https://github.com/camunda/camunda/pull/26971
relates to https://github.com/camunda/camunda-docs/issues/4498

- [x] This on a **specific schedule** and the assignee will coordinate a release with the DevEx team. (apply `hold` label or convert to draft PR)

## PR Checklist


docs/self-managed/operational-guides/update-guide/860-to-870.md
```diff
@@ -59,3 +59,84 @@ In that case, we append the following records:
 - `ASSIGN`
 - `ASSIGNING`
 - `ASSIGNED`
+
+## Prefix migration
+
+If a custom prefix has been defined for Operate and Tasklist (for example, using `camunda.operate.elasticsearch.indexPrefix`), then data must be migrated manually to the new indices.
```
conceptualshark, line 65: I added an example value here; can you ensure it is correct?

docs/self-managed/operational-guides/update-guide/860-to-870.md
```diff
@@ -59,3 +59,84 @@ In that case, we append the following records:
...
+
+- **Kubernetes**: Port-forward your Elasticsearch instance to ensure it is reachable. In the following example, `camunda-platform` can be replaced by the release name:
+
+  ```shell
+  kubectl port-forward svc/camunda-platform-elasticsearch 9200:9200
+  ```
+
+### Create harmonized indices
+
+To begin creating the harmonized indices for migration, first pull the following Docker image:
```
conceptualshark, line 79: Added additional description to these steps, though I understand this process may change entirely to a script, etc.

docs/self-managed/operational-guides/update-guide/860-to-870.md
```diff
@@ -59,3 +59,84 @@ In that case, we append the following records:
...
+
+```shell
+docker run --rm -it --env CAMUNDA_DATABASE_URL="host.docker.internal:9200" --env CAMUNDA_DATABASE_INDEXPREFIX=<prefix> --entrypoint /usr/local/zeebe/bin/schema camunda/zeebe:SNAPSHOT
+```
+
+### Migrate to harmonized indices
+
+Migrating to the harmonized indices will be done through reindexing.
+
+1. In Postman, import the collection `Prefix migration.postman_collection.json`.
```
conceptualshark, line 95: Where will we be hosting this collection?

conceptualshark (commented):
Knowing this is still in progress, I took a quick pass at some minor updates for wording consistency and style. Happy to go over things again once it is out of draft/finalized, or at any other point! I left a few questions and notes as to some of my edits so you could ensure they're still correct.

docs/self-managed/operational-guides/update-guide/860-to-870.md
EuroLew: The additional description makes sense

docs/self-managed/operational-guides/update-guide/860-to-870.md
EuroLew: I would assume in the camunda repo somewhere, but this part is subject to the most change, likely the postman collection will be removed and replaced by a script of some sort.

christinaausley: @EuroLew I oversee Postman, so just let me know if you need access or want me to make any changes for you 😄

docs/self-managed/operational-guides/update-guide/860-to-870.md
ChrisKujawa, line 65: ❗  We should state somehow maybe via a important note or something that it is a prerequisite before doing the actual update.

docs/self-managed/operational-guides/update-guide/860-to-870.md
```diff
@@ -59,3 +59,84 @@ In that case, we append the following records:
...
+
+## Prefix migration
+
+If a custom prefix has been defined for Operate and Tasklist (for example, using `camunda.operate.elasticsearch.indexPrefix`), then data must be migrated manually to the new indices.
+
+### Make the Elasticsearch instance reachable
+
+The migration relies on communicating directly with Elasticsearch. The Elasticsearch instance must be reachable on `http://localhost:9200`.
+
+- **Kubernetes**: Port-forward your Elasticsearch instance to ensure it is reachable. In the following example, `camunda-platform` can be replaced by the release name:
```
ChrisKujawa, line 71:
```suggestion
- **Kubernetes**: Port-forward your Elasticsearch instance to ensure it is reachable. In the following example, `camunda-platform` can be replaced by the service name:
```

ChrisKujawa: I wouldn't reference release name here, as it might be that the customer/user has deployed ES separate

ChrisKujawa, line 71: We can note that our example relies on a kubernetes environment but works also with others.

docs/self-managed/operational-guides/update-guide/860-to-870.md
```diff
@@ -59,3 +59,84 @@ In that case, we append the following records:
...
+
+```shell
+docker pull camunda/zeebe:SNAPSHOT
+```
+
+Run the command below, replacing `CAMUNDA_DATABASE_PREFIX` value with your desired prefix:
+
+```shell
+docker run --rm -it --env CAMUNDA_DATABASE_URL="host.docker.internal:9200" --env CAMUNDA_DATABASE_INDEXPREFIX=<prefix> --entrypoint /usr/local/zeebe/bin/schema camunda/zeebe:SNAPSHOT
+```
```
ChrisKujawa, line 89: ❓ Why is this necessary?

docs/self-managed/operational-guides/update-guide/860-to-870.md
```diff
@@ -59,3 +59,84 @@ In that case, we append the following records:
...
+1. In Postman, import the collection `Prefix migration.postman_collection.json`.
+2. In the Postman client, set the global variables which template the requests, where `oldOperatePrefix` and `oldTasklistPrefix` are the prefixes which were previously configured, and `prefix` is the new prefix for your harmonized indices.
+3. Execute all requests in the collection.
+
+### Migrate archived indices
+
+For each archived index, you must:
+
+- Put the index in read only mode.
+- Clone the index with the changed name.
```
ChrisKujawa, line 104: I think it would be beneficial to explain a bit more why we chose this approach, instead of reindexing again. 

For example because we would need to pre create the indices, read only because we don't want to have any write during cloning etc.

ChrisKujawa: We should also put a warning here that this will increase the disc usage by a lot (I mean it will double it)

docs/self-managed/operational-guides/update-guide/860-to-870.md
```diff
@@ -59,3 +59,84 @@ In that case, we append the following records:
...
+curl -X POST "http://localhost:9200/operate-prefix-job-8.6.0_2025-01-08/_clone/new-prefix-job-8.6.0_2025-01-08" -H "Content-Type: application/json" -d '{
+  "settings": {
+    "index.blocks.write": false,
+    "number_of_replicas": 0
+  },
+  "aliases": {
+    "new-prefix-job-8.6.0_alias": {}
+  }
+}'
+```
```
ChrisKujawa, line 128: I feel I can write a script for such.

```sh
#!/bin/bash

set -euox pipefail

prefix=$1
newPrefix=$2

indices=$(curl "http://localhost:9600/$prefix*" | jq --raw-output "keys[]" | grep -E "$prefix.+_[0-9]+")

echo "Found $indices containing historic data."
echo "Cloning indices to new prefix"

for index in $indices; do

  echo "Set $index read-only"
  curl -X PUT "http://localhost:9200/$index/_settings" -H "Content-Type: application/json" -d '{
    "settings": {
       "index.blocks.write": "true"
    }
  }'


  # Clone index
  newIndexName=$(echo "${index/$prefix/$newPrefix}")

  echo "Clone $index to $newIndexName"
  newAliasName=$(echo "${index%_*}_alias")


  curl -X POST "http://localhost:9200/$index/_clone/$newIndexName" -H "Content-Type: application/json" -d '{
    "settings": {
      "index.blocks.write": false,
      "number_of_replicas": 0
    },
    "aliases": {
      "$newAliasName": {}
    }
  }'

done

echo "Migration to new custom prefix $newPrefix done"
```

@EuroLew can you validate whether this works for you?

docs/self-managed/operational-guides/update-guide/860-to-870.md
```diff
@@ -59,3 +59,84 @@ In that case, we append the following records:
...
+
+With the Elasticsearch migration finished, ensure the following prior to updating to 8.7:
+
+- The prefixes configured for Operate and Tasklist are changed to reflect the new prefix for the harmonized index. If this configuration is not changed, then Operate and Tasklist will continue to refer to the pre-migration indices which will not reflect any new changes.
+
+  For example, given a new prefix of `some-prefix`:
+
+  - The configured `CAMUNDA_OPERATE_ELASTICSEARCH_INDEX_PREFIX` must now be `some-prefix`.
+  - The configured `CAMUNDA_TASKLIST_ELASTICSEARCH_INDEX_PREFIX` must now be `some-prefix`.
+  - The prefix for the [Camunda exporter](/self-managed/zeebe-deployment/exporters/camunda-exporter.md) should be configured using `ZEEBE_BROKER_EXPORTERS_CAMUNDA_ARGS_INDEX_PREFIX` to `some-prefix`.
```
ChrisKujawa, line 142: I guess we should be more explicit that they need to set these configs

ChrisKujawa (commented):
Thanks @EuroLew good first version. I have some comments and thoughts, see below.

docs/self-managed/operational-guides/update-guide/860-to-870.md
EuroLew: will do

docs/self-managed/operational-guides/update-guide/860-to-870.md
EuroLew: thanks for the help it worked :)

docs/self-managed/operational-guides/update-guide/860-to-870.md
ChrisKujawa: Yeah I just later realized that we need to do this, because of the reindex. Could we mention this?

docs/self-managed/operational-guides/update-guide/860-to-870.md
```diff
@@ -59,3 +59,73 @@ In that case, we append the following records:
...
+docker run --rm -it --env CAMUNDA_DATABASE_URL="host.docker.internal:9200" --env CAMUNDA_DATABASE_INDEXPREFIX=<prefix> --entrypoint /usr/local/zeebe/bin/schema camunda/zeebe:SNAPSHOT
+```
+
+### Migrating indices
+
+The migration of the indices assumes that elasticsearch is reachable on
+localhost:9200.
+
+Execute the `.migrate.sh` bash script which has dependencies on `jq` and `curl`.
+It is required that the: operate, tasklist and new prefix are defined.
```
ChrisKujawa, line 103:
```suggestion
We better UX we provide a bash script: `.migrate.sh` to run the migration. The bash script which has dependencies on `jq` and `curl`.
It is required that the: operate, tasklist and new prefix are defined.

If you would like to run the migration on your own, please have a look at the script, what steps are necessary.
```

ChrisKujawa: Actually, can you bring back the description from before? I would like to have it still, the bash is just a good add-on to make it easier, but I think it would be useful to understand what is happening and some might want to ran it manually.

ChrisKujawa: We might want to also add that they should make sure to backup their data before plus testing the script first in dev environment to be safe.

docs/self-managed/operational-guides/update-guide/860-to-870.md
```diff
@@ -59,3 +59,131 @@ In that case, we append the following records:
...
+
+:::note
+Without these configuration changes the update to 8.7 will fail.
+:::
+
+Given a new desired prefix of `some-prefix`
+
+- The configured `CAMUNDA_OPERATE_ELASTICSEARCH_INDEX_PREFIX` must now be `some-prefix`.
+- The configured `CAMUNDA_TASKLIST_ELASTICSEARCH_INDEX_PREFIX` must now be `some-prefix`.
+- The prefix for the [Camunda exporter](/self-managed/zeebe-deployment/exporters/camunda-exporter.md) should be configured using `ZEEBE_BROKER_EXPORTERS_CAMUNDA_ARGS_INDEX_PREFIX` to `some-prefix`.
```
akeller, line 189: Are these changes part of 8.7 or 8.8?

EuroLew: The migration application should be part 8.7

ChrisKujawa: ❌ No 8.8

EuroLew: yep my bad it got delayed, will have to change documentation to reflect this

docs/self-managed/operational-guides/update-guide/860-to-870.md
```diff
@@ -67,116 +67,29 @@ It is a requirement that Operate and Tasklist (for example, using `camunda.opera
...
+2. In the `camunda/bin` directory there is a `prefix-migration` script that will execute the prefix migration,
+   pass the desired prefix using the `CAMUNDA_DATABASE_INDEXPREFIX` environment variable.
 
-Options:
-  --operate <value>     Prefix defined for operate
-  --tasklist <value>    Prefix defined for tasklist
-  --prefix <value>      New prefix for harmonised indices
-  -h, --help            Display this help message.
+```shell
+CAMUNDA_DATABASE_INDEXPREFIX=some-prefix ./prefix-migration
```
ChrisKujawa, line 87: ❓ @EuroLew dont we need also the previous index prefixes?

EuroLew: The script uses `OperateProperties` from the old cluster to retrieve the previous index prefixes, as a result looking at my instructions there's an error instead of stopping the 8.7 cluster it should pause exporting, run migration, then bring the cluster down

EuroLew: Or instead of changing instructions to pause exporting we keep it as is
- stop old cluster
- run migration
- start new cluster

and force the user to provide the old prefixes during the migration, imo the pause exporting approach is more user friendly

ChrisKujawa:
This PR will be replaced with https://github.com/camunda/camunda-docs/pull/5023

PR #4868, opened 2025-01-17
docs(web-modeler): rename milestone to version

wollefitz:
## Description

Part of https://github.com/camunda/web-modeler/issues/12551

## PR Checklist

- [x] I added a redirect for a renamed or deleted page to the .htaccess file.


wollefitz:
Screenshots in this PR are still pending to be updated; will do this after final design review in https://github.com/camunda/web-modeler/pull/12548

akeller:
👋 I'm surprised we didn't have "milestone" in the [glossary](https://docs.camunda.io/docs/reference/glossary/), but we should add it and "version" to help people remap this term in their heads.

wollefitz:
> 👋 I'm surprised we didn't have "milestone" in the [glossary](https://docs.camunda.io/docs/reference/glossary/), but we should add it and "version" to help people remap this term in their heads.

@akeller Thanks for the pointer to the glossary! We've been discussing in the team that at some location within the documentation we should explain the difference between the version concepts of the different components (Web Modeler version vs Zeebe engine version vs process definition version) - would the glossary be a good place for this?

akeller:
> @akeller Thanks for the pointer to the glossary! We've been discussing in the team that somewhere within the documentation we should explain the difference between the version concepts of the different components (Web Modeler version vs Zeebe engine version vs process definition version, etc.) - would the glossary be a good place for this?

It depends! 🙂 Could you pull together a little draft (could be in this PR or a Google doc) so I could see how verbose this would be? My vision with the glossary is this is a quick reference with a term and definition. If we need to go into a lot of details, this may not be the right place.

wollefitz:
> If we need to go into a lot of details, this may not be the right place.

I fear we have to go a bit more into details and suggest doing this in a follow-up PR as we already renamed "milestones" to "versions" in Web Modeler SaaS and I don't want to delay this PR further.

@akeller Would this work for you?

docs/components/modeler/web-modeler/versions.md
wollefitz, line 1: FYI: This page is marked as added by GitHub but is more a copy of `milestones.md` with adjusting the word "milestones".

akeller:
> > If we need to go into a lot of details, this may not be the right place.
> 
> I fear we have to go a bit more into details and suggest doing this in a follow-up PR as we already renamed "milestones" to "versions" in Web Modeler SaaS and I don't want to delay this PR further.
> 
> We will also need to add an announcement on the announcements page.
> 
> @akeller Would this work for you?

Yes, works for me.

mesellings:
> @camunda/tech-writers A thing that I wasn't able to adjust within this PR is https://github.com/camunda/camunda-docs/blob/main/docs/components/modeler/web-modeler/img/process-applications/process-application-milestone-diagram.png which also should say "Create version" instead of "Create milestone". I didn't find the source for recreating this diagram.

I can handle this for you 👍

mesellings:
> > @camunda/tech-writers A thing that I wasn't able to adjust within this PR is https://github.com/camunda/camunda-docs/blob/main/docs/components/modeler/web-modeler/img/process-applications/process-application-milestone-diagram.png which also should say "Create version" instead of "Create milestone". I didn't find the source for recreating this diagram.
> 
> I can handle this for you 👍

@wollefitz I've changed this and committed it 👍

mesellings (dismissed):
Great work @wollefitz, really thorough! 👏 

I've made a few edits for links and a few occurrences of milestone, but all lgtm! 🚀 

Also, before you merge, just to confirm this is what you are after:
- Milestone is changing to version here only in the 'next' docs, which will then become 8.8
- Milestone remains as is for 8.6 here, and thus will also. remain for 8.7

mesellings (approved):
Re-approving, as I edited the image name from milestone to version for completeness 👍

akeller:
> Also, before you merge, just to confirm this is what you are after:
> 
> * Milestone is changing to version here only in the 'next' docs, which will then become 8.8
> * Milestone remains as is for 8.6 here, and thus will also. remain for 8.7

We are still working out the details here, but please assume `next` will be 8.7 as expected until we communicate otherwise. Sorry this is confusing!

wollefitz:
> Milestone is changing to version here only in the 'next' docs, which will then become 8.8

Milestones will be renamed to versions in 8.7 already so we'll need to backport this to the 8.7 versioned docs once they are available.

wollefitz:
> but please assume `next` will be 8.7 as expected until we communicate otherwise. Sorry this is confusing!

Merging now based on Amara's comment. Thank you both for your support!

PR #4875, opened 2025-01-20
connectors: Add hint about merging set cookie headers

ztefanie:
## Description

Added hint about new feature of grouping headers

## PR Checklist

### DIFF
--- a/docs/components/connectors/protocol/rest.md
+++ b/docs/components/connectors/protocol/rest.md
@@ -234,6 +234,10 @@ The following fields are available in the `response` variable:
 Starting from version 8.7.0, the REST Connector supports storing the response as a document. See additional details and limitations in [document handling](/components/concepts/document-handling.md).
 :::
 
+:::note
+If your endpoint returns multiple Set-Cookie headers and you need to capture all of them, set `groupSetCookieHeaders` to `true` to aggregate them into a list.
+:::
+
 ## Output mapping
 
 ### Result variable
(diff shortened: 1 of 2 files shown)


docs/components/connectors/protocol/rest.md
```diff
@@ -234,6 +234,10 @@ The following fields are available in the `response` variable:
 Starting from version 8.7.0, the REST Connector supports storing the response as a document. See additional details and limitations in [document handling](/components/concepts/document-handling.md).
 :::
 
+:::note
+If your endpoint returns multiple Set-Cookie headers and you need to capture all of them, set `groupSetCookieHeaders` to `true` to aggregate them into a list.
+:::
```
johnBgood, line 239: Maybe can we show the returned list as an example (this would help customers figure out the separator we used etc)

johnBgood (approved):

docs/components/connectors/protocol/rest.md
ztefanie: @johnBgood I am not sure if this is really needed as this is like all lists in FEEL, so this should be covered here: https://docs.camunda.io/docs/components/modeler/feel/language-guide/feel-list-expressions/

Or did you mean something else?

johnBgood: Oh, I thought we return a `String` with the values separated by `;` 
If it's a real list, then all good 👍

PR #4878, opened 2025-01-20
docs(self-managed): add download link for Docker Compose

aabouzaid:
## Description

As we move away from cloning repos to using artifacts, I added the download link for Docker Compose.

## PR Checklist


aabouzaid:
@camunda/tech-writers, do we have a function that resolves the version of the docs so there is no need to update the links for every release?

Something like this (or so):

```
https://github.com/camunda/camunda-self-managed/releases/download/docker-compose-<DocsVersion/>/docker-compose-<DocsVersion/>.zip
```

docs/self-managed/setup/deploy/local/docker-compose.md
```diff
@@ -23,8 +23,8 @@ While the [Docker images](/self-managed/setup/deploy/other/docker.md) themselves
 
 To start a complete instance of Camunda 8 Self-Managed environment locally:
 
-1. Clone the [Camunda Self-Managed repository](https://github.com/camunda/camunda-self-managed), and navigate to the directory of the desired Camunda version (for example, `/version/camunda-8.6`).
-2. Run the following command:
+1. Download the [Camunda 8 Docker Compose artifact](https://github.com/camunda/camunda-self-managed/releases/download/docker-compose-alpha/docker-compose-alpha.zip) and extract it.
+2. Enter the extracted directory, and run the following command:
```
mesellings, line 27:
```suggestion
2. Run the following command in the extracted directory:
```
Rewording suggestion 👍

mesellings (approved):
Approved, great work - just one suggestion for rewording 👍

aabouzaid:
@mesellings Thanks for the review :raised_hands: 

If there is no function that resolves the version, I can try to create one.

mesellings:
> @mesellings Thanks for the review 🙌
> 
> If there is no function that resolves the version, I can try to create one.

@aabouzaid I would liaise with @pepopowitz before spending any time on this 👍

conceptualshark (commented):
Thanks for the follow-up! 🙌 

@aabouzaid Re: the release link, there is an ongoing discussion in Distro on how to handle this for C8Run, which has the same issue - I think this could likely be handled there, as opposed to a docs-side fix.

conceptualshark (dismissed):

aabouzaid:
@conceptualshark Conventionally, it's not possible to have the `latest` for each version (e.g. latest for 8.6 and latest for 8.5)
 Also, technically, we cannot have that in Git, as each version has its own tag with the name in it.
 
So doing it in the Docs looks like the most straightforward way to do it.
Because, in the first place, it's a Docs challenge, as we copy the whole dir and move it to the versioned dir.

conceptualshark:
@aabouzaid Looks like the option for C8Run was to link directly to the releases page: https://github.com/camunda/distribution/issues/369 / https://camunda.slack.com/archives/C03NFMH4KC6/p1736961684034949?thread_ts=1736947708.579689&cid=C03NFMH4KC6 but was dropped to route the user more directly.

However, the SM repo isn't as cluttered. Linking to [the SM releases](https://github.com/camunda/camunda-self-managed/releases/) with a message "Download the appropriate artifact for your version", etc, might be more appropriate here? @hisImminence / @akeller tagging you both for a repeat of that convo!

The majority of updating we've had to do with the C8Run artifacts are for patches (8.6.2) and alphas (alpha-2 vs alpha-3 etc), which have no matching docs version, so I am not sure how to resolve those with a docs-side function. 

I'm also happy to get this PR merged as-is and move this conversation! 🚀

aabouzaid:
@conceptualshark we will have more releases there for other components + the whole idea is to avoid redirecting users to the Git repo as it affects the UX (we saw that many times in the past, and customers are more confused when they sent outside the docs).

For Docker Compose, we are using a rolling release, which means there is 1 release per minor version, so it matches the docs versioning (except for `next`, which will be `alpha`).

For example, the current link always looks like this for 8.5 (which is equivalent to `latest` for that version):
```
https://github.com/camunda/camunda-self-managed/releases/download/docker-compose-8.5/docker-compose-8.5.zip
```

So unlike C8Run (and other artifacts), with the dynamic version, we set it once, and no need to change it again when we copy the docs.

aabouzaid:
@conceptualshark To see it in action, I added the change I want to make to better understand what I want to do.

Code:

```
1. Download the artifact for Camunda 8 <DockerCompose/>, then extract it.
```

Rendered:

Screenshot from 2025-01-22 17-35-01


No need to worry about the changes across versions because they are dynamically generated.

src/components/CamundaDistributions/DockerCompose.jsx
```diff
@@ -0,0 +1,22 @@
...
+const DockerComposeBaseURL =
+  "https://github.com/camunda/camunda-self-managed/releases/download";
+
+const DockerCompose = () => {
+  const docsVersion = DocsVersion();
+  const version = docsVersion == "Next" ? "alpha" : docsVersion;
+  return (
+    <a
+      style={{ color: "var(--ifm-link-color)" }}
+      class="theme-doc-version-badge badge badge--secondary"
```
conceptualshark, line 13: I don't think the style/class is necessary; returning the link should format it appropriately.

conceptualshark (commented):
I understand the process here and think there is value in having this available, but I have a few remaining questions:
1. Is a new alpha released at the time a minor is released to ensure the /next/ docs always have something to point to?
2. The docs will have a preemptive 8.7 release cut and a new /next version of the docs generated in February. I believe this will throw off any dynamic versioning until the 8.7 product release in April. 

Given number 2, and that these links are static once out of alpha, I am not sure if it would be best to implement this following the 8.7 release?

src/components/DocsVersion/index.jsx
pepopowitz, line 1: Since this isn't actually emitting any markup for the browser, this function would be better handled as a React hook than a component, named something like `useActiveVersionLabel`. 

But I also don't see that it adds much value beyond using `useActiveVersion` directly. It seems like an unnecessary abstraction. While it might save us from specifying `.label` on the active version, it also introduces the need to handle the exceptional case in a custom way (`if result === "No active version found"`).

aabouzaid: Yeah, that makes sense; it fits more like a hook.

Maybe we should remove it as a whole, given (please correct me if I'm wrong) that `useActiveVersion` will always return a version in Docusaurus's versioned setup.

pepopowitz: I agree, I think that we don't ever intend to use this hook in a situation where the active version can't be identified.

akeller:
> @aabouzaid Looks like the option for C8Run was to link directly to the releases page: [camunda/distribution#369](https://github.com/camunda/distribution/issues/369) / https://camunda.slack.com/archives/C03NFMH4KC6/p1736961684034949?thread_ts=1736947708.579689&cid=C03NFMH4KC6 but was dropped to route the user more directly.
> 
> However, the SM repo isn't as cluttered. Linking to [the SM releases](https://github.com/camunda/camunda-self-managed/releases/) with a message "Download the appropriate artifact for your version", etc, might be more appropriate here? @hisImminence / @akeller tagging you both for any input from that convo!
> 
> The majority of updating we've had to do with the C8Run artifacts are for patches (8.6.2) and alphas (alpha-2 vs alpha-3 etc), which have no matching docs version, so I am not sure how to resolve those with a docs-side function.
> 
> I'm also happy to get this PR merged as-is and move this conversation! 🚀

Let's move this to a separate conversation so we can pull in all the right people and details. @conceptualshark can you create an issue in the backlog? For now, it will likely be a manual update. We need to find a solution that's intentional and as intuitive as possible. I worry version mapping is hugely disruptive to an intuitive experience.

src/components/DocsVersion/index.jsx
aabouzaid: Thanks for your feedback 🙌
Removed :heavy_check_mark:

hisImminence:
> > @aabouzaid Looks like the option for C8Run was to link directly to the releases page: [camunda/distribution#369](https://github.com/camunda/distribution/issues/369) / https://camunda.slack.com/archives/C03NFMH4KC6/p1736961684034949?thread_ts=1736947708.579689&cid=C03NFMH4KC6 but was dropped to route the user more directly.
> > However, the SM repo isn't as cluttered. Linking to [the SM releases](https://github.com/camunda/camunda-self-managed/releases/) with a message "Download the appropriate artifact for your version", etc, might be more appropriate here? @hisImminence / @akeller tagging you both for any input from that convo!
> > The majority of updating we've had to do with the C8Run artifacts are for patches (8.6.2) and alphas (alpha-2 vs alpha-3 etc), which have no matching docs version, so I am not sure how to resolve those with a docs-side function.
> > I'm also happy to get this PR merged as-is and move this conversation! 🚀
> 
> Let's move this to a separate conversation so we can pull in all the right people and details. @conceptualshark can you create an issue in the backlog? For now, it will likely be a manual update. We need to find a solution that's intentional and as intuitive as possible. I worry version mapping is hugely disruptive to an intuitive experience.

The separate discussion is about C8Run, right? 

For Docker Compose, @aabouzaid found a way to automate the update. Could we use this dynamic function and try it out? Every less manual step in the release is a blessing for us ;) 

In case we see issues we can always roll back to manual updates. I would then add a note in our release docs to double the versions for the next docker compose release. Would that be good with you @akeller and @conceptualshark ?

akeller:
@hisImminence this works for me if I understand it correctly.

My comment about separating out the conversation(s) was so we could keep the PR scope to just update the link for Docker Compose. I see we have both content and docs infrastructure changes in the same PR. Do we plan to do something similar to C8Run? In this PR or separate?

hisImminence:
> @hisImminence this works for me if I understand it correctly.
> 
> My comment about separating out the conversation(s) was so we could keep the PR scope to just update the link for Docker Compose. I see we have both content and docs infrastructure changes in the same PR. Do we plan to do something similar to C8Run? In this PR or separate?

Super thanks! Make sense to me to keep the scope of this PR only to docker compose (what is the case afai can see) and have C8Run in a separate discussion/ PR.

So if @conceptualshark agrees too, we can merge (or can I also approve 🤔)? (We are waiting for this to be released to continue the communication and updates in other repos about docker compose 😬).

conceptualshark (commented):
It looks like this will resolve the temporary 8.7 and 8.8 mismatch both to the alpha artifact, so that works for me. 👍 Will distribution be responsible for removing this line at the next minor?

Otherwise, the styling should be updated to match the standard link formatting, and this is good to go.

src/components/CamundaDistributions/DockerCompose.jsx
hisImminence: 👍 removed it!

hisImminence:
> Otherwise, the styling should be updated to match the standard link formatting, and this is good to go.

Ah saw [your comment](https://github.com/camunda/camunda-docs/pull/4878#discussion_r1925673181) about it and removed it - @aabouzaid was also fine with removing!

hisImminence:
Ready to release??!! 🎈 🥳

conceptualshark (approved):
🥳 Thank you so much, good to go!

PR #4902, opened 2025-01-24
Fix improperly formatted internal deep links; prevent them from happening again.

pepopowitz:
## Description

Discovered in #4872.

- Adjusts internal deep-links to use a format that works in both docusaurus 2 and 3.
- Adds a vale rule to prevent these from happening in the future (although once we're on docusaurus 3, the build will just not pass when they're introduced 😅 )

Once this is merged, I'll propagate to my docusaurus 3 branch.

## What's wrong with these links?

The links in question are internal, targeting a document via file path including extension...and they include a slash (/) before the hash (#) of the anchor on the target page.

Docusaurus 2 appears to have no problem rendering links with this format. None of the links I've adjusted in this PR seem to be broken in production.

In Docusaurus 3, the engine does not treat these links as internal, and it does not attempt to translate the file path into the corresponding URL. You can see this happening [in this build](), in my docusaurus 3 working branch. 

Both versions of Docusaurus appear to have no problem with the format where the slash is removed. Thus, we can update these links in `main` now, instead of in the docusaurus 3 branch.

## What it looks like when an infraction is introduced

I temporarily added some infractions to this PR, to generate these comments: 

image

## PR Checklist

### DIFF
--- a/docs/apis-tools/frontend-development/01-task-applications/01-introduction-to-task-applications.md
+++ b/docs/apis-tools/frontend-development/01-task-applications/01-introduction-to-task-applications.md
@@ -13,7 +13,7 @@ Task applications are the interface between humans and Camunda processes to orch
 
 ## What are task applications?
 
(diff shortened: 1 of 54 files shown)


akeller (approved):
😍 🤩 

Ok, I was creeping on this PR earlier and I LOOOOOVE that we can lint for it! 🙌 Thank you for correcting and linting it. See ya never improperly formatted internal deep links!

Issue #4870, opened 2025-01-17
feature request: document harmonized index schema, and update it on demand


pepopowitz:
Originally discussed [in slack](https://camunda.slack.com/archives/C026U8GBNSW/p1734106695856439). Requested by @ChrisKujawa.

> We would like to document our new harmonized index[^1] structure and make this more consumable by users (in general to have something in our docs about it).
>
> Based on the [input](https://camunda.slack.com/archives/C05QBRGMA9F/p1733328060637059?thread_ts=1732547075.013999&cid=C05QBRGMA9F) I understood it would be great to have something similar to what we have [for C7 (ER diagrams) ](https://docs.camunda.org/manual/7.22/user-guide/process-engine/database/database-schema/#entity-relationship-diagrams).
>
> Be aware the indices schemas are in JSON and can be found [here](https://github.com/camunda/camunda/tree/main/webapps-schema/src/main/resources/schema/elasticsearch/create). Ideally, we have something automated (maybe we can brainstorm on this). I did a small research on this and it seems to be possible to either generate images out of JSON or simply generate tables via JS.

## Implementation notes

- [Mermaid](https://mermaid.js.org/) may be a good format for emitting these diagrams.
- We should use [the sync workflow for the c8 REST api docs](https://github.com/camunda/camunda-docs/blob/main/.github/workflows/sync-rest-api-docs.yaml) as an example of pulling a spec from another repo, executing a transformation, and opening a PR for review.
- The transformation itself is the most uncertain part of this process. How can we map the JSON index specs into Mermaid diagrams that are useful? What should the diagrams look like? etc.

[^1]: the harmonized index schema refers to indices that are created and used by the Camunda platform (core) starting with 8.7. See [this document](https://docs.google.com/document/d/1EFZ19Gx8Nf559pP_Bg8ObFMfGYdlq20age8P_WiSBOY/edit?tab=t.0) for more details.

pepopowitz:
marking as `hold` and moving to `blocked` because I'm setting it aside to focus on other things.

ChrisKujawa:
@pepopowitz likely you have lot of other things todo, but I just want to mention this would be really great to have for 8.8, as customers requesting this heavily \cc @aleksander-dytko

pepopowitz:
@ChrisKujawa I think I should be able to work it in before the 8.8 release. The nice thing is, the docs are not subject to a code freeze, so we have until basically the week before 8.8 ships. 

[...]

Issue #1189, opened 2022-08-22
Explore Mermaid for docs diagrams


akeller:
Our docs diagrams are images from various powerpoints and themes/colors associated with Zeebe, Camunda Cloud, and Camunda Platform 8 branding. There are also concerns about these diagrams not being accessible and in most cases, don't include a high-quality text description.

Instead of UML, @Sijoma suggested [Mermaid](https://mermaid-js.github.io/mermaid/#/). While separate from Kubernetes, it's the identified way to make diagrams in the [Kubernetes style guide](https://kubernetes.io/docs/contribute/style/diagram-guide/). 

There are a few issues and requests for Docusaurus to include Mermaid, particularly after [GitHub now supports it](https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/). For now, we can use this - https://github.com/sjwall/mdx-mermaid.

Additionally, there is a Discourse plugin for Mermaid. Theoretically, this would allow our contributor and developer community ecosystem to standardize on Mermaid for diagrams.

akeller:
@pepopowitz thoughts on Mermaid?

pepopowitz:
I think it's neat! We played with it a bit at my previous job, but nothing very deep. I love the idea of diagrams-as-code, though.

The biggest concern I have is if Mermaid fully covers the diagrams we have. I can't tell very easily from their docs if we can/how to extend it if we find that there are shapes/concepts it can't draw that we need to render, say, a BPMN diagram.

But also, would that be a blocking issue? If we can cover 80% of our diagrams with Mermaid but 20% of them can't be drawn, do we leave the 20% as images and consider it a win until we can find/build a way to convert them?

akeller:
Ignore BPMN diagrams. I'm thinking anything non-BPMN we should standardize with something like Mermaid.

akeller:
I'm thinking more about the architecture diagrams like the one [here](https://docs.camunda.io/docs/components/zeebe/technical-concepts/architecture/) or the variety [here](https://docs.camunda.io/docs/components/zeebe/technical-concepts/clustering/).

pepopowitz:
Happy to ignore. It'd be a nice experiment to try to re-draw a few of them with Mermaid at https://mermaid.live/

(Unimportant SEO note: that's not the first result in my search for "mermaid playground". [This is](https://www.amazon.com/Barbie-FXT20-Chelsea-Mermaid-Playset/dp/B07FFWNH9R).

[...]

PR #4918, opened 2025-02-02
[POC] Generate mermaid diagrams from harmonized index schemas

pepopowitz:
Should not be merged as-is. This PR exists for the purpose of soliciting feedback.

## Description

My hackathon project, a proof of concept of #4870. 

Adds a new page containing diagrams of the main & runtime harmonized indexes. Includes scripts to generate those diagrams from the upstream index schemas.

## Preview

https://preview.docs.camunda.cloud/pr-4918/docs/next/self-managed/operational-guides/backup-restore/harmonized-indexes/

image

## What's in this PR

1. Copies the [harmonized index schemas from upstream](https://github.com/camunda/camunda/tree/main/webapps-schema/src/main/resources/schema/elasticsearch/create) for ease of working. Long-term, I'd expect us to check out the `camunda/camunda` repo as part of the workflow.
2. Adds a script (`1-combine-sources`) to combine all the JSON files into one big one (or technically two, to retain the distinction between main & runtime indexes). I'm undecided if this would be useful for a long-term workflow.
3. Adds a script (`2-generate-mermaid-diagrams`) to generate proper markdown to emit mermaid diagrams for all schemas in a JSON file. 
   - Note that I would like **extra hackathon points** for not only **writing unit tests** during a hackathon, but using **TDD** to derive this logic.
4. Adds a script (`2-generate-mermaid-output`) to dump the generated markdown into a file.
5. Adds markdown partials to include that output in [a new page](https://preview.docs.camunda.cloud/pr-4918/docs/next/self-managed/operational-guides/backup-restore/harmonized-indexes/).

## Implementation notes

- I put the new page under self-managed/backup and restore because that seemed like the most likely place someone might be interested in knowing what the schema looked like.
- The schema definitions aren't truly an "Entity Relationship" diagram; that type of diagram seemed like the best fit that I could find.
- The mermaid integration into docusaurus is such that very large diagrams become unreadably small. The diagram is limited to the width of the page, and is itself an SVG, and it gets scaled down to fit the page. 
   - I intentionally left 6 entities in one diagram so you can see this scaling in action, in the 4th row of entities as you scroll down the page.
   - In response, I chose for this prototype to create diagrams of 3 schemas, and stack them on top of each other vertically. I looked briefly into adding scrollbars and making one big scrollable diagram, but I didn't have success with that.
   - We do have the ability to control [the layout algorithm](https://mermaid.js.org/intro/syntax-reference.html#supported-layout-algorithms), but neither of them lays the entities out in a more readable format than what you're seeing.
- There are multiple concepts in the schemas that we'll need to address somehow.  
   - `object` types: an index defines nested types in its schema. I chose to represent these using ERD relationships, extracting the nested type into a separate entity (see `camunda-authorization` index). This is visually inaccurate according to ERDs; it is not a separate index; but it might be a better way to describe things?
      - An example would be grouping a `firstName` and `lastName` under a `name`, or this from our indexes: 
        ```json
        "permissions": {
          "type": "object",
          "properties": {
            "type": {
              "type": "keyword"
            },
            "resourceIds": {
              "type": "keyword"
            }
          }
        }
        ```
   - `join` types: in this case, an item in an index references other items in the same index, through a defined relationship. This is not exactly a concept that a traditional Entity Relationship diagram handles out of the box. I tried joining a couple indexes to themselves (see `camunda-group` and `tasklist-task` indexes) to represent this; I left the others as using the type `join`.
      - Here's an example from a schema: 
         ```json
         "joinRelation": {
            "type": "join",
            "eager_global_ordinals": true,
            "relations": {
              "processInstance": ["activity", "variable"]
            }
          },
          ```
   - There may be some implied "relationships" across indexes, which I did not represent. For example, the `operate-decision` index includes a `decisionRequirementsKey` property, which I _think_ might refer to the `key` of an item in the `operate-decision-requirements` index. 
      - It would make a more complete diagram to include these implied relationships, if I am understanding them correctly.
      - However it would destroy my workaround for mermaid/docusaurus's tendency to squish large diagrams, as I would need all entities in one diagram. That isn't a reason not to do it, but including the relationships would mean I'd have to find a different workaround.
      - The implied relationships are not represented in the indexes. If we chose to represent them in the diagrams, I'd need to maintain a list of relationships here, separate from the upstream source. Again, not a reason not to do it, but it does introduce complexity and fragility.

## Other changes required before "done"

1. The generation of the markdown from JSON schemas would happen in a GitHub workflow.
2. Content for the page needs to be written. https://docs.google.com/document/d/1EFZ19Gx8Nf559pP_Bg8ObFMfGYdlq20age8P_WiSBOY/edit?tab=t.0#heading=h.c447h0byekxu is a good starting point for this. I will likely solicit a technical writer to help me with this 😅 

## Decisions to be made/feedback I'm interested in

Because I think it will be easier for reviewers, I will post a list of questions in a comment, so that you can reply to it with any of your feedback.

Never, at least not in this form!

### DIFF
--- a/docs/self-managed/operational-guides/backup-restore/_harmonized-indexes-main.md
+++ b/docs/self-managed/operational-guides/backup-restore/_harmonized-indexes-main.md
@@ -6,11 +6,11 @@ erDiagram
         keyword ownerType
         keyword resourceType
     }
-    camunda-authorization-permissions {
+    permissions {
         keyword type
         keyword resourceIds
     }
-    camunda-authorization ||--o{ camunda-authorization-permissions : has
+    camunda-authorization ||--o{ permissions : has
 
     camunda-mapping {
         keyword id
--- a/docs/self-managed/operational-guides/backup-restore/_harmonized-indexes-main.md
+++ b/docs/self-managed/operational-guides/backup-restore/_harmonized-indexes-main.md
@@ -1,0 +1,30 @@
+```mermaid
+erDiagram
+    camunda-authorization {
+        keyword id
+        long ownerKey
+        keyword ownerType
+        keyword resourceType
+    }
+    camunda-authorization-permissions {
+        keyword type
+        keyword resourceIds
+    }
+    camunda-authorization ||--o{ camunda-authorization-permissions : has
+
+    camunda-mapping {
+        keyword id
+        long key
+        keyword claimName
+        keyword claimValue
+        keyword name
+    }
+
+    camunda-role {
+        keyword id
+        long key
+        keyword name
+        long memberKey
+        join join
+    }
+```
--- a/.github/workflows/sync-harmonized-indexes/1-combine-sources.js
+++ b/.github/workflows/sync-harmonized-indexes/1-combine-sources.js
@@ -1,0 +1,41 @@
+const fs = require("fs").promises;
+const path = require("path");
+const currentDir = __dirname;
+
+// valid values: `templates` or `indexes`
+const config = {
+  sourceDirectoryPath: currentDir + "/templates",
+  outputFile: currentDir + "/templates.json",
+};
+
+// Combines independent JSON files in a directory into one JSON file,
+//   for ease of processing.
+async function combineSources() {
+  // Read all files from directory
+  const files = await fs.readdir(config.sourceDirectoryPath);
+
+  // Filter for .json files and process each
+  const jsonFiles = files.filter((file) => file.endsWith(".json"));
+
+  const combined = {};
+
+  for (const file of jsonFiles) {
+    const content = await fs.readFile(
+      path.join(config.sourceDirectoryPath, file),
+      "utf8"
+    );
+
+    // Use filename without extension as key
+    const key = path.basename(file, ".json");
+    combined[key] = JSON.parse(content);
+  }
+
+  // Write combined object to new file
+  await fs.writeFile(config.outputFile, JSON.stringify(combined, null, 2));
+
+  return combined;
+}
+
+combineSources();
+
+module.exports = { combineSources };
--- a/.github/workflows/sync-harmonized-indexes/2-generate-mermaid-diagrams.js
+++ b/.github/workflows/sync-harmonized-indexes/2-generate-mermaid-diagrams.js
@@ -38,7 +38,7 @@ function generateEntity(entityName, properties) {
       const [propName, propDefinition] = currentProperty;
       const { type } = propDefinition;
 
-      if (type === "object") {
+      if (type === "object" && propDefinition.properties !== undefined) {
         extraEntities.push(generateEntity(propName, propDefinition.properties));
         extraEntities.push(generateRelationship(propName, entityName));
       } else {
--- a/.github/workflows/sync-harmonized-indexes/2-generate-mermaid-diagrams.spec.js
+++ b/.github/workflows/sync-harmonized-indexes/2-generate-mermaid-diagrams.spec.js
@@ -1,5 +1,5 @@
 // @ts-check
-const { generateMermaidDiagrams } = require("./generate-mermaid-diagrams.js");
+const { generateMermaidDiagrams } = require("./2-generate-mermaid-diagrams.js");
 
 // /** @type {Array<import("./identify-missing-changes").VersionConfig>} */
 // let versionConfig;
--- a/.github/workflows/sync-harmonized-indexes/2-generate-mermaid-output.js
+++ b/.github/workflows/sync-harmonized-indexes/2-generate-mermaid-output.js
@@ -1,0 +1,23 @@
+const fs = require("fs").promises;
+const path = require("path");
+const { generateMermaidDiagrams } = require("./2-generate-mermaid-diagrams");
+const currentDir = __dirname;
+
+// valid values: `templates` or `indexes`
+const config = {
+  sourceFile: currentDir + "/templates.json",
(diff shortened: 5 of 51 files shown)


pepopowitz:
## Things I'm directly seeking feedback on

- Is this better than generating text-only tables? A downloadable diagram?
- What do you think about the `object` representation I described? Do you think an extracted object in the diagram is confusing, since it technically lives on the original index object? Is it clear from the diagram what's going on?
- What do you think about the `join` alternatives I presented in the description? Do you have a preference? Is there another way to represent this that might be better? 
- Do you think it would be better to include the implied relationships I described?
- How many extra hackathon points will you give me for the unit tests & TDD?

And of course, anything else that's on your mind 😄

docs/self-managed/operational-guides/backup-restore/_harmonized-indexes-main.md
```diff
@@ -0,0 +1,232 @@
...
+    camunda-authorization ||--o{ permissions : has
+
+    camunda-group {
+        keyword id
+        long key
+        keyword name
+        long memberKey
+        join join
+    }
+    camunda-group ||--o{ camunda-group: "group:member"
```
pepopowitz, line 24: I forgot to mention this but I snuck these `join` treatments into the generated markdown manually, to see what they would look like. The scripts would need to be updated to accommodate these, if we decide we like them.

akeller:
> Note that I would like extra hackathon points for not only writing unit tests during a hackathon, but using TDD to derive this logic.

🪙 🙌

akeller:
> * The mermaid integration into docusaurus is such that very large diagrams become unreadably small. The diagram is limited to the width of the page, and is itself an SVG, and it gets scaled down to fit the page.

Zoomable images are back again, I see. @Sijoma was interested in Mermaid via [this issue](https://github.com/camunda/camunda-docs/issues/1189). He might have a larger example we could try in our docs to see how the experience would be. I fear most of our diagrams will end up being very large 🥲 . 

> "Harmonized indexes" feels like a thing we call these internally, and I wonder if a user would know what that meant. Is there a better more simplified name for the page?

This is probably a @ChrisKujawa question. It's part of platform unification and (IMO) users only really need to know about it for migrating from the multi-component concept. It's the Camunda platform core indices, but maybe with more capitalization. 
  
> How do you feel about the location of the page? Should it be moved somewhere else?

Good question. I think it's part of the architecture (self-managed/reference-architecture/#architecture) but also part of the update guide (self-managed/operational-guides/update-guide/introduction/). I don't know how often it would be referenced outside the context of updating. 

> Is this better than generating text-only tables? A downloadable diagram?

🤷‍♀️ Immediately, I wondered why we wouldn't offer multiple view/use options, but it doesn't have must to do with this presentation of info.  

> What do you think about the object representation I described? Do you think an extracted object in the diagram is confusing, since it technically lives on the original index object? Is it clear from the diagram what's going on?

I think this is clear, but I also think I cheated by looking at other representations of this data. I also don't have much feedback on the remaining questions because the diagram looks fairly simple...? But maybe I'm missing something.

> How many extra hackathon points will you give me for the unit tests & TDD?

10x

docs/self-managed/operational-guides/backup-restore/_harmonized-indexes-main.md
```diff
@@ -0,0 +1,232 @@
+```mermaid
+erDiagram
+
+    camunda-authorization {
+        keyword id
+        long ownerKey
+        keyword ownerType
+        keyword resourceType
```
ChrisKujawa, line 8: ❓ One thing I was wondering was whether we could turn the values around. That we first have the name of the property then the type? I felt this is somehow more natural. What are your thoughts? 

For example this is also done in the C7 ER diagram https://docs.camunda.org/manual/7.22/user-guide/process-engine/database/database-schema/#entity-relationship-diagrams

ChrisKujawa:
Maybe @aleksander-dytko or @ingorichtsmeier want to give some input here as well :)

aleksander-dytko:
@pepopowitz some thoughts:

> Yeah, I think it should just be like Indicies or maybe "Secondary Storage Schema" something. For C7 we call the page Database Schema

I think we should officially introduce "Primary Storage" and "Secondary Storage" In our docs to describe the data pipeline. This would be useful for further reference e.g. [here](https://docs.google.com/document/d/1xeciKYf4yejXP8y62wdhls1yagjGZuKsI2PaRcixl-o/edit?tab=t.0)

## Visuals
I believe having a visual representation of the schema is useful to quickly orientate in C8. We could first show the list indices and have each line clickable, with the details of the schema. 
image

ChrisKujawa:
Input from @ingorichtsmeier https://camunda.slack.com/archives/C05QBRGMA9F/p1739373311735919?thread_ts=1732547075.013999&cid=C05QBRGMA9F

akeller:
This is now permanently blocked with no resourcing on the Docs Team to move forward.

christinaausley:
@ChrisKujawa Are you comfortable with closing this for now as there is no WIP?

ChrisKujawa:
I guess we can close it, but it is a bit unfortunate as we want to have something for the indicies.

christinaausley:
@ChrisKujawa @MaxTru We do still have the open issue for this, but it is currently unassigned https://github.com/camunda/camunda-docs/issues/4870. 

I don't think @mesellings and I will get to this anytime soon, so I am going to temporarily close this PR. If someone assigns themselves to the issue they can either re-open this PR or open a new one.

PR #4937, opened 2025-02-05
d3: rewrite Metadata override

pepopowitz:
## Description

Part of #4799.

Reintroduces the Metadata override component.

The only notable differences are:

1. This one is now written in TypeScript
2. The path to two hooks changed
3. One property name changed (`unversionedId` to `id`).


.husky/pre-commit
```diff
@@ -1,4 +1,6 @@
 #!/bin/sh
+# Before adding version-manager code here, read https://typicode.github.io/husky/how-to.html#solution.
```
pepopowitz, line 2: Unrelated, but while setting up my new laptop I learned why people occasionally like to add a line here that activates nvm.

src/theme/DocItem/Metadata/determineCanonical.js
pepopowitz, line 1: I was going to convert this file & its tests to also be typescript, but it was too courageous for how quickly I'm trying to move.

PR #4981, opened 2025-02-11
Add back Zeebe REST API to deprecated section in 8.7

akeller:
## Description

Based on my understanding of the 8.7/8.8 split, I think we need to add the Zeebe REST API backed to the deprecated section in the API docs (as it is in 8.6). This PR does that.

## PR Checklist

### DIFF
--- a/versioned_docs/version-8.7/apis-tools/zeebe-api-rest/sidebar-schema.js
+++ b/versioned_docs/version-8.7/apis-tools/zeebe-api-rest/sidebar-schema.js
@@ -1,0 +1,12 @@
+/** @type {import('@docusaurus/plugin-content-docs').SidebarsConfig} */
+
+module.exports = {
+  "Zeebe API (REST)": [
+    "apis-tools/zeebe-api-rest/zeebe-api-rest-overview",
+    "apis-tools/zeebe-api-rest/zeebe-api-rest-authentication",
+    "apis-tools/zeebe-api-rest/zeebe-api-tutorial",
+    {
+      Specifications: require("./specifications/sidebar.js"),
+    },
+  ],
+};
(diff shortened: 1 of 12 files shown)


tmetzke:
In general, I think the whole API part can be copied from 8.6, to be honest.

The only changes in 8.7 really are
1. There are some new endpoints for document handling
2. We're changing the default for key attributes to string

For 1. and 2., we need to regenerate the reference docs from the 8.7 OpenAPI file.
For 2., we also need to adjust the section about key attributes (copy from current "next" (8.8), I think).

That should be it if I am not completely overlooking something else.
I think all other removals/deprecations of API had to be moved to later anyway.

akeller:
> In general, I think the whole API part can be copied from 8.6, to be honest.
> 
> The only changes in 8.7 really are
> 
> 1. There are some new endpoints for document handling
> 2. We're changing the default for key attributes to string
> 
> For 1. and 2., we need to regenerate the reference docs from the 8.7 OpenAPI file. For 2., we also need to adjust the section about key attributes (copy from current "next" (8.8), I think).
> 
> That should be it if I am not completely overlooking something else. I think all other removals/deprecations of API had to be moved to later anyway.

When you say "the whole API part" do you mean other component APIs as well?

tmetzke:
Well, I guess so, since all 8.7 code is just based on 8.6 plus some dedicated single additions here and there. I guess @aleksander-dytko has a better overview of the other APIs but I believe the only changes in 8.7 now are to the C8 REST API (as mentioned above). Not sure if anything changes in Tasklist, Operate, or any other API (e.g. Web Modeler).

aleksander-dytko:
> Well, I guess so, since all 8.7 code is just based on 8.6 plus some dedicated single additions here and there. I guess @aleksander-dytko has a better overview of the other APIs but I believe the only changes in 8.7 now are to the C8 REST API (as mentioned above). Not sure if anything changes in Tasklist, Operate, or any other API (e.g. Web Modeler).

Fully agreed - the 8.7 API is basically the same as 8.6 (with caveats from @tmetzke). All API changes have been [postponed](https://camunda.com/blog/2024/12/api-changes-in-camunda-8-a-unified-and-streamlined-experience/) until 8.8 release.

aleksander-dytko (approved):

Issue #4989, opened 2025-02-12
bug: `allOf` inheritance does not always render all inherited properties


pepopowitz:
Originally discussed [in Slack](), most of these words and examples are from @tmetzke.

We make heavy use of `allOf` inheritance in our C8 REST API schemas, and usually the schema explorer renders all inherited properties. Occasionally, not all properties are rendered -- rendering instead only the properties on a parent item, and excluding anything from an inherited object. 

This appears to happen only in the schema explorer. All inherited properties seem to render in the examples window.

This is currently affecting the 8.6 and 8.7 docs. This will not be an issue anymore starting with 8.8 since we just removed the additional media types. They will only exist in 8.6 and 8.7 for a user migration path (so added with 8.6, flipped default with 8.7, removed with 8.8).

### Activate jobs

https://docs.camunda.io/docs/apis-tools/camunda-api-rest/specifications/activate-jobs/

[Specification](https://github.com/camunda/camunda/blob/c78b7f6c81eda72f78bc133a26865085667560b4/zeebe/gateway-protocol/src/main/proto/rest-api.yaml#L53-L92).

#### Schema explorer (incorrect)

![Image](https://github.com/user-attachments/assets/cb50c764-dcc2-45b0-a75a-c08f3717a940)

#### Examples window (correct)

![Image](https://github.com/user-attachments/assets/3bb3f0d4-1d1d-41b2-967a-839c59374885)

### Find user tasks

https://docs.camunda.io/docs/8.7/apis-tools/camunda-api-rest/specifications/find-user-tasks/

![Image](https://github.com/user-attachments/assets/fc2c89e3-5439-46d3-8c8a-d5c5f558277d)

### Evaluate decision

https://docs.camunda.io/docs/8.7/apis-tools/camunda-api-rest/specifications/evaluate-decision/

![Image](https://github.com/user-attachments/assets/69cbfbaf-43bb-47fc-b041-d996868efd92)

PR #4991, opened 2025-02-12
demo: fix missing properties when `allOf` inheritance is used, sometimes

pepopowitz:
DO NOT MERGE, any fix like this should be made upstream and propagate down here through the sync process.

## Description

Demonstrates a fix for #4989.

## A screenshot of it working

From my localhost, or running at [this preview URL](https://preview.docs.camunda.cloud/pr-4991/docs/apis-tools/camunda-api-rest/specifications/activate-jobs/). Note that I've fixed the issue only in version 8.6, and only for the application/json and number+json responses of the activate jobs endpoint, _not_ the string+json response.

image

## How I fixed it

Basically, I added an empty base schema for the top-level response schema to inherit. 

Note that it does not matter which properties are on that base schema; the mere presence of it triggers the plugin to render all properties correctly.

### wtf is going on???!!!?!

The difference between the endpoints that emit all properties and the endpoints that don't is another `allOf` statement at the top-level response schema. I don't know _why_ this makes a difference, but it does. The presence of an inherited base schema on the top-level response schema somehow convinces the schema explorer to emit all properties on the array items; the lack of an in inherited base schema convinces it to only emit the base properties on the array items.

It's weird and hard to explain, so here are some links and code snippets. Right now, in the upstream spec: 

#### This endpoint emits all properties on the array items (in this case the `evaluatedDecisions` array)

https://docs.camunda.io/docs/8.7/apis-tools/camunda-api-rest/specifications/evaluate-decision/

[Evaluate decision endpoint definition](https://github.com/camunda/camunda/blob/stable/8.7/zeebe/gateway-protocol/src/main/proto/rest-api.yaml#L1167-L1211):

```yaml
  /decision-definitions/evaluation:
    post:
      ...
      responses:
        "200":
          description: The decision was evaluated.
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/EvaluateDecisionResult"
```

[EvaluateDecisionResult definition](https://github.com/camunda/camunda/blob/stable/8.7/zeebe/gateway-protocol/src/main/proto/rest-api.yaml#L3720-L3737) (note the `allOf` that points to a base object): 

```yaml
    EvaluateDecisionResult:
      type: object
      allOf:
        - $ref: "#/components/schemas/EvaluateDecisionResponseBase"
      properties:
        ...
        evaluatedDecisions:
          type: array
          items:
            $ref: "#/components/schemas/EvaluatedDecisionResult"
```

[EvaluateDecisionResponseBase definition](https://github.com/camunda/camunda/blob/stable/8.7/zeebe/gateway-protocol/src/main/proto/rest-api.yaml#L3668C5-L3697): 

```yaml
    EvaluateDecisionResponseBase:
      description: Base properties for EvaluateDecisionResponse
      type: object
      properties:
        ...
```

#### This endpoint does not emit all properties on the array items (in this case the `jobs` array)

https://docs.camunda.io/docs/8.7/apis-tools/camunda-api-rest/specifications/activate-jobs/

[Job activation endpoint definition](https://github.com/camunda/camunda/blob/stable/8.7/zeebe/gateway-protocol/src/main/proto/rest-api.yaml#L53-L78):

```yaml
  /jobs/activation:
    post:
      ...
      responses:
        "200":
          description: The list of activated jobs.
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/JobActivationResult"
```

[JobActivationResult definition](https://github.com/camunda/camunda/blob/stable/8.7/zeebe/gateway-protocol/src/main/proto/rest-api.yaml#L3238-L3245) (note that there is no inheritance of a base schema for this object):
```yaml
    JobActivationResult:
      description: The list of activated jobs. Key attributes as string values.
      type: object
      properties:
        jobs:
          type: array
          items:
            $ref: "#/components/schemas/ActivatedJobResult"
```

To see how I fixed the issue by adding an empty base schema, see the diff on this PR.

## Follow-up work

- me: fix this in an upstream PR for this endpoint
- me: try to identify other endpoints that might suffer from this, and fix them upstream too
- me: open an issue with the generator plugin


tmetzke:
Pure gold, @pepopowitz 🏅🌟

That was also the only difference I could see but I thought 
> Well, this can't be it, why would this make any difference?
Stupid me, always question those things 😆

pepopowitz:
Fixed in #5027.

PR #5149, opened 2025-03-05
chore: add security notice 14

markfarkas-camunda:
## Description

We had a critical vulnerability in Web Modeler. Jira ticket: https://jira.camunda.com/browse/SEC-1285

## PR Checklist

### DIFF
--- a/docs/reference/notices.md
+++ b/docs/reference/notices.md
@@ -1,12 +1,43 @@
 ---
 id: notices
 title: "Security notices"
-description: "Let's take a closer look at security notices, reporting vulnerabilities, and addiitonal security information."
+description: "Let's take a closer look at security notices, reporting vulnerabilities, and additional security information."
 ---
 
 ## Security notices
 
 Camunda publishes security notices after fixes are available.
+
+### Notice 14
+
+#### Publication date
+
+March 6th, 2025
+
+#### Product affected
+
+Camunda Web Modeler
+
+#### Impact
+
+The version of `koa` used by Camunda Web Modeler was affected by the following vulnerability:
+
+- https://nvd.nist.gov/vuln/detail/CVE-2025-25200
+
+#### How to determine if the installation is affected
+
+You are using Camunda Web Modeler version 8.3.16, 8.4.14, 8.5.15, 8.6.7, 8.7.0-alpha4, 8.8.0-alpha1 or previous.
+
+#### Solution
+
+Camunda has provided the following releases which contain the fix:
+
+- [Web Modeler 8.3.17](https://github.com/camunda/web-modeler/releases/tag/self-managed%2F8.3.17)
+- [Web Modeler 8.4.15](https://github.com/camunda/web-modeler/releases/tag/self-managed%2F8.4.15)
+- [Web Modeler 8.5.16](https://github.com/camunda/web-modeler/releases/tag/self-managed%2F8.5.16)
+- [Web Modeler 8.6.8](https://github.com/camunda/web-modeler/releases/tag/self-managed%2F8.6.8)
+- [Web Modeler 8.7.0-alpha5](https://github.com/camunda/web-modeler/releases/tag/self-managed%2F8.7.0-alpha5)
+- [Web Modeler 8.8.0-alpha2](https://github.com/camunda/web-modeler/releases/tag/self-managed%2F8.8.0-alpha2)
 
 ### Notice 13
 
--- a/docs/reference/notices.md
+++ b/docs/reference/notices.md
@@ -12,7 +12,7 @@ Camunda publishes security notices after fixes are available.
 
 #### Publication date
 
-March 6th, 2025
+March 11th, 2025
 
 #### Product affected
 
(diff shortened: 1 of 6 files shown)


docs/reference/notices.md
```diff
@@ -1,13 +1,44 @@
...
+#### Solution
+
+Camunda has provided the following releases which contain the fix:
+
+- [Web Modeler 8.3.17](https://github.com/camunda/web-modeler/releases/tag/self-managed%2F8.3.17)
+- [Web Modeler 8.4.15](https://github.com/camunda/web-modeler/releases/tag/self-managed%2F8.4.15)
+- [Web Modeler 8.5.16](https://github.com/camunda/web-modeler/releases/tag/self-managed%2F8.5.16)
+- [Web Modeler 8.6.8](https://github.com/camunda/web-modeler/releases/tag/self-managed%2F8.6.8)
+- [Web Modeler 8.7.0-alpha5](https://github.com/camunda/web-modeler/releases/tag/self-managed%2F8.7.0-alpha5)
+- [Web Modeler 8.8.0-alpha2](https://github.com/camunda/web-modeler/releases/tag/self-managed%2F8.8.0-alpha2)
```
wollefitz, line 40: ❌ These links are not publicly available. We have to link somewhere else instead (for example the public Docker images `8.6.8`).

wollefitz, line 40: The two alphas can be removed.

wollefitz (changes requested):
Thanks for adding the notice! Additionally to the individual remarks, this has to be backported to all previous doc versions.

docs/reference/notices.md
```diff
@@ -1,13 +1,44 @@
...
+You are using Camunda Web Modeler version 8.3.16, 8.4.14, 8.5.15, 8.6.7, 8.7.0-alpha4, 8.8.0-alpha1 or previous.
+
+#### Solution
+
+Camunda has provided the following releases which contain the fix:
+
+- [Web Modeler 8.3.17](https://github.com/camunda/web-modeler/releases/tag/self-managed%2F8.3.17)
+- [Web Modeler 8.4.15](https://github.com/camunda/web-modeler/releases/tag/self-managed%2F8.4.15)
+- [Web Modeler 8.5.16](https://github.com/camunda/web-modeler/releases/tag/self-managed%2F8.5.16)
+- [Web Modeler 8.6.8](https://github.com/camunda/web-modeler/releases/tag/self-managed%2F8.6.8)
```
wollefitz, line 38: Let's keep this aligned with the naming above and also name the product "Camunda Web Modeler" here.

docs/reference/notices.md
markfarkas-camunda: Where do we have the public docker images?

markfarkas-camunda:
@camunda/tech-writers in this PR we cannot link the docker images for specific versions (which contains the fix) because Web Modeler does not distribute images prior to 8.6 publicly, only on registry.camunda.cloud. In this case what should be the desired format of this security notice?

docs/reference/notices.md
wollefitz: > Where do we have the public docker images?

Answered via https://github.com/camunda/camunda-docs/pull/5149#issuecomment-2703302391

markfarkas-camunda:
> Would it be ready then?

@akeller If you are referring to the docker images, those won't be available, Web Modeler just simply does not offer public Docker images, so we cannot link to previous versions, and we won't be able to in the future either. The question is: can we just skip the linking in this case? (mentioning the versions with the fix but only as plain text and not as links)

> this will go out next week with the official release

That is my understanding as well.

docs/reference/notices.md
```diff
@@ -1,13 +1,42 @@
...
+You are using Camunda Web Modeler version ≤ 8.3.16, ≤ 8.4.14, ≤ 8.5.15, or ≤ 8.6.7.
+
+#### Solution
+
+Camunda has provided the following releases which contain the fix:
+
+- [Camunda Web Modeler 8.3.17](https://github.com/camunda/web-modeler/releases/tag/self-managed%2F8.3.17)
+- [Camunda Web Modeler 8.4.15](https://github.com/camunda/web-modeler/releases/tag/self-managed%2F8.4.15)
+- [Camunda Web Modeler 8.5.16](https://github.com/camunda/web-modeler/releases/tag/self-managed%2F8.5.16)
+- [Camunda Web Modeler 8.6.8](https://hub.docker.com/layers/camunda/web-modeler-webapp/8.6.8/images/sha256-0d12ad2a751d9cc78de25f7a299320c8b7f7fa68f3b2e80f0c826fee926cb547)
```
wollefitz, line 38: 🔧 I don't think linking the webapp image provides any additional value here - my proposal would be to just completely remove the links and only mention the versions.

markfarkas-camunda: I removed the links

wollefitz (dismissed):
Thanks for the adjustments - looks good now!

docs/reference/notices.md
```diff
@@ -1,13 +1,42 @@
...
+You are using Camunda Web Modeler version ≤ 8.3.16, ≤ 8.4.14, ≤ 8.5.15, or ≤ 8.6.7.
+
+#### Solution
+
+Camunda has provided the following releases which contain the fix:
+
+- Camunda Web Modeler 8.3.17
+- Camunda Web Modeler 8.4.15
+- Camunda Web Modeler 8.5.16
+- Camunda Web Modeler 8.6.8
```
akeller, line 38: This should be backported so that when a user looks at their version of the docs, they can see there is a security notice. Tech writers can handle this for you today, so it's ready for tomorrow.

akeller (approved):

PR #5180, opened 2025-03-07
docs(data-purge): adds operational guide for data purge

srsp:
## Description

This PR describes the new cluster data purge functionality described here: [#25861](https://github.com/camunda/camunda/issues/25861)

## PR Checklist

### DIFF
--- a/docs/self-managed/operational-guides/data-purge.md
+++ b/docs/self-managed/operational-guides/data-purge.md
@@ -1,0 +1,268 @@
+---
+id: data-purge
+title: "Data purge"
+description: "Purge data from your cluster"
+---
+
+import Tabs from "@theme/Tabs";
+import TabItem from "@theme/TabItem";
+
+The data purge feature allows you to delete all runtime and historical data from your cluster. This operation resets the cluster to an empty state while maintaining the original topology.
+
+The purge operation performs two main actions:
+
+1. **Runtime Data Deletion**: Removes all live data from brokers, for example process definitions, instances, and jobs.
+2. **Historical Data Purge**: Clears exported data from configured exporters
+
+The data purge feature can be used to:
+
+- Delete data between test runs and therefore enabling reuse of the same cluster for multiple tests.
+- Resetting development or staging environments to a clean state.
+
+## Purge data
+
+You will need access to the Cluster API as described in the [Cluster scaling guide](self-managed/zeebe-deployment/operations/cluster-scaling.md) to perform the purge.
+
+:::danger
+The purge operation is irreversible. It will delete the runtime data in the cluster and the historical data in the exporters! Make sure to back up your data before proceeding.
+:::
+
+### Usage
+
+The purge operation is a cluster-wide, asynchronous operation. Since it is asynchronous, you first launch it by sending a `POST` request to `/actuator/cluster/purge`, and then monitor by polling the topology via `/actuator/cluster` until it is finished.
+
+<Tabs groupId="language" defaultValue="shell" queryString values={
+[
+{label: 'shell', value: 'shell' },
+{label: 'Java', value: 'java' },
+]}>
+
+<TabItem value='shell'>
+
+:::note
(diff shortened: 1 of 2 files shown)


docs/self-managed/operational-guides/data-purge/data-purge.md
```diff
@@ -0,0 +1,189 @@
+---
+id: data-purge
+title: "Data purge"
+description: "Purge data from your cluster"
+---
+
+The data purge feature allows you to delete all runtime and historical data from your cluster. This operation resets the cluster to an empty state while maintaining the original topology.
```
npepinpe, line 7: :wrench: We should emphasize the latter part - while yes it deletes data, it's more that it restores the cluster to a "fresh" state. So there will be _some_ things there after - e.g. the default user (if so configured), default tenant, some schema metadata, etc.

docs/self-managed/operational-guides/data-purge/data-purge.md
```diff
@@ -0,0 +1,189 @@
...
+
+- `changeId`: The ID of the changes initiated to scale the cluster. This can be used to monitor the progress of the scaling operation. The ID typically increases so new requests get a higher ID than the previous one.
+- `currentTopology`: A list of current brokers and the partition distribution.
+- `plannedChanges`: A sequence of operations that has to be executed to achieve scaling.
+- `expectedToplogy`: The expected list of brokers and the partition distribution once the scaling is completed. For the purge feature the expected topology will be the same as the current topology.
+
+<details>
+  <summary>Example response</summary>
+
+```
```
npepinpe, line 54: :question: Is the syntax highlighting automatic, or do we need to add `json` here?

docs/self-managed/operational-guides/data-purge/data-purge.md
```diff
@@ -0,0 +1,189 @@
...
+- Resetting development or staging environments to a clean state.
+
+## Purge data
+
+You will need access to the Cluster API as described in the [Cluster scaling guide](self-managed/zeebe-deployment/operations/cluster-scaling.md) to perform the purge.
+
+:::danger
+The purge operation is irreversible. It will delete the runtime data in the cluster and the historical data in the exporters! Make sure to back up your data before proceeding.
+:::
+
```
npepinpe, line 26: :wrench: Can we add a section either at the top or the end, which puts it all together? For example, I want to purge, so I would send the POST request, grab the change ID, and then poll the cluster via GET until my change ID is the last completed one. I think that's really helpful for users if we can "put it all together" in a runbook like format.

npepinpe (changes requested):
Very nice, just some suggestions :)

docs/self-managed/operational-guides/data-purge/data-purge.md
srsp: So it looks like I have taken one of the only sites as a model, where this was not done :) 
I'll add it.

docs/self-managed/operational-guides/data-purge/data-purge.md
srsp: I agree that the runbook style is good for the user. I had hoped that the current versions already accomplishes that. 
I fear that if I include that, it feels like there is too much repetition, because the current structure is already close to that. 

The screenshot is from a relatively small screen (23"), when scrolled to the "Purge data" paragraph. The runbook should come after the "Danger" block, so if people have already scrolled until there, they can already see both steps involved, even on a smaller screen. 

I could get rid of some of the explanations regarding the output, which I copied from the Cluster Scaling doc, and just reference it. That would make it more compact and less noisy. 

What do you think? 
image

npepinpe:
@saig0 I updated the docs. I'll still wait for a tech writer before merging, but this should be close to the final version. You can use the preview environment for usage.

npepinpe (approved):

conceptualshark (approved):
I made some small wording/grammatical changes, but this looks to be in good shape to me!

PR #5234, opened 2025-03-13
feat: adding extra manifests

hamza-m-masood:
## Description

closes: https://github.com/camunda/camunda-platform-helm/issues/2999

## PR Checklist

### DIFF
--- a/docs/self-managed/setup/guides/add-extra-manifests.md
+++ b/docs/self-managed/setup/guides/add-extra-manifests.md
@@ -1,7 +1,7 @@
 ---
 id: add-extra-manifests
-title: "Add extra manifests to the Camunda 8 Helm chart"
-sidebar_label: "Add extra manifests"
+title: "Add Kubernetes manifests"
+sidebar_label: "Add Kubernetes manifests"
 description: "Learn how to add extra manifests to Helm deployments by injecting arbitrary data in the values.yaml."
 ---
 
--- a/docs/self-managed/setup/guides/add-extra-manifests.md
+++ b/docs/self-managed/setup/guides/add-extra-manifests.md
@@ -1,16 +1,17 @@
 ---
 id: add-extra-manifests
-title: "Add Extra Manifests"
-description: "The ability to add extra manifests by injecting arbitrary data in the values.yaml"
+title: "Add extra manifests to the Camunda 8 Helm chart"
+sidebar_label: "Add extra manifests"
+description: "Learn how to add extra manifests to Helm deployments by injecting arbitrary data in the values.yaml."
 ---
 
 ## Overview
 
-This feature allows users to inject arbitrary data into their `values.yaml` file, enabling the addition of extra manifests. This feature is particularly useful for adding custom manifests by including additional Kubernetes resources such as ConfigMaps, Deployments, or Services.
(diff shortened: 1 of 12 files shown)


hamza-m-masood:
@camunda/distribution @conceptualshark this is good to go for an initial review. This can then be backported when the initial review is done.

hamza-m-masood:
I can't find an 8.3 label
cc @conceptualshark

hisImminence (dismissed):
Nice short documentation for this feature! Thanks @hamza-m-masood 

~Only question: this would be for 8.7 and 8.8 docs, right?~

As we just discussed. Fine for me to merge it first and then backport it to all versions in a follow up PR (in alignment with the HCs. 🚀 )

docs/self-managed/setup/guides/add-extra-manifests.md
```diff
@@ -0,0 +1,39 @@
+---
+id: add-extra-manifests
+title: "Add extra manifests to the Camunda 8 Helm chart"
+sidebar_label: "Add extra manifests"
```
conceptualshark, line 4: I'm wondering if "Add Kubernetes manifests" would be better in both of these locations. When finding this page in the sidebar, would that be more descriptive (and give an indication that it's deployment-specific)? It might also help with searching for this page.

docs/self-managed/setup/guides/add-extra-manifests.md
```diff
@@ -0,0 +1,39 @@
...
+    - |
+      apiVersion: v1
+      kind: ConfigMap
+      metadata:
+        name: example-cm
+      data:
+        test: test
+```
+
+For more information, see the Kubernetes [object documentation](https://kubernetes.io/docs/concepts/overview/working-with-objects/).
```
conceptualshark, line 28: I wanted to provide a little more information here. Is this the best Kubernetes link? 

I was also wondering if we could give any additional context, like:
- Can you add more than one manifest? What is the format for adding more than one? 
- Does the chart automatically create these resources when deployed? 
- What does maintenance look like for updating these objects? (...This might just be my ignorance here.)

Based on https://github.com/camunda/product-hub/issues/2464, it seems like the use case is deployment customization that doesn't involve modifying the Helm chart - I can call that out specifically, too, with any of the answers to the above.

hamza-m-masood: yes, you can add more manfiests. I can make that clearer by adding a second configmap in the array

> Does the chart automatically create these resources when deployed?
Helm will render the manifest and deploy the rendered manifests to the cluster. I don't think this point is worth mentioning int he guide.

> What does maintenance look like for updating these objects?
The customers maintains the extra manfiests themselves. That's why I listed the points in the best practices section. Also, I don't think this is worth adding to the guide.

docs/self-managed/setup/guides/add-extra-manifests.md
hamza-m-masood: Changed both to "Add Kubernetes manifests"

hamza-m-masood:
@conceptualshark I will assume you have approved since I addressed all your comments. I backported to all relevant versions. Please go ahead and merge.

PR #4837, opened 2025-01-10
connectors add hint to skip encoding for rest outbound

ztefanie:
## When should this change go live?

- [ ] This is a bug fix, security concern, or something that needs **urgent release support**.
- [ ] This is already available but undocumented and should be released within a week.
- [ ] This on a **specific schedule** and the assignee will coordinate a release with the DevEx team. (apply `hold` label or convert to draft PR)
- [ ] This is part of a scheduled alpha or minor. (apply alpha or minor label)
- [x] There is **no urgency** with this change and can be released at any time.

## PR Checklist

- [ ] My changes are for an **already released minor** and are in `/versioned_docs` directory.
- [x] My changes are for the **next minor** and are in `/docs` directory (aka `/next/`).

- [x] My changes require an [Engineering review](https://github.com/camunda/camunda-docs/blob/main/howtos/documentation-guidelines.md#review-process), and I've assigned the Engineering DRI or delegate.
- [ ] My changes require a [technical writer review](https://github.com/camunda/camunda-docs/blob/main/howtos/documentation-guidelines.md#review-process), and I've assigned @camunda/tech-writers as a reviewer.

PR #5236, opened 2025-03-13
fix(rest-connector): Add hints about breaking changes for pre encoded…

ztefanie:
… urls in rest connector

## Description

The rest connector does decode and endocde urls leading to pre encoded urls to break.
This was fixed [here](https://github.com/camunda/connectors/pull/3840) and documented [here](https://github.com/camunda/camunda-docs/pull/4837/)
With this PR i document the backport of this feature to 8.6. done [here](https://github.com/camunda/connectors/pull/4274) and add hints about the breaking change, when updating from 8.5 to 8.6. 
This breaking change only happens on self-managed, while on SaaS this was never working and therefore no breaking change was introduced with 8.6. Please verify this is reflected properly with the changes I made here, as I was not super sure where these changes need to be done, if only self manage is affected.
The new feature flag is available on SaaS and SM.

## PR Checklist

### DIFF
--- a/docs/self-managed/operational-guides/update-guide/850-to-860.md
+++ b/docs/self-managed/operational-guides/update-guide/850-to-860.md
@@ -31,7 +31,7 @@ If one of your endpoints returns multiple Set-Cookie headers and you need to cap
 :::
 
 :::caution Breaking change
-If one of your endpoints require pre-encoded URL elements this will break with 8.6.0. Since 8.6.12. the `skipEncoding` in the XML value can be set to `"true"`. This disables the automatic decoding and re-encoding process, ensuring the URL is sent to the server exactly as provided.
+If one of your endpoints requires pre-encoded URL elements, this behavior will change in version 8.6.0. Since 8.6.10, the element template includes the skipEncoding property, which can be set to "true". This disables the automatic decoding and re-encoding process, ensuring the URL is sent to the server exactly as provided.
 :::
 
 ### Kafka Connector supports Avro schema
--- a/docs/self-managed/operational-guides/update-guide/850-to-860.md
+++ b/docs/self-managed/operational-guides/update-guide/850-to-860.md
@@ -30,6 +30,10 @@ The default value of Elasticsearch deployment pods has changed from 2 to 3, and 
 If one of your endpoints returns multiple Set-Cookie headers and you need to capture all of them, set `groupSetCookieHeaders` to `true` in the element template XML. This aggregates the headers into a list. This feature is available since version 8.6.7. The grouping is enabled by default since version 8.6.11.
 :::
 
+:::caution Breaking change
+If one of your endpoints require pre-encoded URL elements this will break with 8.6.0. Since 8.6.12. the `skipEncoding` in the XML value can be set to `"true"`. This disables the automatic decoding and re-encoding process, ensuring the URL is sent to the server exactly as provided.
+:::
+
 ### Kafka Connector supports Avro schema
 
 :::caution Breaking change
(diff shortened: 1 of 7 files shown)


docs/self-managed/operational-guides/update-guide/850-to-860.md
```diff
@@ -30,6 +30,10 @@ The default value of Elasticsearch deployment pods has changed from 2 to 3, and
 If one of your endpoints returns multiple Set-Cookie headers and you need to capture all of them, set `groupSetCookieHeaders` to `true` in the element template XML. This aggregates the headers into a list. This feature is available since version 8.6.7. The grouping is enabled by default since version 8.6.11.
 :::
 
+:::caution Breaking change
+If one of your endpoints require pre-encoded URL elements this will break with 8.6.0. Since 8.6.12. the `skipEncoding` in the XML value can be set to `"true"`. This disables the automatic decoding and re-encoding process, ensuring the URL is sent to the server exactly as provided.
```
sbuettner, line 34: We could rephrase it and mention the `skipEncoding` property of the element template itself.

sbuettner: @ztefanie Wording looks good now. Customers use element-templates (and need to upgrade them) to make use of new features and they rarely change the code in the XML itself.

sbuettner (dismissed):

versioned_docs/version-8.6/components/connectors/protocol/rest.md
```diff
@@ -203,6 +203,12 @@ Secrets are currently not supported in the body of a **REST Connector**.
 }
 ```
 
+### Encoding
+
+In certain scenarios, such as when working with APIs that require pre-encoded URL elements, the REST Connector's default behavior may inadvertently modify encoded segments.
+
+To avoid this, set the `skipEncoding` value to `"true"` in the element template XML. This disables the automatic decoding and re-encoding process, ensuring the URL is sent to the server exactly as provided. This flag is available since 8.6.10.
+
```
conceptualshark, line 211: Should this change also be made in the 8.7 and 8.8 docs?

conceptualshark (commented):
Not a blocking request, but I think linking back to an element template page for both of these changes could be helpful. Do we have the XML options/properties documented anywhere? 

Otherwise I've just left one question on content. As a breaking change, I think these should also be mentioned on the version's announcement pages.

versioned_docs/version-8.6/components/connectors/protocol/rest.md
ztefanie: Thanks for having a look at this, as mentioned in the description of the PR, this PR is only for documenting the backport to 8.6. as the original documentation was done [here](https://github.com/camunda/camunda-docs/pull/4837/) and i think this was merged before 8.8 was created, so this should be present in 8.7. and 8.8. already

ztefanie:
> Not a blocking request, but I think linking back to an element template page for both of these changes could be helpful. Do we have the XML options/properties documented anywhere?

I think I do not understand what exactly you mean. Can you please explain this.

> Otherwise I've just left one question on content. As a breaking change, I think these should also be mentioned on the version's announcement pages.

Ah ok, I though the announcement pages are only for SaaS. I will add it to the announcement pages.

docs/self-managed/operational-guides/update-guide/850-to-860.md
sbuettner: @ztefanie Got you!

conceptualshark (approved):
Thank you for clarifying where the original change came from, and adding this into announcements! This looks good to me. 👍

PR #5249, opened 2025-03-13
Update Algolia crawler instructions for manual triggers

akeller:
## Description

As discussed in the backlog review.

I've also verified that all Docs Team members have the right access and should be able to trigger crawls manually. 

## PR Checklist

These changes are for the docs-on-docs.


mesellings (approved):
Nothing to add other than my thanks for getting this done, this does help us in being able to crawl the site whenever we need to now rather than relying on the scheduled crawl - great work! 🚀 👏

pepopowitz (approved):

PR #5266, opened 2025-03-15
docs(self-managed): update helm version matrix

aabouzaid:
## Description

As we release the alpha Helm charts publically, I've added the version matrix links for those charts.

Also, I added it to previous Camunda docs from 8.3-8.5

## PR Checklist

### DIFF
--- a/docs/reference/supported-environments.md
+++ b/docs/reference/supported-environments.md
@@ -85,9 +85,7 @@ For details on typical volume type usage, refer to the following examples specif
 
 ### Helm charts version matrix
 
-| Camunda version | Helm chart version                                                                      |
-| --------------- | --------------------------------------------------------------------------------------- |
-| Camunda 8.8.x   | [Helm chart 13.x](https://helm.camunda.io/camunda-platform/version-matrix/camunda-8.8/) |
+Camunda Helm chart version `13.x.x` works with Camunda version `8.8.x`. Check the [Helm chart version matrix](https://helm.camunda.io/camunda-platform/version-matrix/camunda-8.8/) for more details.
 
 ## Component requirements
 
--- a/docs/reference/supported-environments.md
+++ b/docs/reference/supported-environments.md
@@ -83,16 +83,16 @@ For details on typical volume type usage, refer to the following examples specif
 - [Microsoft AKS](/self-managed/setup/deploy/azure/microsoft-aks.md#volume-performance)
 - [Google GKE](/self-managed/setup/deploy/gcp/google-gke.md#volume-performance)
 
-### Helm version matrix
+### Helm charts version matrix
 
-As of the Camunda 8.4 release, the Camunda 8 Helm chart version is independent from the application version (for example, the Camunda 8.4 release uses the Helm chart version 9.0.0). The Helm chart is updated with each application release.
-
-| Camunda version | Helm chart version                                                                       |
-| --------------- | ---------------------------------------------------------------------------------------- |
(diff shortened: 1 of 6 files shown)


conceptualshark (commented):
I'm hesitant to include versions higher than the docs version (for example, if you're on 8.5, you should only see the matrix for 8.5 and lower) - it means there's more to remember and keep accurate at future releases. Is this something that is helpful to have documented in this way? @akeller Do you have any opinions on how much to document here?

akeller:
> I'm hesitant to include versions higher than the docs version (for example, if you're on 8.5, you should only see the matrix for 8.5 and lower) - it means there's more to remember and keep accurate at future releases. Is this something that is helpful to have documented in this way? @akeller Do you have any opinions on how much to document here?

I agree with you @conceptualshark. Unless there is a reason you need to see this info for higher versions, let's keep changes to anything current version and lower.

aabouzaid:
> it means there's more to remember and keep accurate at future releases

I believe it's the same. We need to go through all releases and remove the unsupported version for each one.

So when 8.7 is released, we need to go to all versions and remove 8.3
Which is even more work than copying and pasting the same content for all versions.

I'd say either we include the info for the current version or make all versions have the same content instead of having them.

@conceptualshark @akeller, what do you think?

akeller:
> @conceptualshark @akeller, what do you think?

We already remove/add info to the supported environment page for each minor version, so I don't see any differences there. If you don't want to have to deal with removing it manually for the minor, we can add that to our release manager checklist so you and your team just need to review it.

akeller:
@aabouzaid I added the `deploy` label because I'm wondering if we are not understanding each other.

Do we need to add the helm version matrix to the component version table? Or should we not have the helm version as a matrix/table at all? If there is one version of the helm chart for one minor version, we probably don't need a table. I'm having trouble seeing what the best way is to present this information based on the conversation.

aabouzaid:
@akeller That makes sense. Should I remove the table then and make a normal format?

conceptualshark:
> Do we need to add the helm version matrix to the component version table? Or should we not have the helm version as a matrix/table at all? If there is one version of the helm chart for one minor version, we probably don't need a table. I'm having trouble seeing what the best way is to present this information based on the conversation.

We'd thought add it to the [component version table](https://docs.camunda.io/docs/next/reference/supported-environments/#component-version-matrix), but I didn't want to make this table more unwieldy at the time, and it seemed more logical to find in the Self-Managed section. 

We do use a lot of "see the Helm version matrix to determine x y z" in the docs, but I don't know how often needing to know versions outside of the one you're on is necessary (outside of upgrading)? So maybe just having the single version match the docs version/product version is fine here.

aabouzaid:
I've used the normal format instead of the table for the Helm chart version :heavy_check_mark:

aabouzaid:
Ping @akeller @conceptualshark

akeller:
> I've used the normal format instead of the table for the Helm chart version ✔️

@aabouzaid these still look like tables, but I'm fine approving. It's an important update and we can always keep tweaking the format.

akeller (dismissed):

aabouzaid:
@akeller You are correct! I just found out the commit wasn't pushed 🤦‍♂️.

I just pushed it now, and it works as expected.
Screenshot 2025-03-26 at 23-30-35 Supported environments Camunda 8 Docs

akeller (approved):
:shipit:

PR #5338, opened 2025-03-21
Algolia crawler now runs daily

akeller:
## Description

This is the docs-on-docs update to state the Algolia crawler now runs daily. The change to the Algolia config has already happened.

## PR Checklist

Does not apply. These are unversioned docs-on-docs.


conceptualshark (approved):
🚀 !

pepopowitz (approved):

PR #5365, opened 2025-03-25
docs: explain how to use java client with APIs in a c8run environment

pepopowitz:
## Description

This is follow-up to [this discussion](https://github.com/camunda/camunda-docs/pull/4455#discussion_r2000627872).

There is a trick to using the Java client to connect to APIs via cookie auth in a C8Run environment. Thanks to @tmetzke, this PR explains that trick.

It only applies to v8.6 and v8.7, as in v8.8 [we no longer need cookie auth](https://docs.camunda.io/docs/next/self-managed/setup/deploy/local/c8run/#use-camunda-apis) to connect to the APIs.'

## PR Checklist

### DIFF
--- a/versioned_docs/version-8.6/apis-tools/camunda-api-rest/camunda-api-rest-authentication.md
+++ b/versioned_docs/version-8.6/apis-tools/camunda-api-rest/camunda-api-rest-authentication.md
@@ -159,3 +159,7 @@ curl --request GET 'http://localhost:8080/v2/topology'  \
   --cookie cookie.txt \
   --header 'Content-Type: application/json'
 ```
+
+### Authentication via cookie with the Java client
+
+To access Camunda APIs in a Camunda 8 Run environment with the Java client, you'll need to [manually retrieve and send a cookie](/self-managed/setup/deploy/local/c8run.md#use-camunda-apis-with-the-java-client).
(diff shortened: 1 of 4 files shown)


conceptualshark (approved):
This makes sense to me from a TW perspective - thank you for updating the C8Run page as well!

tmetzke (approved):
👍 Love it, great read! 🚀 

💭 The C8Run instructions also mention the gRPC API. I am not the greatest expert in authentication mechanisms in our platform, to be honest, but I believe that cookie authentication is not needed for that protocol. Maybe you can clarify this offline and is this is indeed not needed, we should be clear about that in this C8Run guide. Currently, it somehow reads like it's necessary but there is no example for gRPC calls, so that might confuse users at some point. This non-blocking and can go into a follow-up.

pepopowitz:
> 💭 The C8Run instructions also mention the gRPC API. I am not the greatest expert in authentication mechanisms in our platform, but I believe that cookie authentication is not needed for that protocol. Maybe you can clarify this offline, and if this is indeed not needed, we should be clear about that in this C8Run guide. Currently, it somehow reads like it's necessary, but there is no example for gRPC calls, so that might confuse users at some point. This is non-blocking and can go into a follow-up.

I don't personally have the capability to manage this follow-up right now.

PR #5431, opened 2025-04-02
fix: add another command and some elaboration to ops procedure so it is more general

bkenez:
## Description

The existing dual-region operational procedure guide claims to provide only generic commands regardless of which region had failed but at one point it uses a command as an example that is specific to one region's failure. While this is mentioned, for ease of use I have included example commands for both cases/

## PR Checklist

### DIFF
--- a/versioned_docs/version-8.6/self-managed/operational-guides/multi-region/dual-region-ops.md
+++ b/versioned_docs/version-8.6/self-managed/operational-guides/multi-region/dual-region-ops.md
@@ -298,13 +298,27 @@ In our example, we went with port-forwarding to a localhost, but other alternati
    ```
 
 3. Based on the [Cluster Scaling APIs](../../zeebe-deployment/operations/cluster-scaling.md), send a request to the Zeebe Gateway to redistribute the load to the remaining brokers, thereby removing the lost brokers.
-   In our example, we have lost region 1 and with that our uneven brokers. This means we will have to redistribute to our existing even brokers.
+   Depending on which region had been lost, the load needs to be redistributed to the remaining brokers, either the even, or odd numbered ones. In our example, we have lost region 1 and with that our uneven brokers. This means we will have to redistribute to our existing even brokers. Make sure to only run the correct one based on the surviving region's brokers.
 
-   ```bash
-   curl -XPOST 'http://localhost:9600/actuator/cluster/brokers?force=true' -H 'Content-Type: application/json' -d '["0", "2", "4", "6"]'
-   ```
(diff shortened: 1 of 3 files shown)


versioned_docs/version-8.6/self-managed/operational-guides/multi-region/dual-region-ops.md
```diff
@@ -298,13 +298,27 @@ In our example, we went with port-forwarding to a localhost, but other alternati
...
-   ```
+  <Tabs queryString="lost-region">
+    <TabItem value="redistribute-to-odd" label="Redistribute to odd brokers" default>
+
+    ```bash
+    curl -XPOST 'http://localhost:9600/actuator/cluster/brokers?force=true' -H 'Content-Type: application/json' -d '["1", "3", "5", "7"]'
+    ```
+
+    </TabItem>
+    <TabItem value="region-1-lost" label="Redistribute to even brokers">
```
Langleu, line 311: I'd suggest to make this the default as it also aligns better with the text and UX.

In the text we say, we lost uneven, need to redistribute to even but then the default was odd.

versioned_docs/version-8.7/self-managed/operational-guides/multi-region/dual-region-ops.md
```diff
@@ -331,13 +331,27 @@ In our example, we went with port-forwarding to a localhost, but other alternati
...
+
+  <Tabs queryString="lost-region">
+    <TabItem value="redistribute-to-odd" label="Redistribute to odd brokers" default>
+
+    ```bash
+    curl -XPOST 'http://localhost:9600/actuator/cluster/brokers?force=true' -H 'Content-Type: application/json' -d '["1", "3", "5", "7"]'
+    ```
+
+    </TabItem>
+    <TabItem value="region-1-lost" label="Redistribute to even brokers">
```
Langleu, line 344: same as above about default

Langleu (changes requested):
would suggest changing the default to align for better UX.
Don't forget to add the changes to `docs` current 8.8

versioned_docs/version-8.6/self-managed/operational-guides/multi-region/dual-region-ops.md
bkenez: Right, I did not notice the existence of that flag, changed

bkenez:
Changed the default tab and added changes to /docs as well

Langleu (dismissed):

PR #5477, opened 2025-04-07
docs: add V2 REST API redirects

### LINKED ISSUE #5093: Add redirects for C8 REST API search endpoints
## Context

With https://github.com/camunda/camunda/issues/28791, we adjust the search endpoints to consistently use "search" instead of mixing it with "find" and "query".

This includes adjusting the `operationId`s of endpoints, leading to new URLs.

## Acceptance criteria

* There are redirects from the former "find"-based URLs of C8 REST API endpoints to the new "search"-based URLs.

tmetzke:
## Description

The V2 REST API introduces new URLs for some endpoints due to added operation IDs in the OpenAPI spec.
Adding redirects for "next" docs (8.8 as of writing) to allow navigating correctly.

closes https://github.com/camunda/camunda-docs/issues/5093

## PR Checklist


tmetzke:
⚠️ When I test the redirects on the deployment, they don't work as I would expect them to:
* I browse the URL that should have a redirect now, e.g., `RewriteRule ^docs/next/apis-tools/camunda-api-rest/specifications/get-cluster-topology/?$ ^docs/next/apis-tools/camunda-api-rest/specifications/get-topology/$1 [R=301,L]` should lead to https://preview.docs.camunda.cloud/pr-5477/docs/next/apis-tools/camunda-api-rest/specifications/get-cluster-topology/ redirecting me to https://preview.docs.camunda.cloud/pr-5477/docs/next/apis-tools/camunda-api-rest/specifications/get-topology/.
* This does not work for other previously existing redirects in the file I adjusted either, though. For example, due to `RewriteRule ^docs/self-managed/zeebe-deployment/operations/metrics/?$ /docs/self-managed/operational-guides/monitoring/metrics/ [R=301,L]`, I expect I would be redirected from https://preview.docs.camunda.cloud/pr-5477/docs/self-managed/zeebe-deployment/operations/metrics/ to https://preview.docs.camunda.cloud/pr-5477/docs/self-managed/operational-guides/monitoring/metrics/.

ℹ️ They **do** work on the publicly available docs, so https://docs.camunda.io/docs/self-managed/zeebe-deployment/operations/metrics/ redirects me correctly. I would assume I can simply not easily test this before merging it?

❓ One question regarding behavior: Will this lead to seamless navigation from 8.7 to 8.8? So, for example, when I visit https://docs.camunda.io/docs/8.7/apis-tools/camunda-api-rest/specifications/get-cluster-topology/ and want to navigate to the 8.8 version of it by using the version dropdown in the header, will this redirect me accordingly to https://docs.camunda.io/docs/next/apis-tools/camunda-api-rest/specifications/get-topology/? Or will I land on the docs homepage because https://docs.camunda.io/docs/next/apis-tools/camunda-api-rest/specifications/get-cluster-topology/ does not exist?

pepopowitz:
👋 

> ⚠️ When I test the redirects on the deployment, they don't work as I would expect them to:

I don't actually think the preview environments run any htaccess redirects, so this is not surprising. 

> ❓ I would assume I can simply not easily test this before merging it?

What makes this set of redirects particularly difficult to test is that we don't run any checks of /next URLs. If these were of the current version, then they would be tested by our sitemap checks, which confirm the presence of redirects for URLs that have moved.

Usually if I'm looking to test redirects, I confirm using htaccess.madewithlove.com.

> ❓ One question regarding behavior: Will this lead to seamless navigation from 8.7 to 8.8? So, for example, when I visit https://docs.camunda.io/docs/8.7/apis-tools/camunda-api-rest/specifications/get-cluster-topology/ and want to navigate to the 8.8 version of it by using the version dropdown in the head navigation bar, will this redirect me accordingly to https://docs.camunda.io/docs/next/apis-tools/camunda-api-rest/specifications/get-topology/? Or will I land on the docs homepage because https://docs.camunda.io/docs/next/apis-tools/camunda-api-rest/specifications/get-cluster-topology/ does not exist?

You will land on the docs homepage, because the URL doesn't exist in the 8.8 version.

pepopowitz:
> Should I add a comment to them so `/next` is removed when 8.8 was released as "latest"? How do we handle this usually? I was wondering if we do a complete "scan and replace" for such things for a minor release or if I can already write them in a way that avoids this later?

no, that's okay! I do a scan for them when I cut a new version, but also the sitemap check usually breaks on the ones that I don't remove the `/next` from and that's a pretty good nudge 😅.

> Super helpful 🫶 The htaccess might actually be running in the preview envs but according to htaccess.madewithlove.com, the redirects wouldn't match due to the URL scheme of those envs (having `/pr-xyy/` as the first context path element, and not `/docs/`).

oh yeah, good call! I hadn't thought of that. I don't think there's any easy fix to get past that, if you can think of something let me know. A brute force replacement in the redirect rules seems possible, but also potentially error prone.

> Even with the new redirects? That's a shame, really. It would improve the UX for docs users A LOT, actually. Did we ever look into this?

Yes, unfortunately this is the case. The redirects don't factor into switching versions. Docusaurus keeps a manifest of all docs in each version, and if it can't match by path & document-id when switching versions, it will just kick it to the root. Overriding _display_ logic tends to be possible with Docusaurus swizzling; overriding _functional_ logic like this tends to be difficult, if even possible.

tmetzke:
Thanks, @pepopowitz, for those details 🙏 
I'm good to go then here for the redirects so far.

pepopowitz (approved):

PR #5479, opened 2025-04-07
docs: includeEnabledRecords property documentation

panagiotisgts:
## Description

Document the new introduced `includeEnabledRecords` property for Zeebe Elasticsearch & Opensearch exporter, that enables the exporting into old `zeebe-record` indices for Camunda 8.8 and onwards. By default this property is false.

## PR Checklist

### DIFF
--- a/docs/self-managed/operational-guides/update-guide/870-to-880.md
+++ b/docs/self-managed/operational-guides/update-guide/870-to-880.md
@@ -257,9 +257,10 @@ In that case, we append the following records:
 
 ### Zeebe record types
 
-With the introduction of the Camunda Exporter, Zeebe no longer exports all record types by default. Meaning that not all `zeebe-record` indices will be populated.
+With the introduction of the Camunda Exporter, the Elasticsearch and OpenSearch Exporter no longer exports all record types by default.
+Meaning that not all `zeebe-record` indices will be populated.
 
-::note This only affects records exported with the 8.8 version by the new CamundaExporter. During the migration, left over 8.7 records are exported as normal.
+::note This only affects records exported with the 8.8 version. During the migration, left over 8.7 records are exported as normal.
 ::
 
 The record types that continue to be exported by default are the following:
@@ -273,5 +274,5 @@ The record types that continue to be exported by default are the following:
 
 If there is still need for other record types to be exported you can still do so by setting the configuration property :
 
-- `ZEEBE_BROKER_EXPORTERS_ELASTICSEARCH_ARGS_ZEEBERECORDSEXPORTENABLED=true` for Elasticsearch
-- `ZEEBE_BROKER_EXPORTERS_OPENSEARCH_ARGS_ZEEBERECORDSEXPORTENABLED=true` for OpenSearch
+- `ZEEBE_BROKER_EXPORTERS_ELASTICSEARCH_ARGS_EXPORTLEGACYRECORDS=true` for Elasticsearch
+- `ZEEBE_BROKER_EXPORTERS_OPENSEARCH_ARGS_EXPORTLEGACYRECORDS=true` for OpenSearch
--- a/docs/self-managed/operational-guides/update-guide/870-to-880.md
+++ b/docs/self-managed/operational-guides/update-guide/870-to-880.md
@@ -270,9 +270,6 @@ The record types that continue to be exported by default are the following:
 - `PROCESS_INSTANCE`
 - `VARIABLE`
 - `USER_TASK`
-- `VARIABLE`
+- `INCIDENT`
 
-If there is still need for other record types to be exported you can still do so by setting the configuration property :
-
-- `ZEEBE_BROKER_EXPORTERS_ELASTICSEARCH_ARGS_EXPORTLEGACYRECORDS=true` for Elasticsearch
-- `ZEEBE_BROKER_EXPORTERS_OPENSEARCH_ARGS_EXPORTLEGACYRECORDS=true` for OpenSearch
+If there is still need for other record types to be exported you can still do so see `exportLegacyRecords` for [Elasticsearch](../../zeebe-deployment/exporters/elasticsearch-exporter.md#configuration) or [OpenSearch](../../zeebe-deployment/exporters/opensearch-exporter.md#configuration) exporter configuration respectively.
--- a/docs/self-managed/operational-guides/update-guide/870-to-880.md
+++ b/docs/self-managed/operational-guides/update-guide/870-to-880.md
@@ -257,11 +257,11 @@ In that case, we append the following records:
 
 ### Zeebe record types
 
-With the introduction of the Camunda Exporter, the Elasticsearch and OpenSearch Exporter no longer exports all record types by default.
-Meaning that not all `zeebe-record` indices will be populated.
+With the introduction of the Camunda Exporter, the Elasticsearch and OpenSearch Exporter no longer export all record types by default. Therefore, not all `zeebe-record` indices will be populated.
 
-::note This only affects records exported with the 8.8 version. During the migration, left over 8.7 records are exported as normal.
-::
+:::note
+This only affects records exported with the 8.8 version. During the migration, leftover 8.7 records are exported as normal.
+:::
 
 The record types that continue to be exported by default are the following:
 
@@ -272,4 +272,4 @@ The record types that continue to be exported by default are the following:
 - `USER_TASK`
 - `INCIDENT`
 
-If there is still need for other record types to be exported you can still do so see `includeEnabledRecords` for [Elasticsearch](../../zeebe-deployment/exporters/elasticsearch-exporter.md#configuration) or [OpenSearch](../../zeebe-deployment/exporters/opensearch-exporter.md#configuration) exporter configuration respectively.
+To export other record types, review `includeEnabledRecords` for [Elasticsearch](../../zeebe-deployment/exporters/elasticsearch-exporter.md#configuration) or [OpenSearch](../../zeebe-deployment/exporters/opensearch-exporter.md#configuration) exporter configuration.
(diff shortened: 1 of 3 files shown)


conceptualshark (dismissed):

docs/self-managed/operational-guides/update-guide/870-to-880.md
```diff
@@ -254,3 +254,25 @@ In that case, we append the following records:
 - `ASSIGN`
 - `ASSIGNING`
 - `ASSIGNED`
+
+### Zeebe record types
+
+With the introduction of the Camunda Exporter, Zeebe no longer exports all record types by default. Not all `zeebe-record` indices will be populated.
```
ChrisKujawa, line 260:
```suggestion
With the introduction of the Camunda Exporter, the Elasticsearch and OpenSearch Exporter no longer exports all record types by default. Not all `zeebe-record` indices will be populated.
```

docs/self-managed/operational-guides/update-guide/870-to-880.md
```diff
@@ -254,3 +254,25 @@ In that case, we append the following records:
...
+
+:::note
+This only affects records exported with the 8.8 version by the new Camunda Exporter. During the migration, leftover 8.7 records are exported as normal.
```
ChrisKujawa, line 263:
```suggestion
This only affects records exported with the 8.8 version. During the migration, leftover 8.7 records are exported as normal.
```

docs/self-managed/operational-guides/update-guide/870-to-880.md
```diff
@@ -254,3 +254,25 @@ In that case, we append the following records:
...
+- `PROCESS`
+- `PROCESS_INSTANCE`
+- `VARIABLE`
+- `USER_TASK`
+- `VARIABLE`
+
+Other record types can still be exported by setting the following configuration property:
+
+- `ZEEBE_BROKER_EXPORTERS_ELASTICSEARCH_ARGS_ZEEBERECORDSEXPORTENABLED=true` for Elasticsearch
+- `ZEEBE_BROKER_EXPORTERS_OPENSEARCH_ARGS_ZEEBERECORDSEXPORTENABLED=true` for OpenSearch
```
ChrisKujawa, line 278: TBH I feel the name is a bit sub-optimal. I think we should a bit more to it, like legacy or something.

Because otherwise I would think when setting to false it doesn't export at all 🤔 

But if it is called:

`ZEEBE_BROKER_EXPORTERS_OPENSEARCH_ARGS_ZEEBERECORDSEXPORTLEGACYENABLED` then it might be a bit more clearer. WDYT?

ChrisKujawa (commented):
Thanks @panagiotisgts I'm a bit unhappy with the name. Maybe we can improve that.

furthermore, I think we should also update the ES and OS exporter documentation to reference the new config. Or make a note there.

FYI @RomanJRW - happy for your input as well here

docs/self-managed/operational-guides/update-guide/870-to-880.md
panagiotisgts: @ChrisKujawa  In general I agree that the name is not 100% descriptive but `ZEEBERECORDSEXPORTLEGACYENABLED`  is getting a bit lengthy in my opinion :sweat_smile:. How about `EXPORTLEGACYRECORDS`  ?

ChrisKujawa: Sound great nice and crisp :)

docs/self-managed/operational-guides/update-guide/870-to-880.md
```diff
@@ -254,3 +254,25 @@ In that case, we append the following records:
...
+::
+
+The record types that continue to be exported by default are the following:
+
+- `DEPLOYMENT`
+- `PROCESS`
+- `PROCESS_INSTANCE`
+- `VARIABLE`
+- `USER_TASK`
+- `VARIABLE`
```
RomanJRW, line 273: You have `VARIABLE` listed twice. I think this should be `INCIDENT`? Optimize needs to import incidents, so it should continue to be in the default set, regardless

docs/self-managed/zeebe-deployment/exporters/elasticsearch-exporter.md
```diff
@@ -300,3 +301,19 @@ If you're using containers, you will need to mount the trust store to the contai
...
+Meaning that not all `zeebe-record` indices will be populated.
+
+The record types that continue to be exported by default are the following:
+
+- `DEPLOYMENT`
+- `PROCESS`
+- `PROCESS_INSTANCE`
+- `VARIABLE`
+- `USER_TASK`
+- `VARIABLE`
```
RomanJRW, line 317: same as above

docs/self-managed/zeebe-deployment/exporters/opensearch-exporter.md
```diff
@@ -323,3 +324,19 @@ If you're using containers, you will need to mount the trust store to the contai
...
+Meaning that not all `zeebe-record` indices will be populated.
+
+The record types that continue to be exported by default are the following:
+
+- `DEPLOYMENT`
+- `PROCESS`
+- `PROCESS_INSTANCE`
+- `VARIABLE`
+- `USER_TASK`
+- `VARIABLE`
```
RomanJRW, line 340: same as above

docs/self-managed/operational-guides/update-guide/870-to-880.md
```diff
@@ -254,3 +254,25 @@ In that case, we append the following records:
 - `ASSIGN`
 - `ASSIGNING`
 - `ASSIGNED`
+
+### Zeebe record types
+
+With the introduction of the Camunda Exporter, the Elasticsearch and OpenSearch Exporter no longer exports all record types by default.
+Meaning that not all `zeebe-record` indices will be populated.
```
RomanJRW, line 261: To be specific, does it not mean that the indices will not even be created? We should be as clear as possible here

docs/self-managed/zeebe-deployment/exporters/elasticsearch-exporter.md
```diff
@@ -53,14 +53,15 @@ exporters:
...
+| ------------------- | ---------------------------------------------------------------------------------------- | ----------------------- |
+| url                 | Valid URLs as comma-separated string                                                     | `http://localhost:9200` |
+| requestTimeoutMs    | Request timeout (in ms) for the Elasticsearch. client                                    | `30000`                 |
+| index               | Refer to [Index](#index) for the index configuration options.                            |                         |
+| bulk                | Refer to [Bulk](#bulk) for the bulk configuration options.                               |                         |
+| retention           | Refer to [Retention](#retention) for the retention configuration options                 |                         |
+| authentication      | Refer to [Authentication](#authentication) for the authentication configuration options. |                         |
+| exportLegacyRecords | If `true` all enabled record types will be exported.                                     | `false`                 |
```
RomanJRW, line 64: Hmm I kind of agree with @ChrisKujawa that the name doesn't feel right. I think there are two things I am hesitant about:
- This is already a sub-property of the exporter, so including "export" feels redundant
- Marking these as "legacy" could be premature. It is easily imaginable that Optimize might want some of these records in future, or that they might have another purpose at Camunda

How about something simpler like `includeEnabledRecords`?

RomanJRW (changes requested):
We need to fix the default list so requesting changes, plus I have made a suggestion for a new property name

docs/self-managed/operational-guides/update-guide/870-to-880.md
panagiotisgts: @RomanJRW That is true but only for 8.8 records. The `zeebe-record` indices from 8.7 would likely still be present in ES/OS. Not sure if mentioning that makes it more confusing or not to be honest for the migration steps. 

Maybe we should better include it in the ES/OS exporter configuration. WDYT?

RomanJRW (approved):
Approving now that the record type issue is resolved, but think we should agree on the remaining open threads before merging

docs/self-managed/zeebe-deployment/exporters/elasticsearch-exporter.md
panagiotisgts: I'm onboard with `includeEnabledRecords`, don't know whether @ChrisKujawa  has any objections for this.

panagiotisgts: The check regarding the enabled records is done beforehand so it would not export Optimize related records as well.

christinaausley:
As Cole was originally assigned to this PR, I am happy to take over from the docs side of things. Please let me know when you are ready for final review and if I can help answer any questions.

christinaausley (commented):
This reads like a breaking change -- can we also add this to the announcements page?

panagiotisgts:
@christinaausley @RomanJRW 
> This reads like a breaking change -- can we also add this to the announcements page?

I am not sure what our audience is for this change and what percentage of the customer base is using the ES/OS Zeebe indices for their custom business needs. If you consider it to be, I can include it but not sure about that.

RomanJRW (approved):

RomanJRW:
@christinaausley - do you have a strong opinion on this? I agree with @panagiotisgts that this is only a breaking change for customers with custom use cases for the old exporters. I don't think we normally announce changes here when changes have been made in the past. If you feel like it's better to include, happy to do so.

In the meantime, I have approved from my perspective

christinaausley:
@RomanJRW @panagiotisgts I am fine with leaving this out for now -- if we receive feedback users are missing this information or can't find it, we can always adjust.

christinaausley (approved):

PR #5569, opened 2025-04-15
docs(self-managed): add version arg to Helm install command

aabouzaid:
## Description

Related to: https://github.com/camunda/team-distribution/issues/482

As we don't set the version for the Helm install command, it will use the latest chart version, which doesn't match the Camunda version in the [official docs](https://docs.camunda.io/).

In this PR, I added the following:

- A React.js component to show the corresponding Helm chart version.
- In all places where `helm install` is used, I added `--version $HELM_CHART_VERSION` to clarify that the version should be specified (for all supported Camunda docs versions 8.3-8.8).
- For consistency, unified the Helm chart installation name to always be `camunda` as it's used in most docs.

How the Helm chart alpha (unreleased) will look like:

helm-chart-installation-alpha

How the Helm chart stable will look like:

helm-chart-installation-stable

## PR Checklist


src/components/CamundaDistributions/Helm.jsx
```diff
@@ -0,0 +1,62 @@
...
+  const camundaMinorVersion = camundaVersion.split(".")[1];
+
+  // Starting from the Camunda 8.4 release, we decoupled the Helm chart version and started from version 9.
+  // For more details check the Helm chart version matrix:
+  // https://.camunda.io/docs/reference/supported-environments/#helm-charts-version-matrix
+  if (camundaVersion >= "8.4") {
+    let chartVersion = `${Number(camundaMinorVersion) + 5}`;
+    // Helm --version does not support regex or glob patterns.
+    // So for non-stable versions, we need to set full version name like "13.0.0-alpha".
+    if (["current", "next"].includes(docsVersion.name)) {
```
pepopowitz, line 17: potentially blocking: Is this doing what we think it is doing? 

Based on this line, I would have expected to see `.0.0-alpha` appended in version 8.7, which is the `current` version of the docs. However, [I don't](https://preview.docs.camunda.cloud/pr-5569/docs/self-managed/setup/install/):

image

Am I misunderstanding the code, or is there a different value than `current` for `docsVersion.name` in version 8.7?

pepopowitz (changes requested):
A couple small things I'd like changed/answered, otherwise this new component looks great! I appreciate not having to change this manually in every version.

src/components/CamundaDistributions/Helm.jsx
aabouzaid: @pepopowitz, the code is correct syntactically, but semantically, it's probably not.
It works because our current [Docusaurus config](https://github.com/camunda/camunda-docs/blob/fda491452b0f213792caa81156121a90a780609d/docusaurus.config.js#L529) refers to `8.8` as current.

I'm not sure if we will change that in the docs and go back to the normal flow as we had in the past (so 8.7 is current and 8.7 is next, and there is a bug in the current config) or if it's expected to have the current to be 8.8 version.

If it's a bug, I will update the code.

pepopowitz:
```suggestion
    // Docusaurus refers to vNext as "current".
    if (docsVersion.name === "current") {
```

Ah, I see. I didn't recall that Docusaurus refers to vNext as "current". 

If that's the case, I'd like to see us remove the value from this check that will never be matched, `"next"`, as well as a comment to remind us why we are checking for `"current"`.

pepopowitz (dismissed):
Thanks for making the changes! 

I approve from the infra perspective; I suspect a @camunda/tech-writers probably should review too.

src/components/CamundaDistributions/Helm.jsx
aabouzaid: Fixed :heavy_check_mark:

mesellings (dismissed):
Approved after my last commit

versioned_docs/version-8.3/self-managed/platform-deployment/helm-kubernetes/platforms/redhat-openshift.md
```diff
@@ -0,0 +1,394 @@
...
+    - name: certificate
+      secret:
+        secretName: <SERVICE_CERTIFICATE_SECRET_NAME>
+        items:
+          - key: tls.crt
+            path: tls.crt
+        defaultMode: 420
+```
+
+The actual configuration properties can be reviewed [in the Operate configuration documentation](docs/self-managed/operate-deployment/operate-configuration.md#zeebe-broker-connection).
```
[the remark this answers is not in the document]
mesellings: Fixed

mesellings (approved):
Approved after commit

aabouzaid:
@mesellings That would be great to do :raised_hands: 
So let's merge it.

akeller:
Hi Folks, this docs PR added back the 8.3 docs.... which @pepopowitz removed and unsupported recently. These need to be removed again immediately. I will take care of this now.

PR #5577, opened 2025-04-16
fix(identity-sdk): add property

jonathanlukas:
## Description

This change documents this change:

https://github.com/camunda-cloud/identity/pull/3276/files#diff-d9ae0ba49b24e396544d1635d3d9e83d28a061b5733d274fe89da4c4dbea89d7

## PR Checklist


Ben-Sheppard (changes requested):
Changes look fine to me, we should also add these to the 8.8 docs too @jonathanlukas

jonathanlukas:
@Ben-Sheppard this table is not part of 8.8 as we were not sure whether the identity sdk will still be part of camunda.

Ben-Sheppard:
> @Ben-Sheppard this table is not part of 8.8 as we were not sure whether the identity sdk will still be part of camunda.
> 
> Should we add the whole table to 8.8?

The reason I say this is because management Identity will still exist in 8.8 (it will back web modeler, console, and Optimize) so this value will still be relevant for some of those instances (including management Identity itself). That being said, adding the whole table to the 8.8 docs might be more involved because I'm not sure what that structure would look like. Maybe someone from @camunda/tech-writers could offer an input? 

I'm not opposed to approving this change though and following up on that discussion to not block

jonathanlukas:
So to summarize:

The table would still live in 8.8, but in a section being related to "management-identity" rather than "core-identity" (or whatever you call that now). 

>I'm not opposed to approving this change though and following up on that discussion to not block

I would take that offer :)

jonathanlukas:
Here is the follow-up issue: https://github.com/camunda/camunda-docs/issues/5616

Ben-Sheppard (approved):

PR #5615, opened 2025-04-23
feat: Migration guide for decision instance endpoints

georgios-goulos:

## PR Checklist

### DIFF
--- a/docs/apis-tools/migration-manuals/migrate-to-camunda-api.md
+++ b/docs/apis-tools/migration-manuals/migrate-to-camunda-api.md
@@ -399,6 +399,98 @@ The following conventions apply to all attributes:
 </TabItem>
 </Tabs>
 
+#### Search decision instances
+
+- **V1 endpoint**: `POST http://localhost:8080/v1/decision-instances/search`
+- **V2 endpoint**: `POST http://localhost:8080/v2/decision-instances/search`
+
+<Tabs groupId="search-decision-instances" defaultValue="input-adjustments" queryString values={
+[
+{label: 'Input adjustments', value: 'input-adjustments'},
+{label: 'Output adjustments', value: 'output-adjustments'},
+]
+}>
+
+<TabItem value='input-adjustments'>
+
+- Request structure changes as outlined in [general changes][].
+  - `searchAfter` has been moved under `page`.
+  - `size` is now the `limit` in the `page` object.
+- Renamed attributes in the `filter` object
+  - `id` - Use `decisionInstanceId` instead.
+  - `key` of type `int64` - Use `decisionInstanceKey` of type `string`.
+  - `processDefinitionKey` of type `int64` - This is now of type `string`.
+  - `processInstanceKey` of type `int64` - This is now of type `string`.
+  - `decisionId` - Use `decisionDefinitionId` instead.
+  - `decisionName` - Use `decisionDefinitionName` instead.
+  - `decisionVersion` - Use `decisionDefinitionVersion` instead.
+  - `decisionType` - Use `decisionDefinitionType` instead.
+- Removed attributes in the `filter` object
+  - `result` - Can no longer be used for filtering.
+  - `evaluatedInputs` - Can no longer be used for filtering.
+  - `evaluatedOutputs` - Can no longer be used for filtering.
+
+</TabItem>
+
+<TabItem value='output-adjustments'>
+
+- Response structure changes as outlined in [general changes][].
+  - `total` is moved under the `page` object as `totalItems`.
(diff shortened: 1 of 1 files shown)


georgios-goulos:
Regarding `evaluationDate` in the SEARCH request body, I noticed that in V1 it is a `string` but in V2 it can be a string or something else. I did not include it here as it is my understanding that it does not need to be explicitly migrated in any way, the V1 structure should work in V2.

docs/apis-tools/migration-manuals/migrate-to-camunda-api.md
```diff
@@ -399,6 +399,98 @@ The following conventions apply to all attributes:
...
+  - `key` of type `int64` - Use `decisionInstanceKey` of type `string`.
+  - `processDefinitionKey` of type `int64` - This is now of type `string`.
+  - `processInstanceKey` of type `int64` - This is now of type `string`.
+  - `decisionId` - Use `decisionDefinitionId` instead.
+  - `decisionName` - Use `decisionDefinitionName` instead.
+  - `decisionVersion` - Use `decisionDefinitionVersion` instead.
+  - `decisionType` - Use `decisionDefinitionType` instead.
+- Removed attributes in the objects of the `items` array
+  - `evaluatedInputs` - The endpoint does not serve this information anymore.
+  - `evaluatedOutputs` - The endpoint does not serve this information anymore.
```
tmetzke, line 451: @houssain-barouni, any idea why this is not available anymore?

tmetzke: Is it just not documented but served?

tmetzke: Looking at https://github.com/camunda/camunda/blob/f4fa09a7821a73c6b32911da1b9459402124996d/zeebe/gateway-protocol/src/main/proto/rest-api.yaml#L7392, this is deliberate.

tmetzke (approved):
👍 Looks great to me, nice job 🚀 
ℹ️ Resolving the missing attributes is out of the scope of this PR, we'll create a follow-up if necessary. FYI, @aleksander-dytko, that we have missing attributes in decision instance query endpoints from V1 to V2 currently.

docs/apis-tools/migration-manuals/migrate-to-camunda-api.md
tmetzke: @houssain-barouni, do you still recall if there was any technical reason or was it just not specified in the requirements doc? If there is no technical reason, we need to create a follow-up issue for this to add them to the filters and response values.

houssain-barouni: @tmetzke Sorry I missed the first mention notification
This was definitively not supported in V1 https://github.com/camunda/camunda/blob/2c0cafdf2847e618c64ca0a730e077eb98c8073b/operate/webapp/src/main/java/io/camunda/operate/webapp/api/v1/dao/elasticsearch/ElasticsearchDecisionInstanceDao.java#L109-L144

The V1 endpoints use a `T search(Query<T> query)` pattern, which implies that the filter and output share the same structure. This can be misleading, it may appear that certain filters are supported, even though they have no effect when used. To determine what is actually supported, you must check the underlying implementation. :/

tmetzke: Oh nooooo 🙈 So, we have documented filter fields for Operate V1 that actually don't work? 🤣 This is lovely...

akeller: @tmetzke @houssain-barouni can you please create a follow up issue on this so it can be properly documented?

tmetzke: @houssain-barouni, this is true for filtering. However, we also don't have it in the V2 API results. This is a proper regression, isn't it?

houssain-barouni: Hi @tmetzke , again this is iso-functional with v1
https://github.com/camunda/camunda/blob/d8ee3a18816ae6d65f974d67727747ac0a3f7f66/operate/webapp/src/main/java/io/camunda/operate/webapp/api/v1/dao/elasticsearch/ElasticsearchDecisionInstanceDao.java#L81-L83

`EVALUATED_INPUTS` and `EVALUATED_OUTPUTS` fields are excluded from the fetched documents (right argument is for excluded fields). This is to avoid having a large response payload, I suppose

tmetzke: I see, wow, this is really annoying 😆

PR #5707, opened 2025-05-02
feat: Migration guide for variable endpoints

georgios-goulos:
## Description

Closes https://github.com/camunda/camunda/issues/23748

## PR Checklist

### DIFF
--- a/docs/apis-tools/migration-manuals/migrate-to-camunda-api.md
+++ b/docs/apis-tools/migration-manuals/migrate-to-camunda-api.md
@@ -325,7 +325,7 @@ The following conventions apply to all attributes:
 </TabItem>
 </Tabs>
 
-### Variable
+### Variables
 
 #### Get a variable
 
@@ -349,7 +349,6 @@ The following conventions apply to all attributes:
 
 - Renamed attributes
   - `id` - Use `variableKey` as this refers to the unique system identifier of the variable.
-  - `value` - Use `fullValue` instead.
 - Removed attributes
   - `draft` - Draft variables are not supported in V2 anymore, see also the [save draft variables](#save-task-draft-variables) endpoint for further details.
 
--- a/docs/apis-tools/migration-manuals/migrate-to-camunda-api.md
+++ b/docs/apis-tools/migration-manuals/migrate-to-camunda-api.md
@@ -670,7 +670,9 @@ The following conventions apply to all attributes:
 
 <TabItem value='output-adjustments'>
 
-- Except for the response structure changes, all adjustments from [search variables for process instances](#search-variables-for-process-instances) apply.
+- All adjustments from [search variables for process instances](#search-variables-for-process-instances) apply, with the following exceptions:
+  - Response structure changes.
+  - `truncated` is removed because this endpoint always returns the full variable value.
 
 </TabItem>
 </Tabs>
--- a/docs/apis-tools/migration-manuals/migrate-to-camunda-api.md
+++ b/docs/apis-tools/migration-manuals/migrate-to-camunda-api.md
@@ -139,9 +139,8 @@ The following conventions apply to all attributes:
 - Response structure changes as outlined in [general changes][].
 - Renamed attributes
   - `id` - Use `variableKey` as this refers to the unique system identifier of the variable.
-  - `value` - Use `fullValue` as this represents the full variable value in case the `value` is only a preview due to size constraints. If the `value` is not a preview, the `fullValue` is empty.
-  - `previewValue` - Use `value` as this always represents the variable value. This can be a preview value due to size constraints. In that case, the `fullValue` contains the full variable value.
-  - `isValueTruncated` - Use `isTruncated` as a replacement
+  - `previewValue` - Use `value` as this always represents the variable value. This can be a truncated value due to size constraints.
+  - `isValueTruncated` - Use `isTruncated` as a replacement. If the value of `isTruncated` is `true` and you need the full value, please see the [get a variable](#get-a-variable) endpoint.
 - Removed attributes
   - `draft` - Draft variables are not supported in V2 anymore, see also the [save draft variables](#save-task-draft-variables) endpoint for further details.
 - Added attributes
@@ -647,7 +646,6 @@ The following conventions apply to all attributes:
   - `processInstanceKey` of type `int64` - This is now of type `string`.
   - `scopeKey` of type `int64` - This is now of type `string`.
   - `truncated` - Use `isTruncated` instead.
-  - `value` - If `isTruncated` is `true`, use `value` for the truncated value or `fullValue` for the full value. If `isTruncated` is `false` use `value` for the full value.
 
 </TabItem>
 </Tabs>
--- a/docs/apis-tools/migration-manuals/migrate-to-camunda-api.md
+++ b/docs/apis-tools/migration-manuals/migrate-to-camunda-api.md
@@ -145,7 +145,7 @@ The following conventions apply to all attributes:
   - `draft` - Draft variables are not supported in V2 anymore, see also the [save draft variables](#save-task-draft-variables) endpoint for further details.
 - Added attributes
   - `scopeKey` - Variables belong to a specific scope, for example, the process instance or the element instance of a user task. This value represents the scope the variables is related to.
-  - `processInstanceKey` - A variable belongs to process instance and this value represents the unique system identifier of that instance.
+  - `processInstanceKey` - A variable belongs to a process instance and this value represents the unique system identifier of that instance.
   - `tenantId` - Variables can belong to a dedicated tenant and this value represents the one it belongs to. See [multi-tenancy][] for further details.
 
 </TabItem>
@@ -350,6 +350,9 @@ The following conventions apply to all attributes:
   - `id` - Use `variableKey` as this refers to the unique system identifier of the variable.
 - Removed attributes
   - `draft` - Draft variables are not supported in V2 anymore, see also the [save draft variables](#save-task-draft-variables) endpoint for further details.
+- Added attributes
+  - `scopeKey` - Variables belong to a specific scope, for example, the process instance or the element instance of a user task. This value represents the scope the variables is related to.
+  - `processInstanceKey` - A variable belongs to a process instance and this value represents the unique system identifier of that instance.
 
 </TabItem>
 </Tabs>
--- a/docs/apis-tools/migration-manuals/migrate-to-camunda-api.md
+++ b/docs/apis-tools/migration-manuals/migrate-to-camunda-api.md
@@ -325,6 +325,37 @@ The following conventions apply to all attributes:
 </TabItem>
 </Tabs>
 
+### Variable
+
+#### Get a variable
+
+- **V1 endpoint**: `GET http://localhost:8080/v1/variables/{variableId}`
+- **V2 endpoint**: `GET http://localhost:8080/v2/variables/{variableKey}`
+
+<Tabs groupId="get-a-variable" defaultValue="input-adjustments" queryString values={
+[
+{label: 'Input adjustments', value: 'input-adjustments'},
+{label: 'Output adjustments', value: 'output-adjustments'},
+]
+}>
+
+<TabItem value='input-adjustments'>
+
+- `variableId` - Use `variableKey` as this refers to the unique system identifier of the variable.
+
+</TabItem>
+
+<TabItem value='output-adjustments'>
+
+- Renamed attributes
+  - `id` - Use `variableKey` as this refers to the unique system identifier of the variable.
+  - `value` - Use `fullValue` instead.
+- Removed attributes
+  - `draft` - Draft variables are not supported in V2 anymore, see also the [save draft variables](#save-task-draft-variables) endpoint for further details.
+
+</TabItem>
+</Tabs>
+
 ## Operate API
 
 ### Decision definition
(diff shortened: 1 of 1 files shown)


akeller (dismissed):

docs/apis-tools/migration-manuals/migrate-to-camunda-api.md
```diff
@@ -580,6 +611,73 @@ The following conventions apply to all attributes:
...
+
+- Response structure changes as outlined in [general changes][].
+  - `total` is moved under the `page` object as `totalItems`.
+  - `sortValues` - Use `lastSortValues` in the `page` object instead.
+- Renamed attributes in the objects of the `items` array
+  - `key` of type `int64` - Use `variableKey` of type `string`.
+  - `processInstanceKey` of type `int64` - This is now of type `string`.
+  - `scopeKey` of type `int64` - This is now of type `string`.
+  - `truncated` - Use `isTruncated` instead.
+  - `value` - If `isTruncated` is `true`, use `value` for the truncated value or `fullValue` for the full value. If `isTruncated` is `false` use `value` for the full value.
```
tmetzke, line 651: 💭 We need to understand what the Operate V1 API returns. I believe it only returns the truncated value, which means nothing changes for users. They can still use `value` in V2.

@houssain-barouni, do you know what the Operate V1 API for variable search returns here? The full or truncated value?

docs/apis-tools/migration-manuals/migrate-to-camunda-api.md
```diff
@@ -580,6 +611,73 @@ The following conventions apply to all attributes:
...
+
+<TabItem value='input-adjustments'>
+
+- No input adjustments.
+
+</TabItem>
+
+<TabItem value='output-adjustments'>
+
+- Except for the response structure changes, all adjustments from [search variables for process instances](#search-variables-for-process-instances) apply.
```
tmetzke, line 676: 💭 We need to understand what the Operate V1 API returns. It might return the full value, which makes it more complex for users to switch to V2 until xxx is implemented. If this is the case, we need to use the description from the search endpoint above (but we might remove it from there).

@houssain-barouni, do you know what the Operate V1 API for variable get by key returns here? The full or truncated value? If it's the truncated values as well, how do I ever get the full value from the Operate V1 API? 😅

docs/apis-tools/migration-manuals/migrate-to-camunda-api.md
```diff
@@ -325,6 +325,37 @@ The following conventions apply to all attributes:
...
+
+- `variableId` - Use `variableKey` as this refers to the unique system identifier of the variable.
+
+</TabItem>
+
+<TabItem value='output-adjustments'>
+
+- Renamed attributes
+  - `id` - Use `variableKey` as this refers to the unique system identifier of the variable.
+  - `value` - Use `fullValue` instead.
```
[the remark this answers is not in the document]
tmetzke: FYI, I have tackled the issue linked above yesterday. I believe we can document here as if this is merged already. That implies that we can simply remove this line. It will work as before in V1.

docs/apis-tools/migration-manuals/migrate-to-camunda-api.md
```diff
@@ -325,6 +325,37 @@ The following conventions apply to all attributes:
 </TabItem>
 </Tabs>
 
+### Variable
```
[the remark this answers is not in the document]
georgios-goulos: Apologies, I had the fix locally and I did not push it 🤦

docs/apis-tools/migration-manuals/migrate-to-camunda-api.md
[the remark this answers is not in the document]
georgios-goulos: Thank you for addressing this so quickly

docs/apis-tools/migration-manuals/migrate-to-camunda-api.md
georgios-goulos: @houssain-barouni if you don't know off the top of your head I can try digging through the code.

houssain-barouni: @georgios-goulos yes, I would need to check the code, please have a look and let me know if something is not clear

georgios-goulos: This is the path I found:
* [VariableController calls dao.search(...)](https://github.com/camunda/camunda/blob/main/operate/webapp/src/main/java/io/camunda/operate/webapp/api/v1/rest/VariableController.java#L121) 
* [the dao maps the results without full value ](https://github.com/camunda/camunda/blob/main/operate/webapp/src/main/java/io/camunda/operate/webapp/api/v1/dao/elasticsearch/ElasticsearchVariableDao.java#L96)
* This eventually calls [searchHitToVariable](https://github.com/camunda/camunda/blob/main/operate/webapp/src/main/java/io/camunda/operate/webapp/api/v1/dao/elasticsearch/ElasticsearchVariableDao.java#L126) which does
```
            .setValue((String) searchHitAsMap.get(Variable.VALUE))
            .setTruncated((Boolean) searchHitAsMap.get(Variable.TRUNCATED));
```
So it looks like it's the truncated value for the `search` operation

docs/apis-tools/migration-manuals/migrate-to-camunda-api.md
georgios-goulos: I will make the necessary changes then to distinguish this from the `search` operation. Thanks for looking into this.

docs/apis-tools/migration-manuals/migrate-to-camunda-api.md
tmetzke: OK, makes sense 👍 This is the same now in V2 with my fix PR 🙂

docs/apis-tools/migration-manuals/migrate-to-camunda-api.md
tmetzke: OK, so the new V2 behavior is the same then, compared to V1.

tmetzke:
@georgios-goulos, if you'd like to compare docs pages again for this to be certain, you can use [this PR](https://github.com/camunda/camunda-docs/pull/5732)'s deployed preview, if you like. It contains the variable API changes.

tmetzke:
@georgios-goulos, would you mind having a look at the `Search task variables` section and adjusting this to the new state of the variable API? That would be super helpful to align all variable-related sections 🙏

docs/apis-tools/migration-manuals/migrate-to-camunda-api.md
```diff
@@ -325,6 +324,36 @@ The following conventions apply to all attributes:
...
+]
+}>
+
+<TabItem value='input-adjustments'>
+
+- `variableId` - Use `variableKey` as this refers to the unique system identifier of the variable.
+
+</TabItem>
+
+<TabItem value='output-adjustments'>
```
tmetzke, line 347: 🔧 For the `Search task variables` endpoint, we also document added attributes so users have a full picture of changes. What do you think, should we also add those here? If so, you can copy it from that section, I guess.

tmetzke (approved):
👍 This looks great, thanks for adjusting to the new V2 already and considering the  suggestions here 🙏 
🔧 I left one final suggestion that I just realized about added attributes. I think the previous endpoint PRs didn't have any but maybe we better double-check 😅

docs/apis-tools/migration-manuals/migrate-to-camunda-api.md
georgios-goulos: I am happy to do so, but I thought the migration guide was more about the essential changes users had to make to migrate to the V2 endpoints. The API documentation is there for them to see what else is included. Since we already have these, though, for currently documented endpoints I will add them for the endpoints I documented.

tmetzke: Yeah, that is generally true, added attributes might not be relevant for migration.

However, that also depends a bit on how users consume that API, especially API responses. New search filter properties might not be directly relevant but response attributes might be mapped in a strict way, so users would end up with responses containing new attributes they were not aware of.

They can derive this themselves from the V2 docs, of course. However, coming to the API migration guide to get a full picture of what has changed in the endpoints they are using might be more convenient.

But, ultimately, I also simply don't know 🙂

georgios-goulos: I added the `added attributes` section, but I noticed e.g. for `Search task variables` we do not explicitly specify what is in the `page` object. I know that we refer briefly to it at the top of the page, but in other similar scenarios (e.g. `id` and `key` having the entity as a prefix) we have chosen to state them explicitly for each documented endpoint.

I can also raise a new PR to address the missing `added attributes` section from my previous PRs.

tmetzke: Oh, yeah, fair point 👍
For those general recurring changes, though, I'd prefer to write them once in the general section (but maybe more explicitly than we do now) and refer to them from the single endpoints. What do you think?

georgios-goulos: I agree that writing them once and then simply referring to them is in general a very good approach, which is why I also think that when if we want to have an `added attributes` section we should probably just refer the reader to the API pages, if that is possible.

tmetzke: I see. Trying to look at it from an API user perspective, I would appreciate a self-contained migration guide:
* For endpoint-specific attributes, I would expect the added attributes compared to V1 to be listed directly instead of having a link to the V2 page, for example. Otherwise, I have to go to that V2 page and identify them myself, differentiating them from the ones that are different from those in V1 but simply have been renamed/replaced. That is quite cumbersome, I think. It's definitely possible, it's just not a very nice UX, I think.
* For general changes (e.g., new attributes in `page` in request and response), we can be explicit once in the general changes section of the migration guide and refer to that section from the single endpoints if the changes apply.

This just sounds like a better user experience to me, coming from V1. I don't want to jump between pages if it is avoidable. And the effort on our end is manageable, I guess. What do you think?

@camunda/tech-writers, I'm very interested in your perspective here. You have way more docs UX experience 🙂

georgios-goulos: I should clarify that I may be (selfishly) seeing this purely from an engineer's perspective, so I am trying to apply principles such as DRY to documentation. In my mind any sort of duplicated information increases the cost of maintenance and the risk of inconsistency in the future (e.g. we change the API guide in the future but we forget to update this migration guide).

tmetzke: If I get that right, your point is the following:
We might add more attributes to the V2 endpoints after 8.8 and would have to add them to this guide every time. Is that correct?

If so, I agree. That is quite a pain point I haven't considered much. I see the following possible mitigations:
* We only keep this guide for 8.8. We only need to get it right once then. After the 8.8.0 release, no new attributes should arrive in the endpoints. From 8.9 on, we could consider having an API diff (ideally automated).
* The migration guide does not feature purely added attributes but only focuses on those existing in V1. There is still a risk of an attribute changing again in 8.9+, so this guide would get out of sync again. That risk is substantially smaller, though.

I would also tend not to feature added properties here.

Thanks for bringing that up, @georgios-goulos 👍

georgios-goulos: Let's see what @camunda/tech-writers think of this challenge 😄

akeller: In documentation we can repeat information, but we should be strategic and purposeful about it. We can create a reusable component and reference the same info across multiple pages. This works well and allows us to only update the content in one place, but when the original content is the API reference, we can't use the established pattern. 

What I come back to is the purpose of the guide, which is migrating users to the C8 API. Are the added attributes renamed attributes? Or totally new ones? If they are totally new and we don't need to map them to existing attributes, **I could see adding a crisp sentence and a link to the endpoint in the API reference.** This gets users working with the API reference vs. continuing to stay in the migration guide for new things. They are aware of the changes within the migration guide, but because no action is required to map knowledge between V1 and V2, they can switch their working mode to working directly with the API reference.

If they are renamed, listing them out with their V1 attribute is the way to go.

tmetzke: Thanks a lot, @akeller 🙏 

It's this part:
> If they are totally new and we don't need to map them to existing attributes, I could see adding a crisp sentence and a link to the endpoint in the API reference. This gets users working with the API reference vs. continuing to stay in the migration guide for new things. They are aware of the changes within the migration guide, but because no action is required to map knowledge between V1 and V2, they can switch their working mode to working directly with the API reference.

So, should we go with that, @georgios-goulos? It's basically what you proposed as well, right?

georgios-goulos: Yeah, sounds good to me!
@tmetzke I can raise a PR for this. Should I also capture it in a GH issue?

tmetzke: PR is fine in this case 👍

georgios-goulos: 👋 I created a draft PR for this [here](https://github.com/camunda/camunda-docs/pull/5746). I did not update all the endpoints with added variables as I wanted to get some feedback first.

@tmetzke @akeller feel free to take a look whenever you have time.

PR #5758, opened 2025-05-12
Revert "docs(self-managed): add Opensearch support to elasticsearch prefix indices page"

drodriguez-305:
Reverts camunda/camunda-docs#5658

We will be reverting the docs changes until next release in which we will implement this feature.


aabouzaid (approved):

akeller:
Even though this is a revert of previous changes, you'll need to add redirects because you are removing pages previously indexed by crawlers (like Google). I think removing the ones you added previously makes sense, but you'll need to replace them for the pages you deleted in the context of this PR.

drodriguez-305:
fixed the redirect URLs should be all set now @aabouzaid 

thank you for bringing this up @akeller

hisImminence (approved):

PR #5777, opened 2025-05-14
fix: copy-paste issue in tenant api section

sergiizakharov:
## Description

copy-paste issue in tenant api section, incorrect reference in the sidebar mentioning user instead of the application.

## PR Checklist

### DIFF
--- a/docs/apis-tools/camunda-api-rest/specifications/assign-application-to-tenant.api.mdx
+++ b/docs/apis-tools/camunda-api-rest/specifications/assign-application-to-tenant.api.mdx
@@ -1,8 +1,8 @@
 ---
 id: assign-application-to-tenant
-title: "Assign a user to a tenant"
-description: "Assign a single user to a specified tenant."
-sidebar_label: "Assign a user to a tenant"
+title: "Assign an application to a tenant"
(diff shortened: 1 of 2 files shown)


akeller (commented):
Was this already fixed in the camunda/camunda repo? I would expect changes like this to come in via a PR that looks like https://github.com/camunda/camunda-docs/pull/5759. (via automation). 

FYI @tmetzke

sergiizakharov:
> Was this already fixed in the camunda/camunda repo? I would expect changes like this to come in via a PR that looks like #5759. (via automation).
> 
> FYI @tmetzke

I was not aware about that workflow. I just saw the problem and wanted to fix it. I can close this PR then.

tmetzke:
Hey @sergiizakharov, thanks for taking action here! Those mdx files are all generated, so, as @akeller mentioned, we should not fix this here. 

We need to fix this in the source of truth, which is the [OpenAPI spec](https://github.com/camunda/camunda/blob/main/zeebe/gateway-protocol/src/main/proto/rest-api.yaml) itself. That will be synced periodically to this docs repository automatically.

Would you mind checking that spec for the things you corrected here? If this is also wrong in there, we need to fix it there.

PR #5779, opened 2025-05-14
refactor(self-managed): use camunda download center for c8run artifact

aabouzaid:
## Description

The Self-Managed team unifies the artifacts URL, so now we use the Camunda Download Center instead of redirecting the users to GitHub repos.

## PR Checklist

### DIFF
--- a/src/components/CamundaDistributions/C8Run.jsx
+++ b/src/components/CamundaDistributions/C8Run.jsx
@@ -1,12 +1,13 @@
 import React from "react";
 import { useActiveVersion } from "@docusaurus/plugin-content-docs/client";
 
-const C8RunBaseURL = "https://github.com/camunda/camunda/releases/tag";
+const C8RunBaseURL = "https://downloads.camunda.cloud/release/camunda/c8run";
 
 const getVersion = () => {
   const docsVersion = useActiveVersion();
   if (docsVersion.name == "current") {
-    return "8.8";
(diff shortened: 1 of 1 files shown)


christinaausley (approved):

akeller:
@aabouzaid, I have a couple of questions/comments about this PR.

* This only points to minor versions, no alphas or patches
* If I look in the Download center list, I see what looks like two versions for the same minor. For example, 8.7 and 8.7.0

Is this the intended or expected behavior? No links to alphas or patches?

aabouzaid:
@akeller 

> This only points to minor versions, no alphas or patches

Yes, exact same behaviour, we use the minor version artifact as `latest`, which is updated during the release.

So this artifact:

https://downloads.camunda.cloud/release/camunda/c8run/8.7/camunda8-run-8.7-linux-x86_64.tar.gz

Was updated on 8.7.2 release:

```
May 13, 2025, 11:00:24 AM
````

> If I look in the Download center list, I see what looks like two versions for the same minor. For example, 8.7 and 8.7.0

As explained above, the minor without a patch is treated like `latest` for the minor version.
We still upload the patch version.

The idea here is to link to the latest version, so no need to update the link with every release.
That's a common pattern used everywhere not something we invented in Camunda.

akeller:
@aabouzaid thanks for the context! I've been asked whether these links point to the latest patch, so it's great to see this is handled on your side!

hisImminence (approved):

PR #5953, opened 2025-06-10
fix: rename `camunda.version` to `camunda.dockerImageVersion`

Christian-Broomfield:
## Description

As part of the changes in renaming the property `camunda.version` in favor of `camunda.dockerImageVersion` (see here: https://github.com/camunda/camunda/pull/33515), the documentation also needed to be updated.


saig0 (approved):
:+1:

Issue #4519, opened 2024-10-25
Document that all backup procedures can share the same bucket


jonathanlukas:
As said in the linked thread, all backups can share the same bucket (S3 in that case).

This should be confirmed in detail and then documented:

* S3: can share the same bucket
* Azure store: ?
* GCS: ?

The linked thread: https://camunda.slack.com/archives/C02AWA0RF8A/p1729762762489699

Langleu:
I've handled this as part of the backup & restore rewrite in PR: https://github.com/camunda/camunda-docs/pull/5963
See commit: https://github.com/camunda/camunda-docs/pull/5963/commits/f52b87e09478005675851c3e13a7c8b0b5949fab

All of the applications are capable of using `basePath` or `base_path` for each of the supported backup storages.

PR #5963, opened 2025-06-12
docs: rework of the backup & restore documentation

Langleu:
## Description

~I'll ask for a technical writer review after the first rounds of reviews have been covered to allow for the finishing touches.~
Asked for the tech writer review as the round of reviews has completed and so far not much more movement.

related to https://github.com/camunda/product-hub/issues/2299 and https://github.com/camunda/team-infrastructure-experience/issues/686.

For the reviewers, I've selected you as you had previously interaction and good valued knowledge of this particular documentation and your review would be highly appreciated!

The following is the rework of the backup & restore documentation. It comprises of:

- Sub-page strips to what they really are and move related content to where it belongs
    - e.g. move the configuration where the configuration is of each component
    - moved backup and restore parts to a single page to contain all
    - result: sug-pages are stripped down to what they are - Management API documentation.
        - Personal note: I'd see it more in the API section but outside of this scope, as we currently don't have any central management API documentation. It's partly scattered across the docs.
 - Backup & restore page
     - focus on explaining the motive on why one needs to use the backup & restore feature of Camunda
         - it looks like a lot but generally the steps haven't changed. It's about explaining the procedure and backgrounds. Automation for example would be outside of this particular scope and may be customer environment dependent. 
     - provide extended example outputs and calls
     - general restructure to focus for both backup and restore on the essential high-level steps comprised of sub steps
     - add examples around Kubernetes and the Camunda 8 Helm chart
     - reworked the manual part to be up2date

What not to expect:
- This is not a rework of the procedure itself from a technical side or automation of said procedure. It's solely about creating clarity around backup and restore, to help customers be able to understand the reasoning and execute it themselves.

I've integrated the changes for 8.7 and would after a successful review and rework based on the comments, backport it to all currently supported versions <= 8.7.

For the reviewers, one point I was considering was to integrate the `wait for backup to be completed step` to be part of the create a backup itself as sort of `verify` thing to  reduce the "overall amount of steps". In the end there would just be one general point saying `wait for all webapp backups to be completed based on the verify part of each prior step`. Let me know what you think 🙇.

▶️  Deployed Environment can be found here:
https://preview.docs.camunda.cloud/pr-5963/docs/self-managed/operational-guides/backup-restore/backup-and-restore/

After a successful review from all sides:
- [ ] backport to 8.5
- [ ] backport to 8.6

## PR Checklist

closes https://github.com/camunda/documentation-team/issues/379 https://github.com/camunda/camunda-docs/issues/4519


versioned_docs/version-8.7/self-managed/operational-guides/backup-restore/operate-tasklist-backup.md
```diff
@@ -22,76 +19,14 @@ camunda_operate_123_8.1.0_part_6_of_6
 
 Operate provides an API to perform a backup and manage backups (list, check state, delete). Restore a backup using the standard Elasticsearch API.
 
+<!-- TODO: This note is shit and factual wrong; application HTTP port is 80 or 8080?! -->
```
jonathanlukas, line 22: This note actually makes sense. Port 9600 is used as management port and therefore, the backup api should also be on this port.

versioned_docs/version-8.7/self-managed/operate-deployment/operate-configuration.md
```diff
@@ -478,3 +478,20 @@ camunda.operate:
...
+
+Operate is configured with the snapshot repository name to trigger database snapshots. This is important for coherent backups.
+
+Read more about the procedure and the need to trigger it through Camunda components in the [backup guide](/versioned_docs/version-8.7/self-managed/operational-guides/backup-restore/backup-and-restore.md).
+
+Operate must be configured with the repository name:
+
+| Name                                  | Description                      | Default value |
+| ------------------------------------- | -------------------------------- | ------------- |
+| camunda.operate.backup.repositoryName | ES / OS snapshot repository name | -             |
```
jonathanlukas, line 497: Is this property already described in the Operate configuration? In that case, a reference would make more sense here?

versioned_docs/version-8.7/self-managed/operational-guides/backup-restore/operate-tasklist-backup.md
Langleu: My problem was around this one `The backup API can be reached via the Actuator management port, which by default is the same as application HTTP port`.
Actuator port: 9600
Application HTTP port: 8080 / 80

Langleu: Adjusted it as for me it's misleading to what reality is.

versioned_docs/version-8.7/self-managed/operational-guides/backup-restore/backup-and-restore.md
```diff
@@ -5,73 +5,1287 @@ sidebar_label: "Backup and restore"
...
+- [Operate](/self-managed/operational-guides/backup-restore/operate-tasklist-backup.md#get-backups-list-api)
+- [Optimize](/self-managed/operational-guides/backup-restore/optimize-backup.md#get-backup-info-api)
+- [Tasklist](/self-managed/operational-guides/backup-restore/operate-tasklist-backup.md#get-backups-list-api)
+- [Zeebe](/self-managed/operational-guides/backup-restore/zeebe-backup-and-restore.md#list-backups-api)
+
+This may not be possible in a lot of cases, especially if doing disaster recovery.
+
+#### Available Backups on Elasticsearch / OpenSearch
+
+In that case, follow the described steps above and when you have your Elasticsearch / OpenSearch available, use the snapshot API to list available snapshots and correlate that to available snapshots in your backup bucket (AWS S3, Azure Store, Google GCS). It's important to have the same ID for all backups.
```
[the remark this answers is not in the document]
Langleu: That's what I mention in the beginning:
image

The differentiation is between `you have a Camunda environment` and you're restoring from `0` (disaster recovery).

I haven't looked at the implementation of the backup APIs but assuming they keep their own records of backups then it will bring you nothing as you start from 0. If they just use the snapshot API anyway to figure out backups, then this way is still better.

versioned_docs/version-8.7/self-managed/operational-guides/backup-restore/backup-and-restore.md
```diff
@@ -5,73 +5,1287 @@ sidebar_label: "Backup and restore"
...
+
+<Tabs>
+   <TabItem value="kubernetes" label="Kubernetes" default>
+
+Assuming you're using the official [Camunda Helm chart](/self-managed/setup/install.md), you'll have to adjust your Helm `values.yml` to supply the following temporarily.
+
+It will overwrite the start command of the resulting Zeebe pod, executing a restore script.
+It's important that the backup is configured for Zeebe to be able to restore from the backup!
+
+The following example is possible starting from the Camunda Helm chart version `12.1.0`.
```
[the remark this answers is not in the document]
Langleu: I'll have to double check with the workshop. I also wasn't a big fan of this approach since StatefulSets are not meant to be used this way.
If this is something we can implement in the Helm chart, great but not sure we can backport it for everything, so we would still need some intermediate solution.

hisImminence:
Good work @Langleu!

As a suggestion, how do you think about a diagram with the high level process in the beginning to give users a rough idea/ structure from the start?

image
(as a suggestion - feel free to adjust on the [miro board](https://miro.com/app/board/uXjVL-6SrPc=/))

versioned_docs/version-8.7/self-managed/operational-guides/backup-restore/backup-and-restore.md
```diff
@@ -5,73 +5,1287 @@ sidebar_label: "Backup and restore"
...
+
+Alternatively as a one-line to wait until the state is `COMPLETED` using a while loop and jq to parse the response JSON.
+
+```bash
+while [[ "$(curl -s "$TASKLIST_MANAGEMENT_API/actuator/backups/$BACKUP_ID" | jq -r .state)" != "COMPLETED" ]]; do echo "Waiting..."; sleep 5; done; echo "Finished backup with ID $BACKUP_ID"
+```
+
+### Backup of the Zeebe Cluster
+
+#### 1. Soft pause exporting in Zeebe. Using the [management API](/self-managed/zeebe-deployment/operations/management-api.md)
```
hisImminence, line 549: I know from SaaS that many customer were worried that a backup would stop processing. Should we mention that soft pause will not stop the processing (or link to the glossary what soft pause mean/ do)?

Langleu: Glossary is 1:1 the same explanation as on the linked page.

Thanks, I adjusted the fragment identifier to link directly and pre-select the right tab.
I also repeated the same thing as quick one-liner again as outline on the linked page.

versioned_docs/version-8.7/self-managed/operational-guides/backup-restore/backup-and-restore.md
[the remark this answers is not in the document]
Langleu: As mentioned, it's suggested.
Checking the Workshop, it's also just using the Elasticsearch API instead of the Camunda APIs, which makes sense considering that's what's available.

On Elasticsearch / OpenSearch duplication. While yes, for now kept it for both explicit with links to their documentation. If the general consensus is that we should just keep one of the day, happy to remove it.

versioned_docs/version-8.7/self-managed/operational-guides/backup-restore/backup-and-restore.md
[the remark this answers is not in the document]
Langleu: Had a look through that workshop specific part. First off, cool ideas and approaches trying to fix UX flaws in the process.

I mean having it temporarily persisted does not destroy anything. The restore doesn't overwrite existing data but it's noticeable that it wasn't build around container orchestration.
You can say the same about the job approach on persisting. It's always a two step scenario, deploy zeebe, scale down and then do the job, or values overwrite. If a customer is using GitOps, they would still need to temporarily persist it, roll it out, etc. pp. and the jobs approach would clash with them as well.

As described in the docs the temporary overwrite of `values.yml` would result in a crashloop, so a customer will need to actively remove them otherwise they won't be able to use Camunda 8 as whole anyway.

UX wise, fully agree with you. I think UX wise we have to revamp the whole thing from application and helm chart side. Me introducing now the job topic, would also need to be properly implemented on the Helm chart and tested etc., which wasn't the scope of the ticket. It will be part of my recommendation afterwards, as overall the UX is a bunch of duck tape trying to hide fundamental flaws in the components.

Overall, I'd like in the future if we didn't have to try to fix everything post-mortem with manually described approaches but have a better integrated tested UX from our components.

Langleu:
@ManuelDittmar 
 about

> Another thing that is not covered in this guide: Cleaning up backups

I added a small part about it, but as far as I know, we can't use SLMs because they would also create backups, so we're limited to letting people know they should use our APIs for it regularly + the zeebe extra stuff.
Not sure whether you had something else in mind.

Besides that, thanks again for your review! I tried to assess your comments and adjust the text as much as possible within the scope. Let me know if we should discuss them further to explore acceptable options.

Langleu:
@hisImminence let me know of what you think of this approach for the quick overview:

image

I've decided against a static picture as I'd like to be able to link directly to the sources and not have the "control" of it outside of the docs itself as we are not the maintainer of the page.

We can also reduce the text to just the high level overview. I've reused the card / grid system of Zeebe and added a markdown render. It would look a lot cleaner in comparison.

image

The cards itself link to the chapter, so when you click on it, you end up on the high level `backup / restore process` sort of thing.

We could if wanted also add the direct links to the sub-chapters but refrained from it for now.
image
 
Looking forwards to your feedback!

versioned_docs/version-8.7/self-managed/operational-guides/backup-restore/backup-and-restore.md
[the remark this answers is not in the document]
Langleu: reworked and moved this as part of Deepthi's comment on this.

deepthidevaki:
> We should also make sure that if you backed up with 3 brokers, then scaled to 6, then restore - also only restore to 3 brokers.

This is ok. You can restore to 6 brokers. But the partition count should not change. This was documented somewhere.

mesellings:
@Langleu I've given the API docs a very very brief once-over, but I will need to leave this work now to concentrate on the alpha 6 work and other projects.

- Generally, I think the guide is now in a usable format after the restructure, and would be good enough to publish if need be.
- The Restore Zeebe section needs looking at as it is a mass of notes etc.
- The API reference could do with another TW review, but is probably ok to publish.

Long story short, I think it is ok to publish if you want to get it out for alpha 6, and follow up with another TW review afterwards, otherwise we'll have to look at this again after the release?

Langleu:
@npepinpe can you elaborate on the following?

> That means, all nodes in the cluster have to be stopped (including Operate, Tasklist, and gateways), then you restore Zeebe, and only then can you start Operate/Tasklist/gateways.

What's the reasoning behind Operate / Tasklist / Optimize / Gateway to be stopped during a restore of Zeebe?
If it's a must, we should mention it accordingly

npepinpe:
> What's the reasoning behind Operate / Tasklist / Optimize / Gateway to be stopped during a restore of Zeebe?
> If it's a must, we should mention it accordingly


They are also part of the cluster now, and receive the cluster topology via gossip. This means they are storing the routing table (e.g. which requests goes to which partition leader on which node), and on restore that may now be wrong.

Since we use a very a simple CRDT with a Last-Write-Wins strategy, they would have the highest topology version after a restore, and would disseminate their own invalid topology to the rest of the cluster, potentially breaking routing across the cluster.

Langleu:
@npepinpe, added the info and removed a previous note about having components enabled in https://github.com/camunda/camunda-docs/pull/5963/commits/3ac2abf2dec02fead3cd103c1a9b42d05a02136b .

Langleu:
For 8.5/8.6 no obvious changes needed.
8.8 is out of the scope of the ticket but can still have a look.

Would need an approved review to be able to merge.

PR #6156, opened 2025-07-14
Added source link for Spring SDK

berndruecker:
I added the link to GitHub as I got the same question two or three times now. People sometimes don't easily find the sources and wonder if they are still source available. So I wondered if we simply should add links to the Github sources for components where relevant - to inform people, but also to make sure the source availability gets across as a core value that is important to us. WDYT @camunda/tech-writers


ChrisSchoe:
One reason these questions tend to pop up might be that the pom.xmls in the maven resources ( e.g. https://repo1.maven.org/maven2/io/camunda/spring-boot-starter-camunda-sdk/8.7.7/spring-boot-starter-camunda-sdk-8.7.7.pom ) link to non-public repos ( in this case https://github.com/camunda/camunda/zeebe-parent/spring-boot-starter-camunda-sdk). Changing this, too, might also reduce the amount of questions :)

berndruecker:
@ChrisSchoe If you see specific improvements for the pom.xml - ideally open a PR in https://github.com/camunda/camunda/blob/main/clients/spring-boot-starter-camunda-sdk/pom.xml - but looks like the outdated URL is already removed there for example

PR #6310, opened 2025-07-30
docs(testing): updated docs regarding runtime configuration in a pure java environment

Christian-Broomfield:
## Description

Updated docs regarding runtime configuration in a pure java environment


saig0 (approved):
:+1:

PR #6414, opened 2025-08-12
Phase 1 & 2: Orchestration cluster core settings and features

christinaausley:
## Description

Phases 1 and 2 of https://github.com/camunda/camunda-docs/issues/6314.

# Unifying configuration and conceptual documentation in Orchestration Cluster

This project aims to consolidate outdated or fragmented configuration and conceptual documentation across Operate, Tasklist, and Identity into a centralized, modern structure that reflects how users configure and operate the Orchestration Cluster in 8.8 and beyond.

---

## ✅ Iteration 1: Discovery & structure scaffolding

**Goal:** Define the structure and foundational section for shared configuration and concepts.

### Tasks

- [x] Create a new section under the Orchestration Cluster docs  
  _Proposed names: `Shared configuration and concepts` or `Core settings and features`_
- [x] Define initial TOC/folder structure for:
  - Configuration topics
  - Conceptual topics
  - Migration-specific topics (optional placeholder)
- [x] Draft a README/landing page explaining the purpose of this shared section

### Dependencies

- Align with @felix-mueller and @ThorbenLindhauer on naming and structure
- Confirm alignment with the unified config project

---

## 🚧 Iteration 2: Move & consolidate core configuration topics

**Goal:** Move common configuration content from Operate/Tasklist into the new shared section.

### Tasks

- [x] Migrate the following from Operate/Tasklist:
  - Licensing
  - Webserver
  - Security
  - Logging
  - Example `application.yml` (consider deleting or rewriting)
- [ ] Normalize content for 8.8+ (remove component-specific framing)
- [x] Add redirects or linkbacks to the new shared location
- [x] Update component-specific config index pages with links to shared content

### Dependencies

- None — clearly scoped and can begin immediately after structure is created

## When should this change go live?

- [ ] This is a bug fix, security concern, or something that needs **urgent release support**. (add `bug` or `support` label)
- [ ] This is already available but undocumented and should be released within a week. (add `available & undocumented` label)
- [ ] This is on a **specific schedule** and the assignee will coordinate a release with the Documentation team. (create draft PR and/or add `hold` label)
- [x] This is part of a scheduled **alpha or minor**. (add alpha or minor label)
- [ ] There is **no urgency** with this change (add `low prio` label)

## PR Checklist

- [ ] The commit history of this PR is cleaned up, using [`{type}(scope): {description}` commit message(s)](https://github.com/camunda/camunda-docs/blob/main/CONTRIBUTING.MD#commit-message-header-formatting)

[...]

Issue #6145, opened 2025-07-11
Unified configuration documentation for 8.8


mesellings:
This is a first step towards the 8.8 unified configuration documentation.

@mesellings to complete the following after meeting with @buccarel.

- Create a plan for docs effort required for this project.
- Investigate a structure for presenting the information for old > new.
- Investigate location in docs.

mesellings:
Draft PR created for this structure https://github.com/camunda/camunda-docs/pull/6523

PR #6523, opened 2025-08-26
Unified configuration mapping

mesellings:
## Description

Addresses the need to inform users of the new properties and any mappings involved with the move to a unified configuration for the Orchestration Cluster in 8.8.

Related issue: https://github.com/camunda/camunda-docs/issues/6145

**Note:**

- I have placed it in the sidebar in a different location, awaiting merge of https://github.com/camunda/camunda-docs/pull/6414, where it can then be moved into the core config section.
- This isn't the place to define default values I don't think - these should be defined in the individual config settings docs themselves.

- [X] This is part of a scheduled **alpha or minor**. (add alpha or minor label)

## PR Checklist

- [X] My changes are for **an upcoming minor release** and are in the `/docs` directory (version 8.8).

  - [X] [Engineering team review](https://github.com/camunda/camunda-docs/blob/main/howtos/documentation-guidelines.md#review-process)
  - [X] [Technical writer review](https://github.com/camunda/camunda-docs/blob/main/howtos/documentation-guidelines.md#review-process) via `@camunda/tech-writers` unless working with an embedded writer.


docs/self-managed/react-components/_config-table-data.js
```diff
@@ -0,0 +1,892 @@
...
+    types: ["1-to-1"],
+    area: ["Data"],
+  },
+  {
+    name: "camunda.data.backup.repository-name",
+    legacy: [
+      "camunda.operate.backup.repositoryName",
+      "camunda.tasklist.backup.repositoryName",
+    ],
+    types: ["1-to-1", "Breaking change (double configuration)"],
```
volodymyr-melnykc, line 322: @megglos For the `camunda.data.backup.repository-name` property, we have both "1-to-1" and "Breaking change" types where we seem to contradict if backward compatibility is supported. 
It also says that "_Starting with 8.8, the same repository must be used for both Operate and Tasklist._", do we need this note, since the current documentation page already describes expected properties for 8.8? Would it make sense to mark is just as the "Breaking change"?

docs/self-managed/components/orchestration-cluster/core-settings/configuration/configuration-mapping.md
volodymyr-melnykc, line 1: Can we point to this page from the Self-Managed upgrade guide: https://preview.docs.camunda.cloud/pr-6523/docs/next/self-managed/update/administrators/prepare-for-admin-upgrade/#step-2-assess-camunda-88-changes-and-impact?

volodymyr-melnykc (changes requested):
We have new properties introduced in 8.8, and marked as `new`. Do we need to provide description for them so the user understands what those properties are used for? I see that most of those properties are also not covered on the page that gives an overview of all properties: https://preview.docs.camunda.cloud/pr-6523/docs/next/self-managed/components/orchestration-cluster/core-settings/configuration/properties/

mesellings:
> @megglos @mesellings We have new properties introduced in 8.8, and marked as `new`. Do we need to provide a description for them so the user understands what those properties are used for? I see that most of those properties are also not covered on the page that gives an overview of all properties: https://preview.docs.camunda.cloud/pr-6523/docs/next/self-managed/components/orchestration-cluster/core-settings/configuration/properties/. Do we want to add them there?

Yep, this should go in the all properties page, and not here to avoid content duplication (once they are added we could always link to them in the all properties page as well) - @RomanJRW if I add these into that page I can probably add defaults and some descriptions, but who would be best to then help out with more detail about these config properties please?

docs/self-managed/components/orchestration-cluster/core-settings/configuration/configuration-mapping.md
mesellings: Yep, all done 👍 🚀

mesellings:
> > @RomanJRW if I add these into that page I can probably add defaults and some descriptions, but who would be best to then help out with more detail about these config properties please?
> 
> @mesellings - Do you have a list already of the new properties that need additional info? It would likely be a case by case basis to determine who could provide sufficient context (and also depends on what info you need)

Thanks - I'll add them in and fill in what I can then, and then we can tackle any gaps by pointing specific ones to the most relevant person? I'll make a start on this 👍

docs/self-managed/components/orchestration-cluster/core-settings/configuration/configuration-mapping.md
```diff
@@ -0,0 +1,94 @@
...
+| <span className="badge badge--1-to-1">1-to-1</span>                                                                                                           | <p>New Camunda 8.8 property with an existing equivalent legacy property or set of properties.</p><p>Backwards compatibility is supported as follows:</p><p><ul><li><p>If you have defined the new property, it is used.</p></li><li><p>If you have not defined the new property, the legacy property is used (if you have defined it).</p></li></ul></p>                                                                                                                                                                                                  |
```
theburi, line 37: From: New Camunda 8.8 property with an existing equivalent legacy property or set of properties.
To: New unified Camunda 8.8 property with an existing equivalent legacy property or set of properties.

theburi, line 37: From: If you have not defined the new property, the legacy property is used (if you have defined it).
To: If you have not defined the new property, the legacy property is used .

docs/self-managed/components/orchestration-cluster/core-settings/configuration/configuration-mapping.md
```diff
@@ -0,0 +1,94 @@
...
+| <span className="badge badge--breaking-change-double-configuration-" style={{ width: '150px', fontSize: '75%'}}>Breaking change (double configuration)</span> | <p>New Camunda 8.8 property with an existing equivalent legacy property or set of properties.</p><p>Backwards compatibility is **not** supported.</p><p><ul><li><p>Applies if a legacy configuration is defined with values that match the new configuration default or custom values.</p></li><li><p>If the legacy configuration value and the new configuration value match, the legacy property is accepted, even if not supported.</p></li><li><p>If the values do not match, the application will not start and an error is shown.</p></li></ul></p> |
```
theburi, line 38: I suggest removing the term - "Double configuration "; it is really confusing and needs a lot of extra explanation. I think without it, it's easier to talk about breaking changes. I will suggest how to rephrase to change it below

theburi: Breaking change (double configuration)
Breaking change

theburi: From: <p><ul><li><p>Applies if a legacy configuration is defined with values that match the new configuration default or custom values.</p></li><li><p>If the legacy configuration value and the new configuration value match, the legacy property is accepted, even if not supported.</p></li><li><p>If the values do not match, the application will not start and an error is shown.</p></li></ul></p>
To 
<p><ul><li><p>Move legacy properties marked with breaking change to the new unified set before upgrade to 8.8 </p></li><li><p>You could keep old, legacy properties in your configuration file as long as they match the new unified configuration equivalent. However, this is not recommended as it may lead to misconfiguration. If the values do not match, the application will not start, and an error will be logged.</p></li></ul></p>

docs/self-managed/components/orchestration-cluster/core-settings/configuration/configuration-mapping.md
```diff
@@ -0,0 +1,94 @@
...
+
+### Recommended actions
+
+**As part of upgrading to 8.8, Camunda recommends you replace any legacy properties shown in this table with the equivalent new property.**
```
theburi, line 42: **As part of upgrading to 8.8, replace any legacy properties shown in this table with the equivalent new unified configuration property.**

docs/self-managed/components/orchestration-cluster/core-settings/configuration/configuration-mapping.md
```diff
@@ -0,0 +1,94 @@
...
+
+#### Option 1: Strict migration
+
+The recommended strict migration approach is to remove the legacy properties and add the corresponding new property:
+
+```
+camunda.data.secondary-storage.opensearch.url=http://prod-db.com:54321
+```
+
+#### Option 2: Double configuration
```
theburi, line 69: Lets remove this Option. I don't believe there is a use case for having it.

theburi: And as mentioned above, let's remove double configuration mentions. Customers just need to upgrade them, and we mentioned that. Old properties could be kept as a note.

theburi (commented):
as discussed with @mesellings  a bit of clarification feedback.

docs/self-managed/components/orchestration-cluster/core-settings/configuration/configuration-mapping.md
mesellings: @buccarel @RomanJRW FYI: I've changed to just use "Breaking change" as suggested by @theburi

RomanJRW (dismissed):
Nice improvement - thanks! 👍 

Based on the distribution in [this comment](https://camunda.slack.com/archives/C08E56YUUMS/p1759172128052549?thread_ts=1759155031.238609&cid=C08E56YUUMS), will approve the sections that I checked.

deepthidevaki (changes requested):
Most documentation generated by AI is incorrect. It would be great if you can copy the documentation from the javadoc of each of the property. You can find them in the corresponding classes [here](https://github.com/camunda/camunda/tree/stable/8.8/configuration/src/main/java/io/camunda/configuration). You can rephrase it ofcourse. But please be careful in adding more details because most of the time it is incorrect and also missing important information. 

I reviewed the following sections so far:
```
camunda.cluster.*
camunda.data.snapshot-period
camunda.data.backup.s3
camunda.data.backup.gcs
```

megglos (changes requested):
Finished the `camunda.api.grpc.*` part, it seems we have quite some hallucination in there on descriptions + wrong default values.

Happy to take a different approach to review comments but actually commit changes, wdyt @mesellings ?

buccarel (dismissed):

deepthidevaki (approved):
Thanks a lot for your work.
There are some small inconsistencies. 
* The default value for duration is sometime set with a quotes (`'250ms'`) and sometime without quotues `0s`. My preference is without quotes. I don't think the users have to provide the value with the quotes. 
* The non existing default values are sometime mentioned as "null" and sometime "None". It would be good to use one of them consistently.

PR #6664, opened 2025-09-09
docs: on increasing replica count during upgrade to 8.8

EuroLew:
## Description

Starting in 8.8 the replica settings will default to 1, this is a breaking change which may require cluster changes to bring it back to a green state from yellow where:
Cluster health status logic:

Green = all primaries and replicas allocated.
Yellow = all primaries allocated, but some replicas unassigned.
Red = some primaries unassigned.

It also includes instructions on how to rollback the changes.


docs/self-managed/concepts/pvc-resize-guide.md
```diff
@@ -0,0 +1,163 @@
...
+
+Or for a specific class:
+
+```bash
+kubectl get storageclass <your-storage-class> -o yaml | grep -A5 -B5 allowVolumeExpansion
+```
+
+### Step 1: Check Current PVCs
+```bash
+kubectl get pvc -l app.kubernetes.io/component=data -o wide
```
Langleu, line 17: not generally applicable.
This assumes that you have split up your Elasticsearch setup into their distinctive types à la "master", "data", "cooridnator", ... nodes.

E.g. the Camunda Helm chart only uses master nodes that contain the other roles as well, therefore it's just `app.kubernetes.io/component=master`.
So again, highly dependent on the setup of a user.

docs/self-managed/concepts/pvc-resize-guide.md
```diff
@@ -0,0 +1,163 @@
...
+```
+
+### Step 2: Resize PVCs Directly
+```bash
+# Example: Resize two Elasticsearch data PVCs
+kubectl patch pvc data-elasticsearch-data-0 -p '{"spec":{"resources":{"requests":{"storage":"200Gi"}}}}'
+kubectl patch pvc data-elasticsearch-data-1 -p '{"spec":{"resources":{"requests":{"storage":"200Gi"}}}}'
```
Langleu, line 24: you could either turn it into a small bash script so it's independent of the amount of pvcs or just make it generally applicable requiring a user to actively think and do the change themselves.
You already mentioned above a way to figure out the volumes based on labels.
Again be careful as it must not fit a customers setup 100%.

`data-elasticsearch-data-0` is dependent on your setup, it may be called differently depending on the customers setup or they already have > 200 Gi and downsizing isn't supported.

docs/self-managed/concepts/pvc-resize-guide.md
```diff
@@ -0,0 +1,163 @@
...
+```
+
+
+#### Monitor:
+
+```bash
+kubectl get pvc -l app.kubernetes.io/component=data -w
```
Langleu, line 31: same comment about the first usage of `component=data`, highly depends on the customer's setup.
Not the default in e.g. the Camunda Helm Chart.

docs/self-managed/concepts/pvc-resize-guide.md
```diff
@@ -0,0 +1,163 @@
...
+```
+
+### Step 3 (Optional): Update Helm Values
+```yaml
+# values.yaml for bitnami ES chart
+data:
+    persistence:
+        size: 200Gi
```
Langleu, line 39:
```
Error: UPGRADE FAILED: cannot patch "elasticsearch-data" with kind StatefulSet: StatefulSet.apps "elasticsearch-data" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'ordinals', 'template', 'updateStrategy', 'persistentVolumeClaimRetentionPolicy' and 'minReadySeconds' are forbidden
```

Basically it's a field that can not be updated.
You can scale the PVC and change it's size of the resulting volume but the resulting statefulset will continue mentioning the old initial value.

Excerpt from the statefulset of the data node:
```
  volumeClaimTemplates:
  - apiVersion: v1
    kind: PersistentVolumeClaim
    metadata:
      creationTimestamp: null
      name: data
    spec:
      accessModes:
      - ReadWriteOnce
      resources:
        requests:
          storage: 10Gi
      volumeMode: Filesystem
```

The only workaround for that is to remove it and recreate but sounds like more can go wrong than right depending on a customers setup.

docs/self-managed/concepts/pvc-resize-guide.md
```diff
@@ -0,0 +1,163 @@
...
+```
+
+```bash
+helm upgrade elasticsearch oci://registry-1.docker.io/bitnamicharts/elasticsearch \
+--version 21.6.3 \
+-f values.yaml
```
Langleu, line 45: goes hand in hand with the error mentioned above.
You can not change it in the values.yml unless you remove the statefulset and let it be re-created.

docs/self-managed/concepts/pvc-resize-guide.md
```diff
@@ -0,0 +1,163 @@
...
+    resources:
+        requests:
+            storage: 200Gi
+```
+
+```bash
+kubectl apply -f new-pvc.yaml
+```
+
+## Step 4: Replace Original PVC (Optional Downtime)
```
Langleu, line 136: shouldn't the first step be that you take the system offline to allow for a coherent volume snapshot?
Downtime is required in any case.

docs/self-managed/concepts/pvc-resize-guide.md
Langleu: also tricky again since Camunda doesn't consume the artifact that way but the Camunda Helm Chart includes it.

Langleu (changes requested):
Leaving some comments about the actual content.

My comment remains about considering the general guidance and leaving it to the customer on how to approach this.
- Upstream k8s docs
- Maybe manage service stuff (AWS OpenSearch / Elastic Cloud)

If they need more hand holding --> Consulting and they can then properly evaluate based on their actual environment.

docs/self-managed/concepts/pvc-resize-guide.md
EuroLew: resolved as not giving detailed steps to customer given they will be SM.

docs/self-managed/concepts/pvc-resize-guide.md
EuroLew: resolved as not giving detailed steps to customer given they will be SM.

docs/self-managed/concepts/pvc-resize-guide.md
EuroLew: resolved as not giving detailed steps to customer given they will be SM.

Langleu (dismissed):
thanks for including the suggested changes 👍

zaycodes (approved):
lgtm! Just made a few styling adjustments.

docs/self-managed/components/components-upgrade/870-to-880.md
```diff
@@ -311,3 +311,71 @@ Exporter configuration removal no longer deletes the exporter from the system. I
 ### Email connector
 
 Starting with the 8.8 release, angle brackets (`<` and `>`) are no longer removed from the `messageId`.
+
+## Elasticsearch
```
christinaausley, line 315: Is this a breaking change @EuroLew? If so, we could also add a little note in the announcements. If not, this is fine as-is.

EuroLew: it is a breaking change, if the customers cluster is not set up to handle this, the change will fill up the existing disk storage and completely block Camunda. Where would I mention this as an announcement?

Issue #6875, opened 2025-09-26
Add troubleshooting tips from the support demo to the user documentation


ThorbenLindhauer:
* Move useful configuration and troubleshooting tips from the [support demo FAQ document](https://docs.google.com/document/d/18DARbmL8kfOn18LRlu2H3c1NTGvC74b8qTQN_zvHiK8/edit?tab=t.0) into the public documentation

PR #6876, opened 2025-09-26
add OIDC troubleshooting and advanced config

ThorbenLindhauer:
related to #6875


docs/self-managed/components/orchestration-cluster/identity/debugging-authentication.md
```diff
@@ -0,0 +1,129 @@
...
+- camunda-role
+- camunda-tenant
+- camunda-user
+- camunda-web-session
+
+## Reviewing configuration
+
+To review the effective configuration of your Orchestration Cluster, you can call the [Spring Boot Actuator endpoint](https://docs.spring.io/spring-boot/reference/actuator/endpoints.html#actuator.endpoints) at `<server>:<port>/actuator/configprops`.
+
+For example, with a Camunda 8 Run installation, this endpoint is available at `http://localhost:9600/actuator/configprops`. Note that the actuator port differs from the Orchestration Cluster API port and may not always be accessible, depending on your deployment setup.
```
Ben-Sheppard, line 91: ❓ Should we also note here that the `/actuator/configprops` should be aimed at the Orchestration cluster with context path if applicable? It may make things less clear though/harder to describe.

I mention this because it has come up recently in an internal [request](https://camunda.slack.com/archives/C08MRKHJ0CD/p1758899257601269)

docs/self-managed/components/orchestration-cluster/identity/special-oidc-cases.md
```diff
@@ -0,0 +1,50 @@
...
+<TabItem value="env">
+```
+CAMUNDA_SECURITY_AUTHENTICATION_OIDC_AUTHORIZATIONURI=http://localhost:18080/protocol/openid-connect/auth
+CAMUNDA_SECURITY_AUTHENTICATION_OIDC_TOKENURI=http://<Docker network domain name>:18080/protocol/openid-connect/token
+CAMUNDA_SECURITY_AUTHENTICATION_OIDC_JWKSETURI=http://<Docker network domain name>:18080/protocol/openid-connect/certs
+```
+</TabItem>
+</Tabs>
+
+The exact property values you need to use depend on your OIDC provider.
```
Ben-Sheppard, line 50: 💭 Just a suggestion, with information like this which is important to the text above, I like to draw attention to it via something like a `Tip` block. NABD though

Ben-Sheppard (approved):
Thank you for the addition here, the content looks good to me and I especially appreciate the thought to include debugging/troubleshooting an often challenging area.

I have left a couple of NABD comments, one thing that I wouldn't block this PR on is that for some of these variables we have some nice descriptions however for others, i.e. `CAMUNDA_SECURITY_AUTHENTICATION_OIDC_USERNAMECLAIM` we have a simple description that doesn't convey its complete purpose (IMO). 

We should also add these descriptions to the properties reference page https://docs.camunda.io/docs/next/self-managed/components/orchestration-cluster/core-settings/configuration/properties/#properties (though I'm fine with you creating a follow-up for this to get this PR merged as its been open for a bit now) - given my stance here I'll approve early and leave it to your judgement on the additions to the other pages.

docs/self-managed/components/orchestration-cluster/identity/debugging-authentication.md
ThorbenLindhauer: I'm adding a sentence that people may have to replace the localhost:9600 part, but I won't describe in detail what it is for each setup.

Issue #7207, opened 2025-10-30
Documentation for setting up generic oidc is wrong for 8.8


zalanszak:
In the documentation to connect to an external oidc provider in GENERIC mode:
https://docs.camunda.io/docs/self-managed/components/management-identity/configuration/connect-to-an-oidc-provider/
https://docs.camunda.io/docs/self-managed/deployment/helm/configure/authentication-and-authorization/connect-to-an-oidc-provider/

it is still written that:

```
global:
  identity:
    auth:
      issuer: <URL_OF_ISSUER>
      # this is used for container to container communication
      issuerBackendUrl: <URL_OF_ISSUER>
      tokenUrl: <TOKEN_URL_ENDPOINT>
      jwksUrl: <JWKS_URL>
      type: "GENERIC"
```

However, this is wrong. `issuer` should not be used anymore. Instead, `authUrl` should be used.

Similarly, `issuerBackendUrl` should be omitted - if it's left, it will cause problems, and the redirect will not work.

Related ticket: https://jira.camunda.com/browse/SUPPORT-29817

Issue #7107, opened 2025-10-17
Make the Management Identity OIDC configuration guides independent of Helm


ThorbenLindhauer:
* This page describes how to configure Management Identity against OIDC providers in various setups: https://docs.camunda.io/docs/next/self-managed/components/management-identity/configuration/identity-configuration-overview/
* We have end-to-end Helm OIDC configuration guides at https://docs.camunda.io/docs/next/self-managed/deployment/helm/configure/authentication-and-authorization/ (and are significantly improving them via #6895)
* To reduce redundancies in our docs, I suggest with this ticket:
  * We focus https://docs.camunda.io/docs/next/self-managed/components/management-identity/configuration/identity-configuration-overview/ on the application level only; i.e. remove Helm values examples and Helm explanations from the steps
  * At the very beginning of the guide, we very visibly link to the respective Helm setup guides
  * As needed, we can add more backlinks from the Helm setup guide to the Management-Identity-specific guide
* Note that a small part (external keycloak guide changes) is already in the PR for 6895: https://github.com/camunda/camunda-docs/pull/7083

Issue #7105, opened 2025-10-17
Link to Helm Chart guide from "Connect to an Identity provider" guide


ThorbenLindhauer:
* https://docs.camunda.io/docs/next/self-managed/components/orchestration-cluster/identity/connect-external-identity-provider currently explains how to setup up the OC with an OIDC provider. That includes Helm values, however is not a complete solution to deploy the Helm Chart with the OC connected to an IDP.
* We have end-to-end Helm OIDC configuration guides at https://docs.camunda.io/docs/next/self-managed/deployment/helm/configure/authentication-and-authorization/ (and are significantly improving them via #6895)
* To reduce redundancies in our docs, I suggest with this ticket:
  * We focus https://docs.camunda.io/docs/next/self-managed/components/orchestration-cluster/identity/connect-external-identity-provider/ on the application level only; i.e. remove Helm values examples and Helm explanations from the steps
  * At the very beginning of the guide, we very visibly link to the Helm setup guides
  * As needed, we can add more backlinks from the Helm setup guide to the OC-specific guide (e.g. if we want to point users to the worker setup at the end of the guide)

Issue #6895, opened 2025-09-29
Finalize oidc configuration guide for 8.8


hisImminence:
With the identity changes in 8.8, we introduced a new [authorization and authentication](https://stage.docs.camunda.io/docs/next/self-managed/installation-methods/helm/configure/authentication-and-authorization/) section in the Helm Chart configuration part.

Via https://github.com/camunda/camunda-platform-helm/issues/4206, we introduced a section regarding [basic auth](https://stage.docs.camunda.io/docs/next/self-managed/installation-methods/helm/configure/authentication-and-authorization/#basic-authentication) and [internal keycloak](https://stage.docs.camunda.io/docs/next/self-managed/installation-methods/helm/configure/authentication-and-authorization/#internal-keycloak). We now miss the last part to guide users through a generic/ and entraId configuration.

We have an guide (from 8.7) that was guiding users through this setup that was for only simply copy pasted to 8.8, see [Helm chart OpenID Connect provider setup](https://stage.docs.camunda.io/docs/next/self-managed/installation-methods/helm/configure/connect-to-an-oidc-provider/).

## Scope

With this ticket we want to make sure:
* [ ] As a user I can follow the guide to connect to an generic provider
* [ ] As a user I understand and can see the configuration options for the Helm Charts related to OIDC.
* [ ] The guide is listed in the configuration section of the helm chart (alternatively if time see bonus)

**Bonus**
* [ ] refactor the page to fit into the [authorization and authentication](https://stage.docs.camunda.io/docs/next/self-managed/installation-methods/helm/configure/authentication-and-authorization/) section.
* [ ] As a user I can follow the guide to connect to Entra ID

## Implementation notes

* If possible follow and base the structure of this guide update on the [Internal Keycloack](https://stage.docs.camunda.io/docs/next/self-managed/installation-methods/helm/configure/authentication-and-authorization/#internal-keycloak) guide:
   * General intro and information
   * Prerequisite/ what connection detail I need to have prepared [Create a secrets]
   * Configuration for connection/ management identity/ orchestration identity/ connectors
   * Minimal full config example
   * Connecting to the Camunda cluster
   * Hint/ Table of other configuration params

ThorbenLindhauer:
What's left to do here (based on current state of PR):

* Link to the Helm Chart config guides from https://docs.camunda.io/docs/next/components/concepts/access-control/connect-to-identity-provider/
[...]

PR #7267, opened 2025-11-06
docs(self-managed): finalize oidc configuration updates for 8.8

zaycodes:
## Description

- Update generic OIDC Helm example to use authUrl/publicIssuerUrl instead of issuer - closes https://github.com/camunda/camunda-docs/issues/7207
- Make the Management Identity OIDC configuration guide independent of Helm - closes https://github.com/camunda/camunda-docs/issues/7107
- Link to Helm Chart guide from "Connect to an Identity provider" guide - closes https://github.com/camunda/camunda-docs/issues/7105
- Document offline_access as optional and show CAMUNDA_IDENTITY_AUTH_SCOPES override
- Add prominent admin-access requirement note to external Keycloak Helm guide
- Link Helm guides that create secrets to the Secret management “external Kubernetes secrets” section

Closes https://github.com/camunda/camunda-docs/issues/6895

## PR Checklist

### DIFF
--- a/config-reference/camunda-spring-boot-starter/spring-configuration-metadata.json
+++ b/config-reference/camunda-spring-boot-starter/spring-configuration-metadata.json
@@ -91,6 +91,11 @@
       "description": "Keycloak-specific properties."
     },
     {
+      "name": "management.endpoint.jobworkers",
+      "type": "io.camunda.client.spring.actuator.JobWorkerController",
+      "sourceType": "io.camunda.client.spring.actuator.JobWorkerController"
+    },
+    {
       "name": "zeebe",
       "type": "io.camunda.spring.client.properties.CamundaSelfManagedProperties"
     },
@@ -193,6 +198,12 @@
       "defaultValue": "$HOME\/.camunda\/credentials"
     },
     {
+      "name": "camunda.client.auth.issuer-url",
+      "type": "java.net.URI",
+      "description": "The url of the issuer for the access token. It is used to generate the well-known configuration url from which the `token-url` is retrieved. Only applied if the `camunda.client.auth.well-known-configuration-url` is not set. A default is set by `camunda.client.auth.method: oidc`.",
+      "sourceType": "io.camunda.client.spring.properties.CamundaClientAuthProperties"
+    },
+    {
       "name": "camunda.client.auth.keystore-key-password",
       "type": "java.lang.String",
       "description": "The keystore key password for the OAuth identity provider.",
@@ -244,7 +255,7 @@
     {
       "name": "camunda.client.auth.token-url",
       "type": "java.net.URI",
-      "description": "The authorization server URL from which to request the access token. A default is set by `camunda.client.mode: saas` and `camunda.client.auth.method: oidc`.",
+      "description": "The authorization server URL from which to request the access token. A default is set by `camunda.client.mode: saas`.",
       "sourceType": "io.camunda.client.spring.properties.CamundaClientAuthProperties"
     },
     {
@@ -263,6 +274,12 @@
       "name": "camunda.client.auth.username",
       "type": "java.lang.String",
       "description": "The username to use for basic authentication. A default is set by `camunda.client.auth.method: basic`.",
+      "sourceType": "io.camunda.client.spring.properties.CamundaClientAuthProperties"
+    },
+    {
+      "name": "camunda.client.auth.well-known-configuration-url",
+      "type": "java.net.URI",
+      "description": "The url of the well-known configuration of the issuer. It is used to retrieve the `token-url`. Only applied if `camunda.client.auth.token-url` is not set.",
       "sourceType": "io.camunda.client.spring.properties.CamundaClientAuthProperties"
     },
     {
@@ -328,6 +345,13 @@
       "type": "java.time.Duration",
       "description": "The time interval between keep-alive messages sent to the gateway.",
       "sourceType": "io.camunda.client.spring.properties.CamundaClientProperties"
+    },
+    {
+      "name": "camunda.client.max-http-connections",
+      "type": "java.lang.Integer",
+      "description": "The maximum number of concurrent HTTP connections the client can open.",
+      "sourceType": "io.camunda.client.spring.properties.CamundaClientProperties",
+      "defaultValue": 100
     },
     {
       "name": "camunda.client.max-message-size",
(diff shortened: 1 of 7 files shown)


docs/self-managed/components/management-identity/configuration/connect-to-an-oidc-provider.md
```diff
@@ -17,6 +17,10 @@ Connect Management Identity to an OpenID Connect (OIDC) authentication provider
 - A full list of supported and unsupported features when using an OIDC provider is available in the [OIDC features table](#supported-and-unsupported-oidc-features).
 - To connect to a Keycloak authentication provider, see [connect to an existing Keycloak instance](/self-managed/components/management-identity/configuration/connect-to-an-existing-keycloak.md).
 
+:::info Deploying with Helm?
+If you deploy Camunda 8 Self-Managed with Helm, use the [Helm chart authentication and authorization guides](/self-managed/deployment/helm/configure/authentication-and-authorization/index.md) to configure OIDC and Management Identity.
+
+This guide focuses on application-level Management Identity configuration (for example, environment variables and IdP application settings) and applies regardless of deployment method.
```
ThorbenLindhauer: The part "applies regardless of deployment method" would be a bit confusing for me as a reader, because it sounds like I should follow this guide in any case. Maybe we can replace this with something like "Follow this guide if you want to configure Management Identity on the application level or if you would like to understand how configuration properties apply at the application level" (I'm sure there's a more compact way to say this :))?

docs/self-managed/components/management-identity/configuration/identity-configuration-overview.md
```diff
@@ -7,6 +7,12 @@ description: "Learn more about how Management Identity plays together with Keycl
 
 Configure Management Identity for your Camunda 8 Self-Managed deployment.
 
+:::info Deploying with Helm?
```
ThorbenLindhauer: Should we add the note also the subpages "Connect to an existing Keycloak instance" and "Configure an external IdP using Keycloak"? We could even link directly to the equivalent Helm guides, for example to https://docs.camunda.io/docs/self-managed/deployment/helm/configure/authentication-and-authorization/external-keycloak/ for the existing Keycloak instance page.

docs/self-managed/components/management-identity/configuration/connect-to-an-oidc-provider.md
```diff
@@ -138,6 +96,17 @@ Once set, you cannot update your initial claim name and value using environment
 
 For authentication, the Camunda components use the scopes `email`, `openid`, `offline_access`, and `profile`.
 
+:::tip Optional scopes
+The `offline_access` scope is optional.
```
ThorbenLindhauer: Here we could explain a bit what the purpose of the `offline_access` scope is, so that users can decide more easily if they set it or not.

Something like this:

If the `offline_access` scope is present, your OIDC provider will issue a refresh token to Management Identity on user login. Management Identity will then use the refresh token to renew the user's access token when it expires, so that the user's session is automatically extended. Without the scope present, when the access token expires the user will be redirected to the OIDC provider for new authentication.

And we could link to https://openid.net/specs/openid-connect-core-1_0.html#OfflineAccess as the specification reference.

docs/self-managed/components/management-identity/configuration/connect-to-an-oidc-provider.md
```diff
@@ -169,7 +138,7 @@ Ensure you register a new application for each component.
 You can connect to your OIDC provider through either environment variables or Helm values. Ensure only one configuration option is used.
 :::
 
-<Tabs groupId="optionsType" defaultValue="env" queryString values={[{label: 'Environment variables', value: 'env' },{label: 'Helm values', value: 'helm' }]} >
+<Tabs groupId="optionsType" defaultValue="env" queryString values={[{label: 'Environment variables', value: 'env' }]} >
```
ThorbenLindhauer: We could add a tab `Application.yaml` here that gives the yaml-formatted application config (to make it consistent with how we have it in the Orchestration Cluster config guide, e.g. here: https://docs.camunda.io/docs/next/self-managed/components/orchestration-cluster/identity/connect-external-identity-provider/?optionsType=yaml#step-4-configure-the-oidc-connection-details).

You can create that by translating the env vars into a hierarchical yaml structure, so roughly like:
```
camunda:
  identity:
    type: "MICROSOFT"
    baseUrl: <IDENTITY_URL>
    ...
identity:
  initialClaimName: ..
  ...
spring:
  profiles:
    active: ...
```

ThorbenLindhauer (changes requested):
Thank you for preparing this. I really like to see that we are reducing redundancies/inconsistencies and guiding people to the right place, that's a big improvement.

I left a few detail comments, nothing particularly critical.

docs/self-managed/deployment/helm/configure/authentication-and-authorization/microsoft-entra.md
```diff
@@ -31,6 +31,22 @@ See the [full configuration example](#full-configuration-example) for the comple
...
+
+If your organization restricts this scope for security reasons, you can adjust the scopes with:
+
+```
+CAMUNDA_IDENTITY_AUTH_SCOPES="openid profile email"
+```
+
+This configuration allows login without the `offline_access` scope.
+:::
+
```
ThorbenLindhauer: Here as well, the box is not in the context of configuring a particular component, so the reference to Management Identity and the configuration setting are not meaningful. I suggest we explain here the general principle what `offline_access` is and that it's optional, and give a hint that for each component that it can be left out and how to do that. In the `Configure Orchestration Cluster` section you can already see it and tell people to remove it from the list; for the other components, I don't know off the top of my head where people need to change the config (I assume some components apply the scope per default so we'd have to add an explicit new value to overwrite that). You could validate that with a Self-Managed engineer.

ThorbenLindhauer (changes requested):
Thanks for the quick adjustments. A few more comments, please see below.

docs/self-managed/deployment/helm/configure/authentication-and-authorization/microsoft-entra.md
hamza-m-masood: From what I can see in the Helm chart, scopes are only defined for the connectors and the orchestration cluster. For connectors specifically, we don't provide any default values for the scope:
https://github.com/camunda/camunda-platform-helm/blob/630c4e28fc516465af72e3ce858d0b359f119a8d/charts/camunda-platform-8.8/templates/connectors/files/_application.yaml#L50-L50

I agree that it does not make much sense to add the env var `CAMUNDA_IDENTITY_AUTH_SCOPES="openid profile email"` since it is not clearly related to any specific component. 

I think adding custom scope configuration for specific components would be a new feature in the Helm chart. And  would require more research on how it would work for each component.
In that case let's not mention component-specific configuration in this section for `offline_access` and keep things general, at least for a first draft.

hisImminence (approved):

PR #7334, opened 2025-11-14
Update Javadoc link to the correct version 8.7.17

falko:
## Description

The link to javadoc should be versioned otherwhise users will see the latest alpha.

## PR Checklist

### DIFF
--- a/.github/workflows/build-docs.yaml
+++ b/.github/workflows/build-docs.yaml
@@ -8,7 +8,7 @@ jobs:
     env:
       temporary_link_path: .github/workflows/validate-links
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
       - uses: actions/setup-node@v6
         with:
           node-version: 24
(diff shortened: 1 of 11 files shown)


mesellings (approved):
Approved 🚀

mesellings:
Looking at this more closely then, I'm not sure if this is something we can do - trying to keep the docs up-to-date every time a new patch release is released would be onerous (for example, a new patch came out 8.7.18 during this review). Unless we can automate this somehow, I think the only realistic thing we could do here is to add an admonition/note to this link to say that by default the latest is shown, but that they should select the correct one for their version from the Javadoc page. Wdyt?

mesellings:
@falko what about an approach like this, which saves us from having to try and keep this moving target in sync:

image

We can update this image for other versions (e.g. 8.6)

mesellings (approved):
Approved - only added text for 8.8+ as the page has a different layout, but we can look at this when we give the java client pages a thorough review.

PR #4607, opened 2024-11-15
docs(modeler): add troubleshooting entry for sandboxing issue

nikku:
## Description

This adds a knowledge base / troubleshooting entry for https://github.com/camunda/camunda-modeler/issues/4695.

Related to https://github.com/camunda/camunda-modeler/issues/4695

## When should this change go live?

- [ ] This is a bug fix, security concern, or something that needs **urgent release support**.
- [x] This is already available but undocumented and should be released within a week.
- [ ] This on a **specific schedule** and the assignee will coordinate a release with the DevEx team. (apply `hold` label or convert to draft PR)
- [ ] This is part of a scheduled alpha or minor. (apply alpha or minor label)
- [ ] There is **no urgency** with this change and can be released at any time.

## PR Checklist

- [x] My changes are for an **already released minor** and are in `/versioned_docs` directory.
- [x] My changes are for the **next minor** and are in `/docs` directory (aka `/next/`).

- [x] My changes require an [Engineering review](https://github.com/camunda/camunda-docs/blob/main/howtos/documentation-guidelines.md#review-process), and I've assigned the Engineering DRI or delegate.
- [x] My changes require a [technical writer review](https://github.com/camunda/camunda-docs/blob/main/howtos/documentation-guidelines.md#review-process), and I've assigned @camunda/tech-writers as a reviewer.

PR #7362, opened 2025-11-19
fix: rework generic oidc guide

bkenez:
## Description

Closes: [#4607](https://github.com/camunda/camunda-platform-helm/issues/4607)
Compete rewrite of the generic OIDC setup guide for the helm chart

## PR Checklist

### DIFF
--- a/docs/guides/migrating-from-camunda-7/migration-tooling.md
+++ b/docs/guides/migrating-from-camunda-7/migration-tooling.md
@@ -70,7 +70,7 @@ Click **Analyze and convert**:
 
 On this screen you can now:
 
-- Download the analyzer results as Microsoft Excel file (XSLX)
+- Download the analyzer results as Microsoft Excel file (XLSX)
 - Download the analyzer results as CSV file
 - Download the converted models (individually or as ZIP)
 
@@ -131,8 +131,10 @@ java -jar camunda-7-to-8-migration-analyzer-cli.jar local
 
 Missing required parameter: '<file>'
 Usage: camunda-7-to-8-migration-analyzer-cli local [-dhoV] [--check] [--csv]
-       [--delegate-execution-as-job-type] [--disable-append-elements]
-       [--disable-default-job-type] [--md] [-nr]
+       [--add-data-migration-execution-listener]
+       [--data-migration-execution-listener-job-type=<dataMigrationExecutionListenerJobType>]
+       [--disable-append-elements]
+       [--always-use-default-job-type] [--md] [-nr]
        [--default-job-type=<defaultJobType>]
        [--platform-version=<platformVersion>] [--prefix=<prefix>] <file>
 Converts the diagram from the given directory or file
@@ -144,25 +146,32 @@ java -Dfile.encoding=UTF-8 -jar camunda-7-to-8-migration-analyzer-cli.jar local
 Parameter:
       <file>                 The file to convert or directory to search in
 Options:
+      --add-data-migration-execution-listener
+                             Add an execution listener on blank start events
+                               that can be used for the Camunda 7 Data Migrator
+      --always-use-default-job-type
+                             Always fill in the configured default job type,
+                               interesting if you want to use one delegation
+                               job worker (like the Camunda 7 Adapter).
       --check                If enabled, no converted diagrams are exported
       --csv                  If enabled, a CSV file will be created containing
-                               the conversions results
-      --xslx                 If enabled, a XSLX file will be created containing
-                               the analysis results
+                               the results for the analysis
   -d, --documentation        If enabled, messages are also appended to
                                documentation
+      --data-migration-execution-listener-job-type=<dataMigrationExecutionListen
+        erJobType>
+                             Name of the job type of the listener. If set, the
+                               default value from the 'converter-properties.
+                               properties' is overridden
       --default-job-type=<defaultJobType>
-                             If set, the default value from the
-                               'converter-properties.properties' for the job
-                               type is overridden
-      --delegate-execution-as-job-type, --delegate-expression-as-job-type
-                             If enabled, sets the delegate expression as the
-                               job type
+                             Job type used when adjusting delegates. If set,
+                               the default value from the 'converter-properties.
+                               properties' is overridden
       --disable-append-elements
                              Disables adding conversion messages to the bpmn xml
-      --disable-default-job-type
-                             Disables the default job type
   -h, --help                 Show this help message and exit.
+      --keep-job-type-blank  Sets all job types to blank so that you need to
+                               edit those after conversion yourself
       --md, --markdown       If enabled, a markdown file will be created
                                containing the results for all conversions
       -nr, --not-recursive   If enabled, recursive search in subfolders will be
@@ -174,6 +183,8 @@ Options:
       --prefix=<prefix>      Prefix for the name of the generated file
                                Default: converted-c8-
   -V, --version              Print version information and exit.
+      --xlsx                 If enabled, a XLSX file will be created containing
+                               the results for the analysis
 ```
 
 ### Converting your models
--- a/versioned_docs/version-8.7/guides/migrating-from-camunda-7/migration-tooling.md
+++ b/versioned_docs/version-8.7/guides/migrating-from-camunda-7/migration-tooling.md
@@ -70,7 +70,7 @@ Click **Analyze and convert**:
 
 On this screen you can now:
 
-- Download the analyzer results as Microsoft Excel file (XSLX)
+- Download the analyzer results as Microsoft Excel file (XLSX)
 - Download the analyzer results as CSV file
 - Download the converted models (individually or as ZIP)
 
@@ -131,8 +131,10 @@ java -jar camunda-7-to-8-migration-analyzer-cli.jar local
 
 Missing required parameter: '<file>'
 Usage: camunda-7-to-8-migration-analyzer-cli local [-dhoV] [--check] [--csv]
-       [--delegate-execution-as-job-type] [--disable-append-elements]
-       [--disable-default-job-type] [--md] [-nr]
+       [--add-data-migration-execution-listener]
+       [--data-migration-execution-listener-job-type=<dataMigrationExecutionListenerJobType>]
+       [--disable-append-elements]
+       [--always-use-default-job-type] [--md] [-nr]
        [--default-job-type=<defaultJobType>]
        [--platform-version=<platformVersion>] [--prefix=<prefix>] <file>
 Converts the diagram from the given directory or file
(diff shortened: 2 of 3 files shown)


docs/self-managed/deployment/helm/configure/authentication-and-authorization/generic-oidc-provider.md
```diff
@@ -1,146 +1,784 @@
...
+:::tip Standard OIDC claims
+The OpenID Connect specification defines standard claims:
+
+- **User identification**: `sub` (subject) is always present and uniquely identifies the user. Providers may also include `preferred_username`, `email`, or other profile claims.
+- **Client identification**: `azp` (authorized party) or `client_id` depending on the provider's implementation. Both are standard OIDC claims.
+
+**Camunda's default claim mappings** (used if not explicitly configured):
+
+- `usernameClaim`: `preferred_username` (fallback to `sub` if not present)
+- `clientIdClaim`: `client_id`
```
ThorbenLindhauer: I think we don't have this as a unified concept across all components.

In the orchestration cluster, you can configure username claim and client id claim accordingly (and the defaults are `sub` and `<not set>`, i.e. all accesses are treated as users). In Web Modeler, the user id is always the `sub` claim and then it has a freely configurable (I think?!) user display name. And I'm not sure right now if it even has a concept for a client id. Didn't check the other components now.

In any case, I'm also not sure what this box here helps people with. Why did you include it? Maybe then we can find another way to formulate that.

docs/self-managed/deployment/helm/configure/authentication-and-authorization/generic-oidc-provider.md
```diff
@@ -1,146 +1,784 @@
...
+        redirectUrl: <identity-base-url>
+        secret:
+          existingSecret: oidc-credentials
+          existingSecretKey: identity-client-secret
+        initialClaimName: <user-claim-name>
+        initialClaimValue: <admin-user-claim-value>
+
+identity:
+  enabled: true
+  contextPath: "/identity" # Optional: adjust if using custom context path
```
ThorbenLindhauer: By setting the context path, the redirect URL for the localhost setup from the table above will not be correct, or?

Is there any particular reason that you are including it here? I wonder if we could remove it to keep the config minimal.

docs/self-managed/deployment/helm/configure/authentication-and-authorization/generic-oidc-provider.md
```diff
@@ -1,146 +1,784 @@
...
+```yaml
+global:
+  identity:
+    auth:
+      orchestration:
+        clientId: <orchestration-client-id>
+        audience: <orchestration-audience>
+        secret:
+          existingSecret: oidc-credentials
+          existingSecretKey: orchestration-client-secret
```
ThorbenLindhauer: This is redundant with the section in `orchestration`. Is there a particular reason why you have it in both places? Otherwise I suggest in line with our ambition to limit the use of global values to only have the `orchestration` section. That worked for me when I wrote the Entra/external Keycloak guides.

docs/self-managed/deployment/helm/configure/authentication-and-authorization/generic-oidc-provider.md
```diff
@@ -1,146 +1,784 @@
...
+      oidc:
+        clientId: <orchestration-client-id>
+        audience: <orchestration-audience>
+        secret:
+          existingSecret: oidc-credentials
+          existingSecretKey: orchestration-client-secret
+```
+
+:::info Connectors shares credentials
+Connectors typically uses the same OIDC client as the Orchestration Cluster, as it needs to authenticate to Zeebe Gateway. This is why both reference the same `orchestration-client-secret`.
```
ThorbenLindhauer: The reason why we are configuring it like this is that the OC API needs to accept the client's audience, which it will do by default if they use the same client. Maybe we can make that a bit more explicit (so that users can judge if they want to set up another client for this)?

Also I suggest to not use "Zeebe Gateway" here since with the Orchestration Cluster it's all one component, the Orchestration Cluster (I know the pod is still called Zeebe, but that has technical reasons).

docs/self-managed/deployment/helm/configure/authentication-and-authorization/generic-oidc-provider.md
```diff
@@ -1,146 +1,784 @@
...
+- [ ] Verify tokens contain `preferred_username` and `client_id` claims, or uncomment and configure alternative claim names
+
+## Connect to the cluster
+
+After deploying Camunda with this configuration, use the following `kubectl port-forward` commands to access the APIs and UIs:
 
-<h3>Additional considerations</h3>
+```bash
+# Management Identity
+kubectl port-forward svc/camunda-identity 8084:80 -n camunda
```
ThorbenLindhauer: In the other guides these commands are without the namespace. Let's make it consistent.

docs/self-managed/deployment/helm/configure/authentication-and-authorization/generic-oidc-provider.md
```diff
@@ -1,146 +1,784 @@
...
+
+# Orchestration Cluster
+orchestration:
+  enabled: true
+  security:
+    authentication:
+      method: oidc
+      oidc:
+        clientId: orchestration
+        audience: orchestration
```
ThorbenLindhauer: Should this (and other values in this example) use the placeholders instead? I think otherwise it might confuse people and it's inconsistent with the other OIDC guides.

ThorbenLindhauer (changes requested):
Thanks for preparing this guide. I like it a lot, especially the clear assistance and instructions for people how to set the parameters correctly. I have a few suggestions, please see below.

I also ran the guide locally against a Keycloak that I deployed to the same Kubernetes namespace, i.e. the case that browser and cluster URL for the IdP are different (with the 8.9 Helm Chart from current main, so 8.9.0-alpha1 images). The cluster comes up successfully (all components healthy) and I am able to log into Web Modeler (and I guess also Console), and Connectors seems also able to access the OC REST API. 

But I cannot log into any of the applications that use confidential clients. When the application UI redirects to Keycloak, it uses the cluster-internal URL which is then of course not reachable. And I think the rootcause for that is in Management Identity, particularly [here](https://github.com/camunda/identity/blob/d54110dc3b253a9563e4e0a8c44330198fe9941e/sdk-java/src/main/java/io/camunda/identity/sdk/impl/generic/GenericAuthentication.java#L54-L61). To determine the authorization URL, Management Identity in GENERIC mode always uses the well know configuration endpoint's response (which then only contains cluster-internal URLs).

I guess you tested with an Ingress so didn't run into this issue. Could you please check with an Identity engineer if this understanding of the problem is correct? And if yes, we should decide which way we want to take this problem:

1. Document that the generic setup doesn't support different browser/cluster IdP URLs (and clean up all the conditional explanations for that in the guide)
2. Schedule a fix in Management Identity

I believe to make progress here we should go with option 1 then.

edit: probably obvious, but I only commented on the 8.9 version of the guide; the feedback applies to the 8.8 version, too.

docs/self-managed/deployment/helm/configure/authentication-and-authorization/generic-oidc-provider.md
```diff
@@ -1,146 +1,784 @@
...
+Before you begin, ensure you have:
 
-:::note
-The steps below are a general approach for the Camunda components; it is important you reference the [component-specific
-configuration](#component-specific-configuration) to ensure the components are configured correctly.
+- **An OIDC-compliant provider** already deployed and accessible
+- **Administrative access** to create and configure OIDC clients in your provider
+- **Access to your provider's discovery document** to obtain endpoint URLs
+- **A Kubernetes cluster** with Helm 3.x installed
+- **kubectl** configured to access your cluster
```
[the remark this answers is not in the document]
bkenez: Adjusted 👍

docs/self-managed/deployment/helm/configure/authentication-and-authorization/generic-oidc-provider.md
bkenez: Oh I see, thanks for the explanation.
Rewriting with a better explanation and without mentioning gateway.

docs/self-managed/deployment/helm/configure/authentication-and-authorization/generic-oidc-provider.md
bkenez: True, I figured specific might be a better fit for the complete config, but don't want to diverge from the other guides that much. Adjusting.

docs/self-managed/deployment/helm/configure/authentication-and-authorization/generic-oidc-provider.md
```diff
@@ -1,146 +1,784 @@
...
+    auth:
+      enabled: true
+      type: "GENERIC"
+
+      # OIDC Provider Endpoints (from discovery document)
+      issuer: <issuer-url>
+      publicIssuerUrl: <issuer-url>
+      issuerBackendUrl: <issuer-url-or-internal-dns>
+      tokenUrl: <token-endpoint-url>
+      jwksUrl: <jwks-endpoint-url>
```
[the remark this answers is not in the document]
hamza-m-masood: > That's why in [the external Keycloak guide](https://preview.docs.camunda.cloud/pr-7362/docs/next/self-managed/deployment/helm/configure/authentication-and-authorization/external-keycloak/#full-configuration-example), I decided to not set global.identity.auth.issuer and tell people to set the individual URIs instead.

@ThorbenLindhauer what was the reasoning behind this decision? Wouldn't it make sense to tell customers to set only the issuer? It would save them from having to figure out the other values.

sidebars.js
```diff
@@ -1453,6 +1453,7 @@ module.exports = {
                         "self-managed/deployment/helm/configure/authentication-and-authorization/external-keycloak",
                         "self-managed/deployment/helm/configure/authentication-and-authorization/microsoft-entra",
                         "self-managed/deployment/helm/configure/authentication-and-authorization/generic-oidc-provider",
+                        "self-managed/deployment/helm/configure/authentication-and-authorization/troubleshooting-oidc",
```
afgambin: This new guide should also be added to its landing page: https://docs.camunda.io/docs/next/self-managed/deployment/helm/configure/authentication-and-authorization/external-oidc-provider/

docs/self-managed/deployment/helm/configure/authentication-and-authorization/external-keycloak.md
```diff
@@ -295,3 +295,7 @@ For example:
 - Management Identity: `http://localhost:8084`
 
 Log in with the username `demo` and the password stored in the secret key `identity-firstuser-password`.
+
+## Troubleshooting
```
afgambin: I’d avoid adding this section on all related pages. Users should be able to find it from the OIDC landing page, left-hand nav bar, and via the search bar. Up to you - this isn’t a strong suggestion

docs/self-managed/deployment/helm/configure/authentication-and-authorization/generic-oidc-provider.md
```diff
@@ -1,146 +1,784 @@
...
+| Web Modeler API       | Confidential | Programmatic API access                          |
+| Web Modeler UI        | Public       | User login                                       |
+| Console               | Public       | User login                                       |
+
+**For each client, record:**
+
+- Client ID
+- Client secret (for confidential clients only)
+
+## Configure redirect URIs
```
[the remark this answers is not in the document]
afgambin: Considering this is a how-to guide, it is good practice to break down steps into separate sections that relate to each other, with the goal of making each section as independent as possible.

docs/self-managed/deployment/helm/configure/authentication-and-authorization/generic-oidc-provider.md
```diff
@@ -1,146 +1,718 @@
...
-   - Client ID
-   - Client secret
-   - Audience
-4. Set the following environment variables or Helm values for the component you are configuring an app for:
+
+### Scopes requested by Camunda
+
+Camunda components request OIDC scopes when authenticating users. The default scopes vary by component:
+
+**Management Identity, Optimize, Web Modeler, Console:**
```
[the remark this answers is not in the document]
bkenez: Good point, made the change!

afgambin (dismissed):
@bkenez I finished my review. LGTM now - nice rework!

afgambin (approved):
Re-approving with your latest changes :)

ThorbenLindhauer (approved):
Thanks for writing this guide and making all the changes from the first review. Looks really good to me now!

PR #7530, opened 2025-12-12
docs(self-managed): helm chart configurations are not real

Langleu:
## Description

Just a proposal but the currently defined Helm Chart values are not existing.
Even the ones proposed by me are not fully implemented but atleast align with the other RDBMS configuration page.

Ultimately up to the Distribution team, just want to mark it already as otherwise it may be forgotten.
I'm not a supporter of a 3rd configuration place next to components and Helm Charts but won't fight that. Someone has made that choice.

I'm also not certain that the configuration pieces for the application are 100% complete considering the Helm Chart explicitly enables the RDBMS exporter.

## PR Checklist

### DIFF
--- a/docs/self-managed/concepts/secondary-storage/configuring-secondary-storage.md
+++ b/docs/self-managed/concepts/secondary-storage/configuring-secondary-storage.md
@@ -29,27 +29,57 @@ For the latest list of supported relational databases and versions, see the
 
 <TabItem value="helm">
 
-When deploying with Helm, set the secondary storage type and connection details in your `values.yaml` file:
+When deploying with Helm, set the secondary storage type, connection details, and rdbms exporters in your `values.yaml` file:
 
 ```yaml
-data:
-  secondary-storage:
-    type: rdbms
+orchestration:
+  exporters:
+    camunda:
+      enabled: false
     rdbms:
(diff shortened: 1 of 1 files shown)


christinaausley:
From the docs side, the original draft was written intentionally at a conceptual/configuration-model level, not as a literal reflection of what the Helm chart supports today. The goal was to document the intended secondary storage model (`data.secondary-storage`, exporter enablement, etc.) consistently across Helm, Docker Compose, and manual config, though parts of the Helm implementation were still evolving for 8.9. It was intended to reflect the configuration model the applications expect, and the direction the Helm chart is moving toward, rather than locking docs to a snapshot that’s still in flux.

That said, I agree we shouldn’t imply Helm values that can’t work today. If we keep this content I think we need to:
- clearly label Helm examples as 8.9+ / in progress, or  
- tighten the Helm snippets to only what’s currently implemented and explain the model separately.

On Immi’s point about keeping pages in sync, agreed this is a broader issue. I can open an issue to decide:
- which page is the single source of truth (Helm vs application config)
- how we avoid duplicating Helm values across multiple docs pages
- whether some of this content should be referenced or generated from the Helm docs instead of copied.

Issue #5138, opened 2025-03-04
Search bar placeholder text


mesellings:
Permanent resolution to search placeholder text, instead of using CSS, for better accessibility.

See https://github.com/camunda/camunda-docs/pull/5052#pullrequestreview-2637632832 for details.

PR #7559, opened 2025-12-15
fix(search): Handle placeholder text in the component

alexronquillo:
## Description

Closes: #5138

**(Decision needed)** This solution is trivial and solves the accessibility problem described in the relevant issue. However, it has one side effect: It changes not only the text of the search bar on the homepage, but also the one in the top nav. 

There is another solution in which we eject the component via swizzling (wrapping isn't sufficient, unfortunately) and set the text based on the context. Despite that, this is still the right approach, in my opinion, because swizzling would introduce long-term maintenance work. This is especially concerning, considering there are `TODO` comments in the code for making changes to this very component for Docusaurus v4.

I'd like input from someone else though, before we make a final decision.

## PR Checklist

- [ ] My changes are for **an upcoming minor release** and are in the `/docs` directory (version 8.9).

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -449,7 +449,11 @@ module.exports = {
       apiKey: "68db7725a8410eace68419c29385ad1e",
       appId: "6KYF3VMCXZ",
       indexName: "camunda-v2",
-      placeholder: "Search Camunda 8 docs",
+      translations: {
+        button: {
+          buttonText: "Search Camunda 8 docs",
+        },
+      },
     },
     languageTabs: [
       {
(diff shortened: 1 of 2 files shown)


christinaausley:
From a UX/content standpoint, having the same placeholder text in both the header and homepage search blurs intent a bit. Header search wants to stay minimal (“Search”), while the homepage search can afford to be more descriptive.

Given the current fix technically works but affects both contexts, the alternative (swizzling) adds real long-term maintenance cost, and we already expect homepage search to evolve this year, I’m 👍 on not shipping this change for now and revisiting it when we’re ready to tackle homepage search as a whole.

PR #7728, opened 2026-01-13
fix: Resolve workflow permissions for quality board

mschoe:
Fixes the 'Resource not accessible by integration' error by:
- Replacing GitHub App token with GITHUB_TOKEN
- Adding explicit `issues: write` permission to the workflow.


mschoe:
@mesellings, I hope the usage of the GITHUB_TOKEN with the respective permission config will solve the problem. Unfortunately this is hard to test upfront. So please approve the PR and we will see if the workflow running.

mesellings (dismissed):
Pre-approved to check the workflow can run as requested by @mschoe

mesellings (approved):
Approved 🚀

PR #7732, opened 2026-01-14
chore(migration): be clearer about not manipulating c7 data 

HeleneW-dot:
## Description
related to https://github.com/camunda/camunda-7-to-8-migration-tooling/issues/863

## PR Checklist


tasso94 (changes requested):
Can we make it very specific for the respective modes, runtime, and history? I.e., not document everywhere this requirement to stop C7 but make it very clear that this is problematic for history migration.

tasso94 (changes requested):
Can we add some enablement on why this is the case somewhere to help the user understand the caveats? I think the main issue is for running history instances that get auto-cancelled. If not auto-cancelled, this limitation doesn't exist IMO (i.e., already completed or "naturally" cancelled instances). For on C7 running instances that get auto-cancelled, when changing state between runs this ends up in corrupt C8 data. My call here would be to explain to the user, somewhere, on a more conceptual level than just limiting things. I think we should do both, though.

tasso94 (approved):
That's it! 🚀

PR #8566, opened 2026-04-13
docs(operate): document process instance modification limitation

kristinkomschow:
## Description

Document a limitation for 8.8 and 8.7 when modifying element instances that are within the same flownode instance.

relates to https://github.com/camunda/camunda/issues/37282

## PR Checklist

- [ ] My changes are for **an upcoming minor release** and are in the `/docs` directory (version 8.9).


versioned_docs/version-8.8/components/operate/userguide/process-instance-modification.md
```diff
@@ -129,6 +129,16 @@ If you click the **Apply** button from the summary modal as described [here](#vi
 
 ![applied-modifications](../../../images/operate/modifications/applied-modifications.png)
 
+## Limitations
+
+Maximum token cancelation on one flow node instance:
+
+When canceling or moving all tokens within one flow node instance (e.g. flow nodes nested within a multi-instance body), a maximum of 500 tokens will be canceled due to technical limitations in this version.
```
alexronquillo: > e.g. flow nodes nested within a multi-instance body

I'm a little confused by the wording here. For clarification, is this actually an example, or is this the only case in which the limit applies? I suggest we go with one of the following:

If it _only_ applies to "flow nodes nested within a multi-instance body":

> When canceling or moving all tokens within one flow node instance—for flow nodes nested within a multi-instance body—a maximum of 500 tokens will be canceled due to technical limitations in this version.

If it applies to "flow nodes nested within a multi-instance body", but also other cases:

> When canceling or moving all tokens within one flow node instance—including cases where the flow node is nested within a multi-instance body—a maximum of 500 tokens will be canceled due to technical limitations in this version.

versioned_docs/version-8.8/components/operate/userguide/process-instance-modification.md
```diff
@@ -129,6 +129,16 @@ If you click the **Apply** button from the summary modal as described [here](#vi
 
 ![applied-modifications](../../../images/operate/modifications/applied-modifications.png)
 
+## Limitations
+
+Maximum token cancelation on one flow node instance:
```
alexronquillo: Here, we only mention "cancellation". Later, we also include "moving" the tokens. Maybe we should just say "modification"?

Also, since this is the only listed limitation, should remove this line and renamethis section "Maximum token modification limit" or something like that?

versioned_docs/version-8.8/components/operate/userguide/process-instance-modification.md
alexronquillo: > When **canceling or moving** all tokens within one flow node instance... a maximum of **500 tokens will be canceled** due to technical limitations in this version.

Is there only a limit on the cancellations or is the limit applicable for moving tokens, as well? 

If it applies to cancel and move operations, maybe we can say:

> When **canceling or moving** all tokens within one flow node instance... a maximum of **500 tokens will be affected** due to technical limitations in this version.

Should we also include a note on the prioritization (which 500 modifications are applied)?

versioned_docs/version-8.8/components/operate/userguide/process-instance-modification.md
```diff
@@ -129,6 +129,16 @@ If you click the **Apply** button from the summary modal as described [here](#vi
...
+
+When canceling or moving all tokens within one flow node instance (e.g. flow nodes nested within a multi-instance body), a maximum of 500 tokens will be canceled due to technical limitations in this version.
+
+In this example, canceling or moving all instances would result in only 500 instances being canceled or moved:
```
alexronquillo, line 136: Is it safe for the user to run the same process instance modification multiple times in a row to get through all 2000 tokens? Maybe we should give a recommended course of action after the image?

versioned_docs/version-8.7/components/operate/userguide/process-instance-modification.md
```diff
@@ -129,6 +129,16 @@ If you click the **Apply** button from the summary modal as described [here](#vi
 
 ![applied-modifications](../../../images/operate/modifications/applied-modifications.png)
 
+## Limitations
+
+Maximum token cancelation on one flow node instance:
+
+When canceling or moving all tokens within one flow node instance (e.g. flow nodes nested within a multi-instance body), a maximum of 500 tokens will be canceled due to technical limitations in this version.
+
+In the this example, canceling or moving all instances would result in only 500 instances being canceled or moved:
```
tmetzke:
```suggestion
In this example, canceling or moving all instances would result in only 500 instances being canceled or moved:
```

alexronquillo (changes requested):
Thanks for the contribution! I've left a couple of clarification questions. Please let me know when it's ready for another review.

versioned_docs/version-8.8/components/operate/userguide/process-instance-modification.md
tmetzke: 💭 We can also let people know that repeated invocations are possible, e.g., to cancel all tokens in batches of 500 each.

kristinkomschow: I double-checked this and it should only affect multi-instance bodies because they are the only element type that shares its flowNodeId with another element (since all contained activity instances all get the same flowNodeId as the body wrapper), so I will clarify that.

versioned_docs/version-8.8/components/operate/userguide/process-instance-modification.md
kristinkomschow: > Here, we only mention "cancellation". Later, we also include "moving" the tokens. Maybe we should just say "modification"?

It doesn't apply to all modifications (adding tokens is fine), just those that require cancellation. "Moving" on the technical side is a cancellation + addition in one command, which is why it is also affected.

versioned_docs/version-8.8/components/operate/userguide/process-instance-modification.md
kristinkomschow: It also applies to moving (as mentioned in the other comment, "moving" is a "cancelation" + "addition").

There isn't any specific prioritization for which elements get canceled/moved. It's basically just the first 500 keys that are found in secondary storage by Operate at the time of execution of the Operation. We could add a note that makes this clear and guidance on how to cancel specific instances instead.

versioned_docs/version-8.8/components/operate/userguide/process-instance-modification.md
kristinkomschow: Yes, and my assumption is that this will usually be done since the goal when attempting to cancel all tokens is to have all tokens canceled in the end 😄 
I'll add this explicitly to the documentation so there is no uncertainty around this.

versioned_docs/version-8.8/components/operate/userguide/process-instance-modification.md
kristinkomschow: @alexronquillo remarked the same, so I'll definitely do that. It seemed obvious from my POV so I didn't think about including it, but it may not be clear to users 😄

tmetzke: There's usually no such thing as "obvious" when users are involved 😅

tmetzke (approved):
👍 Sounds good to me, let's see what @alexronquillo thinks here 🙂

kristinkomschow:
@alexronquillo small nudge since I'm still waiting on docs team approval here

christinaausley (approved):

PR #8569, opened 2026-04-13
fix(modeler): correct replace-via-upload versioning note

Skaiir:
## Description

Since 8.8 (camunda/camunda-hub#17787), replace via upload no longer auto-creates a version of the previous content. Instead, an in-app confirmation modal warns the user that the existing content will be replaced and cannot be restored. The docs still claimed a new version would be created, which misled a customer migrating from 8.7 and was reported externally.

This PR:

- **Evergreen docs (8.9):** removes the stale admonition from `docs/components/modeler/web-modeler/modeling/versions.md`. The in-app confirmation modal is sufficient signal for the destructive action, so the docs don't need to double up.
- **8.8 snapshot:** replaces the admonition in `versioned_docs/version-8.8/components/modeler/web-modeler/versions.md` with a caution that explicitly calls out the behavior change. This lives only in the 8.8 snapshot because that's where migrating customers land, and the reverse-knowledge has a shelf life tied to the 8.8 window.

Context on the behavior change and why the feature was removed: internal discussion in \`#team-hub\` (Sep 4, 2025). Re-enabling auto-versioning on upload is being tracked toward 8.10 via \`prj-pdp-3567-decoupled-element-template-lifecycle--governance\` and \`prj-pdp-3175-internal-process-application-versioning-fix\`, but not mentioned here to avoid committing the docs to a roadmap promise.

Refs camunda/camunda-hub#23103

- [x] This is a bug fix, security concern, or something that needs **urgent release support**.

## PR Checklist

- [x] The commit history of this PR is cleaned up, using \`{type}(scope): {description}\` commit message(s)
- [x] My changes are for **an upcoming minor release** and are in the \`/docs\` directory (version 8.9).
- [x] My changes are for an **already released minor** and are in a \`/versioned_docs\` directory.
  - [x] Engineering team review (@wollefitz as DRI for the behavior change)
  - [x] Technical writer review via \`@camunda/tech-writers\`


christinaausley (approved):
Removed this from 8.9 as well 👍

wollefitz (approved):

PR #8585, opened 2026-04-14
docs(build-with-camunda): add manual c8run download option to get-started section

meyerdan:
## Summary

- Adds a collapsible "Download c8run" section below the CLI-based get-started flow for users who don't have npm
- Includes direct download buttons for macOS (Apple Silicon), macOS (Intel), Linux, and Windows (c8run 8.9)
- Notes the OpenJDK 21–25 requirement and links to the full installation guide

## Test plan

- [ ] Verify the collapsible section renders correctly on the Build with Camunda page
- [ ] Confirm all four download links resolve to the correct archives
- [ ] Check the installation guide link points to the correct page


bojtospeter (dismissed):
As we are trying to sunset the bundle, it's OK to have one download for c8run and one for desktop modeler. In case we are keeping the bundle in the future for some reason, we can merge the two collapable download boxes.

mesellings:
Took a quick look, few things to note here (FYI @afgambin):
- I don't like the Web UX of having the two collapsible options just one on top of the other, I know as @bojtospeter said one might disappear, but until then I think we can present it better - however off top of my head not sure what that would be?
- We have links going to the /next docs- now that 8.9 docs are out, we can change all these to point to the normal /docs structure moving forward - then we update the page to point to each minor version (currently 8.9).

afgambin:
- I have updated the links to point to 8.9.
- I have renamed c8run to Camunda 8 Run, which I think reads better in normal text and matches the official product name.
- I played around with the collapsible menus, but I can’t currently find an option that looks better. Anyhow, we’re updating things quite often, so this seems like a good solution for now.

Comment:

We’re currently pointing users to download Desktop Modeler version `5.45`. `5.46` is already available. Do we want to update this?

afgambin (approved):

bojtospeter (approved):
lgtm 🔥

PR #8666, opened 2026-04-21
docs(agentic-orchestration): add new RAG guide

afgambin:
## Description

This PR adds a new guide on how to add long-term memory to AI agents.

## PR Checklist


docs/components/agentic-orchestration/rag-ai-agents.md
```diff
@@ -0,0 +1,120 @@
...
+4. Configure the [**Embedding model**](/components/connectors/out-of-the-box-connectors/embeddings-vector-db.md#embedding-models) with your provider credentials.
+5. Configure the [**Vector store**](/components/connectors/out-of-the-box-connectors/embeddings-vector-db.md#vector-stores) with your database connection details and **index name**.
+6. Set the output **Result variable** to `toolCallResult`.
+7. Give the task a clear **Name** and write a descriptive **Element documentation** to help the LLM understand when to use this tool. The element documentation is passed to the LLM as the [tool description](/components/connectors/out-of-the-box-connectors/agentic-ai-aiagent-tool-definitions.md#tool-definitions).
+
+:::note
+The index name identifies the collection of documents the agent searches. You can use different indexes for different knowledge domains and, if needed, make the index name dynamic using `fromAi()` so the agent selects the appropriate knowledge base for each query.
```
christinaausley: Have we verified making the index name dynamic via `fromAi()` is a supported and recommended configuration for this connector field? If so, should we add guardrails here because dynamic index selection can affect tenant isolation/permissions/retrieval predictability?

docs/components/agentic-orchestration/rag-ai-agents.md
```diff
@@ -0,0 +1,120 @@
...
+:::
+
+### Handle missing or empty results
+
+Add an [**error boundary event**](/components/modeler/bpmn/call-activities/call-activities.md#boundary-events) to the Vector Database connector task to handle cases where the index does not exist or the query returns no results:
```
christinaausley: This appears to conflate two different cases. The Vector Database connector docs say retrieval raises an error when the index is absent, but retrieval otherwise returns an array of chunks. An empty result set doesn't appear to be documented as an error condition, so a boundary error event may not handle the "no results" case, right? We could consider splitting this into "index missing" versus "zero matches returned," and document how the process should detect an empty `chunks` array?

docs/components/agentic-orchestration/rag-ai-agents.md
```diff
@@ -0,0 +1,120 @@
...
+
+- Set the error code to match an `index not found` error.
```
christinaausley: We may need a concrete reference...I couldn't find the exact documented error code or error expression for the "index not found" case in the connector docs, so we could link to the exact error handling contract or replace this with the code/expression users must configure?

docs/components/agentic-orchestration/rag-ai-agents.md
```diff
@@ -0,0 +1,120 @@
...
+- In the boundary event output, set `toolCallResult` to a value such as `"Nothing was found"` so the agent receives a clear, graceful response instead of an unhandled error.
+
+This prevents process failures when the knowledge base is empty or when an index has not yet been created.
```
christinaausley: If the comment above is correct and empty search results are not raised as connector errors, this sentence could come across as too broad. The boundary-event pattern likely prevents failures only for missing-index or other connector-error cases, not for a valid retrieval that returns zero matches, right? I'm not certain here.

docs/components/agentic-orchestration/rag-ai-agents.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Populate the vector database
+
+Knowledge can be loaded into the vector database in two ways:
+
+- **Batch import**: Documents are embedded and stored before the agent starts processing, typically as part of a data preparation process. Use a Vector Database connector task configured with **Operation: Embed document** in a separate BPMN process or script.
+- **Runtime ingestion**: New knowledge is added to the vector database as the agent encounters it. For example, when a human provides an answer that did not previously exist in the database.
```
christinaausley, line 174: We could call out update/idempotency behavior here. The Vector Database connector docs say re-embedding the same document creates duplicate chunks unless existing chunks are deleted first. That seems important for both batch import and runtime ingestion, especially if users plan to re-store corrected answers or refreshed source documents over time?

christinaausley (commented):
Zero grammatical/wording suggestions 🥳 Just a few non-blocking comments as food for thought.

docs/components/agentic-orchestration/rag-ai-agents.md
afgambin: I believe this is natively supported in the AI Agent Connector and the fromAi() function, for example: 

```
indexName: fromAi("indexName", "The name of the knowledge base index to search, e.g. 'support-kb' or 'product-docs'")
query:     fromAi("query", "The user's search query")
```
Can you confirm @maff ?

docs/components/connectors/out-of-the-box-connectors/embeddings-vector-db.md
```diff
@@ -8,19 +8,19 @@ description: Embed, store, and retrieve LLM embeddings and store them in vector
 import Tabs from "@theme/Tabs";
 import TabItem from "@theme/TabItem";
 
-The **vector database connector** allows embedding, storing, and retrieving Large Language Model (LLM) embeddings. This enables building AI-based solutions for your organizations, such as context document search, long-term LLM memory, and agentic AI interaction.
```
maff, line 11: @afgambin should we link back to the guide from this page?

docs/components/agentic-orchestration/rag-ai-agents.md
```diff
@@ -0,0 +1,120 @@
...
+
+The LLM is responsible for generating natural language queries when retrieving, and for deciding what content is worth storing. The actual vector operations (encoding, indexing, and searching) are handled by the connector and the underlying vector store.
+
+To configure either operation, you need:
+
+- A supported [vector store](/components/connectors/out-of-the-box-connectors/embeddings-vector-db.md#vector-stores) and connection credentials.
+- A supported [embedding model](/components/connectors/out-of-the-box-connectors/embeddings-vector-db.md#embedding-models) and its provider credentials.
+- An index name that identifies the collection to read from or write to.
+
+## Add a vector database query tool
```
maff: An additional pattern could be to model the Vector DB retrieval before entering the AI Agent and passing its result as user message, either by concatenating the text result or by adding the documents returned by the retrieval into the user message documents. This would retrieve initial context to the user query before the agent is executed.

Maybe worth adding this as an option?

docs/components/agentic-orchestration/rag-ai-agents.md
maff: From a technical standpoint this would work, but I agree with @christinaausley that this could be dangerous if the AI has free control over it. Maybe we document this in a different way (e.g. dynamic index on some kind of user/account ID variable existing in the process instead of generated via `fromAi`)?

docs/components/agentic-orchestration/rag-ai-agents.md
afgambin: Thanks for the input! I have reworked the section, I hope now it is clearer

docs/components/agentic-orchestration/rag-ai-agents.md
afgambin: Thanks for the input! I have reworked the section, I hope now it is clearer

docs/components/agentic-orchestration/rag-ai-agents.md
```diff
@@ -0,0 +1,120 @@
...
+2. Set **Search query** using the [`fromAi()`](/components/connectors/out-of-the-box-connectors/agentic-ai-aiagent-tool-definitions.md#ai-generated-parameters-via-fromai) function so the LLM generates the query dynamically at runtime:
+
+```feel
+fromAi(toolCall.query, "The query you're making to the vector database.")
+```
+
+3. Set **Max results** to control how many matching documents are returned (the default is five). If the agent doesn't find what it needs on the first attempt, it can re-invoke the tool with a different query.
+4. Configure the [**Embedding model**](/components/connectors/out-of-the-box-connectors/embeddings-vector-db.md#embedding-models) with your provider credentials.
+5. Configure the [**Vector store**](/components/connectors/out-of-the-box-connectors/embeddings-vector-db.md#vector-stores) with your database connection details and **index name**.
+6. Set the output **Result variable** to `toolCallResult`.
```
maff: Currently this returns the list of results as it is returned by the Vector DB connector which contains both text content and documents (containing the same text content), however the AI Agent can't currently properly [deal with the documents](https://github.com/camunda/connectors/issues/7005). Long-term:

- the AI Agent should handle documents properly
- the Vector DB should return either plain text OR documents, not both as it is redundant

The guide as it is now would work, but would return too much data to the AI Agent. Do you think it would introduce too much complexity here it we'd have a result expression filtering only the relevant fields from the results array (e.g. dropping the document)?

Something like this (untested - the expression should work, but `response.chunks` might be to adapt):

```
for chunk in response.chunks return {
    chunkId: chunk.chunkId,
    score: chunk.score,
    content: chunk.content
}
```

docs/components/agentic-orchestration/rag-ai-agents.md
maff: I think this primarily comes from the fact that we use dynamic index names as a "hack" to isolate content - so running into an "index doesn't exist error" is a common case. Isolation could also be done in different ways, such as using metadata stored with the documents to filter retrieved data, but this currently not supported by the Vector DB connector implementation.

For this guide, maybe we make this less prominent, but maybe add a small section on "isolation" describing the dynamic index solution with detail how to handle the non-existing index error?

@mathieu-stennier IIRC we descoped metadata for the first Vector DB implementation, but maybe we should look into it again?

docs/components/agentic-orchestration/rag-ai-agents.md
```diff
@@ -0,0 +1,138 @@
...
+
+- **Batch import**: Documents are embedded and stored before the agent starts processing, typically as part of a data preparation process. Use a Vector Database connector task configured with **Operation: Embed document** in a separate BPMN process or script.
+- **Runtime ingestion**: New knowledge is added to the vector database as the agent encounters it. For example, when a human provides an answer that did not previously exist in the database.
+
+For both approaches, configure the embed task as follows:
+
+1. Set **Operation** to **Embed document**.
+1. Set **Document source** to **Plain text** (or another supported source type).
+1. Provide the text to embed. This can be a process variable, a form output, or any string value.
+1. Configure the same [**Embedding model**](/components/connectors/out-of-the-box-connectors/embeddings-vector-db.md#embedding-models) and [**Vector store**](/components/connectors/out-of-the-box-connectors/embeddings-vector-db.md#vector-stores) settings used by the [retrieval tool](#add-a-vector-database-query-tool) so both operations target the same index.
```
maff: Should we also mention that it needs to have the same model configuration + dimensions as the retrieval part?

docs/components/agentic-orchestration/rag-ai-agents.md
```diff
@@ -0,0 +1,138 @@
...
+
+1. After the content to be stored is available as a process variable, add a user task with a [form](/components/modeler/forms/camunda-forms-reference.md) that displays the content and includes a checkbox or approval field.
+1. Add an [exclusive gateway](/components/modeler/bpmn/exclusive-gateways/exclusive-gateways.md) after the user task with two outgoing paths:
+   - **Approved**: Routes to a vector database connector task with **Operation: Embed document**.
+   - **Rejected**: Skips the embed step and continues the process.
+1. Use the reviewer's output variable as the gateway condition.
+
+This pattern ensures a human remains in the loop for all knowledge base updates, keeping the agent's memory accurate and trustworthy over time.
+
+## Human escalation as a memory source
```
maff: For these modeling examples, maybe an example process (or a screenshot) could help visualizing it and making it less abstract?

docs/components/agentic-orchestration/rag-ai-agents.md
afgambin: I’ll remove the tip then, if this isn’t a recommended pattern. Thanks both!

docs/components/agentic-orchestration/rag-ai-agents.md
afgambin: Added an admonition - great point

docs/components/agentic-orchestration/rag-ai-agents.md
mathieu-stennier: @maff Yes, we could 👍🏽

docs/components/agentic-orchestration/rag-ai-agents.md
afgambin: Added!

docs/components/agentic-orchestration/rag-ai-agents.md
afgambin: Added

docs/components/agentic-orchestration/rag-ai-agents.md
afgambin: I agree, now that the guide is becoming more complex, I think it needs some visuals. I’ll add those.

docs/components/agentic-orchestration/rag-ai-agents.md
afgambin: Added a subsection on this!

maff (dismissed):
LGTM, great content!

docs/components/agentic-orchestration/rag-ai-agents.md
afgambin: Since this somehow surfaces a weakness in the functionality, let's not document it for now. If, afterwards, we get user feedback around this, then we document it. Hopefully, meanwhile we tackle the long-term solution you mention.

afgambin:
@maff I addressed your feedback - thanks as always for the concrete input! I added a section around isolation and dynamic index names, and another one about prefetching context before the agent, as suggested. I see you already approved it, so probably you already saw it, but just double-checking, in case you want to have a final look. Thanks!

I'll backport it to 8.9. Can we also backport it to 8.8? I think so?

maff:
@afgambin Yes I already had a look and have seen your comments 👍 I think everything described is available in 8.8 as well, so backporting to 8.8 and 8.9 should be viable.

afgambin:
@maff @christinaausley seeking re-approval for merging it. Thanks!

maff (approved):

PR #8667, opened 2026-04-21
docs(helm): document webModeler restapi default resource increase in 8.9

hamza-m-masood:
## Summary

- Documents the ~80% increase in `webModeler.restapi` default CPU and ~25% memory increase in 8.9
- The increase is a direct consequence of `webapp` being removed and its workload absorbed by `restapi` (already documented in the same guide)
- Without this note, constrained clusters will silently get a pod stuck in `Pending` after upgrade with no obvious cause
- Adds the section directly after the existing `webapp` removal note and includes a fallback override example

## Test plan

- [ ] Verify the table and values match the actual 8.8 and 8.9 `values.yaml` defaults
- [ ] Confirm the yaml override example renders correctly


jessesimpson36 (approved):
I like notifying people of resource + limit bumps

christinaausley (approved):

christinaausley:
Feel free to merge if these are all the changes @hamza-m-masood -- wasn't sure if this was in draft because you had anything else coming 👍

PR #8668, opened 2026-04-21
Archive 8.6 step 3.3

alexronquillo:
## Description

Update links for archived version 8.6


mesellings:
@alexronquillo thanks for sorting this - looks good, but as we don't tend to touch these docs again once they've been cut into the 8.6 branch, I had a few things that stuck out (new things) that might be good to resolve now?
- The main one is the API explorer/reference docs and removing them completely (e.g. Camunda 8 REST API) - it might be too late, but see https://preview.docs.camunda.cloud/pr-8668/docs/apis-tools/operate-api/overview/#api-explorer for example - how might we still keep the API specs in the 8.6 docs for users?
- The Ask AI button - perhaps we should remove this now, as it doesn't really affect the 8.6 docs as such (liek the search bar)?
- The banner says: "This is documentation for Camunda 8 Docs 8.6, which is no longer actively maintained. For current release documentation, see [latest version](https://preview.docs.camunda.cloud/pr-8668/docs/apis-tools/operate-api/overview/) (8.6)." - should we just remove the (8.6) from the end of this sentence?

alexronquillo:
I would prefer to spend some time trying to get the API docs building, if possible, rather than removing them, but removing them is the easier option. My thought was to create a follow-up ticket to address this for a future release. Do you think it's worth spending the time to do that now? If so, should I prioritize that over the next Hub iteration?

The Ask AI button is in [8.5](https://unsupported.docs.camunda.io/8.5/), so I left it, but I didn't realize it doesn't do anything, so I'll remove it.

mesellings:
> I would prefer to spend some time trying to get the API docs building, if possible, rather than removing them, but removing them is the easier option. My thought was to create a follow-up ticket to address this for a future release. Do you think it's worth spending the time to do that now? If so, should I prioritize that over the next Hub iteration?
> 
> The Ask AI button is in [8.5](https://unsupported.docs.camunda.io/8.5/), so I left it, but I didn't realize it doesn't do anything, so I'll remove it.

Let's create a follow-up ticket for now, and get this archived out of the docs - people can access the openapi spec from https://github.com/camunda/camunda-docs/tree/main/api for now if need be.

mesellings (approved):
Lgtm! 🚀

alexronquillo:
@mesellings Please check the latest commit, which removes ask AI from the header and fixes a few weird link issues.

mesellings (approved):
This looks good @alexronquillo 👍

PR #8677, opened 2026-04-21
pushfeedback admin overview

christinaausley:
## Description

User complaint:
- The page felt too introductory and did not give readers a realistic path from "Admin overview" to a production-ready, configuration-driven setup.
- In particular, the feedback called out the gap between the overview and a "no clickops" deployment model, where teams want to manage admin bootstrap and security setup declaratively rather than through the UI.

What I changed:
- Added explicit guidance that production and no-clickops setups should prefer **Option 3: configuration**.
- Clarified that the UI and Setup API options are useful for manual bootstrap, but are not the recommended path for repeatable, Git-managed deployments.
- Added a short "what to do next" path for production readers:
  - define initial users and role assignments in configuration
  - enable API authentication and authorizations in configuration
  - use Helm application configs and the Helm authentication/authorization guides when deploying with Helm
  - continue to the external IdP guide when using an external identity provider
- The page now works better as an overview because it still introduces the options, but it also tells production users which path to follow and where to go next for a configuration-first setup.

## PR Checklist

### DIFF
--- a/docs/self-managed/components/orchestration-cluster/admin/overview.md
+++ b/docs/self-managed/components/orchestration-cluster/admin/overview.md
@@ -46,6 +46,19 @@ If users are managed within the Orchestration Cluster (that is, without an exter
 - Using the [Setup endpoint of the Orchestration Cluster REST API](#option-2-setup-rest-api)
 - Using the [configuration](#option-3-configuration)
 
+:::info Production and no-clickops setups
+For production environments, prefer **Option 3: configuration** so your admin bootstrap, role assignments, and authentication settings are defined declaratively in `application.yaml`, environment variables, or Helm-managed application configuration.
+
+The UI and Setup API options are useful for manual bootstrap, but they are not the best fit for repeatable, Git-managed deployments.
+
+Typical production next steps are:
+
+- Define initial users and role assignments with the configuration examples on this page.
+- Enable API authentication and authorizations in configuration.
+- If you deploy with Helm, provide these settings through [application configs](/self-managed/deployment/helm/configure/application-configs.md) and the [Helm authentication and authorization guides](/self-managed/deployment/helm/configure/authentication-and-authorization/index.md).
+- If you use an external Identity Provider, continue with [connect Admin to an identity provider](./connect-external-identity-provider.md).
+  :::
+
 :::warning
 After completing the initial setup, ensure at least one user remains assigned to the `admin` role.  
 If no admin user exists, a third party could create a new admin account and gain full access.
@@ -77,6 +90,8 @@ with the following JSON request body:
 ```
 
 This endpoint is only available if **no user is assigned to the `admin` role**.
+
+This option is convenient for scripted bootstrap, but for long-lived production environments we recommend keeping the desired state in configuration.
 
 ### Option 3: Define initial users via the configuration{#option-3-configuration}
 
--- a/docs/self-managed/components/orchestration-cluster/admin/overview.md
+++ b/docs/self-managed/components/orchestration-cluster/admin/overview.md
@@ -42,6 +42,8 @@ To modify the [initial configuration](/self-managed/components/orchestration-clu
 
 If users are managed within the Orchestration Cluster (that is, without an external Identity Provider), you can create an initial user in three ways:
 
+If you use an external Identity Provider instead, continue with [connect Admin to an identity provider](./connect-external-identity-provider.md).
+
 - Using the [Orchestration Cluster UI](#option-1-orchestration-cluster-ui)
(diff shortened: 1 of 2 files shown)


christinaausley:
@ThorbenLindhauer Feel free to re-assign a reviewer if you are not the proper delegate for something like this.

docs/self-managed/components/orchestration-cluster/admin/overview.md
```diff
@@ -46,6 +46,19 @@ If users are managed within the Orchestration Cluster (that is, without an exter
...
 - Using the [configuration](#option-3-configuration)
 
+:::info Production and no-clickops setups
+For production environments, prefer **Option 3: configuration** so your admin bootstrap, role assignments, and authentication settings are defined declaratively in `application.yaml`, environment variables, or Helm-managed application configuration.
+
+The UI and Setup API options are useful for manual bootstrap, but they are not the best fit for repeatable, Git-managed deployments.
+
+Typical production next steps are:
+
+- Define initial users and role assignments with the configuration examples on this page.
```
ThorbenLindhauer: Should this link to the section that explains it?

docs/self-managed/components/orchestration-cluster/admin/overview.md
```diff
@@ -46,6 +46,19 @@ If users are managed within the Orchestration Cluster (that is, without an exter
...
+- Enable API authentication and authorizations in configuration.
```
ThorbenLindhauer: If this is supposed to hint/link to https://docs.camunda.io/docs/self-managed/components/orchestration-cluster/admin/overview/#enable-api-authentication-and-authorizations, then I think it doesn't fit into these production steps: The section talks particularly about Camunda Run, which is not a distribution intended for production deployment. In other setups (like Helm Chart or plan Java application9, authentication and authorization are active by default.

docs/self-managed/components/orchestration-cluster/admin/overview.md
```diff
@@ -46,6 +46,19 @@ If users are managed within the Orchestration Cluster (that is, without an exter
...
+- If you deploy with Helm, provide these settings through [application configs](/self-managed/deployment/helm/configure/application-configs.md) and the [Helm authentication and authorization guides](/self-managed/deployment/helm/configure/authentication-and-authorization/index.md).
+- If you use an external Identity Provider, continue with [connect Admin to an identity provider](./connect-external-identity-provider.md).
```
ThorbenLindhauer: Is this point meaningful here? The section begins with "If users are managed within the Orchestration Cluster (that is, without an external Identity Provider)".

ThorbenLindhauer (changes requested):
Thanks for making this improvement. I have left some detail comments about the content.

More generally, I wonder if this is sufficient to solve the concern mentioned by the feedback or if for example a dedicated best practice guide for production-ready Admin configuration would be better (that focuses on the configuration approach and takes users through configuring a typical production set of Identity entities). I'm genuinely unsure, so don't mean to claim that this would be better. But maybe you can validate with the source of the feedback if the current change is sufficient or something more elaborate is needed?

docs/self-managed/components/orchestration-cluster/admin/overview.md
christinaausley: Good point. I updated this bullet to link directly to the relevant configuration and role-assignment sections, instead of referring to the page more generally.

docs/self-managed/components/orchestration-cluster/admin/overview.md
christinaausley: Agreed. I removed that bullet from the production steps.

The linked section is specifically about Camunda 8 Run, so it was misleading in a general production checklist/authentication and authorization are already enabled by default for other deployment types such as Helm or plain Spring Boot applications.

docs/self-managed/components/orchestration-cluster/admin/overview.md
christinaausley: Agreed. I removed that item from the production checklist and moved the external IdP guidance into the introductory scope-setting text for the section.

christinaausley:
@ThorbenLindhauer thanks for the thoughtful review!

My intent with the current change was to make the existing page more actionable for configuration-first and no-clickops setups, not necessarily to fully cover a broader production-ready Admin setup end to end.

I agree it would be useful to validate this with the source of the original feedback and confirm whether the updated page now addresses the concern, or whether there is still a gap that would justify a dedicated best-practice guide focused on production-ready Admin configuration and a typical set of Identity entities. I can create a backlog ticket to collect additional feedback, but I think generally this is a good iteration in the meantime.

ThorbenLindhauer (approved):
Thanks for making the changes, looks good to me now. And agree, we can make the changes here in any case.

PR #8789, opened 2026-05-06
docs(agentic-orchestration): AI gateway section

### LINKED ISSUE #8555: Surface gateway routing/fallback details in Camunda AO architecture
**Context**
A partner/customer is evaluating n8n and asked about Fallback Model configuration to build more resilient (and potentially sovereign) agents—e.g., when hyperscalers are cut off, customers are blocked, or providers rate-limit.

They noted that while it’s possible to model error boundaries per agent, a configuration-based approach (fallback not only by model but potentially by provider) feels more elegant.

This request appears closer to an AI gateway / routing layer capability than an orchestration-layer capability (examples mentioned: LiteLLM, OpenRouter). The interesting angle for Camunda is likely not implementing fallback routing, but ensuring that when a gateway performs routing/fallback, Camunda can still support production needs at the orchestration level:
- Visibility into which model/provider was used
- Reasons for fallback (rate limit, failure, policy, etc.)
- Ability to reason about this at the process level, not just at an API-call level
This aligns with our broader direction of production readiness / enterprise trust.

Ref thread: https://camunda.slack.com/archives/C08BHU6JQJG/p1775836954005089

**Problem statement**
Externally, this can be perceived as a “feature gap” when compared to n8n’s fallback configuration, unless we clearly communicate:

- where fallback belongs architecturally (gateway vs orchestrator), and
- what Camunda should surface/record to make gateway routing decisions observable and auditable inside workflows.

**Proposal**
- Document our approach, possibly describing the full AO stack and where Camunda focuses, e.g., the gateway owns routing/fallback, and the orchestration layer integrates and surfaces outcomes.
- This could fit within the [Design and architecture guide](https://docs.camunda.io/docs/next/components/agentic-orchestration/ao-design/).

afgambin:
## Description

Closes https://github.com/camunda/camunda-docs/issues/8555

## PR Checklist


docs/components/agentic-orchestration/design-architecture.md
```diff
@@ -52,6 +52,24 @@ Follow these principles when designing your agentic orchestration solution:
...
+
+- Routing requests to a preferred model or provider.
+- Falling back to an alternative model when the primary is unavailable or rate-limited.
+- Enforcing provider-level policies. For example, data residency or cost caps.
+
+Camunda operates at the **orchestration layer**, complementing these gateways by providing production-level visibility and control over the outcomes of routing decisions:
+
+- **Observability**: Track which model and provider handled each agent interaction through process variables and [Operate](/components/operate/operate-introduction.md). When your gateway routes to a fallback model, Camunda can capture this context, including the model used, the provider, and the reason for fallback, making every decision auditable at the process level.
+- **Process-level reasoning**: Use gateway-reported metadata (for example, `model_used`, `fallback_reason`) in downstream process logic. For example, a BPMN gateway can route to a human review step when a fallback model was used, or flag the instance for compliance review.
```
alexronquillo:
```suggestion
- **Process-level reasoning**: Use gateway-reported metadata, like `model_used` and `fallback_reason`, in downstream process logic. For example, a BPMN gateway can route to a human review step when a fallback model was used, or it can flag the instance for compliance review.
```

I suggest not having two "for example" statements in a row.

docs/components/agentic-orchestration/design-architecture.md
```diff
@@ -52,6 +52,24 @@ Follow these principles when designing your agentic orchestration solution:
...
+
+In production agentic systems, model routing and fallback (switching between LLM providers when one is unavailable, rate-limited, or non-compliant) is a common need. This capability belongs to the **AI gateway layer**, not the orchestration layer. Dedicated AI gateways, such as LiteLLM or OpenRouter, handle concerns like:
+
+- Routing requests to a preferred model or provider.
+- Falling back to an alternative model when the primary is unavailable or rate-limited.
+- Enforcing provider-level policies. For example, data residency or cost caps.
+
+Camunda operates at the **orchestration layer**, complementing these gateways by providing production-level visibility and control over the outcomes of routing decisions:
+
+- **Observability**: Track which model and provider handled each agent interaction through process variables and [Operate](/components/operate/operate-introduction.md). When your gateway routes to a fallback model, Camunda can capture this context, including the model used, the provider, and the reason for fallback, making every decision auditable at the process level.
```
alexronquillo: > ...through process variables and [Operate](/components/operate/operate-introduction.md).

Should this say "in Operate"?

docs/components/agentic-orchestration/design-architecture.md
```diff
@@ -52,6 +52,24 @@ Follow these principles when designing your agentic orchestration solution:
...
+- **Process-level reasoning**: Use gateway-reported metadata (for example, `model_used`, `fallback_reason`) in downstream process logic. For example, a BPMN gateway can route to a human review step when a fallback model was used, or flag the instance for compliance review.
+- **Resilience patterns**: Combine Camunda's built-in retry and incident handling with your gateway's routing logic. If all providers fail, Camunda can escalate to a human task or trigger an alternative process path rather than silently failing.
+
+:::tip
+When connecting your AI agent to an LLM through an AI gateway, configure the gateway to return routing metadata (such as which model served the request and why). Store this metadata in process variables so you can observe and act on it in your BPMN process.
```
alexronquillo:
```suggestion
When connecting your AI agent to an LLM through an AI gateway, configure the gateway to return routing metadata, such as which model served the request and why. Store this metadata in process variables so you can observe and act on it in your BPMN process.
```

alexronquillo (commented):
Great work! Left some minor suggestions.

docs/components/agentic-orchestration/design-architecture.md
```diff
@@ -52,6 +52,24 @@ Follow these principles when designing your agentic orchestration solution:
...
+
+- Routing requests to a preferred model or provider.
+- Falling back to an alternative model when the primary is unavailable or rate-limited.
+- Enforcing provider-level policies. For example, data residency or cost caps.
+
+Camunda operates at the **orchestration layer**, complementing these gateways by providing production-level visibility and control over the outcomes of routing decisions:
+
+- **Observability**: Track which model and provider handled each agent interaction through process variables in [Operate](/components/operate/operate-introduction.md). When your gateway routes to a fallback model, Camunda can capture this context, including the model used, the provider, and the reason for fallback, making every decision auditable at the process level.
+- **Process-level reasoning**: Use gateway-reported metadata, like `model_used` and `fallback_reason`, in downstream process logic. For example, a BPMN gateway can route to a human review step when a fallback model was used, or it can flag the instance for compliance review.
```
nikonovd: 🔧 as mentioned in slack, i think we are currnelty unable to extract properties from the raw LLM response, as we strip most of the data out besides some properties that we always use.

we should align to have an issue to extend our functionality or be a bit more defensive here, at least when it's about our AI agent implementation.

Custom model calls could potentially work like that.

afgambin: @nikonovd I have updated the sentence with the conservative approach, i.e., let's not document what is not currently possible.

Is the tip box below accurate, as we also mention " configure the gateway to return routing metadata"?

nikonovd: @afgambin sorry for the delay.

From my perspective we are currently not able to pass-through any AI gateway metadata back to process variables, as we remove this content from the result mostly currently.

@maff do you see any approach currently that i'm missing?

maff: At the moment, we only expose message ID, finish reason and token usage from the returned by the LLM call, but at least adding the model name would be possible - at least OpenAI's and Anthropic's APIs return the used model name and it is available as structured field in Langchain4J's metadata.
Depending on the used gateway, the provider might be part of the model name, but that highly depends on how the gateway is set up (e.g. somebody could name models `bedrock/sonnet-4`, but also just `cheap`, `coding`, ...).

With the AI Agent, I think it needs careful consideration how this information can be consumed. I can imagine flagging a process after the AI Agent ended would somehow work by looking at the conversation history, but that would demand having access to the conversation (e.g. wouldn't work or would be more complex with processes not using variable storage).

Maybe we need something like a FEEL expression which can be used to add additional data into the AI Agent response. This could be evaluated on the AI Agent side (having the configuration and the full conversation in context) to create data like `if any of the AI messages in the conversation returned a model different to the configured one, set a fallbackModelUsed=true flag`? 🤔 

The `if all providers fail` can potentially be modeled already today if the HTTP error returned by the gateway contains this information - e.g. an error expression could be used to escalate depending on the error. The AI Agent doesn't really know about the `all providers`, so it can't decide what happened, but the user might be able to depending on the used gateway.

⏩ I'm not sure if we should add this content at the moment as it may lead to wrong assumptions

afgambin: @maff @nikonovd Ah ok, got it. Thanks for the feedback! I'll archive the PR then for now? I thought to keep the part that Camunda operates at the orchestration layer, but probably worth documenting it once we have some functionality really complementing those gateways, right?

nikonovd: Generally it could make sense to have the general statement in place in the docs. I think this is also the root issue where Mathieu asked about documenting it.

Potentially some of your content is also possible to achieve with, e.g. the OpenAI connector (not our AI Agent). But probably it's wise to leave those statements out until we support modelId / raw response propagation into the variable data.

maff: @afgambin I captured extracting additional metadata from individual requests + from the whole response in https://github.com/camunda/connectors/issues/7204. I think these would be relatively straightforward to do and could support use cases like the one you documented. WDYT?

afgambin: Awesome! SGTM - I'll put this PR on hold until that ticket is addressed. After, we reevaluate the content here and extend it, if needed. Thanks!

PR #8799, opened 2026-05-07
docs: add cancel execution listener event type

romansmirnov:
## Summary

Documents the new `cancel` event type for execution listeners that ships with Camunda 8.10. A `cancel` listener runs when a process instance is terminated and is supported only on the process element.

Related to camunda/product-hub#2768.


konstantinos-mylosis (dismissed):
LGTM 👍

christinaausley (approved):
Looks great! Could consider adding an example in a future version showing when and why you'd use a cancel listener (notifying an external system or recording audit logs etc) to help users understand practical applications, but I think this is great as-is!

PR #8851, opened 2026-05-15
Fix migration manual for Camunda 8.9

tobiasschaefer:
This PR fixes two things:
1. The link to "Camunda Spring Boot Starter" was mistakenly going to "Camunda Process Test (CPT)"
2. Most of the linkes were hard-coded linked to the documentation of Version 8.8. I replaced those with relative links.

Please also merge this to to 8.9 docs (only fixing "next" here in the PR). 8.8 and older are not affected.

This PR is ready for review

@christinaausley : I cannot set the assignee - therefore I'm tagging you here.

### DIFF
--- a/docs/apis-tools/migration-manuals/migrate-to-89.md
+++ b/docs/apis-tools/migration-manuals/migrate-to-89.md
@@ -45,11 +45,11 @@ If you already performed these migrations during your 8.8 upgrade, proceed to [C
 
 | 8.9 status                                                 | Component/Use                                                                                                           | Migrate to                  | Migrate by          |
 | :--------------------------------------------------------- | :---------------------------------------------------------------------------------------------------------------------- | :-------------------------- | :------------------ |
-| <span className="label-highlight yellow">Deprecated</span> | [V1 component APIs](../migration-manuals/migrate-to-camunda-api.md)                                                     | Orchestration Cluster API   | Before Camunda 8.10 |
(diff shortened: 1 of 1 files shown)


christinaausley:
@mesellings Can you remind me if we intentionally made these links to the 8.8 version of docs for a specific purpose? Should this also be completely removed from the `next` docs? Currently it looks like this:
Screenshot 2026-05-19 at 8 38 17 AM

mesellings:
> @mesellings Can you remind me if we intentionally made these links to the 8.8 version of docs for a specific purpose? Should this also be completely removed from the `next` docs? Currently it looks like this: Screenshot 2026-05-19 at 8 38 17 AM

Hmmm, just trying to remember - I know that we had a decision to make about whether to keep these guides in the 8.9 docs as well or just point users to the 8.8 docs and leave them out of 8.9 altogether. 

We decided on the former option, but looks like these links were not updated or were changed erroneously after the decision perhaps.

I would:
- Update the links for the 8.9 docs to point to the 8.9 guides and not the 8.8 versions of them, as we kept the guides in there.
- For 8.10, we will probably have the same issue, as I think (just a quick scan, could be wrong) that some users might only perform the actions in these guides at the point of upgrading to 8.10 - i.e. they might be on 8.9 but actually still not have performed these upgrades. It might be we end up keeping them in 8.10 just as we did in 8.9, and then only remove them in 8.11.

christinaausley (approved):

PR #8860, opened 2026-05-15
docs: rewrite Gateway API setup guide with scenario-based guidance

hisImminence:
## Summary

The previous Gateway API setup page was a thin parameter table with no guidance on when or why to use each setting. This PR replaces it with:

- Four deployment scenarios with example configs: same namespace (default), shared cross-namespace Gateway, pre-existing Gateway, and fully external/manual mode
- A clearer explanation of `global.gateway.namespace` vs `global.gateway.controllerNamespace` — two parameters users frequently will conflate
- Updated parameter descriptions that explain what each setting actually controls
- Guidance for the cross-namespace scenario on what the cluster admin needs to configure on the Gateway side (`allowedRoutes.namespaces`)
- Reorganized TLS and NGINX ProxySettingsPolicy sections

Related to camunda/camunda-platform-helm#6167.

## Test plan

- [ ] Review scenario descriptions for technical accuracy against the Helm chart templates
- [ ] Check all internal links resolve
- [ ] Preview renders correctly in Docusaurus


Ian-wang-liyang (approved):
LGTM

PR #8874, opened 2026-05-19
docs(build): align MCP and Skills card heights

MaxTru:
## Summary
- Pad the MCP terminal block on the Build with Camunda page with trailing blank lines so its card matches the height of the taller Skills card.
- See:

Bildschirmfoto 2026-05-19 um 12 59 34

## Test plan
- [x] `npm run start`, open `/build-with-camunda`, and verify the "Connect to Camunda via MCP" and "Add Camunda skills as Claude plugin" cards line up at the bottom.
- [x] Spot-check at narrow viewport widths where the Skills card lines wrap.

### DIFF
--- a/src/pages/build-with-camunda.js
+++ b/src/pages/build-with-camunda.js
@@ -967,7 +967,10 @@ Available skills:
       "url": "https://camunda-docs.mcp.kapa.ai"
     }
   }
-}`}
+}
+
+
+`}
               </TerminalWindow>
             </div>
           </div>
--- a/src/pages/build-with-camunda.js
+++ b/src/pages/build-with-camunda.js
@@ -835,9 +835,7 @@ $ c8ctl run rocket-launch.bpmn --variables='{"fuelLevel":90}'
               <h4>Manage your clusters</h4>
               <TerminalWindow title="Terminal">
                 {`$ c8ctl cluster start 8.9.0-alpha5
-$ c8ctl cluster stop
-
-`}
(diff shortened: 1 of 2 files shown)


mesellings:
@MaxTru I'll look at doing this via css as that is typically a better way of achieving matching heights for blocks on a web  page rather than hardcoding it in this way

MaxTru:
> @MaxTru I'll look at doing this via css as that is typically a better way of achieving matching heights for blocks on a web page rather than hardcoding it in this way

Thanks. 

Please note that I think we used the blank-line approach in one of the upper boxes as well - see https://github.com/camunda/camunda-docs/blob/main/src/pages/build-with-camunda.js#L839 - so should be fixed there as well

mesellings (approved):
@MaxTru css looks good at different screen sizes, I also added some padding in to avoid content butting up against the edge of the screen on sizes below c.1165px - lgtm!

mesellings:
> > @MaxTru I'll look at doing this via css as that is typically a better way of achieving matching heights for blocks on a web page rather than hardcoding it in this way
> 
> Thanks.
> 
> Please note that I think we used the blank-line approach in one of the upper boxes as well - see https://github.com/camunda/camunda-docs/blob/main/src/pages/build-with-camunda.js#L839 - so should be fixed there as well

I included this as well

mesellings:
@MaxTru released with https://github.com/camunda/camunda-docs/releases/tag/8.9.22

PR #8875, opened 2026-05-19
docs(self-managed): clarify reclaim policy alternatives and fix broken link

leiicamundi:
## Description

Clarifies the Persistent Volume reclaim policy guidance for Camunda 8 Self-Managed in response to a real customer escalation where a platform team could not change the cluster-wide \`StorageClass\` from \`Delete\` to \`Retain\`, and instead enforced volume preservation through alternative mechanisms.

### Changes

- **Add a note on equivalent mechanisms** to both the OpenShift pitfalls section and the production install guide. The note:
  - Lists examples of acceptable alternatives (per-PV \`persistentVolumeReclaimPolicy\` patch, Kyverno / Gatekeeper admission policies enforcing \`Retain\` on dynamically provisioned PVs).
  - Explicitly states that **snapshot- or backup-based strategies are not equivalent** to \`Retain\`, because they don't preserve the original volume binding that a Zeebe broker needs to resume from its existing partition data without manual recovery.
  - Reinforces that the responsibility for validating any alternative mechanism (upgrade + DR scenarios) lies with the operator. We document the best practice but do not sign off on every infrastructure configuration choice.

- **Fix broken relative link** in the OpenShift "Pitfalls to avoid" section: \`[deployment troubleshooting guide](self-managed/operational-guides/troubleshooting.md)\` was missing the leading \`/\`, so it resolved relative to the current page and 404'd. (Already correct in \`version-8.8\`, broken in \`main\` + \`version-8.9\`.)

### Backport

Same change applied to:
- \`docs/\` (current / next release)
- \`versioned_docs/version-8.9/\`
- \`versioned_docs/version-8.8/\`

The link fix only affects \`docs/\` and \`version-8.9/\` (8.8 was already correct).

### Context

Customer escalation — Consulting team, May 2026. Internally aligned position: we document the best practice (\`Retain\` at the StorageClass level), accept equivalent mechanisms, do not own or validate every infra configuration choice, customers must test their setup in a safe environment.

### Preview

- [/self-managed/deployment/helm/install/production/#persistent-volume-reclaim-policy](https://docs.camunda.io/docs/next/self-managed/deployment/helm/install/production/#persistent-volume-reclaim-policy)
- [/self-managed/deployment/helm/cloud-providers/openshift/redhat-openshift/#persistent-volume-reclaim-policy](https://docs.camunda.io/docs/next/self-managed/deployment/helm/cloud-providers/openshift/redhat-openshift/#persistent-volume-reclaim-policy)


alexronquillo (changes requested):
I agree with the copilot language suggestions. Other than those, this looks good to me 👍

leiicamundi:
Thanks @alexronquillo - I applied the suggested changes

alexronquillo (approved):

Issue #7927, opened 2026-02-13
Make minor-by-minor upgrade and latest-patch requirement mandatory in version compatibility docs


giorgionaps:
Strengthen the **Version compatibility checks** docs so that the required upgrade procedure is clearly documented as **mandatory**, not just a soft recommendation.

**Required upgrade procedure (for any version upgrade):**

1. **Always update to the latest patch of your current minor first.**
2. **Never skip a minor version** (perform strictly minor-by-minor upgrades).
3. **Then update to the latest patch of the target minor.**

Current docs:

- Doc: https://docs.camunda.io/docs/self-managed/components/orchestration-cluster/core-settings/concepts/version-compatibility/#recommended-operational-steps  
  Path: `docs/self-managed/components/orchestration-cluster/core-settings/concepts/version-compatibility.md`  
  - Also clarify the **Supported upgrade paths** matrix so it explicitly refers to **latest patch of the minor** for minor upgrades: https://docs.camunda.io/docs/self-managed/components/orchestration-cluster/core-settings/concepts/version-compatibility/#supported-upgrade-paths

Previous (clearer) wording we should mirror:

- Doc (8.5, no longer maintained): https://unsupported.docs.camunda.io/8.5/docs/self-managed/operational-guides/update-guide/introduction/#update-procedure  
  - This page explicitly states the three-step procedure (latest patch of current minor → latest patch of next minor → never skipping minors). Consider reusing this stronger “You must use the following procedure” phrasing for the current docs.

Reported in slack https://camunda.slack.com/archives/C026U8GBNSW/p1770928467358279

PR #8880, opened 2026-05-20
docs(self-managed): make latest-patch update mandatory in version compatibility procedure

afgambin:
## Description

Addresses user feedback that the changes described in #7927 were not fully reflected on the [Version compatibility checks](https://docs.camunda.io/docs/8.8/self-managed/components/orchestration-cluster/core-settings/concepts/version-compatibility/) page. Specifically:

- The **Required upgrade procedure** now reads as a three-step requirement (latest patch of source minor → next minor → latest patch of target minor), instead of phrasing the latest-patch step as a "best stability" recommendation.
- The **Supported upgrade paths** matrix now labels the minor upgrade row as `Minor upgrade (single step, latest patch required)` with the example `8.7.5 → 8.8.3`.
- The **Schema manager rules** matrix mirrors the same label and adds a note that the source minor must be at its latest patch before upgrading.
- The **Recommended operational steps** section no longer duplicates the now-mandatory latest-patch step — it points back to the required procedure.

Updates apply to Next (`/docs/`), `versioned_docs/version-8.8/`, and `versioned_docs/version-8.9/`.

## PR Checklist


giorgionaps (dismissed):
LGTM, thanks for clarifying the process further 👌

alexronquillo:
@afgambin @giorgionaps Not sure if you saw already (please ignore my comment if you did), but it looks like some of these changes were intentionally introduced in another PR: https://github.com/camunda/camunda-docs/pull/8311

Those appear to be addressing a separate concern: https://camunda.slack.com/archives/C026U8GBNSW/p1773766455945909. Just a heads up in case it affects your approach here.

afgambin:
@christinaausley could you have a look here? Since you were working on the [PR Alex mentioned](https://github.com/camunda/camunda-docs/pull/8880#issuecomment-4497995010).

christinaausley:
Hi @afgambin, I looked at both change sets and I think we should keep the broader upgrade guidance from the latest edit, but dial back the parts that make the general path sound stricter than intended.

I'd keep:
- the message that upgrades are sequential minor-by-minor only
- the note that customers can move through dev/test/stage/prod on the same supported minor step
- the "do not skip minors" guidance

I'd roll back/narrow:
- the wording that makes "latest patch of the current minor" a hard requirement in the general supported-upgrade-path section
- the matrix label that says "latest patch required" as the definition of the supported minor upgrade path

If there is a real schema-compatibility constraint that needs a latest-patch baseline, I'd keep that as a narrow note in the schema-specific section only, rather than as the headline rule for all upgrades. That should let us preserve the intent from the earlier feedback without making the general guidance stricter than necessary, i think?

Also happy with your preferences/guidance. WDYT?

afgambin:
@christinaausley sgtm your proposal! Could you take care of that and update the PR accordingly?

christinaausley:
@afgambin I am away on FTO the rest of this week and at a conference all of next week — if you are able to apply the changes as you’d like I can give final approval or revisit this in June once I’m caught back up after my time off?

afgambin:
I asked for clarification, and the tl;dr is that we should keep the page as it was before the tweaks introduced [here](https://github.com/camunda/camunda-docs/pull/8311). Since we currently only test `minor.latest -> minor.latest`, it is better to recommend that approach. More context [here](https://camunda.slack.com/archives/C026U8GBNSW/p1780055995492349?thread_ts=1779247834.782229&cid=C026U8GBNSW).

alexronquillo (approved):
Approving, as a decision was made on the remaining discussion topic (see @afgambin's update above). The rest looks good to me👍

PR #8910, opened 2026-05-22
First draft of hub api sync

alexronquillo:
## Description

Closes: https://github.com/camunda/camunda-hub/issues/24410

- Introduces a new workflow and generation code for the hub API
- Deprecates Web Modeler API v1
- Includes files API spec and generated docs
- Includes the first iteration of the migration guide from v1 to v2

## PR Checklist


nicpuppa:
❓ How will we handle the different versions ? As far as I know Hub doesn't follow the OC versioning. There are not stable branches in Hub, so we have to find a way to highlight the different versions of Hub compared to the OC version I guess. 

cc @jfriedenstab @ThorbenLindhauer

.github/workflows/sync-hub-rest-api-docs.yaml
```diff
@@ -0,0 +1,131 @@
...
+
+      - name: Generate token
+        id: generate_token
+        uses: actions/create-github-app-token@v3
+        with:
+          app-id: ${{ steps.secrets.outputs.APP_ID }}
+          private-key: ${{ steps.secrets.outputs.APP_PRIVATE_KEY }}
+          repositories: camunda-hub
+
+      - name: Check for stable branch
```
nicpuppa: 🔧 This doesn't make sense for Hub, stable branches are not present here

ThorbenLindhauer:
There are maintenance branches, e.g. here: https://github.com/camunda/camunda-hub/tree/self-managed/8.8

For self-managed, Hub versioning works the same as for OC. For SaaS, Hub does continuous deployment from main. So endpoints become available either at the time they get merged or in case they are guarded by a feature flag when the feature flag gets enabled. 

The SaaS case is what we should need to about I guess. Previously by only linking to the production-deployed Swagger UI that was always in sync.

nicpuppa:
Do you think we need to create 2 separate section for SaaS and SM since the versioning is different ? I'd imagine having a workflow for SaaS that it's triggered all the time a PR that touches the specification is merged, and another one for SM that can be weekly based as we release at the same pace of OC @ThorbenLindhauer

ThorbenLindhauer:
I'd like to leave that to @alexronquillo as a docs owner and expert on the docs infrastructure :)

The general idea makes sense to me. As said, I wonder how that will play together with feature flags though.

alexronquillo:
We have separated SaaS and self-managed in the past. So there's precedent for this behavior.

I have a couple of clarification questions:

- How are the feature flags implemented? Would they be represented in the spec or only in the service implementation?
- The SaaS API can be different than Self-Managed, correct? For example, SaaS can get ahead of Self-Managed when Self-Managed gets locked to a maintenance branch. If so, are there other ways to specify an API is SaaS-only or Self-Managed-only? From the original issue description, it looks like there's an idea of using `x-availability` to do this (at least partially).

nicpuppa:
> * How are the feature flags implemented? Would they be represented in the spec or only in the service implementation?

The V2 endpoint are implemented under feature flag, in the spec there is no mention of the feature flag, as it's just a configration 

> * The SaaS API can be different than Self-Managed, correct? For example, SaaS can get ahead of Self-Managed when Self-Managed gets locked to a maintenance branch. If so, are there other ways to specify an API is SaaS-only or Self-Managed-only? From the original issue description, it looks like there's an idea of using `x-availability` to do this (at least partially).

We can use `x-availability` tag to highlights endpoints that are available only either on SaaS or SM

nicpuppa:
> > * How are the feature flags implemented? Would they be represented in the spec or only in the service implementation?
> 
> The V2 endpoint are implemented under feature flag, in the spec there is no mention of the feature flag, as it's just a configration
> 
> > * The SaaS API can be different than Self-Managed, correct? For example, SaaS can get ahead of Self-Managed when Self-Managed gets locked to a maintenance branch. If so, are there other ways to specify an API is SaaS-only or Self-Managed-only? From the original issue description, it looks like there's an idea of using `x-availability` to do this (at least partially).
> 
> We can use `x-availability` tag to differentiate endpoints that are available only either on SaaS or SM

alexronquillo:
> The V2 endpoint are implemented under feature flag, in the spec there is no mention of the feature flag, as it's just a configuration

If it's only a configuration and absent from the spec, we would need to read a configuration file to know whether to render the endpoint in the docs or not?

alexronquillo:
@nicpuppa @ThorbenLindhauer I have pushed commits to edit the workflow. Now, we sync the saas (always from main) and sm (main or self-managed/<version>) separately. These are presented in the documentation the same way we do it for the admin API. There is still an open question of how we handle feature flags for the saas release, but I'm not sure if that should block this getting merged before the alpha2 release. Let me know what you think!

ThorbenLindhauer:
@alexronquillo Yes, that's fine with me, we can solve that as a follow-up. Could we nevertheless add a hint to the SaaS API main page that this is not exposed yet? To give people at least some chance to understand this if they read the 8.10 docs.

alexronquillo:
@ThorbenLindhauer Added 👍 https://preview.docs.camunda.cloud/pr-8910/docs/next/apis-tools/hub-api-saas/overview/

ThorbenLindhauer:
Thanks, Alex. And for Self-Managed it sits behind a feature flag that people can activate if they'd like to try it. So we could mention that too. Sorry for the bit-by-bit info, only realized this after I commented yesterday.

Claude should be able to tell you what the application config property is (and what the env var conversion is if you'd like to add that).

edit: or rather: please double-check if my statement about the feature flag default in self-managed is true, not 100% certain if we have different default configs for SaaS and Self-Managed artifacts

alexronquillo:
@ThorbenLindhauer Yes, I found a comment in the repo that says:

> Endpoints are gated by `camunda.modeler.feature.public-api-v2-enabled`. Additionally, best-effort, they should also be gated on `camunda.modeler.feature.workspaces-enabled` so the V2 projects surface is only exposed when the workspaces vocabulary is live.

As far as I can tell the (and claude can tell me) the workspaces-enabled flag is not yet implemented alongside the API, so I've only included the `camunda.modeler.feature.public-api-v2-enabled` flag

alexronquillo:
@nicpuppa Would you mind taking another look? I've addressed your concerns and @ThorbenLindhauer's above.

nicpuppa (approved):
looks good to me

PR #8918, opened 2026-05-22
docs(howtos): add AI-ready content guidelines

afgambin:
## Description

- Adds [`howtos/ai-ready-guidelines.md`](howtos/ai-ready-guidelines.md): a new contributor guide for writing content that's reliably consumable by AI assistants and RAG pipelines.
- Adds a reference to the new guide from [`.github/instructions/content.instructions.md`](.github/instructions/content.instructions.md) so AI-assisted PRs follow it by default.

## Why this guide

AI assistants are now a primary audience for Camunda docs alongside human readers. The site already publishes machine-readable artifacts (`/llms.txt`, `/llms-full.txt`) and integrates with Kapa.ai via an MCP server. Those handle the **delivery** layer.

The remaining gap is at the **page level**. When agents fetch a page, they extract content by walking headings, reading the first sentences after each heading, and converting tables into structured data. Patterns that work fine for human browsing - configuration described in prose, prerequisites hidden in tabs, parameters shown only in screenshots - silently fail for agents.

This new guide addresses that gap. 

### Why a guide rather than a template or lint rule

- Templates would force every page into a shape that doesn't match reality, and Camunda docs currently do not follow any specific page-type template.
- Lint rules can come later for the high-leverage cases (e.g., connector pages must contain at least one table/follow a template).

## PR Checklist


howtos/ai-ready-guidelines.md
```diff
@@ -0,0 +1,88 @@
...
+
+- **Bad:** A heading "Process instance modification" followed by three paragraphs of history and rationale before defining what process instance modification is.
+- **Good:** A heading "Modify a process instance", followed immediately by: "With process instance modification, you can change the active state of a running process instance without redeploying the process."
+
+### 2. Make every section self-contained
+
+A section must make sense in isolation. RAG systems fetch sections, not whole pages, and agents follow links one at a time.
+
+- Don't write "as mentioned above", "see the previous section", "now that you've configured X".
+- Repeat the prerequisite or context inline. Short, slightly redundant prose beats implicit references.
```
giorgionaps: I would maybe add some examples here, as it seems to me that "short, slightly redundant" is a bit subjective.

howtos/ai-ready-guidelines.md
```diff
@@ -0,0 +1,88 @@
...
+Headings are how agents recover document structure. Vague headings break that recovery.
+
+- **Bad:** "Configuration", "Overview", "Examples", "How it works"
+- **Good:** "Configure the Slack outbound connector", "How Zeebe assigns jobs to workers", "Example: Schedule a recurring timer event"
+
+Reuse the same heading wording when the same concept appears in multiple pages so retrieval can cluster them.
+
+### 4. Keep heading hierarchy consistent
+
+- Use H1 once, for the page title (Docusaurus generates it from frontmatter `title`).
```
giorgionaps: Maybe I'm misunderstanding the guideline, but I would rephrase it slightly. Because, right now, we technically do not use an h1 heading, right? So I am not too sure if that particular bullet encourages the same approach or requires the use of an h1 heading added manually.

howtos/ai-ready-guidelines.md
afgambin: Rephrased and added an example

howtos/ai-ready-guidelines.md
afgambin: Good point! Rephrased it

howtos/ai-ready-guidelines.md
```diff
@@ -0,0 +1,88 @@
...
+- Do not skip levels.
+- Within a page, sibling headings at the same depth should describe sibling concepts.
+
+### 5. Put structured data in tables, not prose
+
+Anything that is fundamentally a list of fields, options, parameters, or values goes in a Markdown table. Prose descriptions of parameters scatter the contract across paragraphs and make field-level extraction unreliable.
+
+### 6. Critical information must appear as text, not only in screenshots or diagrams
+
+Agents can't perform optical character recognition.. If the only place a binding name, field value, or required setting appears is inside an image, that information is invisible to AI assistants.
```
christinaausley:
```suggestion
Agents can't perform optical character recognition. If the only place a binding name, field value, or required setting appears is inside an image, that information is invisible to AI assistants.
```

christinaausley: Otherwise everything looks good!

howtos/ai-ready-guidelines.md
```diff
@@ -0,0 +1,88 @@
...
+
+### 4. Keep heading hierarchy consistent
+
+- Do not add an H1 heading. Docusaurus generates the page title as the only H1 heading from the frontmatter `title`.
+- Do not skip levels.
+- Within a page, sibling headings at the same depth should describe sibling concepts.
+
+### 5. Put structured data in tables, not prose
+
+Anything that is fundamentally a list of fields, options, parameters, or values goes in a Markdown table. Prose descriptions of parameters scatter the contract across paragraphs and make field-level extraction unreliable.
```
alexronquillo: > Prose descriptions of parameters scatter the contract across paragraphs and make field-level extraction unreliable.

I don't really understand what this means. Can you clarify this reasoning?

howtos/ai-ready-guidelines.md
```diff
@@ -0,0 +1,88 @@
...
+
+- Use images as **supplementary illustration**, never as the only source of a value the reader (human or agent) needs to act on.
+- For every diagram, include a short prose description of what it shows.
+- For every screenshot of a UI form, include the field names and their values inline in the text before or after the image.
+
+### 7. Never put critical information only inside a tab
+
+Agents typically see only the active tab when a page is rendered, and Markdown-export pipelines often serialize only the default tab. Content inside non-default tabs becomes invisible.
+
+- **Allowed:** Code examples in different languages across tabs (the concept is repeated, only the syntax changes).
```
alexronquillo: Isn't the syntax important for an Agent?

howtos/ai-ready-guidelines.md
```diff
@@ -0,0 +1,88 @@
...
+
+- A running Self-Managed installation.
+- ...
+```
+
+### 8. Keep code blocks complete and self-contained
+
+- Always close fences (` ``` `).
+- Include the language tag (` ```yaml `, ` ```java `, ` ```bash `).
+- Don't split a single example across multiple unrelated fences with prose in between if the prose isn't part of the example.
```
alexronquillo: > ...if the prose isn't part of the example

Can you expand on this?

alexronquillo (commented):
Thanks @afgambin I left some questions for parts I think needs a little more clarification. Otherwise, nice work!

howtos/ai-ready-guidelines.md
afgambin: If parameter information is hidden in prose instead of shown in a structured format, it’s harder to read/tell which details apply to each field.

Dropping the sentence, since it isn’t really needed to understand this recommendation.

howtos/ai-ready-guidelines.md
afgambin: Agree - updated

howtos/ai-ready-guidelines.md
afgambin: I have rephrased it. I think it reads better now

afgambin:
@alexronquillo Thanks for the review! Ready for another pass

howtos/ai-ready-guidelines.md
```diff
@@ -0,0 +1,85 @@
...
+
+- Use images as **supplementary illustration**, never as the only source of a value the reader (human or agent) needs to act on.
+- For every diagram, include a short prose description of what it shows.
+- For every screenshot of a UI form, include the field names and their values inline in the text before or after the image.
+
+### 7. Never put critical information only inside a tab
+
+Agents typically see only the active tab when a page is rendered, and Markdown-export pipelines often serialize only the default tab. Content inside non-default tabs becomes invisible.
+
+If prerequisites or steps differ by environment or programming language, list both flat in the page body, clearly labeled. For example:
```
alexronquillo: Is this a reasonable expectation? The [properties reference page](https://docs.camunda.io/docs/next/self-managed/components/orchestration-cluster/core-settings/configuration/properties/), for example, is already huge with the tabs. Flattening this content would lead to a page almost twice the size (three times the size for docs that also include helm configs). And for programming languages, if feels like it would really get in users' way if they had to skip over all the code blocks for programming languages that aren't relevant to their stack.

If the LLMs are consuming the markdown export, rather than the generated page, I suggest we: 

1. Try to figure out a solution to make the content in tabs visible.
2. If that fails, I suggest we come up with a new IA that separates content based on these differences (SM vs SaaS, programming languages, etc). If we do that, we should also come up with a solution that reduces content duplication as that could lead to a maintenance nightmare.

howtos/ai-ready-guidelines.md
```diff
@@ -0,0 +1,85 @@
+# AI-ready content guidelines
+
+This guide describes how to write Camunda docs content that's reliably consumable by AI assistants and by RAG/agent pipelines.
+
+It supplements the [Camunda style guide](technical-writing-styleguide.md) and the [contributor documentation guidelines](documentation-guidelines.md).
+
+### 1. Start each page and section with a direct answer
+
+The first sentence under any heading (including the H1) should directly answer "what is this?" or "what does this section/page do?" in one quotable line.
```
alexronquillo:
```suggestion
The first sentence under any heading (including the H1) should be quotable and directly answer "what is this?" or "what does this section/page do?".
```

I'm curious about how strict we want to be with this.

[Here's an example](http://localhost:3000/docs/next/components/agentic-orchestration/evaluate-agents/monitor-ai-agents/#step-2-open-the-process-instance-in-operate):

```md
## Step 2: Open the process instance in Operate

1. Open [Operate](http://localhost:3000/docs/next/components/operate/operate-introduction/).
2. Locate the process instance created by your prompt. See [view a deployed process](http://localhost:3000/docs/next/components/operate/userguide/basic-operate-navigation/#view-a-deployed-process) for more details.
3. Open your process instance view by clicking on its process instance key.
```

There's no lead-in one liner, but the heading, itself, so clearly states what the section is about. What kind of quotable one liner that directly answers "what is this?" or "what does this section/page do?" would we add to this that doesn't:

- Impact readability (isn't repetitive and awkward).
- Violate the don't repeat yourself and omit needless words principles that we apply in most other cases.

mesellings: Anecdotally, it's always been good practice for search engines etc that will use the first sentence under a heading as the shortdesc in a result, so I've always leaned towards including a sentence underneath a heading.

alexronquillo: Ok, thanks for the background! If that's what's best, we can resolve this conversation. 👍

I haven't seen support for that in the past. For example, Google suggests to [create people-first content](https://developers.google.com/search/docs/fundamentals/creating-helpful-content) as the best strategy for SEO and really only mentions the main heading (which we already have guidance for), not subsequent headings. I don't typically see H2s and H3s in search results either. However, I can obviously be wrong about that or missing some data that shows it really does matter.

howtos/ai-ready-guidelines.md
alexronquillo: Just to provide some extra context for my concern. As far as I understand, LLMs will mostly be looking at:

- https://docs.camunda.io/llms.txt
- https://docs.camunda.io/llms-full.txt

for ingesting the docs content. In that case, I am confused why the LLM would have trouble seeing the content in tabs.

howtos/ai-ready-guidelines.md
mesellings: It can be more important in a content reuse model - for example, a section might be surfaced at a higher heading level in another document if you have content used cross-document - although we don't currently have much reuse in docs-as-code/docusaurus, it's still good to future-proof the content in case we move into a CCMS at any point in the future.

howtos/ai-ready-guidelines.md
afgambin: Adding context since Alex and I already discussed this in another channel.

AI agents tend to read Markdown, but that’s only a preference - and only when they know it’s available. They can still fetch content directly from Camunda Docs as HTML, depending on where and how the agent is invoked. These are the cases where flattening tabs might be relevant.

For example, if a user asks something related to Camunda Docs in the standard ChatGPT or Claude browser UI, the LLM isn’t aware of the site instructions behind the scenes and usually fetches the HTML.

However, after testing the case above, the LLM returned results from the different available tabs.

Since we prefer to use Tabs across our docs for better UX and readability, we’re dropping this suggestion from the guidelines.

afgambin:
@alexronquillo I addressed your comments - thanks again!

howtos/ai-ready-guidelines.md
mesellings: Can I add in my two cents worth here - critical info should never be hidden in tabs for human users, so I think this might be a case where the human should override the agent UX as it's good practice on the web - humans should never be given the chance to miss critical info via the UI, or made to hunt for something, so for me tabs should only really be used for code samples and that kind of thing ideally.

afgambin: > critical info should never be hidden in tabs for (human) users

I agree with you, but I think that sort of statement probably fits better in our main style guide than in these guidelines for AI agents, since the former is also consumed by them.

afgambin: @camunda/tech-writers Another question is whether we need these specific AI-ready guidelines, or whether we want to merge them into our main style guide.

zaycodes:
Thanks for putting this together @afgambin. The guidelines are well-reasoned and the reference section is excellent. 

On question before we merge: should this live as a standalone file, or be integrated into the main style guide?

My recommendation is to integrate the majority of these guidelines into howtos/technical-writing-styleguide.md, and keep howtos/ai-ready-guidelines.md as a lightweight companion focused only on AI-specific context.

Here's why, and my suggestion on what that would look like:

The core rules already belong in the style guide

Looking at the existing style guide, there are clear homes for most of what's here:

Under [**Formatting, organization and structure for conceptual pieces**](https://github.com/camunda/camunda-docs/blob/main/howtos/technical-writing-styleguide.md#formatting-organization-and-structure-for-conceptual-pieces):

- **Rules 1 and 3** (direct opening sentences, descriptive headings) extend the existing *Titles, headers, and sidebars* table row, which already says headings should be "descriptive enough", we just need to strengthen that with the direct-answer requirement and examples.

- **Rule 2** (self-contained sections) fits alongside the existing *Separated paragraphs* and *Short sentences* rows.

- **Rule 4** (heading hierarchy, don't skip levels, keep sibling headings parallel) also belongs here alongside Rules 1 and 3, as it's heading guidance.

Under [**Formatting, organization and structure for implementation steps**](https://github.com/camunda/camunda-docs/blob/main/howtos/technical-writing-styleguide.md#formatting-organization-and-structure-for-implementation-steps):

- **Rule 5** (tables over prose for structured data) extends the existing *Visuals* row, which already asks "would this be more visually appealing in a table?". It just needs to be made more prescriptive.

- **Rule 6** (no critical info in screenshots only) extends the existing *Images* row, which already says to "include text above or below the screenshot that includes input text."

- **Rule 7** (complete, self-contained code blocks) fits directly under the existing *Code blocks* row.

Keeping these as a separate file creates a maintenance problem and a discovery problem. Contributors are expected to read the style guide. A second document titled "AI-ready guidelines" will be read by the people already interested in AI, and may be missed by everyone else, which is exactly the opposite of what we want if the goal is to shift baseline authoring quality.

What should stay in ai-ready-guidelines.md

The parts that are genuinely AI-specific and would feel out of place in the main style guide:

- The framing context: why agents are now a primary audience, how they consume content (heading extraction, RAG fetching), what the current delivery layer (llms.txt, Kapa.ai) covers and doesn't cover.

- The links to external benchmarks (Agent Score, agentdocsspec.com, Kapa.ai best practices) — these are reference material, not writing rules.

- Any future guidance specific to agentic use cases that doesn't generalize to human readers.

That version of ai-ready-guidelines.md becomes a short "why this matters and where to learn more" doc that points back to the relevant sections in the style guide rather than duplicating rules that contributors should be following anyway.

afgambin:
@zaycodes I agree that integrating them into our main style guide is probably the best approach, since the suggested additions also apply to humans and aren’t specific enough for AI agents.

I removed the new page, since I don’t think what was left is relevant to keep right now.

Ready for another pass. Thanks!

zaycodes (approved):
Great! Thanks for putting this together @afgambin. LGTM!

PR #8935, opened 2026-05-26
docs(feature): job prioritization

giorgionaps:
## Description

Docs for the new job prioritization feature.

## PR Checklist


docs/components/modeler/bpmn/service-tasks/service-tasks.md
```diff
@@ -38,21 +38,32 @@ Input mappings can be used to transform the variables into a format accepted by
 
 For more information about this topic visit the documentation about [Input/output variable mappings](/components/concepts/variables.md#inputoutput-variable-mappings).
 
+## Job priority
+
+This task type supports `zeebe:jobPriorityDefinition` when implemented as a job worker.
```
georgios-goulos: Service tasks can only be implemented as job  workers.

docs/components/modeler/bpmn/send-tasks/send-tasks.md
```diff
@@ -47,6 +47,13 @@ or use it as a blueprint for your own job worker.
 
 :::
 
+## Job priority
```
georgios-goulos: With these additions it looks like the structure now is:

```
- Additional resource
- Job priority
-- XML representation
```

I take it this was unintentional?

Similarly for script-tasks.md

docs/components/modeler/bpmn/service-tasks/service-tasks.md
```diff
@@ -38,21 +38,32 @@ Input mappings can be used to transform the variables into a format accepted by
...
 ## Additional resources
 
 ### XML representation
 
-A service task with a custom header:
+A service task with a custom header and priority definition:
 
 ```xml
-<bpmn:serviceTask id="collect-money" name="Collect Money">
+<bpmn:process id="order-process" isExecutable="true">
```
georgios-goulos: Is there a reason we are switching to `bpmn:process` instead of just extending the existing `bpmn:serviceTask` with `zeebe:jobPriorityDefinition`?

I am asking because we have similar examples for the rest of the task types and we should probably keep those consistent.

docs/components/concepts/job-workers.md
```diff
@@ -256,6 +256,55 @@ For example, if jobs of a given type are not activated, but a worker is opened f
...
+## Job prioritization
+
+Camunda supports job prioritization for pull-based job activation.
+
+You can define job priority:
+
+- On the process as a default.
+- On supported job-creating tasks as an override, including service tasks, send tasks, and script or business rule tasks implemented as job workers.
+
+Job priority affects the order in which jobs are activated for workers. It does not change retries, failures, or completion semantics.
```
georgios-goulos: opinion: I would move this to the top of this section as it explains what job prioritisation is about before explaining how it can be configured. However, if this is the pattern we typically use feel free to keep it as it is.

docs/components/concepts/job-workers.md
```diff
@@ -256,6 +256,55 @@ For example, if jobs of a given type are not activated, but a worker is opened f
...
+
+Important behavior:
+
+- Higher-priority jobs are activated before lower-priority jobs within a partition.
+- Ordering is best effort across the cluster, not a global guarantee across all partitions.
```
georgios-goulos: How technical do we typically get in these `docs/components/concepts/` pages? Should we add some technical information here to explain this behavior?

docs/components/concepts/job-workers.md
```diff
@@ -256,6 +256,55 @@ For example, if jobs of a given type are not activated, but a worker is opened f
...
+If starvation becomes a problem, increase worker capacity or revise your priority assignments.
+
+Priority values can be static integers or FEEL expressions. FEEL expressions are evaluated when the job is created, and the resulting integer is stored on the job.
+
+Priority uses signed 32-bit integer semantics. Do not assume a fixed `0-99` engine limit.
+
+:::note Compatibility for pre-8.10 jobs
+Jobs created before 8.10 may not have a stored priority value in secondary storage.
+
+For compatibility, Camunda treats these jobs as if they had the default priority value of `0`.
```
georgios-goulos: We should clarify that when users filter/sort using the `priority` field in the relevant job APIs, the pre-8.10 jobs will be filtered out as the searching operation is using the secondary storage and we will not be performing a backfill for that so the `priority` field will not be populated for pre-8.10 jobs. However, the Zeebe engine will still treat such jobs as if they had a priority of 0.

georgios-goulos (commented):
Overall this LGTM! Left some comments.

Is there a "What's new" page for 8.10 that should also be updated? Not necessarily as part of this PR, but it should be planned if such a page exists.

docs/components/concepts/job-workers.md
giorgionaps: @georgios-goulos we typically don't go very in-depth. I'll expand this bullet though

docs/components/modeler/bpmn/service-tasks/service-tasks.md
giorgionaps: @georgios-goulos I'll update the XML snippet here and in other task docs in the next iteration

docs/components/concepts/job-workers.md
giorgionaps: @georgios-goulos I’ll update the compatibility note 👌

bojtospeter (dismissed):
lgtm 🔥

mesellings:
@zaycodes could you review this today/tomorrow please so @giorgionaps is able to make any changes as required before the EOW? 🙏

docs/components/concepts/job-workers.md
```diff
@@ -256,6 +256,57 @@ For example, if jobs of a given type are not activated, but a worker is opened f
 
 If it's not present in the gateway as a client stream, restart your worker. If it's not present as a consumer in one of the brokers, this indicates a bug. As a workaround, restart your gateway, which will cause some interruption in your service, but will force all streams for this gateway to be recreated properly.
 
+## Job prioritization
+
+Camunda supports job prioritization for pull-based job activation. Job priority affects the order in which jobs are activated for workers. It does not change retries, failures, or completion semantics.
```
zaycodes, line 263: Worth adding a sentence or two upfront on when you'd actually need this such as, "Use job prioritization when you have mixed-urgency workloads and want time-sensitive jobs to be processed ahead of lower-priority ones." Right now the section jumps straight into what it is, but readers landing here cold won't know if it's relevant to them.

docs/components/concepts/job-workers.md
```diff
@@ -256,6 +256,57 @@ For example, if jobs of a given type are not activated, but a worker is opened f
...
+
+You can define job priority:
+
+- On the process as a default.
+- On supported job-creating tasks as an override, including service tasks, send tasks, and script or business rule tasks implemented as job workers.
+
+Important behavior:
+
+- If you do not specify job priorities, jobs keep the existing FIFO behavior (order by arrival/creation).
```
zaycodes: A few of these are saying the same thing in different ways. Items 3 and 4 are essentially both saying "ordering is per-partition, not global." Items 1 and 5/7 all touch on FIFO and default priority 0. Worth collapsing those into fewer, cleaner bullets or grouping them into small paragraphs by theme (scope of ordering, defaults, starvation). 9 items in a flat list is a lot to parse.

Suggested rewrite without the important behavior heading:

Job priority is enforced per partition, not globally across the cluster. Within a partition, higher-priority jobs are activated before lower-priority ones. Within the same priority level, jobs activate in FIFO order. If you don't set any priorities, standard FIFO behavior applies throughout.

The default priority is 0. A task-level priority definition overrides any process-level default.

Be aware that low-priority jobs can starve if workers are consistently occupied by higher-priority work. Camunda doesn't provide starvation mitigation, fairness guarantees, or priority aging. If starvation becomes a problem, increase worker capacity or revise your priority assignments.

docs/components/concepts/job-workers.md
```diff
@@ -256,6 +256,57 @@ For example, if jobs of a given type are not activated, but a worker is opened f
...
+- Task-level priority overrides process-level priority.
+- Within the same priority, jobs are activated in FIFO order.
+- Low-priority jobs can starve under sustained high-priority load and may never be activated while workers remain fully occupied by higher-priority jobs.
+- Camunda does not provide fairness, aging, or starvation mitigation in the engine.
+
+If starvation becomes a problem, increase worker capacity or revise your priority assignments.
+
+Priority values can be static integers or FEEL expressions. FEEL expressions are evaluated when the job is created, and the resulting integer is stored on the job.
+
+Priority uses signed 32-bit integer semantics. Do not assume a fixed `0-99` engine limit.
```
zaycodes: Maybe this is just me cos the concept is quite new to me so this comment is optional and you can totally ignore my suggestion: 

This reads oddly to someone like me who isn't coming from a system that uses `0–99` priority scales. The assumption isn't self-evident. Maybe better as a positive statement: "Priority accepts any signed 32-bit integer. There's no fixed upper or lower bound enforced by the engine."

docs/components/concepts/job-workers.md
```diff
@@ -256,6 +256,57 @@ For example, if jobs of a given type are not activated, but a worker is opened f
...
+
+:::warning Compatibility for pre-8.10 jobs
```
zaycodes: This callout is doing two different jobs at once. One is just context (execution defaults to priority 0, no action needed), the other is a genuine gotcha (API queries silently exclude pre-8.10 jobs). I'd split them: keep the execution behavior as plain prose and reserve the callout for the API filtering consequence, which is the part that can actually surprise someone.

Suggested rewrite:

Jobs created before version 8.10 do not have a stored `priority` value. For job execution,
Camunda treats them as having the default priority of `0`.
:::warning
If you use job APIs that filter or sort by `priority`, pre-8.10 jobs are excluded from
results because no priority value is stored for them. Review any priority-based queries
before upgrading.
:::

docs/components/concepts/job-workers.md
```diff
@@ -256,6 +256,57 @@ For example, if jobs of a given type are not activated, but a worker is opened f
...
+Jobs created before 8.10 do not have a populated `priority` value in secondary storage.
+
+For job execution behavior, Camunda treats these jobs as if they had the default priority value of `0`.
+
+However, if you use job APIs that filter or sort by `priority`, pre-8.10 jobs are filtered out because no priority value is stored for them.
+:::
+
+### Validation and runtime failure behavior
+
+Job priority values can be static integers or FEEL expressions.
```
zaycodes: This is a direct repeat of line 281. One of them needs to go.

docs/components/concepts/job-workers.md
```diff
@@ -256,6 +256,57 @@ For example, if jobs of a given type are not activated, but a worker is opened f
...
+
+Invalid FEEL expressions are handled differently depending on when they fail:
+
+- If the expression is invalid at deployment time, BPMN deployment fails and the process is not deployed.
+- If the expression fails at runtime, for example because a referenced variable is missing, Camunda raises an incident and the job is not created.
+
+Current recovery is manual. You can provide the missing variable and then resolve the incident.
```
zaycodes:
```suggestion
To recover, provide the missing variable value and resolve the incident manually in Operate.
```

docs/components/concepts/job-workers.md
```diff
@@ -256,6 +256,57 @@ For example, if jobs of a given type are not activated, but a worker is opened f
...
+
+### Job streaming limitation
+
+Job streaming does not honor job priority ordering.
+If you mix job pull and job streaming, the benefits of prioritization can be reduced.
```
zaycodes: "The benefits of prioritization can be reduced" is a bit vague. Does it mean partially degraded, fully ignored, unpredictable? Worth being specific about what actually happens to priority-ordered jobs when some workers are on streaming. If the practical takeaway is "don't mix pull and streaming if priority ordering matters to you," that's worth stating directly.

docs/components/modeler/bpmn/business-rule-tasks/business-rule-tasks.md
```diff
@@ -100,6 +100,13 @@ Define [variable mappings](/components/concepts/variables.md#inputoutput-variabl
 the [same way as a service task does](/components/modeler/bpmn/service-tasks/service-tasks.md#variable-mappings)
 to transform the variables passed to the job worker, or to customize how the variables of the job merge.
 
+## Job priority
```
zaycodes: This is a top-level ## section, but job priority only applies when the business rule task is implemented as a job worker, not when it's using the DMN decision engine. I'll suggest moving this under ## Job worker implementation as a ### subsection would make the scope clearer and avoid confusing DMN users

zaycodes (changes requested):
The feature coverage is solid. Left a few suggestions inline to tighten it.

zaycodes:
Do we need to backport these changes @giorgionaps or they only apply to 8.10?

bojtospeter:
> Do we need to backport these changes … ?

@zaycodes, we will introduce the whole priority feature in 8.10, no backporting is needed.

giorgionaps:
@zaycodes just pushed some updates to address your comments

zaycodes (dismissed):
lgtm!

christinaausley:
@bojtospeter I see there is a `hold` label -- when would you like this merged?

bojtospeter:
> @bojtospeter I see there is a `hold` label -- when would you like this merged?

hi @christinaausley , the feature is targeted for alpha 3, as the whole functionality is not yet available. Should we merge it before, so we can avoid merge conflicts? (It might be confusing to customers)

christinaausley:
@bojtospeter as "next" isn't an officially "released" version of docs yet and is more of a playground for content, we can go ahead and merge this given we are quite close to alpha3. If you felt like the content would change quite a bit before alpah3 I would say let's hold it, but seems like this is pretty close to good to go?

bojtospeter:
> @bojtospeter as "next" isn't an officially "released" version of docs yet and is more of a playground for content, we can go ahead and merge this given we are quite close to alpha3. If you felt like the content would change quite a bit before alpah3 I would say let's hold it, but seems like this is pretty close to good to go?

Hi @christinaausley, if it's OK to have the documentation live a few weeks before the actual feature is there, then I agree to merge it.

bojtospeter (commented):
1. Process-level priority: XML example missing

The docs say priority can be set "on the process as a default" but never show the XML. From the engine source, process-level uses the same `zeebe:jobPriorityDefinition` element placed inside the process `extensionElements`:

```xml
<bpmn:process id="my-process" isExecutable="true">
  <bpmn:extensionElements>
    <zeebe:jobPriorityDefinition priority="42" />
  </bpmn:extensionElements>
  ...
</bpmn:process>
```

A task-level definition overrides this. Without a process-level example, users won't know the syntax, worth mentioning that.

2. Activation ordering is always-on, which is not explicit in docs

The original spec included a per-request `usePriority` opt-in flag, but per the June 4 progress update that was descoped. The docs say priority "affects the order in which jobs are activated" — true, but users will ask: _do I need to enable anything?_ Suggest adding: "Priority ordering is automatic. No worker-level flag or configuration is required — once a job has a priority set, Zeebe activates higher-priority jobs first, when a job pull method is used"

christinaausley:
@bojtospeter I'll incorporate your feedback and re-request review shortly 👍

christinaausley:
Ok @bojtospeter ready for final review 😄

bojtospeter (approved):
lgtm 🔥

PR #8975, opened 2026-06-01
docs: clarify Helm upload size configuration

eamonnmoloney:
## Summary
- Documents that Helm users should set `global.config.requestBodySize` to align REST multipart limits and Zeebe message size.
- Adds ingress alignment guidance for `nginx.ingress.kubernetes.io/proxy-body-size`.
- Clarifies that Zeebe internal record-batch processing can still reject deployments even when transport limits are aligned.

## Root Cause
Root cause is in the Helm chart: `global.config.requestBodySize` configured Spring multipart upload limits at `10MB`, but Zeebe Gateway/Broker message size remained at the product default `4MB`.

## Validation
- Not run: `npm run format:check` could not run locally because docs dependencies are not installed (`prettier: not found`).
- Content was updated in Next, 8.9, 8.8, and 8.7 docs trees.

## GKE Evidence From Helm PR
Baseline 8.9 chart deployed to `support32927base-89-keyco-inst-gke`:
- Small BPMN upload to `/orchestration/v2/deployments`: HTTP 200.
- ~5MB BPMN upload: HTTP 400, `Request size is above configured maxMessageSize.`

Fixed 8.9 chart deployed to `support32927fix-89-keyco-inst-gke`:
- Runtime config includes `spring.servlet.multipart.max-file-size`, `spring.servlet.multipart.max-request-size`, `camunda.cluster.network.max-message-size`, `camunda.api.grpc.max-message-size`, and legacy `zeebe.broker.network.maxMessageSize` at `10MB`.
- Valid 4.2MB BPMN upload to `/orchestration/v2/deployments`: HTTP 200.
- Valid 5MB BPMN no longer fails with `maxMessageSize`, but still hits Zeebe engine `EXCEEDED_BATCH_RECORD_SIZE`. This appears to be a separate product/engine record-batch limit, not the Helm transport-limit bug.

## Related PRs
- Helm PR: https://github.com/camunda/camunda-platform-helm/pull/6279

## Notes
- Originating nightly run URL: not available in this session; SUPPORT-32927 was reproduced manually on GKE.
- Merge order: Helm first, docs second.

### DIFF
--- a/docs/self-managed/components/orchestration-cluster/orchestration-cluster-api-rest-deployment-configuration.md
+++ b/docs/self-managed/components/orchestration-cluster/orchestration-cluster-api-rest-deployment-configuration.md
@@ -18,7 +18,7 @@ The `maxMessageSize` default value is 4MB for raw Zeebe deployments (10MB for He
 - [Broker config](/self-managed/components/orchestration-cluster/zeebe/configuration/broker.md#zeebebrokernetwork)
 
 If you deploy with Helm, the chart defaults to 10MB via `global.config.requestBodySize`, applying
-this value to the Zeebe message size, the REST multipart file and request size, and the Tomcat HTTP
+this value to the Zeebe Gateway and broker message size, the REST multipart file and request size, and the Tomcat HTTP
 form post size on the Zeebe Gateway. Change `global.config.requestBodySize` only if you need a
 value other than 10MB.
 
@@ -34,7 +34,7 @@ response, not HTTP 413.
 Multipart requests include metadata and boundary overhead in addition to the uploaded file content.
 Set these limits slightly above the largest file you expect users or Connectors to upload.
 
-If you increase this value, you must also adjust the configuration for the deployment REST endpoint to match.
+If you are not using Helm and increase this value, you must also adjust the configuration for the deployment REST endpoint to match.
 
 ### REST API server configuration
 
--- a/docs/self-managed/components/orchestration-cluster/orchestration-cluster-api-rest-deployment-configuration.md
+++ b/docs/self-managed/components/orchestration-cluster/orchestration-cluster-api-rest-deployment-configuration.md
@@ -38,7 +38,7 @@ If you are not using Helm and increase this value, you must also adjust the conf
 
 ### REST API server configuration
 
-If you're using the REST API and submitting files via multipart upload (e.g., using `POST /v2/deployments`), make sure your application is configured to accept the updated request sizes. If you increase the `maxMessageSize` to 10MB, increase these property values to 10MB as well.
+If you're increasing the `maxMessageSize` and using the REST API for multipart uploads (for example, using `POST /v2/deployments`), you must configure your application to accept the updated request sizes. If you increase the `maxMessageSize` to 10MB, increase these property values to 10MB as well.
 
 For Spring Boot applications:
 
--- a/docs/self-managed/components/orchestration-cluster/orchestration-cluster-api-rest-deployment-configuration.md
+++ b/docs/self-managed/components/orchestration-cluster/orchestration-cluster-api-rest-deployment-configuration.md
@@ -12,19 +12,24 @@ This guide walks you through adjusting these settings.
 
 ### Gateway and Broker configuration
 
-The `maxMessageSize` default value is 4MB. You can configure this setting in the:
+The `maxMessageSize` default value is 4MB for raw Zeebe deployments (10MB for Helm). You can configure this setting in the:
 
 - [Gateway config](../orchestration-cluster/zeebe/configuration/gateway.md#zeebegatewaynetwork)
 - [Broker config](/self-managed/components/orchestration-cluster/zeebe/configuration/broker.md#zeebebrokernetwork)
 
-If you deploy with Helm, set `global.config.requestBodySize`. The chart applies this value to
-the Zeebe message size, the REST multipart file and request size, and the Tomcat HTTP form post
-size. If you expose the REST API through Ingress, keep
-`global.ingress.annotations.nginx.ingress.kubernetes.io/proxy-body-size` aligned with the same size.
+If you deploy with Helm, the chart defaults to 10MB via `global.config.requestBodySize`, applying
+this value to the Zeebe message size, the REST multipart file and request size, and the Tomcat HTTP
+form post size on the Zeebe Gateway. Change `global.config.requestBodySize` only if you need a
+value other than 10MB.
+
+If you expose the REST API through Ingress, you must also explicitly set
+`global.ingress.annotations.nginx.ingress.kubernetes.io/proxy-body-size` to the same value.
+This annotation is not propagated automatically from `requestBodySize`.
 
 This setting aligns the HTTP, multipart, Tomcat, Gateway, and Broker transport limits. Deployments
 can still be rejected by Zeebe's internal record batch processing if the deployed resources and
-follow-up records exceed the engine batch limit.
+follow-up records exceed the engine batch limit. This engine-level rejection returns an HTTP 500
+response, not HTTP 413.
 
 Multipart requests include metadata and boundary overhead in addition to the uploaded file content.
 Set these limits slightly above the largest file you expect users or Connectors to upload.
--- a/docs/self-managed/components/orchestration-cluster/orchestration-cluster-api-rest-deployment-configuration.md
+++ b/docs/self-managed/components/orchestration-cluster/orchestration-cluster-api-rest-deployment-configuration.md
@@ -19,7 +19,7 @@ The `maxMessageSize` default value is 4MB. You can configure this setting in the
 
 If you deploy with Helm, set `global.config.requestBodySize`. The chart applies this value to
 the Zeebe message size, the REST multipart file and request size, and the Tomcat HTTP form post
-size. If you expose the REST API through ingress, keep
+size. If you expose the REST API through Ingress, keep
 `global.ingress.annotations.nginx.ingress.kubernetes.io/proxy-body-size` aligned with the same size.
 
 This setting aligns the HTTP, multipart, Tomcat, Gateway, and Broker transport limits. Deployments
--- a/docs/self-managed/components/orchestration-cluster/orchestration-cluster-api-rest-deployment-configuration.md
+++ b/docs/self-managed/components/orchestration-cluster/orchestration-cluster-api-rest-deployment-configuration.md
@@ -6,7 +6,7 @@ description: "Learn how to configure upload limits and multipart handling for Or
 
 ## Deployment configuration
 
-The Orchestration Cluster REST API supports file and multipart uploads when deploying BPMN diagrams and other resources. By default, the maximum allowed request size is 4MB. You can increase this limit in Self-Managed cluster through configuration of the Zeebe Gateway, Broker, and the REST layer (e.g. Spring Boot + Tomcat).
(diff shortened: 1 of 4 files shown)


christinaausley:
Hi @eamonnmoloney is this ready for technical writer review?

docs/self-managed/components/orchestration-cluster/orchestration-cluster-api-rest-deployment-configuration.md
```diff
@@ -17,6 +17,18 @@ The `maxMessageSize` default value is 4MB. You can configure this setting in the
 - [Gateway config](../orchestration-cluster/zeebe/configuration/gateway.md#zeebegatewaynetwork)
 - [Broker config](/self-managed/components/orchestration-cluster/zeebe/configuration/broker.md#zeebebrokernetwork)
 
+If you deploy with Helm, set `global.config.requestBodySize`. The chart applies this value to
+the Zeebe message size, the REST multipart file and request size, and the Tomcat HTTP form post
+size. If you expose the REST API through Ingress, keep
+`global.ingress.annotations.nginx.ingress.kubernetes.io/proxy-body-size` aligned with the same size.
```
hisImminence: **P1 · Ingress annotation is not auto-propagated from requestBodySize, and nginx uses different unit syntax** _(via devils-advocate)_

Two related problems that will cause silent misconfiguration:

1. The chart does NOT propagate `global.config.requestBodySize` to the ingress annotation. The annotation has its own independent default (`10m`) in `values.yaml`. A user who raises `requestBodySize` to `20MB` without separately updating the annotation will still receive nginx 413 errors. The current phrasing 'keep … aligned with the same size' implies the user only needs to check they match — it does not communicate that the chart never updates the annotation automatically and that it is always a separate manual step.

2. nginx ingress uses lowercase size suffixes (`10m`, `20m`) while Spring Boot uses SI notation (`10MB`, `20MB`). A user copying `20MB` from their `requestBodySize` into the annotation value will produce either a silent parse failure (nginx treats unknown suffixes as an error, falling back to the default, which is often unlimited or zero depending on the controller version) or unexpected behavior. The docs give no indication that the format differs.

This same issue applies to `versioned_docs/version-8.8`, `version-8.9`, and `version-8.7` which contain identical ingress annotation language.

```
`global.ingress.annotations.nginx.ingress.kubernetes.io/proxy-body-size` aligned with the same size.
```

hisImminence: **P1 · Component-level ingresses (zeebeGateway.ingress.rest / orchestration.ingress) not mentioned; do not inherit global annotations** _(via devils-advocate)_

The guidance only references `global.ingress.annotations`. However, Helm chart users commonly enable component-specific REST ingress resources: `zeebeGateway.ingress.rest` in the 8.7 chart and `orchestration.ingress` in the 8.8+ chart. These are separate Ingress objects whose annotation maps do not inherit from `global.ingress.annotations`. Neither includes `nginx.ingress.kubernetes.io/proxy-body-size` in their default annotation set (confirmed by chart inspection). A user who enables `zeebeGateway.ingress.rest` or `orchestration.ingress` and follows the current documentation will correctly update the global ingress but leave the component ingress unconfigured, resulting in continued 413 errors through the component ingress path with no obvious connection to the docs they just followed.

The same gap exists in the identical guidance blocks in `versioned_docs/version-8.7`, `version-8.8`, and `version-8.9`.

```
`global.ingress.annotations.nginx.ingress.kubernetes.io/proxy-body-size` aligned with the same size.
```

hisImminence (commented):
**crev review**



_Specialists run: correctness, devils-advocate. Devil's-advocate hypotheses: 0 raised, 0 promoted._

docs/self-managed/components/orchestration-cluster/orchestration-cluster-api-rest-deployment-configuration.md
```diff
@@ -17,6 +17,18 @@ The `maxMessageSize` default value is 4MB. You can configure this setting in the
...
+
+This setting aligns the HTTP, multipart, Tomcat, Gateway, and Broker transport limits. Deployments
+can still be rejected by Zeebe's internal record batch processing if the deployed resources and
```
hisImminence, line 30: **P2 · EXCEEDED_BATCH_RECORD_SIZE produces HTTP 500 not HTTP 413; no remediation path provided** _(via devils-advocate)_

Users reading this page are debugging upload failures. The transport-limit section describes 413 responses, but the `EXCEEDED_BATCH_RECORD_SIZE` rejection from the Zeebe engine produces an HTTP 500 Internal Server Error, not a 413. A user who follows all the transport-limit guidance and still sees failures will likely discount this caveat because their response code does not match 413 patterns.

Additionally: the rejection reason is intentionally left empty in the engine (to avoid exceeding the batch limit on the rejection record itself), so broker logs will show `EXCEEDED_BATCH_RECORD_SIZE` with no detail message. Users checking logs will have no further diagnostic signal.

Finally, the 'engine batch limit' is not controlled by `global.config.requestBodySize` or any exposed Helm value — it is bounded by the Zeebe log segment size. There is no configuration path to raise it documented here, leaving the reader with no actionable next step if they encounter it.

```
can still be rejected by Zeebe's internal record batch processing if the deployed resources and
```

versioned_docs/version-8.7/apis-tools/camunda-api-rest/camunda-api-rest-overview.md
```diff
@@ -34,13 +34,26 @@ You can also directly [download the OpenAPI specification](https://github.com/ca
 
 You can change the `maxMessageSize` default value of 4MB in the [Gateway](../../self-managed/zeebe-deployment/configuration/gateway.md#zeebegatewaynetwork) and [Broker](../../self-managed/zeebe-deployment/configuration/broker.md#zeebebrokernetwork) configuration.
 
+If you deploy with Helm, set `global.config.requestBodySize`. The chart applies this value to
+the Zeebe Gateway, broker message size, REST multipart file and request size, and Tomcat HTTP form
+post size. If you expose the REST API through Ingress, keep
```
hisImminence: **P2 · 8.7 text implies broker has a Tomcat HTTP form post limit; the broker does not run Tomcat** _(via correctness)_

The sentence reads: 'The chart applies this value to the Zeebe Gateway, broker message size, REST multipart file and request size, and Tomcat HTTP form post size.' The phrase 'Tomcat HTTP form post size' is listed as a fourth item alongside broker message size, which implies the broker has a Tomcat form post limit to configure. It does not — the Zeebe broker is a gRPC/network process with no Tomcat REST layer. `server.tomcat.max-http-form-post-size` is a Zeebe Gateway (REST server) setting only. The 8.8/8.9/Next equivalent text avoids this confusion by saying 'Zeebe message size' and 'Tomcat HTTP form post size' in a structure that doesn't tie Tomcat to the broker.

A user reading the 8.7 docs may spend time looking for a way to set `max-http-form-post-size` on the broker deployment, or may believe a limit mismatch exists on the broker when one doesn't.

```
the Zeebe Gateway, broker message size, REST multipart file and request size, and Tomcat HTTP form
```

docs/self-managed/components/orchestration-cluster/orchestration-cluster-api-rest-deployment-configuration.md
```diff
@@ -17,6 +17,18 @@ The `maxMessageSize` default value is 4MB. You can configure this setting in the
 - [Gateway config](../orchestration-cluster/zeebe/configuration/gateway.md#zeebegatewaynetwork)
 - [Broker config](/self-managed/components/orchestration-cluster/zeebe/configuration/broker.md#zeebebrokernetwork)
 
+If you deploy with Helm, set `global.config.requestBodySize`. The chart applies this value to
```
hisImminence: **P2 · Helm default for requestBodySize is already 10MB — docs frame this as a 4MB-default scenario requiring active remediation** _(via devils-advocate)_

The section opens with 'The `maxMessageSize` default value is 4MB' (line 15) and then instructs Helm users to set `global.config.requestBodySize`. The Helm chart default for `global.config.requestBodySize` is `10MB` (confirmed by PR body GKE evidence: 'Runtime config includes ... at 10MB'). A Helm user on the fixed chart who reads this page will believe their baseline is 4MB and that they need to actively configure `requestBodySize`, when in fact they are already at 10MB and all transport limits are already aligned. This creates unnecessary configuration churn. The 4MB default is the bare-product (non-Helm) default; it should be clearly distinguished from the Helm default.

The same framing issue applies to `versioned_docs/version-8.8`, `version-8.9`, and `version-8.7`.

```
If you deploy with Helm, set `global.config.requestBodySize`. The chart applies this value to
```

hisImminence (approved):
Thanks @eamonnmoloney - see comments of crev review

hisImminence (changes requested):
see above

docs/self-managed/components/orchestration-cluster/orchestration-cluster-api-rest-deployment-configuration.md
eamonnmoloney: Fixed. Replaced the passive 'keep aligned' phrasing with an explicit action: 'you must also explicitly set `global.ingress.annotations.nginx.ingress.kubernetes.io/proxy-body-size` to the same value. This annotation is not propagated automatically from `requestBodySize`.' Applied across all four files.

eamonnmoloney: Fixed for 8.7 — added a note that `zeebeGateway.ingress.rest.annotations` also needs `proxy-body-size` set, as that Ingress object does not inherit from `global.ingress.annotations`.\n\nFor 8.8+: confirmed from the chart `values.yaml` that `orchestration.ingress` has no REST sub-key (only gRPC). REST traffic routes through `global.ingress` in 8.8+, so no additional caveat is needed there.

docs/self-managed/components/orchestration-cluster/orchestration-cluster-api-rest-deployment-configuration.md
eamonnmoloney: Fixed. Added 'This engine-level rejection returns an HTTP 500 response, not HTTP 413' to the batch limit caveat in all four files.

versioned_docs/version-8.7/apis-tools/camunda-api-rest/camunda-api-rest-overview.md
eamonnmoloney: Fixed. Reworded to 'the Tomcat HTTP form post size on the Zeebe Gateway' to make clear this is a Gateway-only setting. The broker is gRPC-only with no Tomcat layer.

docs/self-managed/components/orchestration-cluster/orchestration-cluster-api-rest-deployment-configuration.md
eamonnmoloney: Fixed. Rewrote the Helm paragraph to lead with 'the chart defaults to 10MB via `global.config.requestBodySize`' and closes with 'Change `global.config.requestBodySize` only if you need a value other than 10MB.' Also updated the `maxMessageSize` sentence to '4MB for raw Zeebe deployments (10MB for Helm).' Applied across all four files.

eamonnmoloney:
Yes, ready for technical writer review now — all review comments have been addressed.

hisImminence (dismissed):

christinaausley (approved):
Looks great!

PR #8993, opened 2026-06-04
chore(optimize): update optimize import page size docs

HeleneW-dot:
related to [#54410](https://github.com/camunda/camunda/issues/54410)

## Description
Improve documentation on how changing CAMUNDA_OPTIMIZE_ZEEBE_MAX_IMPORT_PAGE_SIZE (zeebe.maxImportPageSize, default 200) affects Optimize behaviour.

## PR Checklist

### DIFF
--- a/docs/components/best-practices/architecture/sizing-your-environment.md
+++ b/docs/components/best-practices/architecture/sizing-your-environment.md
@@ -52,7 +52,7 @@ Optimize is an optional component that provides process analytics and reporting.
 - **Throughput:** Fewer tasks/second achievable at the same hardware level when Optimize is running.
 - **Disk space:** Optimize stores significant amounts of data in Elasticsearch, especially with large payloads. In testing, 128 Gi of ES disk was consumed in under 12 hours at 1 PI/s with the realistic payload (~11 KB) and 30-day retention.
 - **Elasticsearch resources:** More CPU, memory, and disk are needed for ES when Optimize is enabled.
-- **Import time:** Optimize import time increases approximately linearly with payload size. Larger payloads (for example, 11 KB realistic vs. 0.5 KB typical) result in proportionally longer import times.
+- **Individual import latency:** Optimize import time increases approximately linearly with payload size. Larger payloads (for example, 11 KB realistic vs. 0.5 KB typical) result in proportionally larger individual import latency.
 - **Report loading times:** As historical data accumulates in Elasticsearch, Optimize report loading times increase approximately linearly.
 
 #### Mitigations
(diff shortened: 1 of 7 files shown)


christinaausley (dismissed):

docs/components/best-practices/architecture/sizing-your-environment.md
```diff
@@ -61,6 +61,7 @@ Optimize is an optional component that provides process analytics and reporting.
 - Use variable filtering to reduce the amount of data exported/imported by Optimize.
 - Tune retention periods: shorter retention means less data in ES, and better performance.
 - Disable variable import entirely if variables are not needed in Optimize reports.
+- To increase Optimize import throughput, raise `CAMUNDA_OPTIMIZE_ZEEBE_MAX_IMPORT_PAGE_SIZE` so each import cycle fetches more exported records. This helps Optimize keep pace under high load and reduces import lag, at the cost of higher memory use per fetch.
```
ChrisKujawa: 🔧 I would also add that this impacts individual latency of data import right (which is normally not a problem for Optimize users, but still)

❓ Worth mentioning that this might cause issues when we have a lot of data or big variables? I remember Operate/Tasklist importers having issues with this and fail or get stuck when importing too much, does Optimize has a similar issue?

ChrisKujawa: I would also maybe move this up in the list, as it is one of the cheaper options to fine-tune Optimize performance. Wdyt?

ChrisKujawa (commented):
Thanks @HeleneW-dot sorry for the delay.

No blocker - I just would extend this a bit, as described below.

docs/components/best-practices/architecture/sizing-your-environment.md
HeleneW-dot: > I would also add that this impacts individual latency of data import right (which is normally not a problem for Optimize users, but still)

If I am understanding correctly, I think I meant the same with `This helps Optimize keep pace under high load and reduces import lag`, but maybe that phrasing isnt clear enough.

What do you think about `This helps Optimize keep pace under high load and reduces import lag, which can improve Optimize data import latency.`?

HeleneW-dot: > Worth mentioning that this might cause issues when we have a lot of data or big variables? I remember Operate/Tasklist importers having issues with this and fail or get stuck when importing too much, does Optimize has a similar issue?

Hm, so I think this would be the `cost of higher memory use per fetch.`, but lets be more clear about it. How about: `..at the cost of higher memory use per fetch, so you may need to increase memory resources to handle larger amounts of data in each import fetch.`

ChrisKujawa: > If I am understanding correctly, I think I meant the same with This helps Optimize keep pace under high load and reduces import lag, but maybe that phrasing isnt clear enough.
> 
> What do you think about This helps Optimize keep pace under high load and reduces import lag, which can improve Optimize data import latency.?

No it is actually making import latency worse but increases throughput of importing. 


> How about: ..at the cost of higher memory use per fetch, so you may need to increase memory resources to handle larger amounts of data in each import fetch.

Yeah I think this is good to point out 👍🏼  Just to re-iterate over this Operate/Tasklist got stuck when the batches are too large, because ES didn't accepted such big requests. Is this something you experienced with Optimize as well in the past?

HeleneW-dot: >Just to re-iterate over this Operate/Tasklist got stuck when the batches are too large, because ES didn't accepted such big requests. Is this something you experienced with Optimize as well in the past?

I dont recall this specific scenario during my Optimize time, but Optimize import slowing down with large amounts of data is a recurring theme.

I will ping you on slack re. the phrasing instead of going back and forth here

HeleneW-dot: Updated as per slack convo 🙏

ChrisKujawa (approved):
Thanks @HeleneW-dot 👍🏼

PR #9021, opened 2026-06-08
feat(ai-ready): improve AI-ready workflows

afgambin:
## Description

This PR introduces some changes/improvements to our repo AI-ready workflows based on the AI-ready docs benchmark (https://agentdocsspec.com/spec/).

### Changes

**`static/llms.txt`**
- Adds a **Full content** section pointing to `/llms-full.txt` for RAG pipelines and offline indexing. Without this link, `llms-full.txt` is not discoverable by agents — it is not part of the llmstxt.org spec, so agents following the standard only look for `llms.txt`. The section is clearly labeled as not suitable for in-context reading to prevent agents from fetching multiple megabytes inline.

**`docusaurus.config.js`**
- Removes `addLinkTag: false` — not a recognized `docusaurus-plugin-llms` option (was silently ignored).

**`src/theme/Root.tsx`**
- Adds a visually-hidden `<div>` on every page so AI agents parsing HTML can discover `llms.txt` and the `.md` URL pattern. This is a body element (not a `<head>` link), which is what the benchmark checks for.

### Benchmark check addressed

| Check | Before | Expected after |
|---|---|---|
| `llms-txt-directive-html` | FAIL — no body directive | PASS — visually-hidden div present |

### Accepted tradeoffs

**`llms-txt-coverage` check**
`llms.txt` is intentionally a lean, curated entry point that routes agents to section-level indexes rather than listing all 3000+ pages directly. Agents fetch only the section they need, keeping token cost low. 

### Known limitations (follow-up work)

**`.md` URL support (52%) and `markdown-content-parity`**

The `.htaccess` content negotiation rule is correctly configured but only fires when `.md` files physically exist in the build output. `docusaurus-plugin-llms` determines output paths from source file paths, not Docusaurus routes:

- With `docsDir: "docs"` (next/unreleased version), files from `docs/guides/intro.md` land at `build/docs/guides/intro.md`, which accidentally matches current-version URLs ~52% of the time due to path overlap between the next version and 8.9.
- Changing `docsDir` to `versioned_docs/version-8.9` moves generated files to `build/versioned_docs/version-8.9/...` (wrong output path) and breaks URLs in section index files (`llms-guides.txt` etc.), making things worse.

A complete fix requires `docsDir: "versioned_docs/version-8.9"` + `pathTransformation` to correct the section index URLs + a post-build plugin to move `.md` files from `build/versioned_docs/version-8.9/` to `build/docs/`. The effort-to-impact ratio is low given the minor content drift between versions, so this is deferred until `docusaurus-plugin-llms` supports versioned docs path remapping natively.

- **`llms-txt-directive-md`** — adding a blockquote to each generated `.md` file requires a remark plugin or plugin enhancement; out of scope.
- **`markdown-content-parity`** — 6 pages have content drift between `.md` (generated from `docs/` next version) and HTML (served from `versioned_docs/version-8.9/`). Same root cause as above.
- **`page-size`** — 2 pages exceed 100K chars; needs content restructuring.


src/theme/Root.tsx
```diff
@@ -172,6 +172,24 @@ function VersionNotification() {
...
+          margin: "-1px",
+          overflow: "hidden",
+          clip: "rect(0,0,0,0)",
+          whiteSpace: "nowrap",
+          border: 0,
+        }}
+      >
+        For the complete documentation index, see{" "}
+        <a href="/llms.txt">llms.txt</a>. Markdown versions of pages are
+        available by appending .md to any documentation URL.
```
alexronquillo: Does this actually work? I tried this:

- https://docs.camunda.io/docs/next/self-managed/quickstart/developer-quickstart/c8run/install-start/
- https://docs.camunda.io/docs/next/self-managed/quickstart/developer-quickstart/c8run/install-start.md

afgambin: No, it doesn’t. MD availability only works for the current release atm; e.g., this works:

https://docs.camunda.io/docs/self-managed/quickstart/developer-quickstart/c8run/install-start.md

I forgot to remove that sentence. I initially wanted to tackle exactly this problem, but it’s not a trivial fix when using `docusaurus-plugin-llms`, so I’ll investigate it further in follow-up work (I added some context in the PR description).

zaycodes (approved):
LGTM as per https://agentdocsspec.com/spec/#llms-txt-directive-html!

PR #9026, opened 2026-06-08
docs: warn about getMessageKeys() removal in 8.8.22, 8.7.27, and 8.9.1

lenaschoenburg:
## Description

Camunda 8.8.22, 8.7.27, and 8.9.1 unintentionally removed the `getMessageKeys()` method (and the underlying `messageKeys` field) from the public `MessageBatchRecordValue` exporter record. Custom exporters that call `getMessageKeys()` on message batch records fail to compile against, or throw a `NoSuchMethodError` at runtime with, the updated `zeebe-protocol` dependency after upgrading to those patches or any later patch on the same line. The built-in exporters are unaffected. The regression and its fix are tracked in camunda/camunda#54823.

This documents the breaking change in the release announcements of each affected line, following the existing **"8.x.x patch releases"** table convention rather than a top-level banner. Each line gets a `{version}.x patch releases` table row plus a `Breaking change` detail entry:

- **8.8** (`880-announcements.md`): row added to the existing `8.8.x patch releases` table; detail entry under **APIs & tools**.
- **8.7** (`870-announcements.md`): new `8.7.x patch releases` table under **Key changes** plus a detail entry, matching the page's heading style.
- **8.9** (`890-announcements.md`): new `Key changes` > `8.9.x patch releases` table plus a `Breaking change` detail entry under **APIs & tools**.

The same entries are applied to every copy of each announcements page for cross-version consistency: the version-matched page (`version-8.7`, `version-8.8`, `version-8.9`), the copies carried forward into newer version trees, and the `next`/`docs` tree (9 files total).

## PR Checklist

- [ ] I added a DRI, team, or delegate as a reviewer for technical accuracy and grammar/style.

### DIFF
--- a/docs/reference/announcements-release-notes/870/870-announcements.md
+++ b/docs/reference/announcements-release-notes/870/870-announcements.md
@@ -46,6 +46,22 @@ Following the end-of-life of macOS 12, support for Desktop Modeler on macOS 12 h
 
 Collectively, these changes consolidate overlapping functionality, align configuration and client behavior across components, and establish clearer upgrade paths for upcoming releases.
 
+### 8.7.x patch releases
+
+The following key change was also released as part of an 8.7.x patch release.
+
+| Patch release                                                    | Type            | Key change                                                                                              |
+| :--------------------------------------------------------------- | :-------------- | :------------------------------------------------------------------------------------------------------ |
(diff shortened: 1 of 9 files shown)


alexronquillo (approved):
Thanks @lenaschoenburg! I pushed one small change that moved the 8.8.22 patch note to the top of the table since we usually display releases in desc order. Other than that, looks good to me! 🚀

Issue #9038, opened 2026-06-09
Document Camunda Hub Catalog


marstamm:
The rough outline should include  
  
 - Add a new Section in the Camunda Hub Docs for Catalog
 - Explain the supported asset types (Element templates)
 - Add a getting started guide
   - Explain external source of truth, e.g. Git
   - Structure of external repo + metadata needed
   - CI to gather and push to Hub
   - Authentication and Endpointsfor SaaS and SM
 - Add a usage guide for Assets in Hub
   - Discovery in Catalog
   - Discovery in Canvas
   - View for CoE or Administrators

alexronquillo:
FYI: I've already prepped a catalog section in the new Hub docs.

- https://docs.camunda.io/docs/next/components/hub/organization/manage-catalog/
- https://github.com/camunda/camunda-docs/tree/main/docs/components/hub/organization/manage-catalog

PR #9048, opened 2026-06-10
docs(hub): document the Catalog (getting started + usage)

marstamm:
## Description

Documents the Camunda Hub **Catalog** feature in the planned/final 8.10 state (not the current WIP). Builds on the existing `organization/manage-catalog` section prepped by @alexronquillo.

### What's included
- **Manage the catalog** landing — what the Catalog is, supported asset types (element templates only in 8.10), and persona→role mapping (CoE = org administrators; delivery teams = users with the Modeler role).
- **Get started with the Catalog** — collection repository structure + README metadata, two-stage CI/CD, and the ingestion contract for **SaaS and Self-Managed**:
  - `PUT /api/v2/catalog/assets/ingestion`, `multipart/form-data` (`template` + `readme` parts paired via README frontmatter)
  - `204` on success; transactional validation (`400`/`401`/`404`)
  - example collection repo: assets, GitHub Actions workflow, multipart sync script
- **Use Catalog assets in Hub** — Catalog discovery, applying assets in Canvas, and the CoE/administrator view (unpublish vs. delete lifecycle).
- Sidebar wired under **Manage the catalog**.

### Acceptance criteria (#9038)
- [x] New Catalog section in Hub docs
- [x] Supported asset types (element templates)
- [x] Getting started guide (Git source, repo structure + metadata, CI, SaaS + SM auth/endpoints)
- [x] Usage guide (Catalog discovery, Canvas discovery, CoE/admin view)

### Notes for reviewers
- API contract verified against the merged backend (`PubCatalogController`, `CatalogAssetSubmissionResolver`) and the approved specs.
- Documents the **planned `/api/v2`** path. The endpoint currently ships under `/api/experimental` behind the `catalog-enabled` feature flag — not GA yet.
- The browse/Canvas/CoE **UI is still being built** (frontend wiring is a follow-up per spec). Those flows are described as planned behavior under an admonition; labels/screens may change. Screenshots TBD.

Refs: camunda/camunda-docs#9038


marstamm:
@alexronquillo , As promised, I added content for the Catalog docs. Feel free to review (or adjust) as needed :)

docs/components/hub/organization/manage-catalog/getting-started.md
```diff
@@ -0,0 +1,318 @@
...
+```
+
+Each asset directory contains exactly two files:
+
+- **README.md**: A Markdown file with metadata in the frontmatter and a description in the body. The frontmatter references the single element template file and provides structured attributes like category and tags.
+- **Element template file** (`.json`): The element template descriptor. The `id` and `version` fields in this file are authoritative for the asset's identity and version in the Catalog.
+
+### Define the asset README
+
+The README serves as the asset's metadata file. Hub detects an asset by the presence of a README with valid frontmatter:
```
alexronquillo: It's unclear what "valid front matter" means in this context. I suggest we provide a reference table with all valid front matter fields, including:

- name
- type
- description
- required/optional
- default (if applicable)

docs/components/hub/organization/manage-catalog/getting-started.md
```diff
@@ -0,0 +1,318 @@
...
+
+### Usage
+
+Apply this template to a Service task to configure a payment operation.
+```
+
+The Markdown content below the frontmatter is displayed as the asset description when browsing the Catalog in Hub.
+
+:::info
+All files referenced in the metadata (such as the element template) must be included in the API submission. If a referenced file is missing, the submission fails.
```
alexronquillo: Previously, the doc states "Each asset directory contains exactly two files" This sounds contradictory. For example, we mention an asset icon elsewhere, I assume the user will reference a local image file in that case. 

Therefore, I think we should update the other statement to allow for referenced files. Also, if I recall correctly, there was a discussion about how to handle some types of referenced files, such as videos. I suggest we answer as many of these questions up front as possible.

docs/components/hub/organization/manage-catalog/getting-started.md
```diff
@@ -0,0 +1,318 @@
...
+:::
+
+### Version your element templates
+
+The Catalog uses the `id` and `version` fields inside your element template JSON to track versions:
```
alexronquillo: Currently, the [docs](https://docs.camunda.io/docs/next/components/modeler/element-templates/defining-templates/#optional-keys) say `version` is optional. Is it required in this context? Or does the first submission use a default version if omitted? We should clarify this.

docs/components/hub/organization/manage-catalog/getting-started.md
```diff
@@ -0,0 +1,318 @@
...
+  "name": "Payment Connector",
+  "version": 2,
+  "appliesTo": ["bpmn:ServiceTask"],
+  "properties": []
+}
+```
+
+When you submit assets to the Catalog API, Hub compares each submitted template against the latest stored version with the same `id`:
+
+- If the template **content is identical** to the latest stored version, the asset is left unchanged (no new version is created).
```
alexronquillo: I can imagine someone might want to update the markdown file, but not the template. Since the template hasn't changed, it sounds like the asset (including the metadata) is left unchanged. What should happen in this case?

docs/components/hub/organization/manage-catalog/getting-started.md
```diff
@@ -0,0 +1,318 @@
...
+- If the template **content has changed** and the `version` field has been **bumped**, a new version is published and becomes the new latest version.
```
alexronquillo: What happens if the version field decreases instead of increases, but is still unique?

docs/components/hub/organization/manage-catalog/getting-started.md
```diff
@@ -0,0 +1,318 @@
...
+```
+PUT <web-modeler-api-base-url>/v2/catalog/assets/ingestion
+```
+
+The request body is `multipart/form-data` and represents the **complete desired state** of your Catalog. For each asset, the request includes two parts:
+
+- a `template` part containing the element template `.json` file, and
+- a `readme` part containing the `README.md` file.
+
+Each `README.md` is paired with the template it references through the `template:` value in its frontmatter, resolved relative to the README's directory. Because the submission is the full desired state, Hub **unpublishes** any asset that exists in the Catalog but is absent from the submission (see [Step 3](#step-3-handle-asset-lifecycle)).
```
alexronquillo: > resolved relative to the README's directory

This sounds a little bit like you can use a relative path to a template file. I suggest we simplify this and remove ambiguity with something like "The template must be in the same directory as the README".

docs/components/hub/organization/manage-catalog/getting-started.md
```diff
@@ -0,0 +1,318 @@
...
+
+**1. Provide credentials and configuration**
+
+The sync script reads its configuration from environment variables. Store the client credentials securely (for example, as CI/CD secrets) and expose them — together with the environment-specific URLs — as environment variables. The token issuer and the Web Modeler API base URL differ between SaaS and Self-Managed:
```
alexronquillo: We haven't mentioned the sync script before this. I think we can simplify:


```suggestion
Store the client credentials securely, and expose them with the environment-specific URLs as environment variables. The token issuer and the Web Modeler API base URL differ between SaaS and Self-Managed:
```

docs/components/hub/organization/manage-catalog/getting-started.md
```diff
@@ -0,0 +1,318 @@
...
+export CAMUNDA_CONSOLE_CLIENT_ID="<client-id>"
+export CAMUNDA_CONSOLE_CLIENT_SECRET="<client-secret>"
+export CAMUNDA_OAUTH_URL="http://localhost:18080/auth/realms/camunda-platform/protocol/openid-connect/token"
+export CAMUNDA_CATALOG_BASE_URL="http://localhost:8070/api/v2"
+```
+
+</TabItem>
+</Tabs>
+
+See [Web Modeler API authentication](/apis-tools/web-modeler-api/authentication.md) for how to create credentials with the required `create` and `update` permissions.
```
alexronquillo: Should we point this at the new Camunda Hub API docs instead of the deprecated Web Modeler API docs? The authentication instructions differ between [saas](https://docs.camunda.io/docs/next/apis-tools/hub-api-saas/authentication/) and [sm](https://docs.camunda.io/docs/next/apis-tools/hub-api-sm/authentication/).

docs/components/hub/organization/manage-catalog/getting-started.md
```diff
@@ -0,0 +1,318 @@
...
+
+**2. Run the sync script**
+
+The following script obtains an access token, discovers each asset directory (a `README.md` plus a `.json` template), builds the multipart request, and submits the full desired state to the ingestion endpoint. Run it from the root of your asset repository whenever the templates change.
```
alexronquillo: If and when we change this script in the future to account for other types of asset files, all repos with this script will need to be updated. Is there a better way to provide this file or check its compatibility with the latest supported assets than copying and pasting this codeblock?

docs/components/hub/organization/manage-catalog/getting-started.md
```diff
@@ -0,0 +1,318 @@
...
+
+:::note
+The Catalog syncs from a **single repository**. If your element templates already live together in one repository (for example, a monorepo), sync from it directly. If they are spread across several repositories, first consolidate them into one — see [Sync assets from multiple repositories](#sync-assets-from-multiple-repositories).
+:::
+
+## Step 1: Organize your assets
+
+Point the Catalog at a single git repository that holds your element templates. This can be an existing repository (such as a monorepo) or a dedicated one. The sync only looks at element templates and their README files; any other files in the repository — job workers, BPMN processes, forms, DMN decisions — are ignored.
+
+Each asset is grouped in its own directory:
```
alexronquillo: Though the example has the asset directories at the root of the repo, I don't think it's clear from this text that they have to be. If they aren't, we later need to adjust the explanation of where to run the sync script.

docs/components/hub/organization/manage-catalog/getting-started.md
```diff
@@ -0,0 +1,318 @@
...
+If the submission is invalid, the request fails and **no** changes are applied (the ingestion is rolled back as a whole):
+
+| Status             | Meaning                                                                                                                       |
+| ------------------ | ----------------------------------------------------------------------------------------------------------------------------- |
+| `204 No Content`   | Ingestion succeeded. New and updated assets are published; assets absent from the submission are unpublished.                 |
+| `400 Bad Request`  | Validation failed — for example, invalid template JSON, changed content without a version bump, or a duplicate template `id`. |
+| `401 Unauthorized` | No valid access token was provided.                                                                                           |
+| `404 Not Found`    | The token is missing the required `create`/`update` permissions.                                                              |
+
+## Sync assets from multiple repositories
```
alexronquillo: To me, the placement of this section feels awkward. Does it make sense to move this to its own file? I suggest we phrase this guide with more opinionated instructions (single repo) and link to a separate doc for exceptions.

docs/components/hub/organization/manage-catalog/getting-started.md
```diff
@@ -0,0 +1,318 @@
...
+
+if [[ "${HTTP_STATUS}" == "204" ]]; then
+  echo "Catalog sync completed successfully."
+else
+  echo "Catalog sync failed (HTTP ${HTTP_STATUS})."
+  exit 1
+fi
+```
+
+### What the API returns
```
alexronquillo: Do we know when the API docs are coming? I wonder if we even need this section when we have the API docs.

docs/components/hub/organization/manage-catalog/getting-started.md
```diff
@@ -0,0 +1,318 @@
...
+
+## Step 3: Handle asset lifecycle
+
+### Unpublished assets
+
+When you submit assets to the Catalog, any asset that **already exists in Hub but is not included in the submission** is automatically unpublished.
+
+Unpublishing an asset means:
+
+- **For delivery teams** (Web Modeler users): The asset is no longer discoverable in the Catalog and cannot be applied to new diagrams. Projects that already use the template continue to work as before.
```
alexronquillo: > The asset is no longer discoverable in the Catalog and cannot be applied to new diagrams.

Does this only apply to unpublished assets or do older versions of assets also behave this way? In other words, delivery teams can only find and use latest versions.

docs/components/hub/organization/manage-catalog/getting-started.md
```diff
@@ -0,0 +1,318 @@
...
+- **For CoE members** (organization administrators): The asset stays visible so the CoE can track its remaining usage — which diagrams still reference the template — and drive the migration of those diagrams to a newer version or a different template before the asset is removed entirely.
+
+:::info
+Unpublishing is different from deprecation. Deprecation is a property within the element template file itself. Unpublishing is a Catalog-level state that indicates the asset is no longer part of the current submission.
```
alexronquillo: Maybe it would be valuable to include example use cases for each to help illustrate the difference?

docs/components/hub/organization/manage-catalog/getting-started.md
```diff
@@ -0,0 +1,318 @@
...
+:::
+
+### Deleting assets
+
+To permanently remove an asset from the Catalog, use the delete API. Deleting an asset removes it entirely — CoE members can no longer browse it, and element template properties are no longer accessible in diagrams.
```
alexronquillo: Can we provide more details about what happens to diagrams that are currently using the assets?

docs/components/hub/organization/manage-catalog/getting-started.md
```diff
@@ -0,0 +1,318 @@
+---
+id: getting-started-catalog
+title: Get started with the Catalog
```
alexronquillo: I'm not sure if "Get started" is an appropriate title for this guide. This covers the full breadth of capabilities, including unpublishing and permanent deletion. I would consider either renaming it to something more suitable or moving the other lifecycle the content to its own, dedicated guide(s).

docs/components/hub/organization/manage-catalog/catalog-usage.md
```diff
@@ -0,0 +1,59 @@
+---
+id: catalog-usage
+title: Use Catalog assets in Hub
+sidebar_label: Use Catalog assets
+description: "Discover and apply published Catalog assets in the Hub Catalog and Canvas, and manage assets as a Center of Excellence."
+---
+
+Once your Center of Excellence (CoE) has [synced element templates to the Catalog](/components/hub/organization/manage-catalog/getting-started.md), delivery teams discover and apply them while modeling, and the CoE manages their lifecycle.
```
alexronquillo: Since the target user for this guide is the delivery team, we should move it from the Organization level to the Workspace level documentation. Consider here: Workspace > Model business processes > Model

docs/components/hub/organization/manage-catalog/catalog-usage.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Discover assets in the Catalog
+
+Delivery team members browse the Catalog in Hub to find approved, ready-to-use element templates. For each asset, the Catalog shows its name, short description, category, tags, latest version, and icon.
```
alexronquillo: I don't see how to supply an icon in the other guide.

docs/components/hub/organization/manage-catalog/catalog-usage.md
```diff
@@ -0,0 +1,59 @@
...
+
+You can narrow the list to find the right asset:
+
+- **Free-text search** matches against the asset name, short description, category name, and tag names (case-insensitive).
+- **Filter by category** to scope the list to a single area (for example, _Connectors_ or _Human Tasks_).
+- **Filter by tags** to combine related assets across categories.
+- **Sort** by name, creation date, or last updated date.
+
+Assets that were recently added or updated are highlighted so teams can spot new templates at a glance. Only published assets appear in the Catalog — unpublished assets are never shown to delivery teams.
```
alexronquillo:
```suggestion
Assets that were recently added or updated are highlighted so teams can spot new templates at a glance. Only published assets appear in the Catalog — unpublished assets are hidden from delivery teams.
```

"never" doesn't feel right, considering an asset is only unpublished _after_ it was once already published and, therefore, visible. In fact, unpublished assets may actually be actively used if they were applied before unpublication.

docs/components/hub/organization/manage-catalog/catalog-usage.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Apply assets in Canvas
+
+To use a Catalog asset while modeling:
+
+1. Open a BPMN diagram in the Hub modeler (Canvas).
```
alexronquillo: Is "Canvas" the official name for "Hub modeler"? I think we should pick a name and use it in an official capacity rather than using both.

docs/components/hub/organization/manage-catalog/getting-started.md
```diff
@@ -0,0 +1,318 @@
...
+## Step 3: Handle asset lifecycle
+
+### Unpublished assets
+
+When you submit assets to the Catalog, any asset that **already exists in Hub but is not included in the submission** is automatically unpublished.
+
+Unpublishing an asset means:
+
+- **For delivery teams** (Web Modeler users): The asset is no longer discoverable in the Catalog and cannot be applied to new diagrams. Projects that already use the template continue to work as before.
+- **For CoE members** (organization administrators): The asset stays visible so the CoE can track its remaining usage — which diagrams still reference the template — and drive the migration of those diagrams to a newer version or a different template before the asset is removed entirely.
```
alexronquillo: We should also include a guide for tracking asset usage.

docs/components/hub/organization/manage-catalog/catalog-usage.md
```diff
@@ -0,0 +1,59 @@
...
+2. Select an element and apply a published element template from the Catalog.
+3. Configure the element through the properties panel as defined by the template.
+
+When the CoE publishes a newer version of a template you already use, Hub offers an **update** path on elements that reference the older version, so you can move to the latest version on your terms.
+
+## Manage the Catalog as a Center of Excellence
```
alexronquillo: I think we should remove this, as it's already covered in the other guide, and this guide targets the delivery team.

alexronquillo (changes requested):
Thanks for the submission! I've done a first pass, only looking at the content structure and accuracy. I left a lot of questions, so please feel free to reach out if anything doesn't make sense.

Generally, we should keep a few things in mind:

- **Target audience:** Who is the information for? CoE or delivery teams? As far as I understand, these personas are distinct and non-overlapping, so the content shouldn't mix them either.
- **Clarity:** If there are things users might ask questions about, we should go ahead and answer those in the docs.

When you've addressed this pass, I'll move to a language pass, and we can get this merged! 👍

docs/components/hub/organization/manage-catalog/getting-started.md
marstamm: I will add a table with the required/allowed fields. Currently, there are only 3 allowed properties (link to the asset, category and tags). All other information such as title and icon are directly taken from the Element template JSON

docs/components/hub/organization/manage-catalog/getting-started.md
marstamm: True, this is not clear here. We will only allow Markdown + Element Template for this iteration. Images or Videos in the readme are not supported. I will adjust accordingly.

docs/components/hub/organization/manage-catalog/getting-started.md
marstamm: We restrict the element templates allowed in the Catalog ingest step further. We require a version field to exists, as we assume that element templates will be updated over time. This makes the update path easier to understand by having all versions labeled from the start .

docs/components/hub/organization/manage-catalog/getting-started.md
marstamm: That is possible, only an update to the Element template triggers a new version. The Markdown description and metadata can change

docs/components/hub/organization/manage-catalog/getting-started.md
marstamm: It's rejected by the API - the new version needs to be higher than the last submitted

docs/components/hub/organization/manage-catalog/getting-started.md
marstamm: > This sounds a little bit like you can use a relative path to a template file

It is possible to do this. For simplicity in the getting started guide, we can simplify to "in the same directory" and have the details be in the ingestion API spec

docs/components/hub/organization/manage-catalog/getting-started.md
marstamm: Agreed.

docs/components/hub/organization/manage-catalog/getting-started.md
marstamm: Yes, we should! Good catch

docs/components/hub/organization/manage-catalog/getting-started.md
marstamm: We will provide an example repo with Ingest script and example action instead. We can link to it from here

docs/components/hub/organization/manage-catalog/getting-started.md
marstamm: The files can be anywhere. Even if the element templates + readme are not at root, the script will pick them up. We can clarify this

docs/components/hub/organization/manage-catalog/getting-started.md
marstamm: Yes, let's split this up

docs/components/hub/organization/manage-catalog/getting-started.md
marstamm: I'll check with the team on when this is ready. I agree that this should be in the API specs

marstamm: Seems like we already have it! 
https://docs.camunda.io/docs/next/apis-tools/hub-api-sm/specifications/ingest-catalog-assets/

docs/components/hub/organization/manage-catalog/getting-started.md
marstamm: Delivery Teams can only find the **latest, published** version. So yes, this is the same behavior. Never thought about it this way 😅

I'll adjust

docs/components/hub/organization/manage-catalog/getting-started.md
marstamm: Yes. Diagrams will treat it as a missing template: https://docs.camunda.io/docs/components/modeler/desktop-modeler/element-templates/using-templates/#missing-templates

I'll link that page from here

docs/components/hub/organization/manage-catalog/getting-started.md
marstamm: Let's move the lifecycle content in a dedicated page/guide

docs/components/hub/organization/manage-catalog/catalog-usage.md
marstamm: Agreed 👍

docs/components/hub/organization/manage-catalog/catalog-usage.md
marstamm: It is reused from the element template json. We'll make it clearer in the guide

docs/components/hub/organization/manage-catalog/catalog-usage.md
marstamm: I think Hub modeler is the term used in the docs, and canvas the internal team name. Let's call it Hub modeler

docs/components/hub/organization/manage-catalog/getting-started.md
marstamm: This will be done in the follow-up epic https://github.com/camunda/product-hub/issues/3490 and was intentionally left out of iteration. Please let me know if you disagree with this decision

docs/components/hub/organization/manage-catalog/catalog-usage.md
marstamm: Agreed 👍

docs/components/hub/organization/manage-catalog/manage-asset-lifecycle.md
```diff
@@ -0,0 +1,55 @@
...
+
+Before you start, make sure you have [set up the Catalog](/components/hub/organization/manage-catalog/getting-started.md) and can submit assets through your CI/CD pipeline.
+
+## How delivery teams see asset versions
+
+Delivery teams (Web Modeler users) only ever see the **latest published version** of an asset in the Catalog. Older versions are not offered for new diagrams, and unpublished assets are not discoverable at all. Diagrams that already reference a template keep working — the change only affects what teams can newly apply.
+
+## Unpublish an asset
+
+When your CI/CD pipeline submits assets to the Catalog, the submission represents the **full desired state**. Any asset that **already exists in Hub but is not included in the submission** is automatically unpublished. To unpublish an asset, remove its directory from the repository you sync from and run the pipeline again.
```
alexronquillo: Make sure to cover: 

> Removing an element template from submission will not only unpublish the asset, but also deprecate the latest element template version -- ([source](https://camunda.slack.com/archives/C09L79R5AJX/p1781770093112169))

marstamm:
@alexronquillo , I adjusted the PR with the changes requested. I deferred asset usage Documentation for now - we will pick this up in the next epic

docs/components/hub/organization/manage-catalog/getting-started.md
alexronquillo: I agree with your decision 👍

alexronquillo:
@marstamm I am almost done with my language review. I can address these copilot comments on your behalf

alexronquillo (dismissed):
@marstamm Thanks for the draft! I've made some structural changes and line edits throughout. If everything looks ok to you, feel free to merge 👍

docs/components/hub/organization/manage-catalog/getting-started.md
```diff
@@ -0,0 +1,230 @@
...
+]}>
+
+<TabItem value='saas'>
+
+```bash
+export CAMUNDA_HUB_CLIENT_ID="<client-id>"
+export CAMUNDA_HUB_CLIENT_SECRET="<client-secret>"
+export CAMUNDA_OAUTH_URL="https://login.cloud.camunda.io/oauth/token"
+export CAMUNDA_HUB_OAUTH_AUDIENCE="api.cloud.camunda.io"
+export CAMUNDA_HUB_REST_URL="https://hub.cloud.camunda.io"
```
marstamm: While sanity-checking the Guide, I noticed that the Variable names generated by Console are different here:

Image

Should we re-use the env names that are generated from console, even when they do not have the HUB prefix?  We would only need the CAMUNDA_HUB_REST_URL to be manually set


```suggestion
export CAMUNDA_CONSOLE_CLIENT_ID="<client-id>"
export CAMUNDA_CONSOLE_CLIENT_SECRET="<client-secret>"
export CAMUNDA_OAUTH_URL="https://login.cloud.camunda.io/oauth/token"
export CAMUNDA_CONSOLE_OAUTH_AUDIENCE="api.cloud.camunda.io"
export CAMUNDA_HUB_REST_URL="https://hub.cloud.camunda.io"
```

docs/components/hub/organization/manage-catalog/getting-started.md
```diff
@@ -0,0 +1,230 @@
...
+
+<TabItem value='self-managed'>
+
+In Self-Managed, tokens are issued by your [Management Identity](/self-managed/components/management-identity/authentication.md) instance. There is no `audience`, and the Camunda Hub API is served from your own installation, which defaults to `http://localhost:8088`. Adjust the URLs to match your installation:
+
+```bash
+export CAMUNDA_HUB_CLIENT_ID="<client-id>"
+export CAMUNDA_HUB_CLIENT_SECRET="<client-secret>"
+export CAMUNDA_OAUTH_URL="http://localhost:18080/auth/realms/camunda-platform/protocol/openid-connect/token"
+export CAMUNDA_HUB_REST_URL="http://localhost:8088"
```
marstamm: Same as above

```suggestion
export CAMUNDA_CONSOLE_CLIENT_ID="<client-id>"
export CAMUNDA_CONSOLE_CLIENT_SECRET="<client-secret>"
export CAMUNDA_OAUTH_URL="http://localhost:18080/auth/realms/camunda-platform/protocol/openid-connect/token"
export CAMUNDA_HUB_REST_URL="http://localhost:8088"
```

marstamm (commented):
The content looks good! One note on the Environment variable I'd like to clarify before we merge below.

@alexronquillo what is the convention on this? Do we usually use what is created in the UI? I will adjust the Code in the example repo according to the decision we make here

docs/components/hub/organization/manage-catalog/getting-started.md
alexronquillo: Let's go with what's in the UI for now, even if it will change before the 8.10 release. Thanks for calling this out 👍

marstamm: Adjusted ✅

marstamm:
Everything looks good, ready to merge from my side :)

docs/components/hub/organization/manage-catalog/getting-started.md
alexronquillo: Just to clarify: Are these API auth pages also affected? https://preview.docs.camunda.cloud/pr-9048/docs/next/apis-tools/hub-api-saas/authentication/#generate-a-token

marstamm: I think so, at least these are the values I get on current Dev SaaS. I don't know if we plan to rename these before the release

alexronquillo (approved):

PR #9135, opened 2026-06-18
docs: warn that custom exporters must acknowledge record positions

lenaschoenburg:
Adds a warning to the Camunda exporters concept page, in the log stream compaction section, explaining that a custom exporter which never acknowledges record positions prevents log compaction. Because compaction only advances to the lowest acknowledged position across all exporters, one exporter that never advances its position pins the log indefinitely until the disk fills and the broker stops.

This came out of a support incident where a customer's custom exporter did not call `controller.updateLastExportedRecordPosition(...)`, which silently pinned the log and eventually filled the disk. The mechanism was already documented in scattered low-emphasis notes, but the operational consequence was not called out where the compaction behavior is explained. The warning also reminds operators to verify acknowledgment works and to check disk (PVC) sizing before activating a new exporter, since a freshly configured exporter temporarily becomes the lowest acknowledged position until it catches up.

### DIFF
--- a/docs/self-managed/concepts/exporters.md
+++ b/docs/self-managed/concepts/exporters.md
@@ -40,6 +40,19 @@ If no exporters are configured, Zeebe automatically deletes data when it's no lo
 
 :::
 
+:::warning
+A custom exporter that does not acknowledge record positions prevents log compaction and can take down the broker.
+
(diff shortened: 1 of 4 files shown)


ChrisKujawa (approved):
Thanks @lenaschoenburg 👍🏼 

could also link to https://camunda.github.io/zeebe-chaos/2026/06/18/Full-disk-due-to-soft-pause-exporters to show how system would behave when positions are not ack'ed

PR #9179, opened 2026-06-23
perf(apis-tools): bucket TypeScript SDK API reference type aliases in the sidebar to cut build size

lenaschoenburg:
## Description

The TypeScript SDK API reference (`apis-tools/typescript/api-reference`) is TypeDoc-generated and contains **~1,900 type aliases, each as its own doc page**. When a sidebar category sits on the active page's path, Docusaurus renders **every one of its siblings** into that page's static HTML. With a single flat **"Type Aliases"** category, each of the ~1,900 type-alias pages therefore embeds links to all the others — an **N×N blow-up**. The next-version section alone builds to **~1.2 GB** (~640 KB per page).

This PR adds a custom `sidebarItemsGenerator` (`src/tsApiSidebar.js`, wired up in `docusaurus.config.js`) that splits the "Type Aliases" category into **fixed-size alphabetical buckets** (50 per bucket). Each type-alias page then renders only its own bucket (~50 siblings) instead of all ~1,900.

**Why buckets (vs. alternatives I measured):**

| Approach | Section build size | Per-type deep links | Active-highlight + breadcrumbs |
| --- | --- | --- | --- |
| Current (flat category) | ~1.2 GB | ✅ | ✅ |
| One page per module (TypeDoc `router: "module"`) | ~16 MB | ❌ (one 10.6 MB page, URLs change) | n/a |
| Prune type aliases from sidebar | ~101 MB | ✅ | ❌ (no sidebar) |
| Prune + `displayed_sidebar` | ~111 MB | ✅ | ❌ (disconnected nav) |
| **Bucketed split (this PR)** | **~193 MB** | ✅ | ✅ |

Bucketing is the only option that is **camunda-docs-only, keeps every existing page URL, and preserves full navigation** (active-item highlight + breadcrumbs), because the pages stay in the sidebar tree. Measured per-page output drops from ~640 KB to ~85 KB median. Bucket size is near the `√N` optimum, so ~50 is about as small as the total gets.

The generator is scoped to the TypeScript SDK API reference only; every other autogenerated sidebar is returned unchanged. No page URLs change, so no redirects are required.

> [!NOTE]
> This touches docs build infrastructure (`docusaurus.config.js` + a new sidebar-generator module), so it likely needs a **`dx` / docs-infrastructure review**. Sidebar bucket labels are auto-generated alphabetical ranges (e.g. `ActivateAdHoc… – AdvancedMessa…`) — open to polishing if reviewers prefer a different scheme.

## PR Checklist


christinaausley:
Not sure if this was ready for review but went ahead and gave some thoughts if you want to take a look:

Reviewed and verified the numbers by building both ways - everything checks out, and the win is bigger than you claimed, so that's good 😄 

| | You said | I measured |
| --- | --- | --- |
| Type aliases (next) | ~1,900 | 1,892 |
| Section before | ~1.2 GB | 1.2 GB |
| Per page before | ~640 KB | 656 KB |
| Section after | ~193 MB | 187 MB |
| Per page after | ~85 KB | 92 KB |

Perhaps worth adding to the description: because `sidebarItemsGenerator` sits on the docs plugin options, this applies to every version not just Next right?. 8.9 has 1,732 type aliases and drops to 172 MB as well. Total build output is now 1.0 GB, when the Next section alone was 1.2 GB before.

I also simulated the bucketing against all 1,892 + 1,732 real type names looking for duplicate category labels - zero collisions in either version (38 and 35 buckets).

Two changes I pushed to `tsApiSidebar.js`:

1. A guard against silent regression-- because if TypeDoc ever renames the category - `"Type aliases"` is plausible, since our style guide uses sentence case for headings - the generator would quietly no-op and the 1.2 GB blow-up would come back with no signal. It now throws with a message naming the fix, and the label is extracted to a `TARGET_LABEL` constant. Verified it doesn't false-positive on a real build.
2. Defensive `item.items`. The recursive branch used `item.items || []` but the match branch read `item.items.length` directly, so a category with a `link` and no `items` would throw on one path only. Both now use the guarded value 👍 

One thing I'd push back on since you asked for input on the label scheme... the 16-character truncation makes adjacent boundaries ambiguous in my opinion. Bucket 1 ends `AdvancedMessage…` and bucket 2 starts `AdvancedMessage…`, so someone looking for `AdvancedMessageSubscription` can't tell which to open and TypeDoc names share long prefixes constantly. Would you consider truncating to the shortest *distinguishing* prefix instead of a fixed width? Widening to ~22 would help but wouldn't guarantee it. I left it alone since it's a design call, not a defect.

Agree this needs `dx` review before it leaves draft - I've verified behavior not `docusaurus.config.js` conventions.

PR #9180, opened 2026-06-24
Use camelCase for Elasticsearch/OpenSearch exporter configuration keys

ChrisKujawa:
## Description

The Elasticsearch and OpenSearch exporter pages show `args` configuration keys in **kebab-case**. Before Camunda 8.10, exporter `args.*` keys in kebab-case are **silently ignored** — the broker starts with no error and the setting has no effect (the key never matches the camelCase Java field). kebab-case was only made to work in 8.10 via camunda/camunda#51307.

This PR converts the YAML configuration **examples** on these pages to **camelCase**, which binds correctly on **all supported versions** (8.10 accepts both kebab and camelCase; pre-8.10 accepts only camelCase). Values are unchanged.

## Changes

Converted YAML example keys from kebab-case to camelCase in:

- `docs/.../zeebe/exporters/elasticsearch-exporter.md` (current / 8.10)
- `docs/.../zeebe/exporters/opensearch-exporter.md` (current / 8.10)
- `versioned_docs/version-8.9/.../zeebe/exporters/elasticsearch-exporter.md`
- `versioned_docs/version-8.9/.../zeebe/exporters/opensearch-exporter.md`

Examples were the functional copy-paste hazard for 8.9 users. The descriptive **reference tables** and prose still list the canonical kebab option names — happy to convert those too if reviewers prefer full consistency.

## Why this matters

Found on a live 8.9.9 cluster: kebab-cased `variable-name-inclusion-start-with` / `optimize-mode-enabled` silently did nothing, so the Optimize-oriented ES-exporter filters never applied. Switching to camelCase fixed it (verified via the broker's `Exporter configured with …` log).

## Refs

- Bug (8.9 silent-ignore): camunda/camunda#55795
- 8.10 fix that added kebab support: camunda/camunda#51307

### DIFF
--- a/docs/self-managed/components/orchestration-cluster/zeebe/exporters/elasticsearch-exporter.md
+++ b/docs/self-managed/components/orchestration-cluster/zeebe/exporters/elasticsearch-exporter.md
@@ -51,7 +51,7 @@ camunda:
   data:
     exporters:
       elasticsearch:
-        class-name: io.camunda.zeebe.exporter.ElasticsearchExporter
+        className: io.camunda.zeebe.exporter.ElasticsearchExporter
         args:
         # Refer to the table below for the available args options
 ```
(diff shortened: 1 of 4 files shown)


afgambin (approved):
lgtm

PR #9183, opened 2026-06-24
docs(agentic-orchestration): add guide for adding a tool to an AI agent

### LINKED ISSUE #9046: Add guide: How to add a tool to an AI agent
## Summary

A walkthrough for adding tools to AI agents currently exists only as a blog post: https://camunda.com/blog/2025/05/guide-to-adding-tool-ai-agent/ (linked from the docs), but community members and contributors have flagged that this content belongs inside the official docs.

Tool definitions are a core part of the agentic orchestration experience — keeping the walkthrough in a blog post means it's less discoverable, harder to maintain, and not indexed by the docs search or MCP.

## Proposed work

Create a new guide inside the docs that covers:

- What a tool is in the context of an AI agent
- Step-by-step walkthrough for adding a tool to an agent (based on the existing blog post content)
- Keeping it preserved and searchable via docs search and MCP

Reference: 
- Tool definitions guide from the AI Agent connector: https://docs.camunda.io/docs/next/components/connectors/out-of-the-box-connectors/agentic-ai-aiagent-tool-definitions/

afgambin:
## Description

Adds a new dedicated guide, [Add a tool to an AI agent](https://github.com/camunda/camunda-docs/blob/af/9046-add-tool-to-ai-agent-guide/docs/components/agentic-orchestration/add-tool-to-ai-agent.md), inside the Agentic orchestration section of the docs. This brings the walkthrough that previously existed only as a [blog post](https://camunda.com/blog/2025/05/guide-to-adding-tool-ai-agent/) into the official docs, making it discoverable via docs search and MCP.

Closes #9046.

## PR Checklist


docs/components/agentic-orchestration/add-tool-to-ai-agent.md
```diff
@@ -0,0 +1,221 @@
...
+title: Add tools to an AI agent
+sidebar_label: Add tools to an AI agent
+description: "Add BPMN activities as callable tools to your AI agents."
+keywords: ["agentic ai", "AI agents", "tools", "fromAi", "toolCallResult"]
+---
+
+import Tabs from "@theme/Tabs";
+import TabItem from "@theme/TabItem";
+
+Add BPMN activities as callable tools to your AI agents.
```
afgambin: Is **“BPMN activities”** the right term here, or should we use **“call activities”** or just **“activities”**?

docs/components/agentic-orchestration/add-tool-to-ai-agent.md
```diff
@@ -0,0 +1,221 @@
...
+
+## What is a tool
```
afgambin: The idea of this new guide is to explain how to add a tool as a how-to guide, with isolated sections, rather than as a step-by-step tutorial like the getting started guide. So, the “What is a tool?” section feels more like conceptual content and does not really belong here.

I’m wondering whether this Tool definitions (https://docs.camunda.io/docs/next/components/connectors/out-of-the-box-connectors/agentic-ai-aiagent-tool-definitions/) guide should be moved to the Agentic Orchestration section. My idea would be to present it as the conceptual guide for tools: what they are, types, responses, and anything else that is not tied specifically to the connector.

In the connectors section, we would keep only the content related to connector implementation.

wdyt?

docs/components/agentic-orchestration/add-tool-to-ai-agent.md
```diff
@@ -0,0 +1,221 @@
...
+
+</TabItem>
+
+</Tabs>
+
+:::note
+The `toolCallResult` value can be a primitive string, a number, or a complex FEEL context object. Complex objects are serialized to JSON before being passed to the LLM. Prefer returning a structured FEEL context over a raw string when the result has multiple fields, as this gives the LLM more to work with when summarizing the outcome. If `toolCallResult` is not set or is empty after the tool executes, the AI Agent connector returns a constant success string to the LLM.
+:::
+
+## Example: add a REST connector tool
```
afgambin: This example is essentially copied from the getting started guide. Is there another example you think might be more relevant to showcase here?

I think adding an example, even though it is not the common pattern in a how-to guide, helps readers understand the guide content.

docs/components/agentic-orchestration/add-tool-to-ai-agent.md
maff: I think the generic term we use is "elements", similar to the BPMN docs: https://docs.camunda.io/docs/components/modeler/bpmn/. "Call activities" does not apply as that is explicitly an activity delegating to another process.

docs/components/agentic-orchestration/add-tool-to-ai-agent.md
maff: You're right that these pages overlap and I think we should consolidate them. The blurry part is that everything we document around tools is very much tied to the AI Agent connector as no other implementation uses these concepts in that form (e.g. the `toolCall.` and `toolCallResults` conventions). Also this page directly refers to the AI Agent connector.

For me it's currently a bit hard to jump to the right content as I'd need to know if I need to go to the agentic orchestration section  or to the AI Agent connector docs. From that angle I could imagine moving all/most content from the AI Agent connector docs to AO would make sense to have everything in one place.

docs/components/agentic-orchestration/add-tool-to-ai-agent.md
```diff
@@ -0,0 +1,231 @@
...
+- **Input parameters**: values the LLM must supply at call time, declared using the [`fromAi()`](/components/modeler/feel/builtin-functions/feel-built-in-functions-miscellaneous.md#fromaivalue) FEEL function in input mappings.
+- A **result**: the tool's output, returned to the LLM as `toolCallResult`.
+
+The AI Agent connector gathers all root-level activities in the ad-hoc sub-process (those with no incoming flows), builds a tool definition for each, and passes them to the LLM as part of the prompt. The LLM then decides which tool to call, in what order, and with which parameters.
+
+You can use any BPMN activity type as a tool:
+
+| Tool type            | When to use                                                                                                                                          |
+| :------------------- | :--------------------------------------------------------------------------------------------------------------------------------------------------- |
+| REST connector       | Call an external HTTP API.                                                                                                                           |
```
maff: 💭 REST connector is technically not a dedicated activity type, but a service task. Maybe we can frame it as "Connectors" with "REST connector" as example?

docs/components/agentic-orchestration/add-tool-to-ai-agent.md
```diff
@@ -0,0 +1,231 @@
...
+| MCP client connector | Expose tools from an external [MCP server](/components/connectors/out-of-the-box-connectors/agentic-ai-mcp-client.md) as gateway tools to the agent. |
+| Sub-process          | Model a multi-step sub-flow that the LLM triggers as a single tool.                                                                                  |
+
+:::tip
+For the full technical reference on how tool definitions are resolved and the complete `fromAi()` function syntax, see [AI Agent tool definitions](/components/connectors/out-of-the-box-connectors/agentic-ai-aiagent-tool-definitions.md).
+:::
+
+## Add an activity inside the ad-hoc sub-process
+
+Open your process in [Web Modeler](/components/hub/workspace/modeler/index.md) or [Desktop Modeler](/components/modeler/desktop-modeler/index.md).
```
maff: 💭 Do you think having an image/model here (similar to the one on https://docs.camunda.io/docs/next/components/connectors/out-of-the-box-connectors/agentic-ai-aiagent-tool-definitions/) could help understanding "elements without incoming flows"?

docs/components/agentic-orchestration/add-tool-to-ai-agent.md
```diff
@@ -0,0 +1,231 @@
...
+
+:::tip
+You can model a sub-flow inside the ad-hoc sub-process. Only the first activity in the sub-flow (the root node) is exposed to the LLM as a tool; the rest of the flow executes automatically once the LLM selects it.
+:::
+
+## Write a tool name and description
+
+The LLM selects tools based on the activity name and its **Documentation** field. Clear, specific descriptions significantly improve the reliability of tool selection.
+
+1. Give the activity a descriptive **name**. The element ID (derived from the name) is what the LLM receives as the tool name.
```
maff: 🔧 What is meant by "derived from the name"? ID/Name refers to the modeler ID (required) and Name (optional) properties. The ID is always used as "tool name" for the LLM. If documentation is set, it will be used as "tool description" - if the documentation is missing, the element name will be used as fallback (if configured). 

In the case below, the tool name is `SuperfluxProduct` with the documentation as tool description. If the documentation would be missing, the tool description would fall back to `Superflux Product Calculation`. If the element name would also be empty, the tool would have no description.

Image

The table on https://docs.camunda.io/docs/next/components/connectors/out-of-the-box-connectors/agentic-ai-aiagent-tool-definitions/#tool-definitions would already outline these details.

docs/components/agentic-orchestration/add-tool-to-ai-agent.md
```diff
@@ -0,0 +1,231 @@
...
+| Strong | `Resolve customer by company name` | `Use this tool when a document mentions a company and you need its internal customer ID. If multiple matches are returned, request human validation before continuing.` |
+
+A precise description makes the expected behavior explicit and reduces the risk of incorrect tool selection, repeated calls, or hallucinated behavior. Vague descriptions are the most common cause of unreliable agent behavior.
+
+## Declare AI-generated parameters with `fromAi()`
+
+If the tool requires values that the LLM should supply at runtime, such as a search query, a location, or an identifier, declare those parameters using the [`fromAi()`](/components/modeler/feel/builtin-functions/feel-built-in-functions-miscellaneous.md#fromaivalue) FEEL function in input mappings.
+
+The `fromAi()` function marks a value as LLM-provided and generates a JSON Schema parameter definition that is passed to the LLM as part of the tool definition. Always provide a description and type for every parameter, as this significantly reduces hallucinated values.
```
maff: ℹ️  The `type` can be omitted and will fall back to a string if not configured. I think we should document that accordingly to keep the FEEL expressions smaller and more manageable, only adding the type if needed.

docs/components/agentic-orchestration/add-tool-to-ai-agent.md
```diff
@@ -0,0 +1,231 @@
...
+
+| Source            | Target           |
+| :---------------- | :--------------- |
+| `= response.body` | `toolCallResult` |
+
+You can also map a nested field:
+
+| Source                   | Target                  |
+| :----------------------- | :---------------------- |
+| `= response.body.status` | `toolCallResult.status` |
```
maff: 🔧 This variant is actively discouraged by https://preview.docs.camunda.cloud/pr-9183/docs/next/components/concepts/variables/#output-mappings - I think we should either remove that note or only document the `put()` variant here, but with an e2e example writing back to `toolCallResult`. Something like (untested):

Source: `=context put(toolCallResult, "status", response.body.status)`
Target: `toolCallResult`

docs/components/agentic-orchestration/add-tool-to-ai-agent.md
```diff
@@ -0,0 +1,231 @@
...
+
+:::tip Avoid accidental overwrites with multiple outputs
+When your tool task has several output parameters that each contribute a field to `toolCallResult`, prefer nested targets like `toolCallResult.status` over mapping everything directly to `toolCallResult`. Mapping to `toolCallResult` directly replaces the entire variable.
+
+For script tasks, the FEEL `put` function adds a single key to an existing context without replacing it:
+
+```feel
+put(toolCallResult, "Complete", sendEmail)
```
maff: 🔧 Shouldn't this be [`context put`](https://docs.camunda.io/docs/next/components/modeler/feel/builtin-functions/feel-built-in-functions-context/#context-putcontext-key-value)?

docs/components/agentic-orchestration/add-tool-to-ai-agent.md
```diff
@@ -0,0 +1,231 @@
...
+:::
+
+</TabItem>
+
+<TabItem value="script-task">
+
+In a script task, set the variable directly in the script:
+
+```groovy
+toolCallResult = [status: "completed", id: customerId]
```
maff: 🔧 I think that does not work. Instead, the result variable of a script task needs to be set to `toolCallResult` with the expression providing the value.

docs/components/agentic-orchestration/add-tool-to-ai-agent.md
maff: Agree & I think the example is super helpful. Maybe we could pair it with an image so it's easier to imagine?

docs/components/agentic-orchestration/add-tool-to-ai-agent.md
```diff
@@ -0,0 +1,231 @@
...
+
+:::note
+The `toolCallResult` value can be a primitive string, a number, or a complex FEEL context object. Complex objects are serialized to JSON before being passed to the LLM. Prefer returning a structured FEEL context over a raw string when the result has multiple fields, as this gives the LLM more to work with when summarizing the outcome. If `toolCallResult` is not set or is empty after the tool executes, the AI Agent connector returns a constant success string to the LLM.
+:::
+
+## Example: add a REST connector tool
+
+This example adds a tool that fetches current weather conditions using the [Open-Meteo API](https://open-meteo.com/).
+
+1. Inside the ad-hoc sub-process, add a new task, apply the [REST Outbound connector](/components/connectors/protocol/rest.md) using the **Change element** menu, and name the task `Get current weather`.
```
maff: 🔧 As stated above, we should also use a proper ID here which is passed to the LLM (e.g. `Get_Current_Weather`).

docs/components/agentic-orchestration/design-architecture.md
```diff
@@ -72,34 +72,11 @@ This is a typical execution timeline:
 
 ### Define your agent tools
 
-In the AI agent model, each BPMN activity inside an ad-hoc sub-process is a tool exposed to the LLM. The activity name and its documentation are used by the LLM to decide what to do next.
+In the AI agent model, each BPMN activity inside an ad-hoc sub-process is a tool exposed to the LLM. The activity name and its **Documentation** field are used by the LLM to decide which tool to call, in what order, and with which parameters.
```
maff: 🔧 As above regarding ID, name, documentation (discrepancy has been there before).

afgambin:
@maff Thanks for your feedback! I’m putting this on hold for now since I’ve been assigned as release manager for alpha3 and will be acting as [team lead during Mark’s time off](https://camunda.slack.com/archives/C026U8GBNSW/p1783076308534409). I’ll get back to it in a few weeks.

docs/components/agentic-orchestration/add-tool-to-ai-agent.md
afgambin: Totally agree! I’ll keep it out of scope for this PR, but I’ve added it to this related issue: https://github.com/camunda/camunda-docs/issues/9238.

docs/components/agentic-orchestration/add-tool-to-ai-agent.md
afgambin: Yes, I’ve reworked the introduction a bit, moving the “What is a tool?” section to the Tool definitions guide until I move all that content to the AO. This keeps this guide focused on how-to instructions rather than conceptual information.

docs/components/agentic-orchestration/add-tool-to-ai-agent.md
afgambin: Updated

docs/components/agentic-orchestration/add-tool-to-ai-agent.md
afgambin: Added a clarification

docs/components/agentic-orchestration/add-tool-to-ai-agent.md
afgambin: Let's remove it then

docs/components/agentic-orchestration/add-tool-to-ai-agent.md
afgambin: Updated

docs/components/agentic-orchestration/add-tool-to-ai-agent.md
afgambin: Updated

docs/components/agentic-orchestration/add-tool-to-ai-agent.md
afgambin: To keep the guide as focused on the how-to as possible, I’m linking to the example in the getting started guide. And, in this new guide, I’m adding explanatory screenshots for each section, but they aren’t tied to a specific example

afgambin:
@maff @HanselIdes I have addressed all your comments, made more adjustments and added some screenshots. It is ready for another pass, thanks!

docs/components/agentic-orchestration/add-tool-to-ai-agent.md
```diff
@@ -0,0 +1,179 @@
...
+:::
+
+<img src={AddElement} alt="Add a task element to the AI agent ad-hoc sub-process" width="60%"/>
+
+## Write a tool name and description
+
+The LLM selects tools based on the tool element's **ID** and its **Documentation** fields:
+
+- The element's **ID** field is always used as the tool name.
+- The element's **Name** field is a human-readable label shown on the diagram.
```
nikonovd, line 50: 💭 As a new persona not knowing internal details, I'd likely be confused about the ID being used as the tool description instead of the name.

For LLMs this is actually the natural choice, but as a non-experienced person this could confuse me.

WDYT of adding some hint about that discrepancy?

docs/components/agentic-orchestration/add-tool-to-ai-agent.md
```diff
@@ -0,0 +1,179 @@
...
+import TabItem from "@theme/TabItem";
+import AddElement from './img/add-element.png';
+import InputMapping from './img/input-mapping.png';
+import ToolDescription from './img/tool-description.png';
+
+Add BPMN elements as callable tools to your AI agents.
+
+## About
+
+A tool is a BPMN element inside an [ad-hoc sub-process](/components/modeler/bpmn/ad-hoc-subprocesses/ad-hoc-subprocesses.md) that an [LLM](/reference/glossary.md#large-language-model-llm) can choose to invoke to complete a goal.
```
nikonovd: 💭 Isn't it also a flow of bpmn elements?

It's mentioned in the hint below. I'm asking myself, if we should point it out directly, that it's either a single BPMN element or a flow of those. WDYT?

nikonovd (commented):
I did a pass from my subjective opinion. Generally looks great, thanks for your contribution 🚀 

Besides the comments I've got two general questions:

1. This guide is rather generic. Although we show examples, unlike the other pages under `Build with AI` we don't guide the reader using our example process. IMO this could increase the value of this guide.

2. In combination with the point above, we could identify both a connector and a general task in an example to show the differences between their tool configs. 

3. The difference for the tool input definition with `fromAi()` combines two different approaches - direct parameters on element templates and input mappings. IMO we could separate those and show both on an example.
 
4. In addition, reading the last section about combination of tool call results, I would personally prefer to avoid mixing tasks and templated connector configurations in a single section. Instead we could split them up, following similar guides and highlight the disjunctions between those.

That's my food-for-thought. Happy to discuss or state more clearly.

docs/components/agentic-orchestration/add-tool-to-ai-agent.md
afgambin: Excellent point, I’ve refactored the description accordingly

docs/components/agentic-orchestration/add-tool-to-ai-agent.md
afgambin: Good point! I’ve added a note clarifying this

afgambin:
@nikonovd Thanks for the thorough review, helpful as always!

I’ve addressed your comments and points 3 and 4. Regarding points 3 and 4, both sections are now organized by element type using tabs: the `fromAi()` section separates element template fields from input mappings, while the `toolCallResult` combination section distinguishes between connector tasks, regular tasks, and script tasks, as the broader section it belongs to.

Regarding [points 1 and 2](https://github.com/camunda/camunda-docs/pull/9183#pullrequestreview-4826200176), I’d lean toward keeping the guide generic - that is, as a a how-to guide rather than a step-by-step tutorial - as we did with the [RAG with AI agents guide](https://docs.camunda.io/docs/next/components/agentic-orchestration/rag-ai-agents/) in the **Build with AI** section. My rationale is as follows:

- Where possible, we should reduce the maintenance burden and avoid dependencies on the getting started guide or specific examples. That’s why I decided to point readers to the getting started guide (admonition in the About section), where we already explain how to add a tool using a specific example.
- I think we should prioritize how-to guides over tutorials because they tend to be more useful, context-agnostic and with isolated sections, especially in the context of AI consumption: they generally cover all potential use cases, including optional actions, rather than focusing only on the specifics of a particular example (and if we provide an example for every use case in a how-to, we’ll end up with a very long guide that is difficult for readers to digest and follow - here's the compromise). 

We took a different approach with the guides about monitoring in Operate and analyzing in Optimize to showcase simple, quick actions users might take after getting started - more like an intro to the key features available across those platforms. However, those guides weren’t intended to describe every available action on those. That could be documented as a follow-up though - and we’re now doing (planning to) it for the new UI as part of the real-time agent visibility epic.

Let me know what you think / have another look when you can. Thanks!

HanselIdes (commented):
Reviewed with a navigation lens: each page should stay focused on one job, the component page as the deep dive and the guide as the user journey. Five things.

**1. The ID and Documentation behavior is right, and here is the source.** @maff's [correction](https://github.com/camunda/camunda-docs/pull/9183#discussion_r3512744440) matches the connector implementation. [`AdHocToolElement.documentationWithNameFallback()`](https://github.com/camunda/connectors/blob/main/connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/adhoctoolsschema/model/AdHocToolElement.java) returns the documentation when it is non-blank and otherwise falls back to the element name, and [`AdHocToolSchemaGeneratorImpl`](https://github.com/camunda/connectors/blob/main/connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/adhoctoolsschema/schema/AdHocToolSchemaGeneratorImpl.java) keys tool identity off `element.elementId()`. Worth stating with the citation so it stops getting re-litigated.

**2. Two things shipped while this was in review, and both map onto sections of this guide.** The agent modeling-guidance rules merged on 2026-08-03 in #9278: [`agent-tool-documentation`](https://docs.camunda.io/docs/next/components/modeler/reference/modeling-guidance/rules/agent-tool-documentation/), [`agent-fromai-contract`](https://docs.camunda.io/docs/next/components/modeler/reference/modeling-guidance/rules/agent-fromai-contract/), and [`agent-tool-output-key`](https://docs.camunda.io/docs/next/components/modeler/reference/modeling-guidance/rules/agent-tool-output-key/), one per main section here. Assisted tool configuration in Web Modeler merged on 2026-08-10 in #9576, with autofill for a `fromAi()` input, autofill for a `toolCallResult` output, and one-click corrections. The rules report what is wrong and assisted configuration offers to fix it, so a line in each section pointing at them tells a modeler the guidance is enforced in the tool, not just written down.

Note that #9576 edited `agentic-ai-aiagent-tool-definitions.md`, the same file this PR changes. Worth reconciling the new "What is a tool" intro against the new "Assisted tool configuration in Web Modeler" section during the rebase, rather than only resolving the textual conflict.

**3. Two items from the automated review pass on 2026-07-30 are still open on the branch.** Its items 2 and 3 were applied, but item 1 (`Slements` in `design-architecture.md`) and item 4 are both still there. Item 4 is the one worth a second look: this PR adds the heading "Which elements are resolved as tools" and the sentence directly under it still reads "will take all **activities** into account". The heading and its own body now disagree. Both are leftovers of the activities-to-elements sweep.

**4. This page and the getting-started guide now point at each other as the complete source.** This page says the getting-started walkthrough is "complete, step-by-step" and "applies all the steps below". Getting-started says this page has "full guidance". A reader landing on either is sent to the other, with no signal about which to read.

The wording is symmetric but the content is not. As the automated pass noted in its item 5, the `toolCallResult` to `toolCallResults` aggregation explanation left getting-started and did not arrive here. That page got thinner at the same moment its link started calling this one complete.

Suggestion is to have each link advertise what is different rather than what is more, so getting-started owns the journey and this page owns the deep dive:

- From here: "For this in the context of a full working agent, see step 4 of the getting-started guide."
- From getting-started: "To add tool types beyond this connector, such as user tasks, script tasks, sub-flows, or MCP, see Add tools to an AI agent."

**5. Backport the link swap to 8.9.** This answers the docs-only flag from 2026-06-24 that is still open. 8.9 already documents `fromAi`, element template fields being handled as input mappings, and gateway tool definitions, so nothing here promises functionality 8.9 does not have. 8.9 is also where the Web Modeler link-outs land today, which is the practical reason to do it. One caveat: 8.9 links `fromAi` as `feel-built-in-functions-ai-agent.md#fromaivalue` while this page uses `feel-built-in-functions-miscellaneous.md#fromaivalue`, so the swap cannot be copy-pasted.

On @nikonovd's first point about anchoring the guide in an example: the [AI Agent Chat quick start](https://marketplace.camunda.com/en-US/apps/587865/ai-agent-chat-quick-start) blueprint is zero-config, so it works as the try-it entry point without turning this page into a second tutorial.

nikonovd:
@afgambin Thanks for adapting! Nothing open from my side of my PR review.

As @HanselIdes left some comments I'll leave the approval to him.

docs/components/agentic-orchestration/add-tool-to-ai-agent.md
```diff
@@ -0,0 +1,260 @@
...
+</Tabs>
+
+### Combine multiple outputs into a single `toolCallResult`
+
+When your tool produces several values that each contribute a field to `toolCallResult`, do not add one output mapping per field targeting `toolCallResult.<field>`:
+
+- Output mappings and result variables containing a period are discouraged. See [output mappings](/components/concepts/variables.md#output-mappings).
+- Mapping to `toolCallResult` directly replaces the entire variable, so several mappings targeting it overwrite each other.
+
+Whether you can add fields to `toolCallResult` one at a time depends on the element type. Use the approach that matches your tool's element type:
```
maff, line 223: 💭 I feel this could be a bit abstract for people new to modeling AI Agents as it does not directly mention multiple elements/sub-flows. WDYT of also adding an example model/image here?

afgambin: Added

afgambin:
@HanselIdes @maff All comments addressed. Ready for a final review / approval. Thanks!

Question: Do we want to backport it to 8.9 and 8.8?

nikonovd:
> Question: Do we want to backport it to 8.9 and 8.8?

Would make sense from my perspective, as many customers are still on those versions and getting started with AO usecases.

docs/apis-tools/csharp-sdk/api-reference/models.md
```diff
@@ -765,7 +765,7 @@ public sealed class AdvancedActorTypeFilter
...
 | `In`     | `List<AuditLogActorTypeEnum>`     | Checks if the property matches any of the provided values.                                                                                                                                                                                                |
-| `Like`   | `Nullable<LikeFilter>`            | Checks if the property matches the provided like value. Supported wildcard characters are: * `*`: matches zero, one, or multiple characters. * `?`: matches one, single character. Wildcard characters can be escaped with backslash, for instance: `\*`. |
+| `Like`   | `Nullable<LikeFilter>`            | Checks if the property matches the provided like value. Supported wildcard characters are: _ `_`: matches zero, one, or multiple characters. * `?`: matches one, single character. Wildcard characters can be escaped with backslash, for instance: `\*`. |
```
maff: Are the changes in this file intended?

docs/components/agentic-orchestration/add-tool-to-ai-agent.md
maff: I think the individual sections are OK and would also work without the images in addition to the code blocks, I was rather referring to this into section and the general possibility of multiple elements/"produces several values". I think just from reading the text, it might not be directly obvious to users that it's possible to combine sub-flows/multiple elements into a tool and why it would be needed to combine `toolCallResult`. Sorry if unclear in the initial comment!

WDYT if we'd add an image showing a flow of multiple elements to this section? An example (we can also think of a better one, e.g. including a connector):

image

HanselIdes: I think this would help get the point across. If it fits existing patterns, I would use a textAnnotation calling out the data that's being added - that way you can show that data doesn't conflict in the image itself

docs/components/agentic-orchestration/design-architecture.md
```diff
@@ -53,53 +53,30 @@ Follow these principles when designing your agentic orchestration solution:
 In Camunda agentic orchestration, decision-making and orchestration are intentionally split:
 
 - **LLM responsibility**: Interprets the system prompt, current user prompt, and available tool descriptions. It decides which tool to call, in what order, and with which parameters.
-- **Camunda responsibility**: Executes the selected BPMN activities, stores process state, applies retries and incident handling, and coordinates user tasks and other deterministic workflow logic.
```
HanselIdes, line 56: This activity->element rename looks unrelated to this PR

HanselIdes (commented):
@afgambin At a high level the guide reviews well. I added one non-blocking idea about how to improve the visualization.
The only ❌ I see is that this PR has some unrelated changes that likely need a different set of reviewers

docs/apis-tools/csharp-sdk/api-reference/models.md
afgambin: Good call! It seems Prettier format-check ran in the background while I was working with Claude and introduced these changes. I’ve reverted them.

docs/components/agentic-orchestration/design-architecture.md
afgambin: These renaming updates are intended for this page (We discussed this at the beginning of the PR).

docs/components/agentic-orchestration/add-tool-to-ai-agent.md
afgambin: @maff Ah, I understand now. I also felt that the code blocks covered the subsections well enough. So, before making any changes, let’s agree on the final example.

Based on both of your feedback, I imagine something like this:

Image

Then, we show, using code snippets or so, not in the screenshot, how `toolCallResult` accumulates outputs from those elements as the process evolves. For example,

Image

Wdyt? Is this the right assumption?

maff: That sounds great! 👍 

In the first step (ask human) it could even be a direct assignment (e.g. `toolCallResult = { approved: true }`), with the 2 subsequent steps adding additional content.

afgambin: Added a new section.

afgambin:
Backported this change to 8.9 and 8.8 (commit c3ebfb7fa), applied on top of the same branch:

- Added `add-tool-to-ai-agent.md` + its 4 images to both `versioned_docs/version-8.9/` and `versioned_docs/version-8.8/`, and registered it in both `versioned_sidebars/*-sidebars.json`.
- Applied the same content edits to `ai-agents.md`, `design-architecture.md`, `agentic-ai-aiagent-tool-definitions.md`, `getting-started-agentic-orchestration.md`, and the `_ao-card-data.js` recommendation card in both versions.

Content adaptations for version parity (not present in the source PR):
- The Web Modeler link uses the pre-8.10-IA-restructure path (`/components/modeler/web-modeler/index.md`) in both versioned copies, since the `hub/workspace/modeler` path is unreleased-only.
- The 8.8 copy of the "Tool type" table omits the **MCP client connector** row, since the MCP Client connector doc doesn't exist in 8.8.
- Links to the Web Modeler "autofill" assisted-configuration feature (`#autofill-a-fromai-input`, `#autofill-a-toolcallresult-output`) are omitted in both 8.9 and 8.8, since that content (from #9576) hasn't been backported yet.

`npm run build` passes with no new broken links/anchors beyond the pre-existing baseline.

afgambin:
@maff @HanselIdes I addressed your latest comments and backported all the PR changes to 8.9 and 8.8. I believe it’s good to go!

HanselIdes (approved):
LGTM- the accumulating variable example is very clear

PR #9190, opened 2026-06-24
docs: add Kubernetes version support policy to supported environments

hisImminence:
## Summary

Two doc changes to align and cross-link our Kubernetes version support policy.

**Context:** A question came up in a Helm chart PR ([camunda/camunda-platform-helm#6027](https://github.com/camunda/camunda-platform-helm/pull/6027)) about whether a change requiring Kubernetes 1.23 features would break customers on older versions. Our answer: no — we follow the Kubernetes official support cycle and 1.23 is already EOL within it. However, that policy was only stated in `reference-architecture/kubernetes.md`, not in `supported-environments.md` where customers and support engineers look first.

**Changes (all versions 8.7–8.10/current):**

1. **`reference/supported-environments.md`:** Merged the Kubernetes version support policy into the existing Deployment options paragraph — no new block, just tighter wording that makes the policy discoverable in the right place.

2. **`self-managed/reference-architecture/kubernetes.md`:** Simplified the `#infrastructure` version statement and replaced the external `kubernetes.io/releases` link with a link to `supported-environments.md` as the single authoritative reference.

### DIFF
--- a/docs/reference/supported-environments.md
+++ b/docs/reference/supported-environments.md
@@ -55,6 +55,8 @@ We recommend running Camunda 8 Self-Managed in a Kubernetes environment. We prov
 
 With the correct configuration, Camunda 8 Self-Managed can be deployed on any [Certified Kubernetes](https://www.cncf.io/training/certification/software-conformance/#benefits) distribution (cloud or on-premises). However, we officially test and support a specific list of platforms.
 
+Camunda 8 is not tied to a specific Kubernetes version. The Helm chart supports the Kubernetes [official support cycle](https://kubernetes.io/releases/).
+
 The following are tested and supported deployment options for Kubernetes, Docker, and manual installation:
 
 - [Stock Kubernetes](/self-managed/deployment/helm/install/quick-install.md)
--- a/docs/reference/supported-environments.md
+++ b/docs/reference/supported-environments.md
@@ -53,9 +53,7 @@ We recommend running Camunda 8 Self-Managed in a Kubernetes environment. We prov
 
 ### Deployment options
 
-With the correct configuration, Camunda 8 Self-Managed can be deployed on any [Certified Kubernetes](https://www.cncf.io/training/certification/software-conformance/#benefits) distribution (cloud or on-premises). However, we officially test and support a specific list of platforms.
-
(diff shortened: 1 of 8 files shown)


docs/reference/supported-environments.md
leiicamundi, line 1: @hisImminence Let's remove `However, we officially test and support a specific list of platforms.`, I don't think it makes sense anymore

versioned_docs/version-8.7/reference/supported-environments.md
leiicamundi, line 1: Same removal suggestion: "However, we officially test and support a specific list of platforms."

Also, does it means for a customer that install 8.7 with a kube version not supported today anymore that he is not supported?

leiicamundi:
@hisImminence left few comments, applicable to all versions.

~I'm also wondering if we should not consolidate kube requirements in https://docs.camunda.io/docs/reference/supported-environments/#component-requirements? It looks like the right place to put it nah?~

See https://github.com/camunda/camunda-docs/pull/9190/changes#r3474507291

versioned_docs/version-8.7/self-managed/reference-architecture/kubernetes.md
```diff
@@ -124,7 +124,7 @@ For more details on multi-region configurations, especially dual-region setups,
 
 We recommend using an officially [certified Kubernetes](https://www.cncf.io/training/certification/software-conformance/#benefits) distribution.
 
-Camunda 8 is not tied to a specific Kubernetes version. To simplify deployment, we provide a [Helm chart](/self-managed/setup/install.md) for easy installation on Kubernetes. The latest Helm chart is typically compatible with Kubernetes' [official support cycle](https://kubernetes.io/releases/).
+Camunda 8 is not tied to a specific Kubernetes version. To simplify deployment, we provide a [Helm chart](/self-managed/setup/install.md) that follows the Kubernetes official support cycle. See [supported environments](/reference/supported-environments.md) for tested and supported platforms.
```
leiicamundi, line 127: 2nd review: I think we should avoid repeating the what we follow and just refereing to the supported environment. which would translates as:

> that follows our [Kubernetes support statement](/reference/supported-environments.md).

eamonnmoloney (approved):
I like this but @leiicamundi has comments

leiicamundi:
@eamonnmoloney your approval triggered the automerge :D I'll create a follwoup pr to remove the duplication

PR #9204, opened 2026-06-26
docs: clarify removing an exporter via config requires an explicit actuator disable

eamonnmoloney:
## Description

Follow-up to https://jira.camunda.com/browse/SUPPORT-33221 ("PVC usage grows, is the exporter disfunctional?") and broker defect [camunda/camunda#52260](https://github.com/camunda/camunda/issues/52260).

When an operator disables Optimize (8.8+), the Helm chart removes the legacy Elasticsearch/OpenSearch exporter from the broker's **static** configuration, removes the exporter from static configuration and does not remove it from dynamic configuration blocking log compaction and growing the Zeebe PVC until it was disabled manually via `POST actuator/exporters/elasticsearch/disable`.

The Zeebe management API page documented `disable`/`delete` but never told operators that a configuration-level removal (such as disabling Optimize through Helm) does **not** stop the running exporter and that an explicit actuator disable + convergence check is required. This PR closes that gap.

This complements the chart-side note added in camunda/camunda-platform-helm#6455.

## Changes

Adds one paragraph to the "Disable an exporter" section of the Zeebe **management API** page, explaining that:

- removing an exporter from config (e.g. disabling Optimize) only updates static config and does not fully deactivate the running exporter;
- operators must explicitly disable it via the management API and confirm every broker reports `DISABLED` (links to the existing "Monitor an exporter" section).


docs/self-managed/components/orchestration-cluster/zeebe/operations/management-api.md
```diff
@@ -149,6 +149,8 @@ POST actuator/exporters/{exporterId}/disable
 
 After disabling the exporter, no records will be exported to this exporter. Other exporters continue exporting.
 
+If you remove an exporter from the cluster configuration through Helm values — for example, by disabling Optimize, which removes the Elasticsearch or OpenSearch exporter — the exporter is only dropped from the static configuration, and the running exporter is not stopped automatically. The broker reconciles the removal into its dynamic cluster configuration only on partition leaders and relies on cluster gossip to converge across all brokers. If reconciliation does not converge, the exporter can remain active on some brokers, which prevents log compaction and increases disk usage. After removing an exporter through configuration, explicitly disable it using the request above, and confirm that every broker reports the exporter as `DISABLED` (see [Monitor an exporter](#monitor-an-exporter)).
```
houssain-barouni: I think we don't need to put lot of details

```suggestion
If you remove an exporter from the cluster configuration through Helm values — for example, by disabling Optimize, which removes the Elasticsearch or OpenSearch exporter — the exporter is only dropped from the static configuration, and the exporter is still declared in the dynamic cluster configuration, which prevents log compaction and increases disk usage. After removing an exporter through configuration, you need to explicitly disable it using the request above, and confirm that every broker reports the exporter as `DISABLED` (see [Monitor an exporter](#monitor-an-exporter)).
```

docs/self-managed/components/orchestration-cluster/zeebe/operations/management-api.md
houssain-barouni, line 1: @eamonnmoloney can we put a note in https://github.com/camunda/camunda-docs/blob/e959728819346600a15ddb16518561669ea742fb/docs/self-managed/deployment/helm/configure/enable-additional-components.md?plain=1#L136 and link it to this page? Helm users would check that page
Also should we add this doc in Helm documentation https://artifacthub.io/packages/helm/camunda/camunda-platform#optimize-parameters?

houssain-barouni (commented):
Thank you @eamonnmoloney for taking care of this
I have left some comments

docs/self-managed/components/orchestration-cluster/zeebe/operations/management-api.md
houssain-barouni: This is addressed in https://github.com/camunda/camunda-docs/pull/9204/changes/6d67ae38e43e3dfebee8976f0a0c33d14d4a9723

Helm docs are handled in https://github.com/camunda/camunda-platform-helm/pull/6455

zaycodes (changes requested):
Quick check on the exporter wording across versions to make sure we're consistent.

The new paragraph in `management-api.md` is worded differently between versions, and I want to confirm whether that's intentional before this merges.

docs/next (8.10):

"...the exporter is only dropped from the static configuration, and the exporter is still declared in the dynamic cluster configuration, which prevents log compaction..."

8.8 and 8.9:

"...the exporter is only dropped from the static configuration, and the running exporter is not stopped automatically. The broker reconciles the removal into its dynamic cluster configuration only on partition leaders and relies on cluster gossip to converge... If reconciliation does not converge, the exporter can remain active on some brokers..."

The 8.10 version reads as though a manual disable is always required (the exporter is simply left in the dynamic config), whereas the 8.8/8.9 version reads as though the broker usually removes it automatically and it only stays active if reconciliation doesn't converge, implying the manual step is more of a fallback.

Is this difference intentional? Does the behavior actually differ between 8.8/8.9 and 8.10?
If it's not intentional, should we align all three versions on one wording so readers get the same guidance?

CC @eamonnmoloney @houssain-barouni

houssain-barouni:
> Quick check on the exporter wording across versions to make sure we're consistent.
> 
> The new paragraph in `management-api.md` is worded differently between versions, and I want to confirm whether that's intentional before this merges.
> 
> docs/next (8.10):
> 
> "...the exporter is only dropped from the static configuration, and the exporter is still declared in the dynamic cluster configuration, which prevents log compaction..."
> 
> 8.8 and 8.9:
> 
> "...the exporter is only dropped from the static configuration, and the running exporter is not stopped automatically. The broker reconciles the removal into its dynamic cluster configuration only on partition leaders and relies on cluster gossip to converge... If reconciliation does not converge, the exporter can remain active on some brokers..."
> 
> The 8.10 version reads as though a manual disable is always required (the exporter is simply left in the dynamic config), whereas the 8.8/8.9 version reads as though the broker usually removes it automatically and it only stays active if reconciliation doesn't converge, implying the manual step is more of a fallback.
> 
> Is this difference intentional? Does the behavior actually differ between 8.8/8.9 and 8.10? If it's not intentional, should we align all three versions on one wording so readers get the same guidance?
> 
> CC @eamonnmoloney @houssain-barouni

sorry, it was not intentional, I will align the 3 versions

zaycodes (approved):

PR #9130, opened 2026-06-18
docs: make diagnostic script take java thread dumps from pods

lilspikey:
java thread dumps can provide a lot of useful information, so it makes sense to include them in the diagnostic scripts.

this is done only for pods that have java and jattach (and pgrep) available, which means usually we will only get thread dumps for the camunda pods. older images might not have jattach available, so we might not always be able to get thread dumps via this script.  potentially we can use `kill -SIGQUIT` for that, but we do not want to use that indiscriminately (as it could cause non-Java process to exit).

example output when running the command:

```
  - Collecting Java threaddumps ...
    - Collecting threaddump from pod: camunda-0, java process: 1
    - Collecting threaddump from pod: camunda-1, java process: 1
    - Collecting threaddump from pod: camunda-2, java process: 1
    - Skipping pod: connectors-775fbcbcb-tbg7d as jattach not installed
    - Skipping pod: customer-notification-77b776876b-chvhr as jattach not installed
    - Skipping pod: dispute-process-request-get-vendor-info-8568797df5-6tbs6 as jattach not installed
    - Skipping pod: dispute-process-request-proof-from-vendor-b89fd84f6-clxvp as jattach not installed
    - Skipping pod: elastic-0 as jattach not installed
    - Skipping pod: elastic-1 as jattach not installed
    - Skipping pod: elastic-2 as jattach not installed
    - Skipping pod: extract-data-from-document-845468cb8f-zx22z as jattach not installed
    - Skipping pod: identity-548769bf95-vfhdf as jattach not installed
    - Skipping pod: identity-548769bf95-zfn9x as java not installed
    - Skipping pod: inform-about-successful-claim-9dfbc6bd5-w9bdm as jattach not installed
    - Skipping pod: keycloak-0 as jattach not installed
    - Skipping pod: keycloak-postgresql-0 as java not installed
    - Skipping pod: leader-balancer-29696270-jvpj8 as java not installed
    - Skipping pod: leader-balancer-29696280-s28pd as java not installed
    - Skipping pod: leader-balancer-29696290-wz5tz as java not installed
    - Skipping pod: metrics-exporter-749fdff578-8sxgd as java not installed
    - Skipping pod: optimize-88bd46974-mwlkk as jattach not installed
    - Skipping pod: prom-els-exporter-864fc59c5f-6jkvn as java not installed
    - Skipping pod: refunding-646d878967-bj9tx as jattach not installed
    - Skipping pod: starter-7bf9fd7c57-7ctcp as jattach not installed
```

[...]

PR #9210, opened 2026-06-26
docs: collect Zeebe actuator outputs in the diagnostics script

### LINKED ISSUE #9209: Extend the diagnostics collection script to capture Zeebe actuator outputs
## Context

The Self-Managed diagnostics collection script (`camunda-collect-diagnostics.sh`) is defined inline in `docs/self-managed/deployment/helm/operational-tasks/diagnostics.md` (and the versioned 8.7/8.8/8.9 copies). It gathers Kubernetes resources, pod logs, Helm values, and — since #9130 — JVM thread dumps via `kubectl exec … jattach`.

It currently collects **no actuator data**. When Support investigates engine/cluster issues, the Zeebe management actuators are often the fastest way to see partition roles, exporter state, cluster topology, flow control, and a metrics snapshot — without asking the customer to run extra commands. This issue adds that to the standardised bundle so it arrives with the first diagnostics zip.

## What to collect

For each Camunda orchestration-cluster / Zeebe pod, query the management endpoint (default port `9600`) and save each response as a file in the pod's diagnostics output.

**Custom Zeebe actuators (read-only `GET`):**

- `GET /actuator/partitions` — per-broker partition roles (leader/follower), health, processing/exporting state, snapshot positions.
- `GET /actuator/cluster` — dynamic cluster topology, broker/partition assignment, routing state, pending operations.
- `GET /actuator/exporters` — configured exporters and their status (enabled/disabled, position).
- `GET /actuator/flowControl` — current flow-control configuration per partition.
- `GET /actuator/jobstreams` — active client and remote job-activation streams.
- `GET /actuator/backupRuntime/state` — current backup status.

**Generic Spring Boot actuators:**

- `GET /actuator/prometheus` — full metrics snapshot in Prometheus text format.
- `GET /actuator/configprops` — effective configuration properties. Secrets are already masked server-side via `management.sanitize.keywords` (user, pass, secret, license, accessKey, token, …).

Only the read-only `GET` operations listed above should be collected.

## Explicitly excluded (do not call)

These custom actuators are **mutating** (POST/DELETE/PUT) and must not be invoked by a diagnostics script, as they change cluster state: `exporting` (pause/resume), `banning` (ban instance), `rebalance` (trigger rebalancing), the `backupRuntime`/`backups` take/delete operations, and the write operations on `cluster` (scale, purge, partition-distribution), `exporters` (enable/disable/delete), and `partitions` (pause/resume/snapshot).

## Implementation notes (for whoever picks this up)

- **Reaching the endpoint:** the management port (9600) is not the main service port. Two options:
  - `kubectl port-forward <pod> 9600:9600` then `curl` locally — no dependency on in-pod tooling, works on distroless images. Recommended.
  - `kubectl exec <pod> -- curl/wget http://localhost:9600/actuator/<id>` — consistent with the existing thread-dump step, but depends on curl/wget being present in the image (not guaranteed).
- **Scope per pod:** the custom Zeebe actuators only exist on the orchestration-cluster / broker / gateway pods. `prometheus` and `configprops` exist on all Camunda Spring Boot components. Attempt each endpoint and tolerate non-200 / connection-refused responses silently (skip + continue), exactly as the current log/exec steps tolerate failures — do not abort the whole run.
- **Version differences:** ports/paths differ for pre-8.8 separate Operate/Tasklist apps. Keep the change focused on the Zeebe/orchestration management endpoint (9600) and degrade gracefully elsewhere.
- **Output:** write one file per endpoint per pod, e.g. `<pod>-actuator-partitions.json`, `<pod>-actuator-prometheus.txt`, `<pod>-actuator-configprops.json`.

## Data privacy

`configprops` masks credential-like keys but does **not** redact connection URLs, hostnames, or bucket names. This is already covered by the script's existing PII review step before sharing — extend that caution note to mention `configprops` explicitly.

## Acceptance criteria

- The script collects the read-only custom actuators and the `prometheus` + `configprops` generic actuators from the relevant pods.
- No mutating actuator endpoint is ever called.
- Failures for pods/endpoints that don't expose a given actuator do not abort the run.
- The data-privacy note mentions `configprops`.
- All versioned copies of the docs page are updated consistently.

## References

- Docs page: `docs/self-managed/deployment/helm/operational-tasks/diagnostics.md`
- Adjacent prior work: PR #9130 (thread dumps via `kubectl exec jattach`)
- Endpoint sources (camunda/camunda): `BrokerAdminServiceEndpoint.java`, `ClusterEndpoint.java`, `ExportersEndpoint.java`, `FlowControlEndpoint.java`, `JobStreamEndpoint.java`, `BackupEndpoint.java`; management config in `dist/src/main/resources/application.properties`

lenaschoenburg:
## Description

Extends the Self-Managed diagnostics collection script (`camunda-collect-diagnostics.sh`) to capture Zeebe/Spring Boot actuator outputs, so they arrive with the first diagnostics zip instead of being requested separately.

A new step port-forwards the management endpoint (default port `9600`) for each pod and collects only **read-only `GET`** actuators, one file per endpoint per pod:

- Custom Zeebe: `partitions`, `cluster`, `exporters`, `flowControl`, `jobstreams`, `backupRuntime/state`
- Generic: `prometheus`, `configprops`

Mutating endpoints (exporting, banning, rebalance, backup take/delete, cluster/exporters/partitions writes) are intentionally **not** called. Pods that don't expose the port — and a missing `curl` — are tolerated and skipped, matching the script's existing behaviour.

### Reliability

- **Ephemeral local port** chosen by `kubectl` (`:9600`) instead of a hardcoded one, so it can't clash with a port already bound on the operator's machine (which would otherwise silently yield empty actuator files).
- **Waits for the tunnel** by parsing kubectl's `Forwarding from …` line instead of a fixed `sleep`, and bails early if the forward dies (pod has no management port).
- **`trap … EXIT INT TERM`** tears down the port-forward even on Ctrl-C, so no `kubectl port-forward` processes are leaked.
- **`curl`-missing breadcrumb**: if `curl` isn't installed, the script writes an `actuator-collection-skipped.txt` into the archive. This lets Support distinguish "newer script ran, curl was absent, actuator data simply not attempted" from a failed/partial collection or an older script — avoiding an unnecessary re-run request.

Also updates the "What the script collects" list and extends the data-privacy note to mention that `configprops` masks credentials but not URLs/hostnames.

Applied consistently to all four copies of the page (current, 8.9, 8.8, 8.7).

Closes #9209

## Notes for reviewers

- Implemented against current `main`. The thread-dump work (#9130) is still open and touches a different part of the loop, so no overlap.

### DIFF
--- a/docs/self-managed/deployment/helm/operational-tasks/diagnostics.md
+++ b/docs/self-managed/deployment/helm/operational-tasks/diagnostics.md
@@ -10,7 +10,7 @@ description: "Get diagnostics and logs from a Helm chart deployment."
 This script automates the process of gathering logs and diagnostics from a Camunda Helm chart deployment running in a Kubernetes cluster. The script collects all relevant information (including pod logs, events, and resource details) into a single directory, and outputs it in a .zip file to make it easier to share this information with the Camunda Support team.
 
 :::caution Data privacy notice
-Before sharing the generated diagnostics file with Camunda Support, review and remove any sensitive information such as passwords, API keys, personal data, or business-sensitive data from the collected logs and configuration data.
+Before sharing the generated diagnostics file with Camunda Support, review and remove any sensitive information such as passwords, API keys, personal data, or business-sensitive data from the collected logs and configuration data. This includes the collected actuator outputs: the `configprops` endpoint masks credential-like values, but connection URLs, hostnames, and bucket names are not redacted.
 :::
 
 ### What the script collects
@@ -23,6 +23,7 @@ The script outputs the following data from your namespace and creates a zip file
 - **Cluster Nodes**: Node descriptions.
 - **Network Resources**: Services, endpoints, and ingresses.
 - **Configuration**: Config map information.
+- **Actuator endpoints**: Read-only Zeebe and Spring Boot actuator outputs from Camunda pods (partition status, cluster topology, exporters, flow control, job streams, backup state, Prometheus metrics, and configuration properties).
 
 ### Usage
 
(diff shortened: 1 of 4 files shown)


docs/self-managed/deployment/helm/operational-tasks/diagnostics.md
```diff
@@ -156,6 +157,80 @@ else
...
+  cleanup_port_forward() {
+    [[ -n "$pf_pid" ]] && kill "$pf_pid" 2>/dev/null
+    [[ -n "$pf_pid" ]] && wait "$pf_pid" 2>/dev/null
+    [[ -n "$pf_log" ]] && rm -f "$pf_log"
+    pf_pid=""
+    pf_log=""
+  }
+  trap cleanup_port_forward EXIT INT TERM
+
+  for pod in $(kubectl get pod -n "$namespace" --no-headers -o custom-columns=":metadata.name"); do
```
ChrisKujawa: ❓ Do we really want to iterate overall pods in the namespace? 

🔧 We could to something like: `-l app=camunda-platform`

docs/self-managed/deployment/helm/operational-tasks/diagnostics.md
```diff
@@ -156,6 +157,80 @@ else
...
+    # Let kubectl choose a free local port to avoid clashing with anything already bound.
+    pf_log=$(mktemp)
+    kubectl port-forward -n "$namespace" "$pod" ":${management_port}" > "$pf_log" 2>&1 &
```
ChrisKujawa: 🔧 This can fail right? What is the expected behavior here?

ChrisKujawa: We might want to retry this?

ChrisKujawa (dismissed):
Thanks @lenaschoenburg 👍🏼 I feel we might want to make it a bit more fault tolerant - like with retries as example see https://github.com/camunda/camunda/blob/main/load-tests/docs/scripts/verify-test.sh#L126

But it is already a good improvement - i leave it to you - thanks for your work 👍🏼

docs/self-managed/deployment/helm/operational-tasks/diagnostics.md
lenaschoenburg: I've added a retry for setting up the port-forward :+1:

ChrisKujawa (approved):
👍🏼

PR #9211, opened 2026-06-26
Hub api migration guide

alexronquillo:
## Description

Update migration guide. I tried to apply everything from the issue description and cross reference that with the new docs and the old docs. There was a lot of stuff missing, so I added it to the best of my ability. I also ran claude against it to check my work. However, please triple check to make sure I've captured all the necessary changes.

Closes: https://github.com/camunda/camunda-hub/issues/25128

## Follow-up questions

- I did not include code samples for all endpoint, because I thought the migration guide was long enough as it is and might be confusing to readers. Instead, they can compare what they have in their codebase with the _new_ docs. Do you think this is an acceptable approach?
- Are the cursor fields in the search responses always present? Even when using limit/offset pagination?
- I left out code samples for calculating time periods. IMO, this is a client-side implementation detail and is entirely the customer's responsibility. Do you think this is an acceptable approach?
- The issue description says "V1 projectId and V2 workspaceKey refer to the same underlying entity — the same UUID works in both APIs". However, the file and folder APIs translate V1 projectId to V2 projectKey. So if projectId = projectKey = workspaceKey, I'm not sure how that works, because I file can't be placed at the workspace level. Can you clarify this?

## PR Checklist


docs/apis-tools/hub-api-saas/authentication.md
```diff
@@ -11,20 +11,20 @@ import PageDescription from '@site/src/components/PageDescription';
 
 ## The process
 
-Generate a [JSON Web Token (JWT)](https://jwt.io/introduction/), and include it in every request.
+Generate a [JSON Web Token (JWT)](https://jwt.io/introduction/), and include it in every request. If you already have a Web Modeler API v1 token, you can use the same token for this API.
```
ThorbenLindhauer:
```suggestion
Generate a [JSON Web Token (JWT)](https://jwt.io/introduction/), and include it in every request. If you already have a client or token for the Web Modeler API v1, then you can reuse it for this API.
```

Tokens are short-lived. I think it's more important for users to know that they don't need to recreate their API clients.

docs/apis-tools/hub-api-sm/authentication.md
```diff
@@ -11,7 +11,7 @@ import PageDescription from '@site/src/components/PageDescription';
 
 ## The process
 
-Generate a [JSON Web Token (JWT)](https://jwt.io/introduction/), and include it in every request.
+Generate a [JSON Web Token (JWT)](https://jwt.io/introduction/), and include it in every request. If you already have a Web Modeler API v1 token, you can use the same token for this API.
```
ThorbenLindhauer, line 14: Here we can consider making a similar point as with https://github.com/camunda/camunda-docs/pull/9211/changes#r3492788671. However I think that also depends on the migration story for https://github.com/camunda/camunda-hub/issues/21467 (which we are not working on yet). So if you want to be defensive, you can leave the content here unchanged for now.

docs/apis-tools/migration-manuals/migrate-from-web-modeler-to-hub-api.md
```diff
@@ -18,6 +18,14 @@ Web Modeler API v1 was the REST API for Web Modeler, a standalone product for mo
 
 [Camunda Hub API v2](/apis-tools/hub-api-sm/overview.md) is the successor API for the broader Camunda Hub platform. Camunda Hub unifies organizational management, workspace governance, and process modeling into a single platform. As a result, the conceptual model and architecture of the API have changed.
 
+## Deprecation timeline
+
+| Version  | Action                                                                                                                                                                                                                               |
+| -------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| **8.10** | Camunda Hub API v2 ships alongside Web Modeler API v1. Web Modeler API v1 is marked as **deprecated**, and all responses include `Deprecation`, `Sunset`, and `Link` HTTP headers. The v1 OpenAPI spec is marked `deprecated: true`. |
```
ThorbenLindhauer: We have decided against adding the headers, see https://github.com/camunda/camunda-hub/issues/25124#issuecomment-4797243417

docs/apis-tools/migration-manuals/migrate-from-web-modeler-to-hub-api.md
```diff
@@ -89,21 +133,148 @@ Instead of specifying the number of pages to skip, you specify the index to star
 
 In addition to the different pagination model, the default page size has changed. In v1, the default page size was 10. In v2, the default limit is 100.
 
-### Projects
+#### Cursor pagination
```
ThorbenLindhauer: As mentioned above, no cursor pagination in Hub V2 API for all I know.

docs/apis-tools/migration-manuals/migrate-from-web-modeler-to-hub-api.md
ThorbenLindhauer: Unrelated to your PR here, but I think if you choose two different values here (e.g. 50 and 20) then the explanation becomes even easier to follow.

docs/apis-tools/migration-manuals/migrate-from-web-modeler-to-hub-api.md
```diff
@@ -89,21 +133,148 @@ Instead of specifying the number of pages to skip, you specify the index to star
...
+
+In addition to offset pagination, Camunda Hub API v2 offers cursor pagination. With cursor pagination, you use a `page` object with two fields:
+
+- `page.after` specifies the cursor to start from. You receive this from a prior response.
+- `page.limit` limits the number of items returned.
+
+```json title="Camunda Hub API v2"
+{
+  "page": {
+    "after": "{endCursor from previous response}",
```
ThorbenLindhauer: If we wanted to keep this we should have a real example of a cursor. With the current explanation the concept of the cursor is very abstract for the reader.

docs/apis-tools/migration-manuals/migrate-from-web-modeler-to-hub-api.md
```diff
@@ -39,28 +47,64 @@ See the [Camunda Hub API authentication guide](/apis-tools/hub-api-sm/authentica
...
 }
 ```
 
 ### Pagination
 
+Offset pagination in Camunda Hub API v2 is different from Web Modeler API v1. Additionally, v2 introduces cursor pagination.
+
+#### Offset pagination
+
 In Web Modeler API v1, you used two fields to paginate items in a response:
```
ThorbenLindhauer: > to paginate items in a response

Maybe just me as a non-native speaker: This sounds like this doesn't talk about the request. It took me a little bit to understand that the offset and cursor pagination only talk about the requests and that there's then a separate response section.

docs/apis-tools/migration-manuals/migrate-from-web-modeler-to-hub-api.md
```diff
@@ -89,21 +133,148 @@ Instead of specifying the number of pages to skip, you specify the index to star
...
+   └─ File        # root-level
+
+```
+
+In Camunda Hub API v2, _workspaces_ are a new organizational level above projects:
+
+```
+Organization
+└─ Workspace         # new top-level container, groups projects
+   ├─ Project        # same as v1 project, now inside a workspace
```
ThorbenLindhauer: No, this is a V1 process application.

docs/apis-tools/migration-manuals/migrate-from-web-modeler-to-hub-api.md
```diff
@@ -89,21 +133,148 @@ Instead of specifying the number of pages to skip, you specify the index to star
...
+
+```
+Organization
+└─ Project        # top-level container
+   ├─ Folder
+   │   ├─ File
+   │   └─ Folder  # nested
+   └─ File        # root-level
+
+```
```
ThorbenLindhauer: This mixes two concerns: Terminology (that changes between V1 and V2) and organization of data (that changes between 8.9 and 8.10; so in 8.10 it also affects the V1 API).

I suggest to similarly explain this in two steps to readers: First just the terminology changes: Project is now Workspace. Process application (that wasn't exposed directly in the V1 API - just as folders) become projects.

Then, for the change of the organization of data you could talk about https://github.com/camunda/camunda-hub/issues/24222 (how we adapt V1 API behavior to the new schema). We have a dedicated docs issue for that, though: https://github.com/camunda/camunda-hub/issues/25045 (could do it at once or later)

docs/apis-tools/migration-manuals/migrate-from-web-modeler-to-hub-api.md
```diff
@@ -89,21 +133,148 @@ Instead of specifying the number of pages to skip, you specify the index to star
...
+
+In Camunda Hub API v2, _workspaces_ are a new organizational level above projects:
+
+```
+Organization
+└─ Workspace         # new top-level container, groups projects
```
ThorbenLindhauer: This is called "project" in V1 and 8.9.

docs/apis-tools/migration-manuals/migrate-from-web-modeler-to-hub-api.md
```diff
@@ -89,21 +133,148 @@ Instead of specifying the number of pages to skip, you specify the index to star
...
+   ├─ Project        # same as v1 project, now inside a workspace
+   │  ├─ Folder
+   │  │   ├─ File
+   │  │   └─ Folder  # nested
+   │  └─ File        # root-level
+   └─ Project
+      └─ File        # root-level
+```
+
+In the workspace API, the v1 `projectId` is now called the v2 `workspaceKey`. The same UUID works in both. In files and folders, the v1 `projectId` maps to the v2 `projectKey`.
```
ThorbenLindhauer: The first sentence is correct: `projectId` becomes `workspaceKey`. The third sentence is not: `projectKey` in the V2 API now points to the 8.10 concept "project". So that's a new thing that has no direct equivalent in V1.

Process applications in V1 were accessible via the folder API. In V2, they have their own dedicated project API and are no longer accessible via the folders API. Talking here about the target state we are building with https://github.com/camunda/camunda-hub/issues/24664, so not yet merged.

docs/apis-tools/migration-manuals/migrate-from-web-modeler-to-hub-api.md
```diff
@@ -113,42 +284,373 @@ All v1 files endpoints have a direct v2 equivalent:
...
+| ------------------ | ------------------ | ---------------- | --------------------------------------------------------------------------------------------------- |
+| `fileType`         | `type`             | Request/response | Renamed. `connector_template` is no longer supported. `element_template` is now `element-template`. |
+| `folderId`         | `folderKey`        | Request/response | Renamed                                                                                             |
+| `projectId`        | `projectKey`       | Request/response | Renamed                                                                                             |
+| -                  | `content`          | Response         | New field                                                                                           |
+| `id`               | `fileKey`          | Response         | Renamed                                                                                             |
+| `canonicalPath`    | `canonicalPath`    | Response         | In v1, `canonicalPath` was an object. In v2, it's a string.                                         |
```
ThorbenLindhauer: Two things here:

* Let's mention that `/` is the delimiter for the path elements in the String
* There's another behavioral difference here coming with https://github.com/camunda/camunda-hub/issues/24664: A project (ex process application aka folder) is included in the canonical path in the V1 API. But in the V2 API it is not anymore, so the canonical path only includes the keys of the actual folders and the project is cleanly a separate thing.

docs/apis-tools/migration-manuals/migrate-from-web-modeler-to-hub-api.md
```diff
@@ -113,42 +284,373 @@ All v1 files endpoints have a direct v2 equivalent:
...
-| ------------------ | ---------------------------- | --------------------------------------------------------------------- |
-| `filter`           | `filter`                     | Now supports advanced operators, including `$eq`, `$in`, and `$like`. |
-| `total` (response) | `page.totalItems` (response) | Moved into the `page` response object.                                |
+In addition to the [general field changes](#file-api-field-mapping), the following request fields have changed:
+
+| Web Modeler API v1       | Camunda Hub API v2  | Notes                                                                                           |
+| ------------------------ | ------------------- | ----------------------------------------------------------------------------------------------- |
+| `filter`                 | `filter`            | Now uses [advanced operators](#search-filters), including `$eq`, `$in`, and `$like`             |
+| `filter.projectId`       | `filter.projectKey` | Renamed                                                                                         |
```
ThorbenLindhauer: As mentioned above, this now returns the key of the project, not the workspace (ex `projectId)`. So this is more than a renaming.

docs/apis-tools/migration-manuals/migrate-from-web-modeler-to-hub-api.md
```diff
@@ -113,42 +284,373 @@ All v1 files endpoints have a direct v2 equivalent:
...
+| `filter.folderId`        | `filter.folderKey`  | Renamed                                                                                         |
```
ThorbenLindhauer: Renamed and `null` when the file is directly in a project (which returned the process application id as the folder id in V1).

docs/apis-tools/migration-manuals/migrate-from-web-modeler-to-hub-api.md
```diff
@@ -113,42 +284,373 @@ All v1 files endpoints have a direct v2 equivalent:
...
+| Delete a folder | `DELETE /v1/folders/{folderId}` | `DELETE /v2/folders/{folderKey}` |
+
+### Field mapping {#folder-api-field-mapping}
+
+The following fields have changed across all folder endpoints:
+
+| Web Modeler API v1 | Camunda Hub API v2 | Application      | Notes   |
+| ------------------ | ------------------ | ---------------- | ------- |
+| `parentId`         | `parentFolderKey`  | Request/response | Renamed |
+| `projectId`        | `projectKey`       | Request/response | Renamed |
```
ThorbenLindhauer: Same as for files, `projectKey` is in V2 the reference to the project, not the workspace.

docs/apis-tools/migration-manuals/migrate-from-web-modeler-to-hub-api.md
```diff
@@ -113,42 +284,373 @@ All v1 files endpoints have a direct v2 equivalent:
...
+| Update a folder | `PATCH /v1/folders/{folderId}`  | `PATCH /v2/folders/{folderKey}`  |
+| Delete a folder | `DELETE /v1/folders/{folderId}` | `DELETE /v2/folders/{folderKey}` |
+
+### Field mapping {#folder-api-field-mapping}
+
+The following fields have changed across all folder endpoints:
+
+| Web Modeler API v1 | Camunda Hub API v2 | Application      | Notes   |
+| ------------------ | ------------------ | ---------------- | ------- |
+| `parentId`         | `parentFolderKey`  | Request/response | Renamed |
```
ThorbenLindhauer: Similar as for files, this returns `null` for folders directly in process applications (same case as https://github.com/camunda/camunda-docs/pull/9211/changes#r3493158487).

docs/apis-tools/migration-manuals/migrate-from-web-modeler-to-hub-api.md
```diff
@@ -113,42 +284,373 @@ All v1 files endpoints have a direct v2 equivalent:
...
+
+In Web Modeler API v1, folder data in the response was nested under a `metadata` key:
+
+```json title="Web Modeler API v1"
+{
+  "metadata": {
+    "id": "f-456",
+    "name": "processes",
+    "projectId": "p-123",
+    "parentId": null,
```
ThorbenLindhauer: This is a case that after the 8.10 migration cannot occur anymore (as it represents a folder that sits directly in a workspace, without a project in between). So better set some parent id, too.

And in the V2 example below the same constellation is correct: Because `projectKey` there references the process application. Just make sure to use the same value that you choose for `parentId` here. `p-123` is wrong because that becomes the workspace key in 8.10/V2.

docs/apis-tools/migration-manuals/migrate-from-web-modeler-to-hub-api.md
```diff
@@ -113,42 +284,373 @@ All v1 files endpoints have a direct v2 equivalent:
...
+
+| Operation            | Web Modeler API v1                                     | Camunda Hub API v2                           |
+| -------------------- | ------------------------------------------------------ | -------------------------------------------- |
+| Create a version     | `POST /v1/versions`                                    | `POST /v2/versions`                          |
+| Get a version        | `GET /v1/versions/{versionId}`                         | `GET /v2/versions/{versionKey}`              |
+| Update a version     | `PATCH /v1/versions/{versionId}`                       | `PATCH /v2/versions/{versionKey}`            |
+| Delete a version     | `DELETE /v1/versions/{versionId}`                      | `DELETE /v2/versions/{versionKey}`           |
+| Search versions      | `POST /v1/versions/search`                             | `POST /v2/versions/search`                   |
+| Restore a version    | `POST /v1/versions/{versionId}/restore`                | `POST /v2/versions/{versionKey}/restoration` |
+| Compare two versions | `GET /v1/versions/compare/{version1Id}...{version2Id}` | Does not exist                               |
```
ThorbenLindhauer: Maybe link here to the "compare two versions" subsection, too.

docs/apis-tools/migration-manuals/migrate-from-web-modeler-to-hub-api.md
```diff
@@ -113,42 +284,373 @@ All v1 files endpoints have a direct v2 equivalent:
...
+2. Insert the keys into one of the following URL patterns:
+
+| Resource type    | Template URL                                                                     |
+| :--------------- | :------------------------------------------------------------------------------- |
+| BPMN             | `{baseUrl}/diagrams/{fileKey}/versions/{versionKey1}...{versionKey2}`            |
+| Element template | `{baseUrl}/connector-templates/{fileKey}/versions/{versionKey1}...{versionKey2}` |
+| Form             | `{baseUrl}/forms/{fileKey}/versions/{versionKey1}...{versionKey2}`               |
+| RPA              | `{baseUrl}/rpa-scripts/{fileKey}/versions/{versionKey1}...{versionKey2}`         |
+
+Replace `{baseURL}` with the Web Modeler base URL. The version keys must be for the same file.
```
ThorbenLindhauer: Shall we give a full example for Camunda SaaS here? We only have one Hub Prod instance, so the URL would be static. Only in Self-Managed I have to know my host.

ThorbenLindhauer (changes requested):
Thanks for preparing this. Please see my feedback below. Most of it is around some misunderstanding between the changes of workspaces/projects/process applications/folders. If you read this and still have doubts, please reach out. We can have a Zoom chat, that may be quicker to resolve everything.

To your questions:

* Re 1: Yes, I think that is fine. I suggest to have code examples only where we really think it helps people understand the change. E.g. in cases where a field is renamed, we don't need an example. In cases in which the structure of the JSON has changed, an example helps a lot.
* Re 2: Cursor pagination is currently not implemented. It's a stretch goal and I haven't seen an issue for it: https://camunda.slack.com/archives/C0AEAM0S8LU/p1780057229480079. It's in the schema (e.g. here: http://localhost:3000/docs/next/apis-tools/hub-api-saas/specifications/search-projects/) but in the current implementation always `null`. Please check with Nic.
* Re 3: Agree.
* Re 4: Yeah, that's a set of bugs I am currently fixing with https://github.com/camunda/camunda-hub/issues/24664. So the original claim "V1 projectId and V2 workspaceKey refer to the same underlying entity" is the correct one. Plus that with the new information architecture of Workspace -> Project -> File/Folder, it's intended that you cannot create a file directly in a workspace. You must always have a project.

Other comments:

* In the introduction of the guide, we can add an explanation that the Hub V2 API guidelines adopt the OC V2 API conventions to a large degree. That will give people who are already familiar with the OC API an intuitive understanding of what has changed and how things work.
* Not directly related to your changes here but something I wanted to share: We can make the authentication guides between OC API and Hub API more consistent. At least for SaaS the process to create a client and obtain a token is very similar. I think by having a similar content structure and formulations we make it easier for users to understand that. So we could have more consistency between https://docs.camunda.io/docs/next/apis-tools/orchestration-cluster-api-rest/orchestration-cluster-api-rest-authentication/#using-a-token-oidcjwt and 
* Btw the V2 API is exposed in SaaS dev, so feel free to try it out there and see what's there

ThorbenLindhauer:
I have also posted a brief summary of how our target data model looks like here: https://camunda.slack.com/archives/C07USKREKDL/p1782751395620439?thread_ts=1782751388.520769&cid=C07USKREKDL

Maybe that helps for orientation.

docs/apis-tools/migration-manuals/migrate-from-web-modeler-to-hub-api.md
alexronquillo: I'm confused by this. This was included as a deliverable in the issue I'm working on. It's also included in the published API docs search responses: https://docs.camunda.io/docs/next/apis-tools/hub-api-saas/specifications/search-files/ Do you mean it's not there _yet_ or it won't be there by launch?

Edit: Somehow I missed your comment about this above.

docs/apis-tools/migration-manuals/migrate-from-web-modeler-to-hub-api.md
alexronquillo: yes, my mistake. this comes directly from the issue description, but I missed it in my edits.

docs/apis-tools/migration-manuals/migrate-from-web-modeler-to-hub-api.md
alexronquillo: yes, my mistake.

docs/apis-tools/migration-manuals/migrate-from-web-modeler-to-hub-api.md
ThorbenLindhauer: I understand your confusion :)

@nicpuppa can you please clarify the state of cursor pagination now and if anything is planned to change until the release?

nicpuppa: @alexronquillo @ThorbenLindhauer cursor pagination is not implemented at the moment, as of now I don't think will be implemented for the 8.10 release

We can also remove it from the specification if it creates confusion.

docs/apis-tools/hub-api-sm/authentication.md
alexronquillo: Yeah, there's a lot going on here. Let's keep it as-is for now and address it with the other issue. I'll note this for review when we get closer to release.

docs/apis-tools/migration-manuals/migrate-from-web-modeler-to-hub-api.md
alexronquillo: I think we should remove it from the spec. Will you take care of that @nicpuppa?

nicpuppa: Yes, I can

nicpuppa: here's the PR https://github.com/camunda/camunda-hub/pull/25910

docs/apis-tools/migration-manuals/migrate-from-web-modeler-to-hub-api.md
alexronquillo: I've removed cursor-related examples.

docs/apis-tools/migration-manuals/migrate-from-web-modeler-to-hub-api.md
alexronquillo: I've edited this section to try to clear this up with the following changes:

- Focus on the **product-level** data structure and terminology changes.
- Leave the application of these changes in the API out of this section. 
- Instead, explain how the product-level changes apply to the API in context throughout the rest of the migration guide.

After consideration, I still used the approach of:

1. Data organization changes
2. Terminology changes

My reasoning is:

- The concept of "process applications" wasn't exposed explicitly in v1. So I felt it made more sense to provide the data organization so the terminology changes would follow and depend on that context. 
- This section now serves as background information, rather than addressing API-specific changes. So it doesn't call out API-level terminology changes anymore.

Please let me know after reading if you still disagree. Happy to change it if you think it will be better the other way around.

alexronquillo:
@nicpuppa The issue says the error format has changed, giving this as the old format:

```
{ "message": "File not found", "status": 404 }
```

However, the v1 response I get is as described [here](https://modeler.camunda.io/swagger-ui/index.html#/Files/getFile):

```bash
GET /api/v1/files/12345a67-89b0-1c23-4567-890d12345678

{
    "title": "Not Found",
    "status": 404,
    "detail": "Resource not found.",
    "instance": "/api/v1/files/12345a67-89b0-1c23-4567-890d12345678"
}
```

Can you give a little more explanation here? Thanks!

nicpuppa:
@alexronquillo looks like an oversight from my side, sorry for that. The V1 error message looks like has the same structure of V2

docs/apis-tools/migration-manuals/migrate-from-web-modeler-to-hub-api.md
```diff
@@ -113,42 +301,661 @@ All v1 files endpoints have a direct v2 equivalent:
...
-| ------------------ | ------------------ | -------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `folderId`         | `folderKey`        | Renamed.                                                                                                                                                 |
-| `projectId`        | `projectKey`       | Renamed. Use to move files between projects within a Camunda Hub workspace.                                                                              |
-| `revision`         | `revision`         | Now required. Fetch the current revision from a get or create response, and include it in your update request to prevent overwriting concurrent changes. |
+A `revision` is now required to prevent overwriting concurrent changes. Fetch the current revision from a get or create response, and include it in your update request:
+
+```json title="Camunda Hub API v2"
```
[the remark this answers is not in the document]
alexronquillo: This is a meta comment used for highlighting the UI codeblock.

alexronquillo:
@ThorbenLindhauer Thank you for the thorough review. I've addressed all your comments over many commits. To highlight some of the bigger changes:

- Tested almost every request to make sure everything was accurate, and made updates accordingly.
- Added codeblocks where non-trivial changes are introduced.
- Renamed the collaborators API to "members"
- Consistently used v2 naming conventions and gave v1 terminology where appropriate
- Updated the intro to call out similarities between the OC APIs and Hub APIs
- Updated the SaaS auth doc to match the actual current state
- Fixed mistakes confusing v1/v2 terminology vs pre-8.10/8.10 terminology

I also created [another issue](https://github.com/camunda/camunda-docs/issues/9330) to address the inconcistencies between the Hub API and the OC API auth docs.

This is ready for another technical review. @nicpuppa @wollefitz

christinaausley (dismissed):
I don't have any direct edits to make, but two non-blocking comments:

1. The SaaS auth guide grants the client access to the Web Modeler API scope, but the Self-Managed guide grants the Camunda Hub API scope. Which is correct for v2? And is a reused Web Modeler API v1 token with audience `api.cloud.camunda.io` actually accepted by `hub.cloud.camunda.io/api/v2`?
2. Does Camunda Hub API v2 support the full operator set shown for Modeler resources (`$neq`, `$notIn`, `$exists`, `$or`), or is that the generic Orchestration Cluster API v2 list where some may not apply to files/folders/workspaces?

alexronquillo:
@christinaausley Thanks for the review! 

1. As far as I understand, `Camunda Hub API` will be the final scope. Originally, I had Camunda Hub API for both, but I realized by trying to get ahead of the product (writing `Camunda Hub API` while the product still currently uses `Web Modeler API`) it can be confusing for people using the current state, so I changed it. Ultimately, I guess we can make a case for either direction. I'll push another commit to use "Web Modeler" for both until it gets updated in the UI/backend.
2. I tested the operators quite a bit, but I didn't test every possible permutation. It seems to me like the full set is supported in all search APIs in the Camunda Hub API. @nicpuppa please correct me if I'm wrong.

docs/apis-tools/migration-manuals/migrate-from-web-modeler-to-hub-api.md
```diff
@@ -9,14 +9,72 @@ import PageDescription from '@site/src/components/PageDescription';
...
+Web Modeler API v1 is the REST API for Web Modeler, a standalone product for modeling and managing process diagrams. It exposes resources like projects, folders, files, and collaborators as they exist within Web Modeler.
 
-[Camunda Hub API v2](/apis-tools/hub-api-sm/overview.md) is the successor API for the broader Camunda Hub platform. Camunda Hub unifies organizational management, workspace governance, and process modeling into a single platform. As a result, the conceptual model and architecture of the API have changed.
+[Camunda Hub API v2](/apis-tools/hub-api-saas/overview.md) is the successor API for the broader Camunda Hub platform. Camunda Hub unifies organizational management, workspace governance, and process modeling into a single platform. As a result, the conceptual model and architecture of the API have changed.
+:::tip
+Camunda Hub API v2 adopts the Orchestration Cluster API v2 conventions. If you're already familiar with the Orchestration Cluster API v2, you will recognize patterns such as the offset-based pagination model, explicit filter operators, and flat response structures used throughout Camunda Hub API v2.
+:::
```
ThorbenLindhauer, line 23: Thanks for adding this.

"Orchestration Cluster API v2" could also link to the respective page.

docs/apis-tools/migration-manuals/migrate-from-web-modeler-to-hub-api.md
```diff
@@ -35,75 +93,154 @@ In Camunda 8 Self-Managed, the URLs depend on your configuration. The URLs and p
...
+```json title="Web Modeler API v1"
+{
+  "filter": {
+    "name": "my-process",
+    "type": "bpmn"
+  }
+}
+```
+
+Camunda Hub API v2 extends this behavior with a set of explicit operators:
```
ThorbenLindhauer: Maybe we can be a bit more elaborate on the "extends" part. It's easy for readers to skip over this word (so they might misunderstand that this is not an extension but a change) and get worried that they now have to rewrite all their query filters, even if they just test equality.

A statement that the simple equality filters exist as before could make this clear.

docs/apis-tools/migration-manuals/migrate-from-web-modeler-to-hub-api.md
```diff
@@ -113,42 +250,661 @@ All v1 files endpoints have a direct v2 equivalent:
...
+In Camunda Hub API v2, the collaborators API has been renamed to `members`. The following sections cover changes that apply to the member API endpoints.
+
+### Endpoint mapping
+
+All member API endpoints have a Camunda Hub API v2 equivalent:
+
+| Operation             | Web Modeler API v1                             | Camunda Hub API v2                                     |
+| --------------------- | ---------------------------------------------- | ------------------------------------------------------ |
+| Add a collaborator    | `PUT /v1/collaborators`                        | `POST /v2/workspaces/{workspaceKey}/members`           |
+| Remove a collaborator | `DELETE /v1/collaborators/{projectId}/{email}` | `DELETE /v2/workspaces/{workspaceKey}/members/{email}` |
```
nicpuppa: ❌ The v1 path for removing a collaborator is transposed. The actual v1 endpoint is:

```
DELETE /v1/projects/{projectId}/collaborators/{email}
```

Source: `PubV1CollaboratorController#deleteCollaborator` → `@DeleteMapping("/projects/{projectId}/collaborators/{email}")` with base path `/api/v1`.

docs/apis-tools/migration-manuals/migrate-from-web-modeler-to-hub-api.md
```diff
@@ -113,42 +250,661 @@ All v1 files endpoints have a direct v2 equivalent:
...
+| `filter`                 | `filter`            | Now uses [advanced operators](#search-filters), including `$eq`, `$in`, and `$like`                                                                                                                                                                                                                                                                               |
+| `filter.projectId`       | `filter.projectKey` | In v1, `projectId` refers to the ID of the "workspace" ([called the "project" before Camunda 8.10](#structure-and-terminology)). In v2, `projectKey` refers to the key of the "project". The concept of a "project" was [called a "process application" before Camunda 8.10](#structure-and-terminology), but you can't filter on process application data in v1. |
```
nicpuppa: ❌ There is no `filter.projectKey` for file search. `FileFilter` (`files.yaml`) exposes only `name`, `folderKey`, `type`, `created`, `updated`, `createdBy`, `updatedBy` — filtering files by project isn't supported in v2 (consistent with project data not being resolved in search results, see the note on line 395). Suggest dropping this row or marking it as not available in v2.

docs/apis-tools/migration-manuals/migrate-from-web-modeler-to-hub-api.md
```diff
@@ -113,42 +250,661 @@ All v1 files endpoints have a direct v2 equivalent:
...
+| Get info  | `GET /v1/info`     | `GET /v2/info`     |
+
+### Get info
+
+The following response fields have changed:
 
-Only changed fields are listed:
+| Web Modeler API v1 | Camunda Hub API v2 | Notes   |
+| ------------------ | ------------------ | ------- |
+| `createPermission` | -                  | Removed |
```
nicpuppa: 🔧 Optional: consider also noting that the `version` field value changes from `"v1"` to `"v2"` (`InfoResult` returns `"v2"` in `info.yaml`). It's the one behavioral change to the retained fields and was called out in the source issue.

docs/apis-tools/migration-manuals/migrate-from-web-modeler-to-hub-api.md
```diff
@@ -113,42 +250,661 @@ All v1 files endpoints have a direct v2 equivalent:
...
+| ------------------ | ------------------ | ---------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `fileType`         | `type`             | Request/response | Renamed. `connector_template` is no longer supported. `element_template` is now `element-template`.                                                                                                                                                                                                                                                                     |
```
nicpuppa: 💭 Minor wording: "`connector_template` is no longer supported" is accurate for **creation** (it's not in the `FileCreateRequest` type enum), but existing connector-template files are still returned — surfaced as `type: "element-template"` (`V2FileResponseMapper#mapFileTypeValue` maps `CONNECTOR_TEMPLATE` → `element-template`). Consider: "can no longer be created; existing files are reported as `element-template`".

docs/apis-tools/migration-manuals/migrate-from-web-modeler-to-hub-api.md
```diff
@@ -113,42 +250,661 @@ All v1 files endpoints have a direct v2 equivalent:
...
+| `id`               | `fileKey`          | Response         | Renamed                                                                                                                                                                                                                                                                                                                                                                 |
+| `canonicalPath`    | `canonicalPath`    | Response         | In v1, `canonicalPath` is a list of objects. In v2, it's a [string](#canonical-path).                                                                                                                                                                                                                                                                                   |
+
+### Canonical path
```
nicpuppa, line 267: 💭 FYI (spec bug, not the guide): this v2 `canonicalPath` format — folder keys only, `/`-joined, no leading slash, no file key — matches the actual `V2FileResponseMapper#buildCanonicalPath` output. The `FileResult` **example** in `files.yaml` (`"/fld-012-ghi/file-789-xyz"`) is the inconsistent one (leading slash + file key). Might be worth a fix against the spec example rather than the guide.

ThorbenLindhauer (dismissed):
Thanks for taking up my suggestions. I find the guide very clear and easy to follow now, as well as easily browsable (e.g. if I am looking what changed for request X, I can easily find that).

Two small suggestions below.

docs/apis-tools/migration-manuals/migrate-from-web-modeler-to-hub-api.md
alexronquillo: Thanks 👍 no action taken with respect to this migration guide.

alexronquillo: Created a PR to handle the spec changes: https://github.com/camunda/camunda-hub/pull/26209

alexronquillo:
@nicpuppa @ThorbenLindhauer I've addressed all the comments. This PR is ready for another pass.

nicpuppa (dismissed):
good job, thanks

alexronquillo:
@nicpuppa I had to resolve a merge conflict. Can you please approve one more time? 🙏

nicpuppa (approved):

PR #9215, opened 2026-06-26
Announce Tasklist V1 candidate group task visibility regression

fabiopaini-camunda:
## Description

Adds an announcement entry to the 8.8 and 8.9 release announcement pages documenting a candidate group task visibility regression introduced in patch releases 8.8.26 and 8.9.7.

The regression occurs because the Zeebe engine resolves candidate group names to IDs at task creation time, while the Tasklist V1 API resolves the authenticated user's group IDs back to names before comparison. When a group's name differs from its ID, this mismatch causes tasks to be invisible to all members of that group.

The announcement covers:
- What the bug is and when a customer is affected (using Tasklist V1 API with user task access restrictions, with candidate groups whose name differs from their ID)
- The workaround (set `ZEEBE_BROKER_EXPERIMENTAL_ENGINE_CACHES_CANDIDATEGROUPNAMERESOLUTION=false` and reference candidate groups by name in BPMN)
- The fix (8.8.29 and 8.9.10) which also retroactively fixes tasks created while the regression was active

All 5 announcement files are updated (two in `docs/`, three in `versioned_docs/`) following the same pattern used for the multi-instance regression announcement.

## Checklist

- [x] I have read the [contributing guidelines](https://github.com/camunda/camunda-docs/blob/main/CONTRIBUTING.md).
- [x] I have checked that the content is accurate and follows the style guide.

## Related issues

Relates to [camunda/camunda#55576](https://github.com/camunda/camunda/issues/55576)
More info in incident channel: [#inc-support-33391-tasklist-candidate-group-issue-8-9gen9](https://camunda.slack.com/archives/C0BBPKK12SZ)


mesellings:
@afgambin can you take a look at this for me please - there also appears to be an issue with two versions of the Engine multi-instance announcement in 8.8 and 8.9 which might be worth fixing in this issue or separately? 🙏

afgambin (approved):
@fabiopaini-camunda I did some minor text teaks, lgtm now!

@mesellings I addressed the engine entry duplication directly here.

Issue #8949, opened 2026-05-28
Update workflows to use vault secrets


alexronquillo:
For several of our workflows, we're currently using GH secrets instead of vault secrets. This is an anti-pattern, so we should migrate to vault secret usage ASAP. Especially considering that camunda-docs is a public repo.

[Slack reference](https://camunda.slack.com/archives/C5AHF1D8T/p1779954089421679?thread_ts=1779444488.671519&cid=C5AHF1D8T)

PR #9243, opened 2026-06-29
Vault secrets

alexronquillo:
## Description

Use existing vault credentials instead of GH credentials. These credentials are already used in other workflows, so just making everything consistent here. I don't expect to see any issues from this change.

The second part of this migration (new secrets that only currently exist in GH) will come in follow-up PRs.

Refs: https://github.com/camunda/camunda-docs/issues/8949

## PR Checklist

### DIFF
--- a/.github/workflows/add-to-project.yaml
+++ b/.github/workflows/add-to-project.yaml
@@ -9,11 +9,24 @@ jobs:
     if: github.event.requested_reviewer.login == 'christinaausley' || github.event.requested_reviewer.login == 'mesellings' || github.event.requested_team.name == 'Tech Writers'
     runs-on: ubuntu-latest
     steps:
+      - name: Import Secrets
+        id: secrets
+        uses: hashicorp/vault-action@v4.0.0
+        with:
+          url: ${{ secrets.VAULT_ADDR }}
+          method: approle
+          roleId: ${{ secrets.VAULT_ROLE_ID }}
+          secretId: ${{ secrets.VAULT_SECRET_ID }}
+          exportEnv: false
+          secrets: |
+            secret/data/products/camunda-docs/github/apps/camunda-docs-pr-automation APP_ID;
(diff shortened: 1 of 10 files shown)


.github/workflows/add-to-project.yaml
```diff
@@ -9,11 +9,24 @@ jobs:
     if: github.event.requested_reviewer.login == 'christinaausley' || github.event.requested_reviewer.login == 'mesellings' || github.event.requested_team.name == 'Tech Writers'
     runs-on: ubuntu-latest
     steps:
+      - name: Import Secrets
+        id: secrets
+        uses: hashicorp/vault-action@v4.0.0
```
afgambin: I was comparing this with the recent workflow we introduced, `.github/workflows/claude-review.yml`. I also found `preview-env-deploy.yml` and `preview-env-teardown.yml`, and they all pin vault-action to a commit SHA (79632e33d6953d190b940ffa440bf97821cabd80), which seems to be the most secure approach because a SHA is immutable.

Here, we are using `hashicorp/vault-action@v4.0.0` instead, which is a mutable tag. If HashiCorp were to move or overwrite that tag, the action code could change silently. Low risk anyhow, but it is probably better to be consistent across all workflows, and use whichever approach you think is best.

afgambin (dismissed):
@alexronquillo lgtm, left a non-blocking comment

.github/workflows/add-to-project.yaml
alexronquillo: Thanks for raising this. 

I guess we should probably update our internal guidance: https://docs.int.camunda.com/docs/vault/using-vault-secrets-in-github-actions/#basic-usage @Kerruba @kellervater

afgambin (approved):
lgtm!

.github/workflows/add-to-project.yaml
kellervater: good point. feel free to raise a PR in camunda/infra-core (Copilot or Claude could be of great help here) and request a review in Slack (#epo-reviews)

PR #9244, opened 2026-06-30
docs: collect Elasticsearch/OpenSearch _cat/indices in the diagnostics script

lenaschoenburg:
## Description

Follow-up to the actuator collection added in #9210. @epollum asked that the diagnostics bundle also include the secondary-storage index list, which is useful when investigating storage growth or exporter issues.

Adds a step that port-forwards the datastore HTTP port (`9200`) and saves `_cat/indices?v` from the first reachable Elasticsearch/OpenSearch node into `<pod>-cat-indices.txt`. The listing is cluster-wide, so it stops after the first node that answers. It reuses the same ephemeral-port + bounded-retry + trap-based cleanup machinery as the actuator step.

Secured datastores (TLS/auth) are skipped — we have no credentials in this context — and the docs note they must be collected manually.

Also adds a matching "Database indices" entry to the "What the script collects" list.

Applied consistently to all four copies of the page (current, 8.9, 8.8, 8.7).


epollum (approved):
looks good! I'll be sure to let support know about this change, and the actuator additions.

PR #9245, opened 2026-06-30
docs: add Spring Boot version support policy to starter getting-started page

aleksander-dytko:
## Description

Adds a `## Version support policy` section to the Camunda Spring Boot Starter getting-started page, below the existing version compatibility table.

The section covers:

- **Supported versions** - Camunda supports two Spring Boot major versions at a time: the current major (latest stable minor, default artifact) and the previous major (pinned to its last stable minor, dedicated artifact).
- **Minor version updates** - Camunda tracks only the latest stable minor within the current major. New minors (e.g. 4.1) are adopted in the next Camunda release after Spring Boot minor GA. No separate artifacts per minor.
- **New major version support** - The new major becomes default in the Camunda release before the previous major's OSS support ends (example: Spring Boot 4.x became default with 8.9 before Spring Boot 3.5 OSS ended in June 2026).
- **Deprecating older major versions** - Previous major maintained (last stable minor) until end of Spring Commercial support, then no further updates.
- **Current version status table** - 3.5.x and 4.0.x with OSS and Commercial support dates and Camunda status.

**Context:** Recurring customer questions about Spring Boot minor version support (e.g. "is 4.1 planned?") and the Spring Boot 4.0 OSS EOL in December 2026. This policy gives a clear, referenceable answer.

**Review needed from:** Marcos Vieira / Sebastian Bathke (Meggle) for accuracy against internal engineering commitments before merging.

## PR Checklist

  - [ ] [Engineering team review](https://github.com/camunda/camunda-docs/blob/main/howtos/documentation-guidelines.md#review-process) - pending Marcos / Meggle alignment
  - [ ] [Technical writer review](https://github.com/camunda/camunda-docs/blob/main/howtos/documentation-guidelines.md#review-process) via `@camunda/tech-writers`


megglos:
@aleksander-dytko upfront, does this duplicate/partially replace https://docs.camunda.io/docs/reference/announcements-release-notes/release-policy/#spring-bootframeworksecurity-updates ? or could we link there 

the current setup is we have overall update policies wrt to Spring defined here https://docs.camunda.io/docs/reference/announcements-release-notes/release-policy/#spring-bootframeworksecurity-updates 
and compatibility with spring minor versions of the SDK documented here https://docs.camunda.io/docs/apis-tools/camunda-spring-boot-starter/getting-started/#version-compatibility

The latter is probably hard to discover rn so at least linking from the SDK page would be beneficial.

docs/apis-tools/camunda-spring-boot-starter/getting-started.md
```diff
@@ -71,6 +71,42 @@ To use the Spring Boot 3 module, replace the default dependency in your project:
 </dependency>
 ```
 
+## Version support policy
```
megglos: 🔧 Most of this section is general policy that is not tied to a specific version. The best home for "how Camunda handles Spring support windows" is release-policy.md → **Client libraries and SDKs** (`#spring-bootframeworksecurity-updates`) — but that section does not yet answer this question for the starter. Suggestion: move the general rules (Supported versions / New major / Deprecating older majors) there, and replace this heading with a short link, for example: _"For the full Spring Boot version support policy, see [Release policy → Spring Boot/Framework/Security updates](...)."_ This also fixes the "hard to find" concern — the policy page already links back down to the compatibility table here.

docs/apis-tools/camunda-spring-boot-starter/getting-started.md
```diff
@@ -71,6 +71,42 @@ To use the Spring Boot 3 module, replace the default dependency in your project:
...
+Camunda aligns Spring Boot Starter version support with the [Spring Boot support timeline](https://spring.io/projects/spring-boot#support).
+
+### Supported versions
+
+Camunda supports two Spring Boot major versions at a time:
+
+- **The current major version** - shipped as the default `camunda-spring-boot-starter` artifact, always bundled with the latest stable minor of that major (for example, once Spring Boot 4.1 is available, `camunda-spring-boot-starter` will bundle 4.1.x instead of 4.0.x).
+- **The previous major version** - shipped as a dedicated `camunda-spring-boot-{n}-starter` artifact, pinned to the last stable minor of that major line. Supported until the end of [Spring Commercial support](https://spring.io/projects/spring-boot#support) for that major.
+
+### Minor version updates
```
megglos: ❌ This is incorrect, see https://docs.camunda.io/docs/reference/announcements-release-notes/release-policy/#client-libraries-and-sdks For a given Camunda minor release we'll always stay on the Spring boot minor our minor got originally released with. Otherwise updating the default in a patch could be a breaking change. However we test for support with newer spring minor releases and update the compatibility table accordingly, it's then up to the user to update to the new spring minor.

docs/apis-tools/camunda-spring-boot-starter/getting-started.md
```diff
@@ -71,6 +71,42 @@ To use the Spring Boot 3 module, replace the default dependency in your project:
 </dependency>
 ```
 
+## Version support policy
+
+Camunda aligns Spring Boot Starter version support with the [Spring Boot support timeline](https://spring.io/projects/spring-boot#support).
+
+### Supported versions
```
megglos: 🔧 These bullets repeat the **Dedicated Spring Boot 3 and 4 modules** section already on this page (line 48): the default `camunda-spring-boot-starter` plus the dedicated `-3`/`-4` artifacts. If the general rule moves to release-policy.md, this part is no longer needed here.

docs/apis-tools/camunda-spring-boot-starter/getting-started.md
```diff
@@ -71,6 +71,42 @@ To use the Spring Boot 3 module, replace the default dependency in your project:
...
+
+:::note Example
+Spring Boot 3.5 OSS support ended June 2026. Camunda made Spring Boot 4.x the default with Camunda 8.9 (released April 2026), the last Camunda release before the 3.5 OSS end date.
+:::
+
+### Deprecating older major versions
+
+Once a new major becomes the default, the previous major is maintained as a dedicated artifact (pinned to its last stable minor) until the end of Spring Commercial support for that major. After that point, the dedicated artifact receives no further updates.
+
+### Current version status
```
megglos: 🔧 This is a second table organized by Spring Boot version, and it overlaps the **Version compatibility** table at the top of the page (which already maps each artifact to its bundled/compatible Spring Boot version). Keeping two tables means they can easily fall out of sync — the top table already shows 8.10.x with 4.0.x, while the "Camunda status" column here talks about moving to 4.1. Suggestion: add the useful columns (Spring OSS support, Camunda status) to the existing table and remove this one. Also, "until the end of Spring Commercial support" now appears about 4 times on the page (caution box + two subsections + this table).

megglos (changes requested):
Thanks for tackling this — it fills a real gap. Right now neither page clearly answers "what happens to the starter when Spring's free support ends before Camunda's support window."

My main question is **where** this content should live. Most of the new section describes general rules that apply to every version, and those rules repeat — and in one place disagree with — the existing [release-policy Spring section](https://docs.camunda.io/docs/reference/announcements-release-notes/release-policy/#spring-bootframeworksecurity-updates), see the inline ❌  comment. The new status table also overlaps the Version compatibility table already on this page.

**Suggested split:**

_Move the general rules to `release-policy.md` → Client libraries and SDKs_ (that is the main page for support policy, and it is easier to find there):
- Camunda always ships a numbered artifact for each Spring Boot major (for example `camunda-spring-boot-3-starter`, `camunda-spring-boot-4-starter`), so users can pin to a major and stay on it; the plain `camunda-spring-boot-starter` just follows the current default major. Two majors are supported at a time, the previous one until the end of Spring Commercial support.
- New major becomes the default before the previous major's free support ends
- Make the minor-version wording match the existing "one month before the support window ends" rule, so both pages say the same thing (see inline)

_Keep on this page_ (the practical, hands-on info a developer needs):
- The **Version compatibility** table as the one place that maps each artifact to its Spring Boot version (you could add "Spring OSS support" / "Camunda support" columns) — and drop the new second table
- The Dedicated modules section + dependency snippets
- The edited 3.5.x caution box (the fixes there are good)
- **A link up to the policy page** — this solves the "hard to find" problem raised earlier; the policy page already links down to the table here, so the two would point at each other

@afgambin as you previously did an amazing job on making this page better, can you help here again? 😃

aleksander-dytko:
@megglos thanks for the thorough review of the PR! 

@afgambin could i hand over the PR to you to complete it?

docs/apis-tools/camunda-spring-boot-starter/getting-started.md
afgambin: Good point! General policy rules moved to release-policy.md ([Spring Boot/Framework/Security updates](https://github.com/camunda/camunda-docs/pull/...)). The major version support rules now live under "Client libraries and SDKs"

docs/apis-tools/camunda-spring-boot-starter/getting-started.md
afgambin: Removed

docs/apis-tools/camunda-spring-boot-starter/getting-started.md
afgambin: Good call, added a new "Spring OSS support" column in the existing compatibility table, keeping only one table

docs/apis-tools/camunda-spring-boot-starter/getting-started.md
```diff
@@ -39,11 +39,11 @@ With the Camunda Spring Boot Starter, you can build:
 
 ## Version compatibility
 
-| Camunda Spring Boot Starter artifact | Camunda Spring Boot Starter version | JDK  | Bundled Spring Boot version | Compatible Spring Boot version(s) |
-| ------------------------------------ | ----------------------------------- | ---- | --------------------------- | --------------------------------- |
-| `camunda-spring-boot-starter`        | 8.10.x                              | ≥ 17 | 4.0.x                       |                                   |
-| `camunda-spring-boot-4-starter`      | 8.10.x                              | ≥ 17 | 4.0.x                       |                                   |
-| `camunda-spring-boot-3-starter`      | 8.10.x                              | ≥ 17 | 3.5.x                       |                                   |
+| Camunda Spring Boot Starter artifact | Camunda Spring Boot Starter version | JDK  | Bundled Spring Boot version | Compatible Spring Boot version(s) | Spring OSS support |
```
afgambin: I see the `Compatible Spring Boot version(s)` column is empty. Is that on purpose?

afgambin:
> could i hand over the PR to you to complete it?

@aleksander-dytko Yes, happy to!

@megglos I addressed all your comments and did a review pass. LGTM now! I left a non-blocking, possibly PR-unrelated question. Ready for another pass, thanks!

docs/apis-tools/camunda-spring-boot-starter/getting-started.md
megglos: yes as of now there wasn't any newer release to support, this will change shortly though as 4.1 got releases and well add support for that.

For the `camunda-spring-boot-3-starter` no new entries will be added as there will be no new 3.x minor from spring. 3.5 was the last.

docs/reference/announcements-release-notes/release-policy.md
```diff
@@ -158,3 +158,12 @@ However, Camunda tests against newer Spring Boot minor releases and declares com
 If the library is compatible with a newer Spring version with an active OSS support window, Camunda declares compatibility one month before the support window ends, at the latest.
 
 With this policy, you can safely update your applications to a new Spring Boot version by overriding the default Spring release with another compatible version.
+
+**Spring Boot major versions**
+
+Camunda supports two Spring Boot major versions at a time:
+
+- The **current major** is shipped as the default `camunda-spring-boot-starter` artifact.
+- The **previous major** is shipped as a dedicated `camunda-spring-boot-{n}-starter` artifact, pinned to the last stable minor of that major. It remains supported until the end of Spring Commercial support for that major, after which no further updates are released.
```
megglos: ❌ This says a numbered artifact ships only for the *previous* major, but Camunda actually ships a numbered artifact for **every** supported major, so users can pin to a major and stay on it. It also disagrees with the getting-started page, which lists `camunda-spring-boot-4-starter` for the current major and calls it "Identical to `camunda-spring-boot-starter`". Suggestion: say that each supported major has its own `camunda-spring-boot-{n}-starter`, and the plain `camunda-spring-boot-starter` just follows the current default major. For example:

> - A numbered `camunda-spring-boot-{n}-starter` artifact is published for each supported major, so you can pin to a major and stay on it.
> - The plain `camunda-spring-boot-starter` always follows the current default major.
> - The previous major stays supported until the end of Spring Commercial support for that major, after which it receives no further updates.

docs/reference/announcements-release-notes/release-policy.md
```diff
@@ -158,3 +158,12 @@ However, Camunda tests against newer Spring Boot minor releases and declares com
 If the library is compatible with a newer Spring version with an active OSS support window, Camunda declares compatibility one month before the support window ends, at the latest.
 
 With this policy, you can safely update your applications to a new Spring Boot version by overriding the default Spring release with another compatible version.
+
+**Spring Boot major versions**
```
megglos: 🔧 This is bold text, while its siblings under *Client libraries and SDKs* are real headings (`#### Orchestration Cluster`, `#### Client libraries and SDKs`). As bold it won't show up in the table of contents or be deep-linkable. A `#####` heading would be more consistent.

docs/apis-tools/camunda-spring-boot-starter/getting-started.md
megglos: 🔧 Seeing this I actually am in favor of omitting that fully, I don't see a strong need to maintain spring periods here as well and it bears the potential to be misinterpreted as our support window. Furthermore as soon as we list multiple compatible Spring Boot versions, each will come with their own OSS period. I would at most link to the spring support window page being https://spring.io/projects/spring-boot#support wdyt?

megglos (changes requested):
Thanks for the rework — this addresses the earlier feedback well. The general rules moved to `release-policy.md`, the second table is gone (folded into the compatibility table as a "Spring OSS support" column), the duplicated bullets are removed, and the crosslink is in place. 👍

Three follow-up points, one of them worth fixing before merge (❌); the other two are minor (🔧).

docs/reference/announcements-release-notes/release-policy.md
afgambin: Yes, good call. Tbh, we don’t have clear guidelines around deeply nested headings, but internally we’ve tried not to go deeper than `####`. Anyhow, I agree with you that this is a better approach for cross-linking. Updated!

docs/reference/announcements-release-notes/release-policy.md
afgambin: Text updated following your suggestion

docs/apis-tools/camunda-spring-boot-starter/getting-started.md
afgambin: Yes, SGTM. I added a sentence with the link below the table.

afgambin:
@megglos Thanks for your feedback, addressed. Ready for another pass!

megglos (dismissed):
Thanks for the iterations, this looks good now. ✅

Both review rounds are addressed:
- The support rules now live in `release-policy.md`, and the getting-started page links to them. One home for the policy, no duplication across pages.
- The second table is gone. The compatibility table stays the single reference, with a link out for the Spring dates.
- The major-version wording is correct now: a numbered `camunda-spring-boot-{n}-starter` ships for each supported major, and the plain `camunda-spring-boot-starter` follows the current default major.
- `##### Spring Boot major versions` is a proper heading.

A reader can now clearly see what we support and how we handle Spring support windows closing before ours. Approving.

megglos:
ah one more thing! we need to also apply this to the versioned docs

afgambin:
@megglos Backported. Please have a look and reapprove if everything looks good, thanks!

megglos (approved):
Amazing, thank you!

PR #9257, opened 2026-06-30
resolve remaining operate tasklist api references

christinaausley:
## Description

Resolves https://camunda.slack.com/archives/C08P3BN5QM7/p1782811143273599.

## PR Checklist


tanjinyi:
🔧 Just an addendum to the ones I identified: https://docs.camunda.io/docs/next/apis-tools/zeebe-api-rest/zeebe-api-rest-overview/

I believe the entire `Zeebe API (REST)` section here should also be removed as they would have been part of the V1 API.

docs/self-managed/concepts/multi-region/dual-region.md
```diff
@@ -23,58 +23,44 @@ Before implementing a dual-region setup, ensure you understand the topic, the [l
...
+| <p align="left">**Orchestration Cluster**</p> |               | ✅ Required              |                           |         |
+|                                         Zeebe | Active-active | ✅ Required              | Both regions process data | 0       |
+|                                         Admin | Active-active | ✅ Required              | Cluster-level identity    | 0       |
+|                                       Operate | Active-active | ✅ Required              | Both regions serve users  | 0       |
+|                                      Tasklist | Active-active | ✅ Required              | Both regions serve users  | 0       |
+|         <p align="left">**Elasticsearch**</p> | Active-active | ✅ Required              | Data replicated to both   | 0       |
 
 :::important
 
-**All components in both regions must be fully operational at all times.** "Passive" refers only to user traffic routing, not system operation. Both regions actively participate in data processing and replication.
```
leiicamundi, line 39: @christinaausley the reason why we added a reminder of what "passive/active" is for each components is related to customer reports not understanding exactly the scope. Could we keep it? wdty?

leiicamundi (changes requested):
see my comments regarding multi region

docs/self-managed/concepts/multi-region/dual-region.md
```diff
@@ -273,12 +246,12 @@ If the primary region is lost:
...
+- **Data loss**: No data loss — all data is replicated via the Camunda Exporter and survives region loss (see [Active-active](#active-active)).
 
 #### Recovery steps for primary region loss
 
 1. **Temporary recovery:** Follow the [operational procedure](/self-managed/deployment/helm/operational-tasks/dual-region-ops.md#failover-phase) for temporary recovery to restore functionality and unblock the process automation engine (zeebe).
-2. **Traffic rerouting:** With v2 APIs (default in 8.9+), remove the failed region from serving traffic (for example, via DNS or load balancer health checks). With v1 APIs, redirect user traffic to the secondary region (now primary).
+2. **Traffic rerouting:** Remove the failed region from serving traffic (for example, via DNS or load balancer health checks).
 3. **Data and task management**:
    - Reassign uncompleted tasks lost from the previous primary region.
    - Recreate batch operations in Operate.
```
leiicamundi: The intro and architecture sections now describe a fully symmetric active-active setup ("no primary/secondary distinction at the UI layer", "No data loss"), but this **Region loss** section still uses the old primary/secondary active-passive framing and contradicts them:

- L247 "User traffic is unavailable" — in active-active the surviving region keeps serving traffic.
- L256-257 "Reassign uncompleted tasks lost / Recreate batch operations" — contradicts L249 "No data loss". If everything is replicated via the Camunda Exporter, there is nothing lost to reassign or recreate.

Suggested symmetric rewrite for these lines:

```suggestion
- **Zeebe halt**: Processing stops due to quorum loss (half of the brokers become unreachable).
- **User traffic**: The surviving region keeps serving user traffic. Remove the failed region from the traffic pool so requests are not routed to it.
- **No data loss**: All data is replicated to both regions via the Camunda Exporter and survives region loss (see [active-active](#active-active)).

#### Recovery steps

1. **Temporary recovery:** Follow the [operational procedure](/self-managed/deployment/helm/operational-tasks/dual-region-ops.md#failover-phase) for temporary recovery to restore functionality and unblock the process automation engine (Zeebe).
2. **Traffic rerouting:** Remove the failed region from serving traffic (for example, via DNS or load balancer health checks).
3. **Data recovery:** No manual recovery is required — all Operate and Tasklist data is replicated via the Camunda Exporter and remains available in the surviving region.
```

The rest of the section is outside this PR's diff, so it can't be part of this suggestion, but it needs the same treatment for consistency:

- **L234** "User interfaces become unavailable if primary region is lost" — the surviving region keeps serving the UI; drop the "if primary region is lost" clause.
- **L243 `### Primary region loss` / L260 `### Secondary region loss`** — under Zeebe quorum, losing either region is symmetric. Merge into a single region-agnostic subsection (for example, `### Region loss impact`) instead of a primary/secondary split.
- **L258** "create a new secondary region" → "recreate the lost region".
- **L272-273** note "Unlike primary region loss, no user-facing data is lost" — contradicts L249; remove or reword, since neither region loss causes data loss.

This ties into @leiicamundi's question in https://github.com/camunda/camunda-docs/pull/9257#discussion_r3504315648 — see my reply there on picking one direction consistently (and the 8.9 vs 8.10 boundary).

docs/self-managed/concepts/multi-region/dual-region.md
leiicamundi: Agree it's worth keeping the active/passive reminder **if** that nuance still applies in 8.9.

One thing to reconcile first: this PR targets `/docs` (8.9), but `authorizations.md` in this same PR states Tasklist V1 was removed in **8.10**. If V1 still exists in 8.9, the active-passive behavior for Operate/Tasklist on v1 APIs is still real in 8.9, and removing the per-component reminder here (and the v1 caveats across the dual-region page) would be premature for this version.

So it comes down to two consistent options:

1. **Keep** the active/passive reminder (your suggestion) — then the dual-region page must also keep the v1/active-passive caveats it currently removes.
2. **Fully commit** to active-active for 8.9 — then it must be consistent everywhere, including the **Region loss** section, which still uses primary/secondary framing (I left an inline suggestion on it).

Either way the page shouldn't be half active-active / half active-passive. Could you confirm the actual v1 availability in 8.9 so we can pick one direction?

christinaausley:
@tanjinyi 

> I believe the entire Zeebe API (REST) section here should also be removed as they would have been part of the V1 API.

Given that the API still works and there are tutorial + spec pages users may need during migration, keeping it under `Deprecated` with the warning seems like the safer documentation choice vs. full deletion -- WDYT? It is currently already under Deprecated in the sidebar with a warning banner, or do you want full removal?

docs/self-managed/concepts/multi-region/dual-region.md
christinaausley: Fair note on needing consistency — the docs target Next (post-8.10), not 8.9. Since V1 was removed in 8.10, the fully active-active framing in this PR is correct and consistent for that version, right?

The 8.9 versioned docs are not touched by this PR, so the active/passive per-component reminder remains intact there. Does that resolve the concern, or would you like me to also update the :::important block to clarify the version boundary?

docs/self-managed/concepts/multi-region/dual-region.md
christinaausley: Done 👍 applied your suggested rewrite. Merged the two subsections into a single ### Region loss impact, updated the bullet framing to be symmetric (Zeebe halt, user traffic survives in surviving region, no data loss), removed the manual task recovery steps, and dropped the contradictory note about secondary region loss. Also addressed the other consistency items you flagged outside the diff.

tanjinyi:
> @tanjinyi
> 
> > I believe the entire Zeebe API (REST) section here should also be removed as they would have been part of the V1 API.
> 
> Given that the API still works and there are tutorial + spec pages users may need during migration, keeping it under `Deprecated` with the warning seems like the safer documentation choice vs. full deletion -- WDYT? It is currently already under Deprecated in the sidebar with a warning banner, or do you want full removal?

Ah, whoops 😅 you're right on this part. I thought they were removed alongside the changes I made for V1, but I just checked against the latest version and it's still present. If that's the case, we should maintain the authentication page as well, but modify the reference such that it hits one of the existing APIs instead of the removed v1 process instance API. You can treat this comment as a pre-approval after restoring the authentication page and modifying the API example.

docs/self-managed/concepts/multi-region/dual-region.md
leiicamundi: Thanks @christinaausley

docs/self-managed/concepts/multi-region/dual-region.md
leiicamundi: @christinaausley we could put a little note head of the page to indicates this changed eventually (a reminder linked to the changenote)

leiicamundi (dismissed):

christinaausley:
Thank you both for the thorough review!

@tanjinyi:

- **Recreated** zeebe-api-rest-authentication.md with the curl example updated from the removed `/v1/process-instances` endpoint to `GET $ZEEBE_BASE_URL/v2/topology`
- **Restored** the `zeebe-api-rest-authentication` entry in sidebar-schema.js and sidebars.js
- **Updated** auth links in zeebe-api-rest-overview.md and tutorial.md to point back to the restored local auth page

@leiicamundi

- **Added** a `:::note` block at the top of dual-region.md explaining that this page reflects Next (post-8.10) fully active-active behavior following V1 API removal, with a link to the 8.10 announcements

Ready for final approval!

tanjinyi (approved):
Thanks! Looks good.

leiicamundi (approved):
lgtm! thanks @christinaausley

PR #9263, opened 2026-07-01
docs(self-managed): document EKS Pod Identity as a supported AWS auth method alongside IRSA [ready]

leiicamundi:
## Description

Related to https://camunda.slack.com/archives/C03UR0V2R2M/p1782810486794029

Clarifies that Camunda 8 on Amazon EKS supports **EKS Pod Identity** in addition to IRSA for authenticating workloads to AWS services (S3, Aurora PostgreSQL, OpenSearch). Pod Identity support was already effective (both mechanisms resolve credentials through the AWS SDK default credentials provider chain) but was not stated in the docs, which described IRSA only.

Applied to `amazon-eks/irsa.md` in **Next, 8.9, 8.8, and 8.7** (8.7 uses the legacy `setup/deploy` path). Content-only, no structural changes.

Changes per file:

- Added a page intro that presents IRSA and EKS Pod Identity as the two supported IAM-role mechanisms, and states that Camunda supports both because credential resolution is handled entirely by the AWS SDK.
- Added a `:::warning` that you must not configure IRSA and Pod Identity on the same Kubernetes service account: the SDK default chain resolves the IRSA web identity token before the Pod Identity credentials, so IRSA takes precedence and the Pod Identity association is ignored.
- Added an **Authenticate with EKS Pod Identity instead** subsection to the S3 document store guide, explaining what differs from the IRSA setup (keep `irsa.enabled: true` to skip static credential injection, create a Pod Identity association instead of the `role-arn` annotation, use the `pods.eks.amazonaws.com` trust principal, expect `AWS_CONTAINER_CREDENTIALS_FULL_URI`).

Camunda validates its deployments and reference architecture with IRSA; Pod Identity is not tested directly but is supported because it relies on the same AWS SDK behavior.

## PR Checklist


aabouzaid (dismissed):

zaycodes (commented):
Made a couple of small tweaks:

- Reworded "the same S3 permission policy shown above" to "…as the IRSA setup," so it still reads right if someone jumps straight to that section.
- Added a pointer on the Pod Identity bullet linking the [eksctl guide](https://docs.camunda.io/docs/next/self-managed/deployment/helm/cloud-providers/amazon/amazon-eks/eks-eksctl/#domain-deployment-requirements), since clusters created with it default to Pod Identity. That way the two pages connect.

One question:
- Since you're clarifying that `irsa.enabled` really just turns off static-credential injection (and Pod Identity needs it too), the param's description on the [document handling Helm config page](https://docs.camunda.io/docs/next/self-managed/concepts/document-handling/configuration/helm/) still explains the flag purely in terms of IRSA and doesn't mention Pod Identity. Should we tweak that for consistency?

Happy to update the param description if you're okay with it.

leiicamundi:
Good call, @zaycodes — agreed, and I've updated it for consistency rather than leaving the page describing the flag purely in IRSA terms.

I reworded the `global.documentStore.type.aws.irsa.enabled` `@param` on the document handling Helm config page across **Next, 8.9, 8.8, and 8.7**:

> When set to true, no static AWS credentials are injected into pods, so the AWS SDK resolves credentials through its default provider chain. Despite its name, this supports both IRSA and EKS Pod Identity. When false (default), credentials are required via existingSecret or accessKeyId/secretAccessKey configuration.

This mirrors the framing in the `irsa.md` Pod Identity section (the flag only stops static credential injection; Pod Identity relies on the same behavior). Pushed in de1c3a2.

One thing to flag for later: this `@param` block is a snapshot of the chart's `values.yaml`, so the same description also lives upstream in [`camunda/camunda-platform-helm`](https://github.com/camunda/camunda-platform-helm). To keep them from drifting, we should make the matching change there too — I'm happy to open that PR separately.

Thanks for the eksctl cross-link and the S3 rewording!

zaycodes (approved):
Thanks @leiicamundi! LGTM!

hisImminence (approved):

versioned_docs/version-8.8/self-managed/concepts/document-handling/configuration/helm.md
```diff
@@ -32,7 +32,7 @@ global:
                 enabled: false
                 ## @extra global.documentStore.type.aws.irsa IRSA (IAM Roles for Service Accounts) configuration for AWS authentication.
                 irsa:
-                  ## @param global.documentStore.type.aws.irsa.enabled When set to true, no AWS credentials are injected into pods, allowing IRSA to be used for authentication. When false (default), credentials are required via existingSecret or accessKeyId/secretAccessKey configuration.
+                  ## @param global.documentStore.type.aws.irsa.enabled When set to true, no static AWS credentials are injected into pods, so the AWS SDK resolves credentials through its default provider chain. Despite its name, this supports both IRSA and EKS Pod Identity. When false (default), credentials are required via existingSecret or accessKeyId/secretAccessKey configuration.
```
leiicamundi, line 35: @hisImminence FYI, an example of duplication of the source of truth of the chart parameters

leiicamundi:
Merging.

@hisImminence (cc @eamonnmoloney ) as you can see, we have another evidence of the duplication symptom of the doc here (exists in two repos, multiple files) 

Can you approve https://github.com/camunda/camunda-platform-helm/pull/6519?

Thanks

PR #9273, opened 2026-07-01
docs(agentic-orchestration): surface open-weight model story and cost trade-off

HanselIdes:
## What changed

- Rewrote the **Supported models** intro paragraph in the getting started guide to frame the frontier vs. open-weight model choice upfront
- Added a `:::tip Choosing the right model and model provider` callout explaining the per-token vs. fixed-infrastructure cost structure
- Added a new **Choosing by use case** subsection to the LLM recommendations page covering the capability/control/cost trade-off
- Updated the OpenAI-compatible provider description to name open-weight models (Qwen, Llama, Mistral) and Ollama explicitly

All changes backported to `versioned_docs/version-8.9` (current release) and `versioned_docs/version-8.8` (introduction version).

## Why

The CEO didn't know Camunda already supports open-weight models via the OpenAI Compatible connector or Ollama. This is a visibility gap, not a product gap. These changes make the open-weight story part of the default framing rather than a buried footnote.

## Notes for reviewer

- No new provider dropdown entries added (explicit non-goal per VP guidance)
- The tip callout sits between the intro paragraph and the use-case table — not stacked against the existing `:::important` hardware note
- The 8.8 version of `model-recommendations-agentic.md` had slightly different wording in the plain-text I/O paragraph; the new subsection was inserted at the same logical position


afgambin (approved):
@HanselIdes I did a review pass with minor text tweaks. lgtm now!

I see you’re off today, so I’m going ahead and merging it, since I don’t see any dependencies or blockers.

PR #9286, opened 2026-07-02
docs: add output mapping composite variable behavior change announcement for 8.7, 8.8, 8.9

remcowesterhoud:
## Description

Adds a new announcement entry — "Output mapping behavior change for composite variables" — to the 8.7, 8.8, and 8.9 release announcements, alongside the existing multi-instance sub-process regression entry.

Patches 8.7.28–8.7.32, 8.8.23–8.8.27, and 8.9.1–8.9.8 changed output mapping behavior for composite (object) variables from replace to merge. The fix patches (8.7.33, 8.8.28, 8.9.9) reverted this change. Neither the introduction nor the reversion was documented, leaving users who upgraded into or out of the affected range without an explanation for silent behavioral changes in their processes.

The new entry explains the merge-vs-replace distinction, identifies the affected version ranges, clarifies that replace is the intended long-term behavior, and gives users concrete action items for both the affected patches and the upgrade to the fix.

Relates to [camunda/camunda#56133](https://github.com/camunda/camunda/issues/56133).

## PR Checklist

### DIFF
--- a/docs/reference/announcements-release-notes/870/870-announcements.md
+++ b/docs/reference/announcements-release-notes/870/870-announcements.md
@@ -102,7 +102,7 @@ Before 8.7.28 and from 8.7.33+, assigning an object literal to a variable replac
 
 Example: task A sets `result.a = 1`, then task B sets `result = {b: 2}`:
 
-- _Replace_ (before 8.7.28 and 8.7.33+): `result = {"b": 2}` — task A's value is overwritten.
+- _Replace_ (before 8.7.28 and from 8.7.33+): `result = {"b": 2}` — task A's value is overwritten.
 - _Merge_ (8.7.28–8.7.32): `result = {"a": 1, "b": 2}` — task A's value is preserved.
 
 Replace is the intended long-term behavior. The merge behavior in the affected patches was an unintended regression.
--- a/docs/reference/announcements-release-notes/870/870-announcements.md
+++ b/docs/reference/announcements-release-notes/870/870-announcements.md
@@ -53,7 +53,7 @@ The following key changes were also released as part of an 8.7.x patch release.
 | Patch release                                                    | Type            | Key change                                                                                                       |
 | :--------------------------------------------------------------- | :-------------- | :--------------------------------------------------------------------------------------------------------------- |
 | [8.7.28](https://github.com/camunda/camunda/releases/tag/8.7.28) | Regression      | [Multi-instance sub-process output mapping variable scope regression](#multi-instance-output-mapping-regression) |
-| [8.7.28](https://github.com/camunda/camunda/releases/tag/8.7.28) | Regression      | [Output mapping behavior change for composite variables](#output-mapping-behavior-change)                        |
+| [8.7.28](https://github.com/camunda/camunda/releases/tag/8.7.28) | Regression      | [Output mapping behavior change for object variables](#output-mapping-behavior-change)                           |
 | [8.7.27](https://github.com/camunda/camunda/releases/tag/8.7.27) | Breaking change | [`getMessageKeys()` removed from the exporter record](#getmessagekeys-removed-from-the-exporter-record)          |
 
 ### `getMessageKeys()` removed from the exporter record {#getmessagekeys-removed-from-the-exporter-record}
@@ -90,11 +90,11 @@ Under these conditions:
 - Before the fix is available: ensure all variable names inside the multi-instance sub-process are unique and do not reuse names that exist on the parent scope.
 - After upgrading to the fixed patch: bugs #11789 and #35251 are reintroduced by the fix. If you previously had adaptations in place to work around these bugs and removed them, reapply those adaptations.
 
-### Output mapping behavior change for composite variables {#output-mapping-behavior-change}
+### Output mapping behavior change for object variables {#output-mapping-behavior-change}
 
 **Affected versions:** 8.7.28–8.7.32. Fixed in 8.7.33.
 
-Patches 8.7.28–8.7.32 changed how output mappings behave when writing to composite (object) variables. Upgrading to 8.7.33+ reverts this change, which can alter the behavior of your running processes.
+Patches 8.7.28–8.7.32 changed how output mappings behave when writing to object variables. Upgrading to 8.7.33+ reverts this change, which can alter the behavior of your running processes.
 
 Before 8.7.28 and from 8.7.33+, assigning an object literal to a variable replaces the variable entirely. In 8.7.28–8.7.32, the behavior changed to _merge_: existing keys in the variable are preserved and new keys are added.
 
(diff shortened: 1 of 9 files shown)


docs/reference/announcements-release-notes/870/870-announcements.md
```diff
@@ -89,6 +90,28 @@ Under these conditions:
...
+
+**Affected versions:** 8.7.28–8.7.32. Fixed in 8.7.33.
+
+Patches 8.7.28–8.7.32 changed how output mappings behave when writing to composite (object) variables. Upgrading to 8.7.33+ reverts this change, which can alter the behavior of your running processes.
+
+**What changed**
+
+Before 8.7.28 and from 8.7.33+, assigning an object literal to a variable replaces the variable entirely. In 8.7.28–8.7.32, the behavior changed to _merge_: existing keys in the variable are preserved and new keys are added.
+
+Example: task A sets `result.a = 1`, then task B sets `result = {b: 2}`:
```
eppdot: ❓ why do we use different syntax here?

docs/reference/announcements-release-notes/870/870-announcements.md
```diff
@@ -89,6 +90,28 @@ Under these conditions:
...
+
+- _Replace_ (before 8.7.28 and 8.7.33+): `result = {"b": 2}` — task A's value is overwritten.
```
eppdot: 🔧 we should use same phrases, e.g. like above "Before 8.7.28 and from 8.7.33+"

eppdot (dismissed):
I only commented on the 8.9 announcements, but feedback applies also to the lower versions.

You also added lots of empty lines 🤔 was that intended?

Pre-approving.

docs/reference/announcements-release-notes/870/870-announcements.md
remcowesterhoud: These release announcements are older and were never reshaped to the other format. I'm not sure if there's a specific reason for it, but I don't see changing this as part of the scope of this PR.

remcowesterhoud:
> You also added lots of empty lines 🤔 was that intended?

That must've been auto-formatting or something 🤔. I don't think it hurts having these. Unless the tech-writer reviewer thinks otherwise I'll leave it in.

docs/reference/announcements-release-notes/870/870-announcements.md
eppdot: I meant how the expression/variable is set `result.a = 1`, then task B sets `result = {b: 2}` 😅

remcowesterhoud: Oh haha, because this is what the bug is.

Before A would result in variable:
```json
{
  "result": {
    "a": 1
  }
}

```

The B would replace it to:

```json
{
  "result": {
    "b": 2
  }
}

```

That is the proper behavior. In these few versions we had a merge behavior and it'd result in:

```json
{
  "result": {
    "a": 1,
    "b": 2
  }
}

```

If you changed the syntax of B to `result.b = 2`, you wouldn't run into the bug as it'd be merged into `result.b`, which doesn't exist on the parent yet.

konstantinos-mylosis: 💭 I also had the same thought, maybe we could rephrase it:

``` md
The regression affects processes where one task sets a single key on an object variable, for example `result.a = 1` — and a later task overwrites the whole variable by assigning a new object to it, for example `result = {b: 2}`. The second assignment is supposed to replace result entirely, but in the affected versions it merged the values instead.

Example:
- Task A output mapping: `result.a = 1` → variable becomes `{"result": {"a": 1}}`
- Task B output mapping: `result = {b: 2}` → assigns an object literal to the parent variable

_Replace_ (before 8.7.28 and from 8.7.33+):
`result = {"b": 2}` — task B's literal replaces result entirely; task A's key is gone.

_Merge_ (8.7.28–8.7.32):
`result = {"a": 1, "b": 2}` — task A's key is unexpectedly preserved.

If task B had set `result.b = 2` instead of `result = {b: 2}`, this regression would not apply. Writing `result.b = 2` only touches the b key — it never overwrites the rest of result. The problem only occurs when assigning a whole object `result = {b: 2}`, which is supposed to replace result entirely.
```

docs/reference/announcements-release-notes/870/870-announcements.md
```diff
@@ -89,6 +90,28 @@ Under these conditions:
...
+
+- _Replace_ (before 8.7.28 and from 8.7.33+): `result = {"b": 2}` — task A's value is overwritten.
+- _Merge_ (8.7.28–8.7.32): `result = {"a": 1, "b": 2}` — task A's value is preserved.
+
+Replace is the intended long-term behavior. The merge behavior in the affected patches was an unintended regression.
+
+**Action**
+
+- **Running 8.7.28–8.7.32:** your processes use merge behavior. Review any process that writes to composite variables from multiple tasks.
+- **Upgrading to 8.7.33+:** output mappings revert to replace behavior. Test affected processes before upgrading to production.
```
konstantinos-mylosis: 💭 Would it make sense to expand the Action section a little?

``` suggestion
- **Running 8.7.28–8.7.32:** your processes use merge behavior. Identify any process where one task writes to a sub-key of a variable and a later task assigns an object literal to the same parent. If found, either switch the later task to path notation `result.b = 2` or include all required keys explicitly in its object literal.
- **Upgrading to 8.7.33+:** replace behavior is restored. The same processes identified above will behave differently after upgrading. If your process was relying on earlier tasks' values being kept, you need to fix it before upgrading: instead of assigning a whole object `result = {a: 1, b: 2}`, make sure it includes all the keys it needs explicitly — or write each key separately `result.a = 1, result.b = 2`
```

konstantinos-mylosis (commented):
Left a few comments. 

IMO, we could add a little more detail in a couple of places.

docs/reference/announcements-release-notes/870/870-announcements.md
eppdot: why not write it like this then:
- `result = {a: 1}`
- `result = {b: 2}`

does it require this little awkward syntax `result.a = 1`?

remcowesterhoud: > awkward syntax

I don't agree that it's awkward 😄. It's proper FEEL that users will know. But here it doesn't matter so I've changed it. I've also applied Mylo's suggestion.

docs/reference/announcements-release-notes/870/870-announcements.md
remcowesterhoud: Done 👍

remcowesterhoud:
@zaycodes Do you have an indication when this can be merged and released?

zaycodes:
@remcowesterhoud I will review today. Was waiting for the technical review to get done.

zaycodes (dismissed):
Small terminology question. Should this use "object variable" instead of "composite variable"?

I checked and "composite variable" isn't used anywhere else in the docs, whereas "object variable" is the term we use elsewhere. You've already hedged it as "composite (object) variables," so it's clear, I just want to make sure we're consistent, since the heading and the table link both say "composite."

If "object variable" works, I'd suggest switching to it throughout the entry (heading, table link text, and body). If "composite" is the term the engineering issue uses on purpose, happy to leave it as is

konstantinos-mylosis (approved):
Nice 🚀

PR #9287, opened 2026-07-02
doc: update default number of replicas and shards for exporters

mdii:
## Description

Updates Elasticsearch/OpenSearch and Camunda Exporter documentation to align with actual code defaults:

- **8.10 (next):** `number-of-replicas` default changed `0` → `1` for ES/OS Exporter indices (camunda/camunda#55366)
- **All versions:** `number-of-shards` for ES/OS exporters default corrected to `varies` with a note explaining that most indices use `1` shard, while `zeebe-record-job`, `zeebe-record-process-instance`, and `zeebe-record-user-task` templates use `3`
- **All versions:** `numberOfReplicas` for Camunda Exporter corrected to `1` (already the code default)
- **8.10 upgrade guide:** added note about the replica default change and action required for single-node clusters

> ⚠️ This PR should be merged after camunda/camunda#56419.

## PR Checklist


mdii:

docs/self-managed/components/orchestration-cluster/zeebe/exporters/camunda-exporter.md
```diff
@@ -72,7 +72,7 @@ Helm property path prefix for these options:
 | Option                | Description                                                                                                                                                                                 | Default |
 | --------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------- |
 | numberOfShards        | The number of [shards](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules.html#_static_index_settings) used for each created index.                              | 1       |
-| numberOfReplicas      | The number of shard [replicas](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules.html#dynamic-index-settings) used for created index.                           | 0       |
```
RomanJRW, line 75: this relates to the Camunda Exporter rather than the ES/OS ones, so I wouldn't expect to see a change in this PR. However, do we really only default this to 0 for the Camunda Exporter? Or is it higher and you are fixing this in the docs too?

docs/self-managed/components/orchestration-cluster/zeebe/exporters/elasticsearch-exporter.md
```diff
@@ -123,8 +123,8 @@ For Optimize-specific filtering, see [Optimize export filtering](/self-managed/c
...
+| number-of-replicas                  | The number of shard [replicas](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules.html#dynamic-index-settings) used for each new record index created.                                                                                                                                                                                                                                        | 1               |
+| number-of-shards                    | The number of [shards](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules.html#_static_index_settings) used for each new record index created.                                                                                                                                                                                                                                                | varies          |
```
RomanJRW, line 127: @camunda/tech-writers - the default value for this property depends on the index that it is applied to, and can be one of `1` or `3`. Can you suggest a way to capture this in the table or is `varies` with the note under the table sufficient?

docs/self-managed/upgrade/components/890-to-8100.md
```diff
@@ -270,3 +270,14 @@ DELETE /api/v1/clusters/{clusterId}
...
+## Elasticsearch/OpenSearch Exporter
+
+### Default replica count changed
+
+In 8.10, the default value of `number-of-replicas` for Elasticsearch/OpenSearch Exporter indices changed from `0` to `1`.
+
+This applies to new indices created after upgrading. Existing indices are not affected.
+
+- **Single-node clusters:** Set `number-of-replicas: 0` explicitly in your exporter configuration to avoid yellow cluster health. On a single node, replicas cannot be assigned and remain unassigned, which causes the cluster to report yellow health and may trigger monitoring alerts.
+- **Multi-node clusters:** No action required. The new default improves fault tolerance.
```
RomanJRW: Hmm no action required, but more disk space will be consumed. Do you think it's worth noting that?

RomanJRW (commented):
Generally ok I think, but some things to clarify. Also want to perspective of @camunda/tech-writers

docs/self-managed/components/orchestration-cluster/zeebe/exporters/camunda-exporter.md
mdii: The default number of replicas for Camunda Exporter has been 1 for some time but the docs haven't been updated. It is not related to ES/OS ones but I am fixing that in this PR too. Maybe I should update the PR title.

docs/self-managed/upgrade/components/890-to-8100.md
mdii: Good idea 👍

docs/self-managed/components/orchestration-cluster/zeebe/exporters/camunda-exporter.md
RomanJRW: Got it, in that case definitely makes sense to update here 👍

RomanJRW (approved):
Thanks for clarifying the Camunda Exporter change 👍 

Approved from my perspective, but I added @camunda/tech-writers as a reviewer in case they can suggest a better way to frame the ES/OS exporter defaults

christinaausley (commented):
A few questions after looking through the content 👍 None I necessarily view as blocking.

1. For `number-of-shards` default...instead of `varies` in the Default column, would `1 or 3` (with a note listing the exceptions) be more clear? or do you prefer `1` with the note carrying the 3-shard exceptions? And is the list (`zeebe-record-job`, `zeebe-record-process-instance`, `zeebe-record-user-task`) the complete set, or just examples?

The "1 or 3" behavior is applied to 8.8, 8.9, and 8.10. Have we confirmed the per-template split is accurate in 8.8/8.9 and isn't actually 8.10-only? (The replica `0 to 1` change is scoped to 8.10, so just want to be sure the shard note isn't retroactively describing newer behavior.)

The note says setting `number-of-shards` overrides the template defaults "including the three listed above." Should we add a caution that setting a low value (like `1`) de-shards the high-volume `job`/`process-instance`/`user-task` indices, which can affect performance at scale? It reads as neutral to me

2. For camunda exporter -- confirming the `numberOfReplicas` `0 to 1` change is a docs-only correction to the code default (not a behavior change in 8.10), and is correct to apply across 8.8, 8.9, and next?

mdii:
> A few questions after looking through the content 👍 None I necessarily view as blocking.
> 
> 1. For `number-of-shards` default...instead of `varies` in the Default column, would `1 or 3` (with a note listing the exceptions) be more clear? or do you prefer `1` with the note carrying the 3-shard exceptions? And is the list (`zeebe-record-job`, `zeebe-record-process-instance`, `zeebe-record-user-task`) the complete set, or just examples?
> 
> The "1 or 3" behavior is applied to 8.8, 8.9, and 8.10. Have we confirmed the per-template split is accurate in 8.8/8.9 and isn't actually 8.10-only? (The replica `0 to 1` change is scoped to 8.10, so just want to be sure the shard note isn't retroactively describing newer behavior.)

The per-template split is indeed the be existing behavior in versions >= 8.8.

> The note says setting `number-of-shards` overrides the template defaults "including the three listed above." Should we add a caution that setting a low value (like `1`) de-shards the high-volume `job`/`process-instance`/`user-task` indices, which can affect performance at scale? It reads as neutral to me
> 

We could but we already have a dedicated [page](https://docs.camunda.io/docs/self-managed/concepts/secondary-storage/managing-secondary-storage/) for managing the secondary storage.

> 2. For camunda exporter -- confirming the `numberOfReplicas` `0 to 1` change is a docs-only correction to the code default (not a behavior change in 8.10), and is correct to apply across 8.8, 8.9, and next?

Yes, it is a doc-only correction. The replica count has been changed from 0 to 1 since 8.8.

christinaausley (approved):

PR #9344, opened 2026-07-09
docs: document skipVariableWriteWithoutUserTasks exporter option and its migration limitation

EuroLew:
## Description

Documents the Camunda Exporter option `skipVariableWriteWithoutUserTasks` (added in camunda/camunda#47843, available since 8.8) on the Camunda Exporter configuration page, in `docs/` (next), `versioned_docs/version-8.9/`, and `versioned_docs/version-8.8/`.

It also documents a known limitation as a warning note: when the option is enabled and a process instance is later migrated to a process definition version that contains user tasks, previously skipped variable data is not retroactively backfilled — process-level variable filters in task search only match process instances created on a version that already contained user tasks. Per the decision in camunda/camunda#49963, we document this limitation instead of implementing a backfill, which proved too complex and brittle for a niche scenario.

Re-raises the previously closed #8396 with the limitation note added.

Relates to camunda/camunda#49963.

- [ ] This is on a specific schedule and the assignee will coordinate a release with the DevEx team. (create a new issue to track the work)
- [ ] This can go live whenever. (add `hold` label)

## PR Checklist

- [x] My changes are for an already released minor and are in `versioned_docs` directory.
- [x] My changes are for the next minor and are in `docs` directory (aka `/next/`).

### DIFF
--- a/docs/self-managed/components/orchestration-cluster/zeebe/exporters/camunda-exporter.md
+++ b/docs/self-managed/components/orchestration-cluster/zeebe/exporters/camunda-exporter.md
@@ -155,10 +155,7 @@ indices. The history can be configured as follows:
 
 <TabItem value="tasklist">
 
-Helm property path prefix for this option:
-`camunda.data.secondary-storage.{elasticsearch|opensearch}.`
-
-Tasklist-specific export options:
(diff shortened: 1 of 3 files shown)


alexronquillo (approved):
After a minor language change, it LGTM 🚀

PR #9384, opened 2026-07-14
docs(message-events): document boundary event correlationKey scope change for 8.10

### LINKED ISSUE #4799: The variable scope for a boundary event is the activity that it is attached to
**Describe the bug**
If a boundary event is attached to an activity then the variables of the boundary event expression (i.e. timer expression) are accessed from the scope of the activity instead of its flow scope activity (e.g. the workflow instance scope).  

If the activity has an input variable mapping that overlaps a variable that is used by a boundary event expression then it accesses the value of the local variable of the activity. This may lead to an incorrect event subscription or an incident.

This happens for all boundary event expressions except the message correlation-key expressions. 

**To Reproduce**
* deploy a workflow with
  * a service task with the input mapping
    * source: `="foo"` target: `x`
  * a timer boundary attached to the service task with 
    * a duration expression `=x`
* create an instance of this workflow with variable `{"x": "PT1M"}`
* verify that an incident is created  (`Expected result of the expression 'x' to be parsed to a duration, but was 'foo'`)

Try the same for message correlation key, and notice that it behaves differently.

**Expected behavior**

We want the same behavior for message correlation key expression evaluation when subscribing an activity to a message boundary event. Evaluate the expression in the context of the activity that is subscribing to the event.

**Root cause**

We explicitly use the flow scope instead of the element instance, but we shouldn't.

https://github.com/camunda/camunda/blob/89fd3db77ece80b872cc0631f82f425fb4a8f0fc/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/common/CatchEventBehavior.java#L290-L293

Notice how the scope key for the message name expression is always just element instance key:

https://github.com/camunda/camunda/blob/89fd3db77ece80b872cc0631f82f425fb4a8f0fc/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/common/CatchEventBehavior.java#L270

**Log/Stacktrace**

No.

**Environment:**
 - OS: [e.g. Linux]
 - Zeebe Version: 0.23.0, 0.24.0-SNAPSHOT
 - Configuration: [e.g. exporters etc.]

### LINKED ISSUE #9380: Document message boundary event correlation-key scope fix (8.10) + evaluateBoundaryEventCorrelationKeyInActivityScope flag
camunda/camunda#56196 fixes a long-standing inconsistency in how the `correlationKey` expression of a **message boundary event** is evaluated, closing camunda/camunda#4799 (open since 2020). This lands in 8.10 (unreleased at time of writing, no backport), and needs documenting across both the current docs and the affected versioned docs.

## Behavior change (8.10+)

**Before (existing behavior in 8.7, 8.8, 8.9, and all prior versions):**
For a message boundary event, the `correlationKey` FEEL expression is evaluated against the **flow scope** (the parent/enclosing scope) rather than the activity the boundary event is attached to. This is inconsistent with timer duration, message name, and signal name expressions on boundary events, which are all evaluated in the activity's own (element instance) scope. Consequences:
- The `correlationKey` expression cannot see variables introduced via input mappings on the activity the boundary event is attached to.
- If the `correlationKey` expression fails to evaluate, the resulting incident is raised on the flow scope (not the activity) — so a user must resolve it by setting a variable on the parent scope rather than directly on the activity, unlike other incidents on that activity.

**After (new default behavior from 8.10 onward):**
Correlation-key evaluation for message boundary events now uses the activity's own scope (element instance), consistent with timer/message-name/signal-name expressions. Incidents from a failing `correlationKey` expression now report the activity's own scope key rather than the parent/process instance key.

## New configuration flag

- **Name:** `zeebe.broker.experimental.features.evaluateBoundaryEventCorrelationKeyInActivityScope`
- **Env var:** `ZEEBE_BROKER_EXPERIMENTAL_FEATURES_EVALUATEBOUNDARYEVENTCORRELATIONKEYINACTIVITYSCOPE`
- **Default:** `true` (new/fixed behavior is active by default)
- **When set to `false`:** restores the legacy (pre-8.10) flow-scope behavior. This is a kill-switch/rollback mechanism, not a gradual opt-in.
- **Requires a broker restart** — no hot-reload.
- **Breaking-change note:** low-risk but real — affects processes where a boundary event's `correlationKey` expression references a variable shadowed by an input mapping on the activity, or local to a sub-process/call activity the boundary is attached to.
- Should be added to the config reference table rendered at `/self-managed/components/orchestration-cluster/zeebe/configuration/broker.md` (source data: `docs/self-managed/react-components/_config-table-data.js`).

## Task checklist

**8.10 (current/next docs — `docs/` tree, unversioned):**
- [ ] Update `docs/components/modeler/bpmn/message-events/message-events.md` (Message boundary events section / correlationKey description) to describe correlation-key expressions now evaluating in the activity's own scope, consistent with timer/signal expressions, and mention the kill-switch flag.
- [ ] Add the new flag to `docs/self-managed/react-components/_config-table-data.js` (rendered into `docs/self-managed/components/orchestration-cluster/zeebe/configuration/broker.md`).
- [ ] Update `docs/components/modeler/bpmn/events.md` (general boundary events section) if a cross-reference or clarification is warranted there.

**8.7–8.9 (versioned docs — document existing behavior as-is, no fix):**
- [ ] `versioned_docs/version-8.9/.../message-events.md` — accurately describe the current flow-scope correlation-key evaluation behavior (as it exists today, not a fix).
- [ ] `versioned_docs/version-8.8/.../message-events.md` — same.
- [ ] `versioned_docs/version-8.7/.../message-events.md` — same.

## Open question for engineering

As of the source PR, the new flag has **no unified `camunda.*` configuration name** — only the legacy `zeebe.broker.experimental.features.*` path exists. Please confirm with engineering whether a unified name is planned before finalizing the config reference table entry.

## References

- Source PR: camunda/camunda#56196
- Bug: camunda/camunda#4799

koevskinikola:
## Description

Documents the message boundary event `correlationKey` scope fix from [camunda/camunda#56196](https://github.com/camunda/camunda/pull/56196) (closes [camunda/camunda#4799](https://github.com/camunda/camunda/issues/4799)), landing in 8.10 and backported to 8.9 (disabled by default there).

- **`docs/` (8.10, unreleased):** describes that a message boundary event's `correlationKey` expression now evaluates in the activity's own (element instance) scope, consistent with the timer duration, message name, and signal name expressions, and documents the `evaluateBoundaryEventCorrelationKeyInActivityScope` kill-switch flag (default `true` in 8.10+, requires a broker restart to change).
- **`versioned_docs/version-8.9`:** the fix was backported via [camunda/camunda#57700](https://github.com/camunda/camunda/pull/57700), shipping **disabled by default**. Documents the default flow-scope behavior plus the opt-in flag to get the activity-scope behavior early.
- **`versioned_docs/version-8.8`, `version-8.7`:** documents the existing (pre-fix) flow-scope evaluation behavior as-is — a clarification, not a fix, since these versions aren't getting the change.
- **Config reference:** adds `camunda.processing.evaluate-boundary-event-correlation-key-in-activity-scope` to the unified-configuration mapping table and the processing properties reference (app properties + env var tabs), for both `docs/` (default `true`) and `versioned_docs/version-8.9` (default `false`), mapped to the legacy `zeebe.broker.experimental.features.evaluateBoundaryEventCorrelationKeyInActivityScope` path.

**Note:** the source PR only added the legacy config path, with no unified `camunda.*` name, on either `main` or `stable/8.9`. I filed [camunda/camunda#57698](https://github.com/camunda/camunda/issues/57698) to wire up the `camunda.processing.evaluate-boundary-event-correlation-key-in-activity-scope` name in code on both branches — this should land before this docs PR merges so the property actually works as documented.

Closes #9380

## PR Checklist


christinaausley (dismissed):
This looks good!

As far as camunda/camunda#57698, it looks like we're treating `camunda.processing.evaluate-boundary-event-correlation-key-in-activity-scope` as the primary property name, but in the PR description we say that name isn't wired in code yet -- only the legacy `zeebe.broker.experimental.*` path actually works.

Are we planning to ship out 57698 before this docs PR merges, so anyone who follows the documented `camunda.processing.*` name won't set a property that's silently ignored, and the behavior won't change? Alternatively, the note could flip the emphasis: lead with the `zeebe.broker.*` path as the confirmed one and reference `camunda.processing.*` as the unified name once #57698 lands?

You're the expert, so happy with whatever works best.

koevskinikola:
I raised a PR for the missing code here: https://github.com/camunda/camunda/pull/59638

danielkelemen:
Core PRs:
Main: https://github.com/camunda/camunda/pull/59638 merged
8.9: https://github.com/camunda/camunda/pull/59650 queued to merge

> Are we planning to ship out 57698 before this docs PR merges, so anyone who follows the documented camunda.processing.* name won't set a property that's silently ignored, and the behavior won't change?

@christinaausley, yeah, that's a good point. It's probably not a bad idea to not publish docs before the code changes of `camunda.processing.evaluate-boundary-event-correlation-key-in-activity-scope` are live.

That would mean we need to hold back this PR until next patch.

I might split this up to two parts: docs change & new config. The docs part can be merged and the other one wait.

danielkelemen:
Forgot this a bit due to other priorities. I'll just resolve the conflict and merge the PR.

christinaausley (approved):

PR #9453, opened 2026-07-22
docs(hub): update postman collection via workflow

alexronquillo:
## Description

Introduce a new workflow to update the Hub API postman collection.

Closes: https://github.com/camunda/camunda-hub/issues/25129

## PR Checklist


afgambin:
@alexronquillo I was reviewing this with Claude, and I think these two items are worth double-checking:

1. "Missing npm dependencies — `sync-hub-api.js` requires `@apidevtools/json-schema-ref-parser` and `openapi-to-postmanv2` (`sync-hub-api.js:5–6`), but neither package is declared in `package.json`. The workflow runs `npm ci` and then `node ./hacks/postman/sync-hub-api.js` (`update-postman.yaml:48–59`), so the job might fail with a `Cannot find module` error unless the packages are added to `package.json` or installed by the workflow." However, I understand that you tested it successfully. Could these dependencies be provided elsewhere?

2. "Silent failures during creation or updates — `getAllCollections` checks `res.ok` and throws an error if the request fails (`sync-hub-api.js:31–37`), but `sendCollectionRequest`, which is used by both `createCollection` and `updateCollection`, does not check `res.ok` before returning `res.json()` (`sync-hub-api.js:230–236`). If a POST or PUT request to Postman fails—for example, because of invalid authentication, rate limiting, or a server error—the script logs the JSON error response through `console.log(result)` but exits successfully. As a result, the workflow and any downstream failure notification could report success even though the collection was not synchronized."

Let me know if these points are irrelevant / not applicable, and I’ll approve the PR. I understand that you tested it with a dry run and it worked, so I’m unsure whether these suggestions apply. Great work!

alexronquillo:
@afgambin thanks for the feedback! Yes, it's better to install these deps independently (they're already sub-dependencies of the docusaurus OpenAPI package) just to protect against the unlikely situation where the parent package drops those dependencies silently in the future and makes us have to fix it at that time. For the error handling, I agree it's better to raise the error 👍 This should be ready for another pass.

alexronquillo:
@claude[agent] review

afgambin (approved):
approving!! :)

mesellings:
Great work here @alexronquillo  - make sure you add this to your kudos board 👏

PR #9492, opened 2026-07-28
docs(management-identity): document userIdClaim, usernameClaim, groupsClaim

p-wunderlich:
## Summary
Documents three Camunda Identity SDK properties in the Management Identity configuration variables reference, for the current (unreleased) docs and 8.9:

- `CAMUNDA_IDENTITY_USERIDCLAIM` / `camunda.identity.user-id-claim` — new property added in camunda/identity#4865, closing camunda/identity#4864. Lets customers migrating from Keycloak directly to Azure AD/Entra ID configure the `oid` claim as the user ID, since `sub` changes across that migration while `oid` stays stable.
- `CAMUNDA_IDENTITY_USERNAMECLAIM` / `camunda.identity.username-claim` — pre-existing property (camunda/identity#3375) that `userIdClaim` was explicitly modeled on, previously undocumented in this table.
- `CAMUNDA_IDENTITY_GROUPSCLAIM` / `camunda.identity.groups-claim` — pre-existing property (camunda/identity#3386) also previously undocumented in this table.

## Scope notes
- `userIdClaim` is intentionally **not** added to `version-8.8`/`version-8.7` docs — the backport (camunda/identity#4873) only went to `v8.9`.
- `usernameClaim`/`groupsClaim` do exist in those older versions too, but backfilling their docs there is left as a possible follow-up rather than folded into this PR.

## Verification
- `npx prettier --check` passes on both edited files.
- The two edited files remain byte-identical to each other (as they were before this change).
- No other files touched.


versioned_docs/version-8.9/self-managed/components/management-identity/miscellaneous/configuration-variables.md
```diff
@@ -32,20 +32,23 @@ The core configuration variables are as follows:
...
+| `CAMUNDA_IDENTITY_GROUPSCLAIM`       | `camunda.identity.groups-claim`         | Defines the claim used to read a user's groups directly from the token. If unset, groups are instead resolved via a call to the Identity groups API.                                                                                                                                                                                | -                                                    |
+| `CAMUNDA_IDENTITY_USERIDCLAIM`       | `camunda.identity.user-id-claim`        | Defines the claim used to identify a user. If unset, or if the configured claim is missing from both the token and the userinfo response, falls back to the token's `sub` claim. Useful when migrating between identity providers, for example using Azure AD's `oid` claim to preserve user identities carried over from Keycloak. | -                                                    |
```
megglos: 🔧 let's add a note that this is only available from `8.9.7` onwards

megglos (dismissed):
Only non-blocking comments.

One note on adding the info from when the new prop is available.

> usernameClaim/groupsClaim do exist in those older versions too, but backfilling their docs there is left as a possible follow-up rather than folded into this PR.

I'd still consider it worth to do that here while we are there. Let's assign copilot.

versioned_docs/version-8.9/self-managed/components/management-identity/miscellaneous/configuration-variables.md
p-wunderlich: Added — noted `userIdClaim` is available from `8.9.7` onwards in this version's row. Left the backfill of `usernameClaim`/`groupsClaim` to 8.8/8.7 as a separate follow-up per your comment.

megglos (dismissed):

p-wunderlich:
Backfilled `usernameClaim`/`groupsClaim` to 8.8 and 8.7 as suggested — verified both properties have existed since tag `8.7.2` in camunda/identity, so no availability caveat is needed for either version. `userIdClaim` itself stays out of scope for 8.8/8.7 since it genuinely doesn't exist there (only from 8.9.7 onwards).

megglos (approved):

PR #9517, opened 2026-07-30
Tenant Issue 9: Connectors Runtime Multi-Tenant Support

### LINKED ISSUE #8895: Tenant Issue 9: Connectors Runtime Multi-Tenant Support
**Review Stakeholders:** @johnBgood @sbuettner @deepthidevaki

- [#3640 Physical Tenant - Multi-Tenant Connector Runtime and Java Client](https://github.com/camunda/product-hub/issues/3640)
- [#3040 Centralized Secret Resolution via Zeebe](https://github.com/camunda/product-hub/issues/3040)

**Deliverables**
- Connectors runtime multi-tenant overview page covering:
  - One runtime instance, multiple Physical Tenants
  - No separate deployment per tenant needed
  - Job worker registration per tenant
  - Authorization and secret isolation model
- Configuration guide covering:
  - Runtime registration with Orchestration Cluster
  - Tenant specification in runtime config (all or subset)
  - Secret resolution per tenant (centralized secret management)
  - Authentication headers/credentials for multi-tenant requests
- Outbound connectors documentation covering:
  - How connectors know which tenant triggered job
  - Tenant context propagation to external systems
  - Per-tenant secret access (API keys isolated per tenant)
  - Example: HTTP connector calling different endpoints per tenant
- Inbound connectors documentation covering:
  - Inbound connector path/routing strategy
  - How inbound webhooks routed to correct tenant
  - Known limitations for inbound connectors
  - Security implications of shared inbound endpoint
- Operational considerations covering:
  - Monitoring job execution per tenant
  - Secret rotation across tenants
  - Scaling strategy (one runtime vs multiple)
  - Failure handling and retry semantics per tenant

**Context We Have**
- One Connectors runtime can register job workers for multiple Physical Tenants
- Per-tenant secret isolation via centralized secret management (epic #3040)
- Outbound: per-tenant secret access, context-aware routing
- Inbound: routing/path strategy still being finalized
- No separate runtime deployment per tenant required
- Secret isolation dependency on epic #3040

**Engineering Questions**
1. How does Connectors runtime know which Physical Tenant a job came from? (Header injection, context variable, explicit configuration?)
2. For inbound connectors (webhooks), what is routing strategy to determine which tenant a request is for? (URL path, header, or other?)
3. If secret updated in centralized secret manager, does runtime refresh immediately, on next job, or on schedule?
4. For inbound connectors, what is recommended security model to prevent webhook from one tenant accidentally triggering job in another?

christinaausley:
## Description

Closes https://github.com/camunda/camunda-docs/issues/8895.

- Adds `connectors-runtime.md` to the Physical Tenants doc set covering multi-tenant Connectors runtime support
- Adds sidebar entry under Physical Tenants

**What's covered:**
- Architecture diagram: one runtime instance, fan-out to per-PT job workers and inbound connectors
- `camunda.clients.*` multi-client configuration with YAML example
- Outbound connector tenant context propagation and per-tenant secret access
- Inbound webhook path routing (`APPEND_ENGINE_AND_TENANT_TO_WEBHOOK_PATH`) and security implications
- Authorization, scaling, monitoring limitations, secret rotation, and failure handling

**Pending (TODOs in file, not blocking merge):**
- Inbound flag configuration mechanism + webhook auth model — awaiting Nic Puppa
- Secret isolation connector-specific behavior — awaiting Berkay Canbolat (#3040)
- MS Teams App Integration PT support — awaiting Igor Dultsev

## PR Checklist

### DIFF
--- a/docs/self-managed/concepts/physical-tenants/connectors-runtime.md
+++ b/docs/self-managed/concepts/physical-tenants/connectors-runtime.md
@@ -1,0 +1,171 @@
+---
+id: connectors-runtime
+title: "Connectors runtime: Physical Tenant support"
+sidebar_label: "Connectors runtime"
+description: "Configure one Connectors runtime instance to serve multiple Physical Tenants, with per-tenant job workers, secret isolation, and inbound webhook routing."
+---
+
+One Connectors runtime instance can register job workers and serve inbound connectors for multiple Physical Tenants. No separate runtime deployment per tenant is required.
+
+:::note Related pages
+
+- **[Configuration reference](/self-managed/concepts/physical-tenants/configuration-reference.md)** — General tenant configuration
+- **[Authorization model](/self-managed/concepts/physical-tenants/authorization-model.md)** — Roles and permissions per tenant
+- **[API routing](/self-managed/concepts/physical-tenants/api-routing.md)** — How requests route to Physical Tenants
+
+:::
+
+## Architecture
+
+```mermaid
+graph TD
+    runtime["Connectors Runtime\n(single instance)"]
+
+    subgraph ptA["Physical Tenant A"]
+        workerA["Job Worker\n(type, tenanta)"]
+        inboundA["Inbound Connector\n/inbound/tenanta/..."]
+    end
+
+    subgraph ptB["Physical Tenant B"]
+        workerB["Job Worker\n(type, tenantb)"]
+        inboundB["Inbound Connector\n/inbound/tenantb/..."]
+    end
+
+    subgraph ptDefault["Physical Tenant default"]
+        workerDefault["Job Worker\n(type, default)"]
+        inboundDefault["Inbound Connector\n/inbound/default/..."]
+    end
+
+    runtime --> workerA
+    runtime --> workerB
+    runtime --> workerDefault
+    runtime --> inboundA
+    runtime --> inboundB
+    runtime --> inboundDefault
+```
+
+## How the runtime identifies the Physical Tenant
+
+The activated job record carries the `physicalTenantId` propagated through the broker request. The runtime uses this value to determine which Physical Tenant a job belongs to.
+
+- `@JobWorker`-annotated methods fan out to **all configured clients** automatically — there is no `client` attribute to bind a worker to one tenant.
+- Workers are keyed by `(client, type)`, so each Physical Tenant gets its own registered worker per job type.
+- Fan-out is over **statically configured** `camunda.clients.*` entries only. A Physical Tenant only gets a job worker if it is explicitly configured as a client.
+- gRPC requests use the `Camunda-Physical-Tenant` header to target the correct tenant.
+
+## Configuration
+
+Use the `camunda.clients.*` multi-client configuration to connect the runtime to multiple Physical Tenants. The shared `camunda.client.*` block sets base connection details inherited by all entries; per-client entries override only what differs (typically `physical-tenant-id` and `auth.*`).
+
+```yaml
+camunda:
+  client:
+    # Shared base — inherited by all client entries
+    grpc-address: https://your-cluster.example.com:26500
+    rest-address: https://your-cluster.example.com
+  clients:
+    default:
+      physical-tenant-id: default
+      auth:
+        client-id: connector-default
+        client-secret: ${SECRET_DEFAULT}
+      primary: true # resolves @Autowired CamundaClient and the default @JobWorker target
+    tenanta:
+      physical-tenant-id: tenanta
+      auth:
+        client-id: connector-tenanta
+        client-secret: ${SECRET_TENANTA}
+    tenantb:
+      physical-tenant-id: tenantb
+      auth:
+        client-id: connector-tenantb
+        client-secret: ${SECRET_TENANTB}
+```
+
+- The client name (`tenanta`) is a free-form label and does not have to match the `physical-tenant-id`.
+- `physical-tenant-id` must be lowercase alphanumeric, maximum 64 characters.
+- Mark one entry `primary: true` when configuring multiple clients. With a single client, it is the primary implicitly.
+- Existing single-client deployments using `camunda.client.*` are unaffected — `camunda.client.*` transparently maps to `camunda.clients.default.*` at startup.
+
+## Outbound connectors
+
+### Tenant context propagation
+
+The `physicalTenantId` is available in the activated job context. Use it to route requests to tenant-specific endpoints or apply tenant-specific configuration.
+
+**Example — HTTP connector with per-tenant routing:**
+
+Set the HTTP connector URL as a FEEL expression referencing the `physicalTenantId` process variable:
+
+```
+= "https://api.example.com/" + physicalTenantId + "/orders"
+```
+
+Tenant A's jobs call `.../tenanta/orders`; tenant B's jobs call `.../tenantb/orders`. The URL resolves dynamically — no per-tenant connector configuration needed.
+
+### Per-tenant secret access
+
+Secrets are scoped per Physical Tenant. The `SecretContext` includes the `physicalTenantId`, so the gateway resolves secrets against the correct tenant's store. One tenant's secrets cannot be accessed from another tenant's job context.
+
+Secret references:
+
+- **Canonical:** `camunda.secrets.MY_SECRET` — FEEL expression form
+- **Legacy:** `{{secrets.MY_SECRET}}` — string template, supported indefinitely
+
+Resolved values are never written to the variable store, Operate, Tasklist, logs, or the command log. Only the reference name appears in exports — the reference is a pointer, not the value.
+
+Secret resolution requires `SECRETS:REVEAL` on the `SECRET` resource type, scoped per Physical Tenant.
+
+<!-- TODO: Confirm connector-specific behavior with Berkay Canbolat before publishing: does the runtime call POST /v2/secrets/resolve per PT, and is transparent resolution available in the connector context in alpha5? (#3040) -->
+
+## Inbound connectors
+
+Each Physical Tenant gets its own `ImportSchedulers` instance with a dedicated `CamundaClient`, so inbound connector events and state are isolated per tenant.
+
+### Webhook path routing
+
+Enable the `APPEND_ENGINE_AND_TENANT_TO_WEBHOOK_PATH` flag to activate namespaced paths:
+
+```
+/inbound/<physicalTenantId>/<tenantId>/<path>
+```
+
+Without this flag, the first registered inbound connector matching a path claims the request regardless of Physical Tenant — enable namespaced paths in any multi-tenant deployment.
+
+<!-- TODO: Confirm with Nic Puppa: exact mechanism to enable this flag (env var, Spring property, or Helm value), whether per-tenant webhook authentication is enforced at the routing layer, and whether the inbound routing strategy is final for alpha5. -->
+
+### Known limitations
+
+- Each inbound request targets exactly one Physical Tenant. Routing a single webhook event to multiple tenants is not supported.
+- Cross-tenant inbound routing is not supported.
+
+## Authorization
+
+| Permission                      | Resource            | Required for                   |
+| ------------------------------- | ------------------- | ------------------------------ |
+| `SECRETS:REVEAL`                | `SECRET`            | Resolving secret references    |
+| Standard job worker permissions | Process definitions | Activating and completing jobs |
+
+Permissions are scoped per Physical Tenant. Grant `SECRETS:REVEAL:*` (wildcard) to the Connectors runtime service account to preserve the pre-8.10 implicit-trust scope while keeping authorization explicit.
+
+## Operational considerations
+
+### Scaling
+
+One runtime instance per cluster is standard. Deploy multiple instances only when you need runtime-layer resource isolation (for example, performance SLA separation) — Physical Tenants already isolate data structurally.
+
+### Monitoring
+
+Per-tenant job execution metrics are not available in 8.10. Micrometer metrics are tagged by job type and action only — no `physicalTenantId` or client dimension. Per-tenant filtering is a planned follow-up.
+
+### Secret rotation
+
+When a secret is rotated, the gateway cache expires within the configured TTL (default 20 seconds). No restart required. New job activations after the TTL receive the updated value.
+
+### Failure handling
+
+A failure in one tenant's job workers does not affect workers registered for other tenants.
+
+<!-- TODO (Igor Dultsev): MS Teams App Integration has a Physical Tenant support PR with investigation of multiple Keycloak per tenant. Request current status and add a section once finalized. -->
+
+<p class="link-arrow">[Physical Tenant isolation model](/self-managed/concepts/physical-tenants/index.md)</p>
--- a/docs/self-managed/concepts/physical-tenants/connectors-runtime.md
+++ b/docs/self-managed/concepts/physical-tenants/connectors-runtime.md
@@ -114,9 +114,11 @@ Secret references:
 
 Resolved values are never written to the variable store, Operate, Tasklist, logs, or the command log. Only the reference name appears in exports — the reference is a pointer, not the value.
 
-Secret resolution requires `SECRETS:REVEAL` on the `SECRET` resource type, scoped per Physical Tenant.
+Secret resolution requires `SECRETS:REVEAL` on the `SECRET` resource type, scoped per Physical Tenant. The runtime calls `POST /v2/secrets/resolve` scoped to the correct Physical Tenant for each job.
 
-<!-- TODO: Confirm connector-specific behavior with Berkay Canbolat before publishing: does the runtime call POST /v2/secrets/resolve per PT, and is transparent resolution available in the connector context in alpha5? (#3040) -->
+:::note
+Transparent secret resolution (automatic resolution of `{{secrets.MY_SECRET}}` references in the connector context) is subject to Connector team capacity for 8.10. If unavailable at release, secret references must be resolved explicitly via the API. Check the 8.10 release notes for final availability.
+:::
 
 ## Inbound connectors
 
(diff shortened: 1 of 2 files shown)


deepthidevaki:
@christinaausley, @nicpuppa will review when he is back.

docs/self-managed/concepts/physical-tenants/connectors-runtime.md
```diff
@@ -0,0 +1,181 @@
...
+```
+
+## How the runtime identifies the Physical Tenant
+
+The activated job record carries the `physicalTenantId` propagated through the broker request. The runtime uses this value to determine which Physical Tenant a job belongs to.
+
+- `@JobWorker`-annotated methods fan out to **all configured clients** automatically — there is no `client` attribute to bind a worker to one tenant.
+- Workers are keyed by `(client, type)`, so each Physical Tenant gets its own registered worker per job type.
+- Fan-out is over **statically configured** `camunda.clients.*` entries only. A Physical Tenant only gets a job worker if it is explicitly configured as a client.
+- gRPC requests use the `Camunda-Physical-Tenant` header to target the correct tenant.
```
nicpuppa: **Two corrections in this list.**

The `@JobWorker` bullet describes Spring SDK behavior, not the Connectors runtime. There is no `@JobWorker` annotation anywhere in the connectors codebase — outbound workers are registered programmatically by `OutboundConnectorManager`. Readers of this page will go looking for an annotation that isn't part of the picture.

"Workers are keyed by `(client, type)`" — they're keyed by `(physicalTenantId, connectorType)` (`OutboundConnectorManager.java:74`, `InstanceCacheKey`). This matters because when `physical-tenant-id` is unset the fallback differs by code path: `OutboundConnectorManager` falls back to the *client name* (`:162-163`) while `OutboundConnectorRuntimeConfiguration` falls back to `"default"` (`:490-491`). So the practical guidance for readers is: always set `physical-tenant-id` explicitly.

I'm not sure here we should mention how gRPC works, in my opinion is not related to connectors but more a internal implementation

Suggestion:

```markdown
The activated job record carries the `physicalTenantId`, propagated through the broker request. The
runtime uses this value to determine which Physical Tenant a job belongs to.

- The runtime registers one job worker per configured client per connector type, so each Physical
  Tenant gets its own worker for each job type.
- Registration covers **statically configured** `camunda.clients.*` entries only. A Physical Tenant
  only gets a job worker if it is explicitly configured as a client.
- Always set `physical-tenant-id` explicitly on each client entry. If it is omitted, the runtime
  falls back to a derived value and per-tenant attribution in metrics and the connector instance
  listing becomes unreliable.
```

docs/self-managed/concepts/physical-tenants/connectors-runtime.md
```diff
@@ -0,0 +1,181 @@
...
+      physical-tenant-id: tenantb
+      auth:
+        client-id: connector-tenantb
+        client-secret: ${SECRET_TENANTB}
+```
+
+- The client name (`tenanta`) is a free-form label and does not have to match the `physical-tenant-id`.
+- `physical-tenant-id` must be lowercase alphanumeric, maximum 64 characters.
+- Mark one entry `primary: true` when configuring multiple clients. With a single client, it is the primary implicitly.
+- Existing single-client deployments using `camunda.client.*` are unaffected — `camunda.client.*` transparently maps to `camunda.clients.default.*` at startup.
```
nicpuppa, line 90: The example omits `mode`, which every multi-client wiring test sets per entry (`camunda.clients.engine-a.mode=self-managed` — see `MultiClientPhysicalTenantWiringTest`, `DocumentPhysicalTenantIsolationWiringTest`). Readers copying this block verbatim may not get a working config.

Also, the inline comment on line 72 references `@JobWorker` — see my comment above; that annotation isn't used by this runtime, so "the default `@JobWorker` target" won't mean anything to a reader of this page.

docs/self-managed/concepts/physical-tenants/connectors-runtime.md
```diff
@@ -0,0 +1,181 @@
...
+
+**Example — HTTP connector with per-tenant routing:**
+
+Set the HTTP connector URL as a FEEL expression referencing the `physicalTenantId` process variable:
+
+```
+= "https://api.example.com/" + physicalTenantId + "/orders"
+```
+
+Tenant A's jobs call `.../tenanta/orders`; tenant B's jobs call `.../tenantb/orders`. The URL resolves dynamically — no per-tenant connector configuration needed.
```
nicpuppa: **This example won't evaluate.** `physicalTenantId` is not a process variable and is not exposed on `JobContext`. It lives on the activated-job record and is threaded only into `SecretContext` (`JobHandlerContext.java:98`) and into metrics tags. FEEL evaluates against process variables, so a bare `physicalTenantId` reference fails on an unresolved variable.

I'd drop the example. If per-tenant endpoint routing is a use case we want to show, the honest version is to set a tenant-specific variable at process start and reference that. 

Suggestion: 

```markdown
### Tenant context propagation

The runtime resolves each job's Physical Tenant from the activated job record and uses it to select
the per-tenant secret provider, document store, and metrics attribution for that job.

The `physicalTenantId` is not exposed as a process variable, so it cannot be referenced directly
from a FEEL expression in a connector's element template. To route a connector to a tenant-specific
endpoint, set a variable at process start and reference that variable instead.
```

docs/self-managed/concepts/physical-tenants/connectors-runtime.md
```diff
@@ -0,0 +1,181 @@
...
+- **Canonical:** `camunda.secrets.MY_SECRET` — FEEL expression form
+- **Legacy:** `{{secrets.MY_SECRET}}` — string template, supported indefinitely
+
+Resolved values are never written to the variable store, Operate, Tasklist, logs, or the command log. Only the reference name appears in exports — the reference is a pointer, not the value.
+
+Secret resolution requires `SECRETS:REVEAL` on the `SECRET` resource type, scoped per Physical Tenant. The runtime calls `POST /v2/secrets/resolve` scoped to the correct Physical Tenant for each job.
+
+:::note
+Transparent secret resolution (automatic resolution of `{{secrets.MY_SECRET}}` references in the connector context) is subject to Connector team capacity for 8.10. If unavailable at release, secret references must be resolved explicitly via the API. Check the 8.10 release notes for final availability.
+:::
```
nicpuppa, line 118: ### :warning: Blocking

This section describes a mechanism that doesn't exist in 8.10. None of the following appear anywhere in the connectors codebase: `SECRETS:REVEAL`, a `SECRET` resource type, `POST /v2/secrets/resolve`, or `camunda.secrets.MY_SECRET` as a FEEL form. There is no gateway-side per-tenant secret store involved.

What actually ships is name composition in the environment secret provider:

- `camunda.connector.secretprovider.environment.physicaltenantaware`, **default `false`** (`ConnectorsAutoConfiguration.java:97`)
- When enabled, the lookup key becomes `${prefix}${physicalTenantId}_${name}` (`EnvironmentSecretProvider#composeSecretNameWithPrefix`)
- `SecretContext` does carry `physicalTenantId` (`JobHandlerContext.java:98`) — that part is right — but it only feeds the composed name

The consequence is the opposite of what line 108 says: with the flag at its default, clients pointed at different Physical Tenants share one flat environment namespace, so `{{secrets.FOO}}` in tenant B resolves the same `SECRET_FOO` as tenant A. "One tenant's secrets cannot be accessed from another tenant's job context" is **not true by default**, and this is the one place on the page where getting it wrong has a security consequence. It needs to read as a required configuration step, not a guarantee.

Two smaller points:

- `{{secrets.X}}` is the *only* supported form. Labelling it "legacy" against a "canonical" form that doesn't exist will send readers down a dead end.
- Line 115 ("never written to the variable store, Operate, Tasklist, logs, or the command log") is true of the runtime's own handling, but nothing prevents a connector from returning a resolved secret in its own result variables. Suggest softening to a statement about the runtime rather than an absolute guarantee.

Suggested replacement for lines 106-117:

````markdown
### Per-tenant secret access

Per-tenant secret isolation is **opt-in and disabled by default**. Enable it explicitly in any
multi-tenant deployment:

```yaml
camunda:
  connector:
    secretprovider:
      environment:
        physicaltenantaware: true
```

Equivalent environment variable:
`CAMUNDA_CONNECTOR_SECRETPROVIDER_ENVIRONMENT_PHYSICALTENANTAWARE`.

With this enabled, the runtime resolves each secret against a name scoped to the job's Physical
Tenant: `${prefix}${physicalTenantId}_${name}`. With the default secret prefix `SECRET_`, a
reference to `MY_SECRET` from a job on `tenanta` resolves the environment variable
`SECRET_tenanta_MY_SECRET`.

:::warning Secrets are shared across tenants by default
With `physicaltenantaware` left at its default of `false`, all configured clients resolve secrets
from a single flat namespace. A reference to `{{secrets.MY_SECRET}}` resolves the same
`SECRET_MY_SECRET` value regardless of which Physical Tenant the job belongs to. Enable
`physicaltenantaware` in any deployment where tenants must not share secret values.
:::

Reference secrets in connector properties using `{{secrets.MY_SECRET}}`. The runtime replaces
references when it binds the job's variables, and does not write resolved values back to the
variable store, Operate, Tasklist, or logs.

Independently of tenant scoping, `camunda.connector.secret-resolver.secret-filter.mode` controls
whether a connector element may reference secret keys it has not declared. It defaults to
`DISABLED`, meaning no key-level restriction is enforced. Set it to `LAX` or `STRICT` to restrict
each element to the secret keys declared in its process definition.
````

File: docs/self-managed/concepts/physical-tenants/connectors-runtime.md [hunk 5]
```diff
@@ -0,0 +1,181 @@
...
+
+## Inbound connectors
+
+Each Physical Tenant gets its own `ImportSchedulers` instance with a dedicated `CamundaClient`, so inbound connector events and state are isolated per tenant.
```
nicpuppa: Inverted. There is a **single** `ImportSchedulers` instance holding a `Map<physicalTenantId, SearchQueryClient>`, which polls every configured tenant concurrently on each scheduled tick (`ProcessDefinitionImportConfiguration.java:51`, `ImportSchedulers.java:48`) — not "its own instance per tenant".

`ImportSchedulers` is also an internal class name that doesn't belong in user-facing docs; what the reader needs is the behavior.

```suggestion
The runtime polls each configured Physical Tenant for process definitions using that tenant's own client, and tracks inbound connector state per tenant, so inbound executables and their state stay isolated across tenants.
```

docs/self-managed/concepts/physical-tenants/connectors-runtime.md
```diff
@@ -0,0 +1,181 @@
...
+Without namespaced paths, the first registered inbound connector matching a path claims the request regardless of Physical Tenant — enable namespaced paths in any multi-tenant deployment.
+
+:::warning Isolation relies on path uniqueness, not tenant-scoped credentials
+The routing layer performs pure path-segment matching on `physicalTenantId/tenantId/path` — no credential or signature check happens at the routing layer. HMAC and signature verification is handled per-connector downstream, inside each webhook element's own logic, and has no awareness of `physicalTenantId`. Tenant isolation depends entirely on each tenant's webhook paths being distinct and unguessable.
+:::
+
+### Known limitations
+
+- Each inbound request targets exactly one Physical Tenant. Routing a single webhook event to multiple tenants is not supported.
+- Cross-tenant inbound routing is not supported.
```
nicpuppa, line 149: :white_check_mark: **Verified — no changes needed here.** Flagging explicitly since the PR description lists this as awaiting my input.

docs/self-managed/concepts/physical-tenants/connectors-runtime.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Authorization
+
+| Permission                      | Resource            | Required for                   |
+| ------------------------------- | ------------------- | ------------------------------ |
+| `SECRETS:REVEAL`                | `SECRET`            | Resolving secret references    |
+| Standard job worker permissions | Process definitions | Activating and completing jobs |
+
+Permissions are scoped per Physical Tenant. Grant `SECRETS:REVEAL:*` (wildcard) to the Connectors runtime service account to preserve the pre-8.10 implicit-trust scope while keeping authorization explicit.
```
nicpuppa: ### :warning: Blocking

I don't think there is anything like this implemented in the connectors at the moment

Suggest replacing the section with what the runtime actually requires — per-client credentials plus standard job-worker permissions, scoped per tenant:

```markdown
## Authorization

Each `camunda.clients.*` entry authenticates independently using its own `auth.*` credentials, so
grant permissions per Physical Tenant to the identity configured for that client.

Each client's identity needs the standard permissions to activate, complete, and fail jobs for the
process definitions it serves, plus read access to process definitions for inbound connector
polling. See [Authorization model](/self-managed/concepts/physical-tenants/authorization-model.md)
for how permissions are scoped per Physical Tenant.
```

If there *is* a planned secrets permission for 8.10 that I'm not seeing, let's confirm the resource type and endpoint with whoever owns it before it goes in — as written it reads like settled fact.

docs/self-managed/concepts/physical-tenants/connectors-runtime.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Operational considerations
+
+### Scaling
+
+One runtime instance per cluster is standard. Deploy multiple instances only when you need runtime-layer resource isolation (for example, performance SLA separation) — Physical Tenants already isolate data structurally.
+
+### Monitoring
+
+Per-tenant job execution metrics are not available in 8.10. Micrometer metrics are tagged by job type and action only — no `physicalTenantId` or client dimension. Per-tenant filtering is a planned follow-up.
```
nicpuppa: ### :warning: Blocking

Every connector metric carries a `physicalTenantId` tag:

- Tag constant — `ConnectorMetrics.java:33`
- Outbound invocations counter and execution-time timer — `ConnectorMetrics.java:164`, `:183`
- Outbound gauges (last-completed, last-failed, max-execution-time) — `ConnectorOutboundMetrics.java:135`
- All inbound meters (activations, triggers, last-activated, last-triggered) — `ConnectorsInboundMetrics.java:64` onward
- Tenants that can't be resolved fall back to the tag value `default`, so the tag is never absent

There's also a `physicalTenantIds` query parameter on the `/outbound` metrics endpoints (`OutboundConnectorsRestController.java:122`) for filtering the aggregate by tenant.

```suggestion
### Monitoring

Every connector metric is tagged with `physicalTenantId`, so outbound and inbound activity can be filtered and grouped per Physical Tenant. Two tenants running the same connector type report separately rather than collapsing into one series.

Tagged metrics include outbound invocation counts and execution times, the last-completed and last-failed timestamps, and inbound activation and trigger counts. Jobs handled by a single-client deployment that configures no `physical-tenant-id` are reported under the tag value `default`.
```

docs/self-managed/concepts/physical-tenants/connectors-runtime.md
```diff
@@ -0,0 +1,181 @@
...
+
+### Secret rotation
+
+When a secret is rotated, the gateway cache expires within the configured TTL (default 20 seconds). No restart required. New job activations after the TTL receive the updated value.
```
nicpuppa: ### :warning: Blocking

The 20-second TTL belongs to a different component. The only 20s value in the runtime is the SaaS Console secret provider — `ConnectorsAutoConfiguration.java:184`, `new ConsoleSecretProvider(consoleSecretApiClient, Duration.ofSeconds(20))`. It's hardcoded rather than configured, and its API client throws unless `mode=saas`, so it cannot apply to a Self-Managed Physical Tenants deployment — which is what this page is about.

With the environment secret provider, rotating a secret means changing the environment variable, which requires a restart. "No restart required" is the reverse of the truth here.

For completeness: the only cache in this path is the secret-*key* cache (`camunda.connector.secret-resolver.secret-filter.cache.*`), which is size-bounded with no TTL, and caches declared secret keys read from the process definition XML — not secret values.

```suggestion
### Secret rotation

Secrets resolved through the environment secret provider are read from the runtime's environment, so rotating a value requires updating the environment variable and restarting the runtime.

Because a single runtime instance serves all configured Physical Tenants, rotating one tenant's secret restarts the shared runtime and therefore briefly interrupts every configured tenant. Plan rotations accordingly.
```

docs/self-managed/concepts/physical-tenants/connectors-runtime.md
```diff
@@ -0,0 +1,181 @@
...
+
+### Failure handling
+
+A failure in one tenant's job workers does not affect workers registered for other tenants.
```
nicpuppa, line 177: True for outbound job workers, but too broad as a blanket statement. 

Suggestion:

```suggestion
### Failure handling

A failure in one tenant's job workers does not affect workers registered for other tenants.

On the inbound side, a polling failure for one Physical Tenant is logged and does not stop the other tenants from completing their imports in the same cycle. The runtime's readiness signal is shared across tenants, however, so a persistent polling failure for a single tenant marks the whole runtime instance as not ready.
```

docs/self-managed/concepts/physical-tenants/connectors-runtime.md
```diff
@@ -0,0 +1,181 @@
+---
+id: connectors-runtime
+title: "Connectors runtime: Physical Tenant support"
+sidebar_label: "Connectors runtime"
+description: "Configure one Connectors runtime instance to serve multiple Physical Tenants, with per-tenant job workers, secret isolation, and inbound webhook routing."
```
nicpuppa: The description advertises "secret isolation", which — per my comment on the secrets section — is opt-in and off by default. Suggest wording that doesn't imply it's automatic:

```suggestion
description: "Configure one Connectors runtime instance to serve multiple Physical Tenants, with per-tenant job workers, opt-in secret scoping, and inbound webhook routing."
```

nicpuppa (changes requested):

docs/self-managed/concepts/physical-tenants/connectors-runtime.md
christinaausley: Thanks! updated to match your suggestion.

For context: this was written based on the general multi-client job-worker pattern used elsewhere in our Camunda docs (where `@JobWorker` is the standard registration mechanism), without direct confirmation against the connectors runtime source at the time. Good catch that the runtime's registration model is actually distinct... appreciate you tracing it down to the exact fallback behavior -- much more meaningful!

docs/self-managed/concepts/physical-tenants/connectors-runtime.md
christinaausley: Fixed -- the example was adapted from a single-client config snippet that doesn't need `mode` set explicitly, so it got missed when extending to the multi-client case. Thanks for pointing to the wiring tests!

docs/self-managed/concepts/physical-tenants/connectors-runtime.md
christinaausley: Updated 👍 the assumption was that since `physicalTenantId` is available on the activated job record, it would be exposed the same way other job-context values are to FEEL expressions.

docs/self-managed/concepts/physical-tenants/connectors-runtime.md
christinaausley: Updated -- this assumed the secrets model would follow the same resource/permission pattern established elsewhere for Physical Tenants (e.g., the `BACKUP`/`EXPORTER` resource types), and that the flagged "subject to Connector team capacity" note covered the remaining uncertainty - it didn't, since the whole mechanism described wasn't real to begin with, not just pending. Good catch!

docs/self-managed/concepts/physical-tenants/connectors-runtime.md
christinaausley: Updated ...this described the per-tenant isolation *outcome* but guessed at the underlying mechanism as one-instance-per-tenant rather than a single instance keyed by tenant. Fair point on dropping the internal class name too 👍

docs/self-managed/concepts/physical-tenants/connectors-runtime.md
christinaausley: Updated -- same root cause as the secrets section above...this table assumed a secrets permission model would ship alongside the resource-scoped pattern used elsewhere for Physical Tenants, without confirming it against the connectors implementation.

docs/self-managed/concepts/physical-tenants/connectors-runtime.md
christinaausley: Updated -- the 20-second TTL detail was picked up from the SaaS Console secret provider and assumed to apply to secret rotation generally, without confirming it's hardcoded to that specific SaaS-mode provider and irrelevant to the Self-Managed path this page covers

christinaausley:
Resolved all -- thanks @nicpuppa for the thorough review 👍

nicpuppa (approved):
lgtm, thanks

PR #9531, opened 2026-08-03
docs(self-managed): document multiple Optimize Helm releases for Camunda 8.10

### LINKED ISSUE #6239: Document supported pattern for deploying multiple Optimize instances against a single orchestration cluster (8.10)
## Description

Document the supported pattern for deploying **multiple Optimize instances against a single orchestration cluster** for Camunda 8 Self-Managed (8.10).

Today the Helm chart supports exactly one Optimize deployment per release. `optimize.replicas` horizontally scales that single deployment but those replicas share a single config, a single Identity client, a single Elasticsearch index prefix, and a single ingress route — so its not ideal for customers that want two independent Optimizes (e.g. per team, per tenant, with separate dashboards / index lifecycles / auth).

The longer-term direction is to make Optimize multi-tenant at the application level (one deployment, N tenants — similar to Web Modeler). Rather than extending the Helm chart to template multiple Optimize deployments inside one release (added chart complexity for a workaround that Optimize itself will obsolete), we should **document and support the existing two-release pattern**: deploy a second Helm release of the same chart in the same namespace with only `optimize.enabled: true`, pointing it at the first release's Elasticsearch + Identity + Keycloak via standard values.

## Expected vs Actual Behavior

_N/A — task/documentation issue, not a bug._

## Steps to Reproduce

_N/A — task/documentation issue._

## Motivation / Use Case

- Several customers have asked how to run more than one Optimize against a single Camunda Self-Managed installation (per-team / per-tenant Optimize, separate dashboards, independent index lifecycles, separate Identity clients).
- The current "horizontally scale Optimize" answer doesn't actually solve it — it's still one Optimize with one config.
- The two-release approach is achievable today with the chart as it ships, but undocumented; customers shouldn't have to discover it by trial-and-error.
- Investing in chart-level `optimize.instances[]` would be wasted once Optimize becomes multi-tenant — guidance is the right place to invest for 8.10.

## Acceptance Criteria

- [ ] Reference scenario files added in the chart for testing:
  - [ ] `values-platform.yaml` — full Self-Managed stack with one Optimize. Includes an `identity.clients` entry that provisions a second Keycloak client (e.g. `optimize-tenant-b`) for the extra Optimize.
  - [ ] `values-optimize-only.yaml` — second release of the same chart with **only** `optimize.enabled: true`. All other components disabled. Points at the platform release's Elasticsearch (`global.elasticsearch.url.host`), Keycloak (`global.identity.keycloak.url`), and Identity (`global.identity.service.url`). Uses the second Identity client (`global.identity.auth.optimize.clientId` + matching secret), `CAMUNDA_OPTIMIZE_ELASTICSEARCH_SETTINGS_INDEX_PREFIX` to keep its indices distinct, and a distinct `optimize.contextPath` so the combined ingress route doesn't collide.
- [ ] Documentation page added under Self-Managed Helm docs covering:
  - [ ] When to use this pattern (and the explicit note that Optimize multi-tenancy is the long-term direction; this is a stopgap).
  - [ ] Prerequisites (shared Elasticsearch, shared Keycloak realm, same namespace).
  - [ ] Step-by-step: install platform release → verify extra Keycloak client + assign `Optimize` role to relevant users → install Optimize-only release → verify both Optimize URLs are reachable through ingress.
  - [ ] Notes on index-prefix isolation, Identity client + role mapping, ingress merging on a shared host, and resource sizing per release.
- [ ] A user following the docs can deploy two Optimizes against one orchestration cluster end-to-end, each on its own context path, each authenticating with its own Identity client.
- [ ] Process data from the orchestration cluster reaches both Optimizes; each writes to its own ES index prefix (verifiable with `_cat/indices`).
- [ ] All resources are created by `helm install` — no `kubectl apply` of hand-written manifests, no `global.extraManifests` workaround.

## Additional Context

Internal proof-of-concept in namespace `hamza-optimize` (cluster `distro-ci`, GKE):

- Release `integration` — full Camunda 8.9 stack (orchestration cluster, Identity, bundled Keycloak, bundled Elasticsearch, first Optimize at `/optimize`). Identity provisions an extra Keycloak client `optimize2` via `identity.clients[]`.
- Release `optimize2` — same chart, `optimize.enabled: true` only. Points at release 1's ES / Keycloak / Identity. Uses client `optimize2`. Served at `/optimize2`.

Verified end-to-end:

- Both `/optimize` and `/optimize2` redirect to Keycloak with the correct `client_id` and render the login page.
- BPMN deployed to the orchestration cluster, started an instance, completed the job — process completed cleanly.
- Elasticsearch shows two independent index sets: 21 `optimize-*` indices and 22 `optimize2-*` indices.
- Two Ingress objects on the same host (one per release) — nginx-ingress merges them.

Key non-default values used in the second release:

```yaml
global:
  elasticsearch:
    enabled: true
    url:
      host: "integration-elasticsearch"  # release 1's ES service
  identity:
    keycloak:
      url:
        protocol: http
        host: "integration-keycloak"     # release 1's Keycloak
        port: "80"
      contextPath: "/auth"
      realm: "/realms/camunda-platform"
    service:
      url: "http://integration-identity:80"
    auth:
      optimize:
        clientId: optimize2              # extra client provisioned by release 1
        secret:
          inlineSecret: "<matches identity.clients[optimize2].secret>"

identity:        { enabled: false }
identityKeycloak:{ enabled: false }
identityPostgresql: { enabled: false }
elasticsearch:   { enabled: false }
orchestration:   { enabled: false }
connectors:      { enabled: false }
console:         { enabled: false }
webModeler:      { enabled: false }

optimize:
  enabled: true
  contextPath: "/optimize2"
  env:
    - name: CAMUNDA_OPTIMIZE_ELASTICSEARCH_SETTINGS_INDEX_PREFIX
      value: "optimize2"
```

These values files (`hamza-optimize-values.yaml`, `hamza-optimize-2-values.yaml`) will be the basis for the reference examples shipped with the docs.

## Management Identity Configuration

The second Optimize authenticates as a **distinct Keycloak client** so each instance has its own client ID, secret, and redirect URI. The client is provisioned by Management Identity itself when the platform release boots — no Keycloak Admin UI clicks required.

### 1. Provision the extra Optimize client (release 1 only)

The platform release's `values-platform.yaml` adds an entry to `identity.clients[]`. Management Identity reads this on startup and creates the client in the `camunda-platform` realm via the Keycloak Admin API:

```yaml
identity:
  clients:
    - id: optimize2
      name: Optimize2
      type: confidential
      secret:
        # Either inlineSecret (non-prod) or existingSecret/existingSecretKey (prod).
        inlineSecret: "<strong-random-secret>"
      # rootUrl is the absolute base; redirectUris is resolved relative to it.
      rootUrl: "https://camunda.example.com/optimize2"
      redirectUris: "/api/authentication/callback"
      # Client-level permissions: what audiences this client may obtain tokens for.
      permissions:
        - resourceServerId: optimize-api
          definition: write:*
        - resourceServerId: camunda-identity-resource-server
          definition: read:users
```

Notes:
- `resourceServerId: optimize-api` matches the audience the chart's `component-presets.optimize` registered for the first Optimize, so **no separate API/resource server needs to be defined** — the existing one is reused. The Optimize backend will accept tokens with audience `optimize-api` regardless of which client minted them.
- `resourceServerId: camunda-identity-resource-server` + `read:users` lets the second Optimize resolve user info from Identity (display names, roles).
- For confidential clients, Identity expects a Kubernetes Secret it can read at boot. Use `secret.existingSecret` / `existingSecretKey` for anything other than a quick test.

### 2. Verify the client landed in Keycloak

After `helm install` of release 1, check the client exists with the right redirect URI:

```bash
PW=$(kubectl -n <ns> exec <release>-keycloak-0 -- \
  cat /opt/bitnami/keycloak/secrets/identity-keycloak-admin-password)
TOKEN=$(kubectl -n <ns> exec <release>-keycloak-0 -- curl -s -X POST \
  http://localhost:8080/auth/realms/master/protocol/openid-connect/token \
  -d username=admin -d password="$PW" -d grant_type=password -d client_id=admin-cli \
  | jq -r .access_token)
kubectl -n <ns> exec <release>-keycloak-0 -- curl -s -H "Authorization: Bearer $TOKEN" \
  "http://localhost:8080/auth/admin/realms/camunda-platform/clients?clientId=optimize2" \
  | jq '.[0] | {clientId, enabled, redirectUris}'
```

Expected (with `rootUrl: https://camunda.example.com/optimize2` + relative `redirectUris: /api/authentication/callback`):

```json
{
  "clientId": "optimize2",
  "enabled": true,
  "redirectUris": ["https://camunda.example.com/optimize2/api/authentication/callback"]
}
```

### 3. User access (role mapping)

The chart's `component-presets.optimize` already creates an **`Optimize` role** in the realm with permissions:

- `audience: optimize-api`, `definition: write:*`
- `audience: camunda-identity-resource-server`, `definition: read:users`

Because both Optimize clients share the `optimize-api` audience, **any user with the `Optimize` role can use either Optimize instance**. The chart assigns this role to the initial demo user automatically when `identity.firstUser.enabled: true && optimize.enabled: true`. For real users, assign the `Optimize` role in the Management Identity UI (Users → User → Assigned roles) or via Identity's user API.

> **Per-tenant isolation:** if Team A should see only Optimize 1 and Team B only Optimize 2, the two Optimize instances need **different audiences** (not just different `client_id`s) and **separate roles** in Identity. The chart's current presets register one `optimize-api` audience and one `Optimize` role, so this isn't directly expressible via values today — it's a future requirement to call out in the docs and is one of the things Optimize multi-tenancy is expected to handle natively.

### 4. Point the second release's Optimize at the client

In `values-optimize-only.yaml`, the optimize component's auth is set to the `optimize2` client. The `inlineSecret` here must match the value set in the `identity.clients[]` entry on release 1:

```yaml
global:
  identity:
    auth:
      optimize:
        clientId: optimize2
        audience: optimize-api
        redirectUrl: "https://camunda.example.com/optimize2"
        secret:
          inlineSecret: "<same-as-release-1>"
```

### Summary

| | Release 1 (platform) | Release 2 (optimize-only) |
|---|---|---|
| Keycloak `client_id` | `optimize` (chart preset) | `optimize2` (custom, via `identity.clients[]` on release 1) |
| Client secret source | `global.identity.auth.optimize.secret` | Must match release 1's `identity.clients[optimize2].secret` |
| Token audience | `optimize-api` | `optimize-api` (shared) |
| User role | `Optimize` | `Optimize` (shared — same audience) |
| Redirect URI | `/optimize/api/authentication/callback` | `/optimize2/api/authentication/callback` |
| ES index prefix | `optimize` (default) | `optimize2` (via `CAMUNDA_OPTIMIZE_ELASTICSEARCH_SETTINGS_INDEX_PREFIX`) |
| Ingress path | `/optimize` | `/optimize2` |

## References

- Slack thread: `#prj-sm-user-journeys-8_10`
- Related: #4870 (prior PoC validating multi-Optimize + shared management identity with separate confidential clients)
- Long-term direction: Optimize multi-tenancy (PM investigation) — once available, this pattern can be deprecated.

## Environment

- **Platform:** GKE (validation cluster `distro-ci`, ingress-nginx, external-dns)
- **Helm CLI version:** 3.20.x
- **Chart version:** `camunda-platform` 14.1.0 (appVersion 8.9.x) — guidance to target 8.10
- **Kubernetes version:** v1.33

bkenez:
## Description

Documents the Camunda 8.10 two-release Helm pattern for running multiple Optimize instances against one orchestration cluster. The guide covers release boundaries, existing-secret setup, Keycloak and generic OIDC clients, ingress callbacks, authorization limitations, index-prefix isolation, sizing, verification, troubleshooting, and cleanup.

Closes https://github.com/camunda/camunda-platform-helm/issues/6239.

Companion Helm reference-values PR: https://github.com/camunda/camunda-platform-helm/pull/6761

Validation:

- `npx prettier --check docs/self-managed/deployment/helm/operational-tasks/deploy-multiple-optimize-instances.md docs/self-managed/components/optimize/configuration/shared-elasticsearch-cluster.md sidebars.js`: passed.
- `CI=true npm run build`: passed and generated the new page. Existing repository-wide broken-anchor warnings remained; none originated from this guide.

Live GKE verification used two Helm releases on one ingress-nginx host. Both Optimize URLs authenticated with their intended confidential clients in fresh browser contexts, a process instance completed, and both isolated Optimize index families imported matching process documents. The unique namespace was deleted afterward.

- `deploy-camunda matrix run --versions 8.10 --shortname-filter eske --shortname-exact --flow-filter install --platform gke ...`: passed (`1/1` matrix entry).
- `helm upgrade integration ... -f values-platform.yaml --wait`: deployed revision 2; all platform and companion pods ready.
- `helm install optimize-team-b ... -f values-optimize-only.yaml --wait`: deployed; exactly one workload (`Deployment/optimize-team-b`) ready.
- OIDC redirect and fresh Chromium login checks: both Optimize clients authenticated and landed on their intended context paths.
- Orchestration REST deployment and process-instance creation: process completed successfully.
- Elasticsearch `_search`: two matching process documents under each isolated Optimize prefix.
- Namespace cleanup: completed.

A brand-new independent verifier checked every operational claim against the current Camunda 8.10 chart and reported no material findings.

## PR Checklist

### DIFF
--- a/docs/self-managed/components/optimize/configuration/shared-elasticsearch-cluster.md
+++ b/docs/self-managed/components/optimize/configuration/shared-elasticsearch-cluster.md
@@ -14,6 +14,8 @@ This is due to the fact that a dedicated cluster provides the highest reliabilit
 
 The following illustration demonstrates this use case with two Optimize instances that connect to the same Elasticsearch/OpenSearch cluster but are configured with different `*.settings.index.prefix` values. This results in different indexes and aliases created on the cluster, strictly isolating the data of both Optimize instances, so no instance accesses the data of the other instance.
 
+To deploy two Optimize instances against one Camunda 8.10 orchestration cluster with Helm, follow [deploy multiple Optimize instances with Helm](/self-managed/deployment/helm/operational-tasks/deploy-multiple-optimize-instances.md).
+
 :::note Warning
 Changing the value of `*.settings.index.prefix` after an instance was already running results in new indexes being created with the new prefix value. There is no support in migrating data between indexes based on different prefixes.
 :::
--- a/docs/self-managed/deployment/helm/operational-tasks/deploy-multiple-optimize-instances.md
+++ b/docs/self-managed/deployment/helm/operational-tasks/deploy-multiple-optimize-instances.md
@@ -37,7 +37,7 @@ Both releases must use the same Camunda 8.10 chart and application version. Inst
 Before you install the releases, prepare the following:
 
 - A production-ready Kubernetes cluster and a Helm CLI version supported by the Camunda 8.10 chart.
-- A single-region Camunda deployment. For a multi-region deployment, explicitly enable `orchestration.exporters.zeebe.enabled` and validate the exporter topology before using this pattern.
+- A single-region Camunda deployment. This pattern relies on Management Identity, which [dual-region deployments don't support](/self-managed/concepts/multi-region/dual-region.md#limitations) — Optimize itself isn't supported there either.
 - A namespace with enough capacity for the platform and a second Optimize Deployment.
 - An external Elasticsearch or OpenSearch cluster configured for the platform release. Follow the [Elasticsearch](/self-managed/deployment/helm/configure/database/elasticsearch/using-external-elasticsearch.md) or [OpenSearch](/self-managed/deployment/helm/configure/database/using-external-opensearch.md) guide.
 - An external Keycloak or supported OIDC provider and a Management Identity configuration. The reference files use the [external Keycloak setup](/self-managed/deployment/helm/configure/authentication-and-authorization/external-keycloak.md).
@@ -260,7 +260,7 @@ Using the same host for multiple Ingress objects is controller-specific. This gu
 
 ## Verify process data and index isolation
 
-Both Optimize instances read the same `zeebe-record` indices produced by the Orchestration Cluster. In this single-region reference topology, the 8.10 chart automatically enables the legacy Zeebe exporter when Optimize and its Elasticsearch or OpenSearch connection are enabled. Deployments with two or more regions must explicitly set `orchestration.exporters.zeebe.enabled: true`.
+Both Optimize instances read the same `zeebe-record` indices produced by the Orchestration Cluster. The 8.10 chart automatically enables the legacy Zeebe exporter when Optimize and its Elasticsearch or OpenSearch connection are enabled. This pattern applies to single-region deployments only — [dual-region deployments don't support Optimize](/self-managed/concepts/multi-region/dual-region.md#limitations).
 
 1. Deploy a test process to the shared Orchestration Cluster.
 1. Start and complete at least one process instance.
--- a/docs/self-managed/deployment/helm/operational-tasks/deploy-multiple-optimize-instances.md
+++ b/docs/self-managed/deployment/helm/operational-tasks/deploy-multiple-optimize-instances.md
@@ -41,7 +41,7 @@ Before you install the releases, prepare the following:
 - A namespace with enough capacity for the platform and a second Optimize Deployment.
 - An external Elasticsearch or OpenSearch cluster configured for the platform release. Follow the [Elasticsearch](/self-managed/deployment/helm/configure/database/elasticsearch/using-external-elasticsearch.md) or [OpenSearch](/self-managed/deployment/helm/configure/database/using-external-opensearch.md) guide.
 - An external Keycloak or supported OIDC provider and a Management Identity configuration. The reference files use the [external Keycloak setup](/self-managed/deployment/helm/configure/authentication-and-authorization/external-keycloak.md).
-- An ingress-nginx controller, one DNS host, and a TLS Secret for the host. The example relies on ingress-nginx merging paths from two Ingress objects with the same host and ingress class.
+- An `ingress-nginx` controller, one DNS host, and a TLS Secret for the host. The example relies on `ingress-nginx` merging paths from two Ingress objects with the same host and Ingress class.
 - A production values file for the full platform, including external PostgreSQL, datastore authentication and TLS, image pull credentials, and component resources. Follow the [production installation guide](/self-managed/deployment/helm/install/production/index.md).
 - Existing Kubernetes Secrets for every credential referenced by the values files. Follow the [secret management guide](/self-managed/deployment/helm/configure/secret-management.md#method-2-external-kubernetes-secrets-recommended-for-all-versions).
 
@@ -72,7 +72,7 @@ Update these values in both files before installation:
 | Value                                                    | Required change                                                     |
 | -------------------------------------------------------- | ------------------------------------------------------------------- |
 | `global.host`                                            | Set the shared browser-facing host.                                 |
-| `global.ingress.className`                               | Set the ingress-nginx class used by both releases.                  |
+| `global.ingress.className`                               | Set the `ingress-nginx` class used by both releases.                |
 | `global.ingress.tls.secretName`                          | Set the existing TLS Secret for the shared host.                    |
 | `global.identity.keycloak.*`                             | Set the in-cluster Keycloak service, port, context path, and realm. |
 | `global.identity.auth.publicIssuerUrl`                   | Set the issuer URL reachable from a user's browser.                 |
@@ -220,9 +220,9 @@ kubectl get deployments,statefulsets --namespace "$NAMESPACE" \
 
 The output must contain only the `optimize-team-b` Deployment and no StatefulSet.
 
-## Verify ingress and authentication
+## Verify Ingress and authentication
 
-The two releases create separate Ingress objects with the same host and different paths. ingress-nginx merges these rules into one virtual host:
+The two releases create separate Ingress objects with the same host and different paths. `ingress-nginx` merges these rules into one virtual host:
 
 | Release           | Path               | Callback URL                                                 |
 | ----------------- | ------------------ | ------------------------------------------------------------ |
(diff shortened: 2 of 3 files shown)


christinaausley (commented):
Nice! I have just two questions:

1. The prerequisites and "Verify process data and index isolation" sections both said multi-region deployments only need `orchestration.exporters.zeebe.enabled: true` to use this pattern. But `dual-region.md`'s limitations table states Optimize (and Management Identity, which this pattern depends on) isn't supported in dual-region at all. I updated both spots to cross-link the Dual-Region limitations table instead of suggesting it's reachable with one extra flag, but let me know if that's not the intended framing and should be adjusted across all of these docs?

2. The Helm PR that provides `values-platform.yaml`/`values-optimize-only.yaml` mentioned in the description isn't merged yet. I assume this won't merge until that does, right? Or can you confirm those files already exist at the referenced release-tag path before this merges, so the `curl` commands in "Prepare the reference values" don't 404 for early readers?

bkenez:
Thanks for the review and the fix!

1. Your dual-region framing is correct — the [dual-region limitations table](https://docs.camunda.io/docs/next/self-managed/concepts/multi-region/dual-region/#limitations) lists both Management Identity and Optimize as unsupported, so the old wording implying it was reachable with one exporter flag was wrong. Happy with the cross-links as you wrote them.

2. No ordering problem: the Helm PR (https://github.com/camunda/camunda-platform-helm/pull/6761) merged on Aug 5 and the values files are already included in the `camunda-platform-8.10-15.0.0-alpha4` release tag — I verified the raw URLs the guide constructs return HTTP 200 for that tag. Any 8.10 chart tag from alpha4 onward works, and by the time this page goes live with the 8.10 minor, the stable `15.0.0` tag will exist.

christinaausley (approved):
Thanks a ton!

PR #9499, opened 2026-07-29
docs(self-managed): explain Identity as the root cause of Kind OIDC discovery failures

leiicamundi:
## Description

Adds a troubleshooting entry for the failure reported in [camunda-deployment-references#2686](https://github.com/camunda/camunda-deployment-references/issues/2686), and corrects the memory requirement for the Kind guide.

In domain mode, Identity provisions the `camunda-platform` Keycloak realm. When Identity does not become ready, the realm is never created and every other component crash-loops while reporting the *missing realm* rather than the missing Identity:

- Orchestration: `503 Service Unavailable` on `https://camunda.example.com/auth/realms/camunda-platform`
- Connectors: `Failed to retrieve well known configuration ... status code 404 and message Not Found`

The reporter spent several weeks chasing Keycloak, DNS, and TLS because nothing pointed at Identity. In their logs Keycloak was healthy and the `master` realm resolved correctly, while `camunda-identity` sat in `CrashLoopBackOff` for over 100 minutes.

Changes:

- **New troubleshooting section**, `Components fail OIDC discovery with 404 or 503 (Domain mode)`: names Identity as the root cause, gives the commands to confirm the realm is missing and to inspect the Identity pod, and explains that restarting Orchestration or Connectors cannot help while Identity is down.
- **Prerequisites**: the stated 8 GB is not enough for `SECONDARY_STORAGE=elasticsearch`, which additionally runs Elasticsearch, Keycloak, and three PostgreSQL clusters. The guide now asks for 12 GB or more in that mode.
- **Insufficient resources**: aligned with the new figure, and now explains how to tell a container memory limit (`Reason: OOMKilled`) from a host resource shortage.

Both `/docs` (Next) and `/versioned_docs/version-8.9` received an identical patch, as 8.9 is the current version and the page the reporter followed.

Sister PR in the reference architectures repository, which raises the Kind Identity resource limits and makes the deployment scripts report the blocking component instead of looping silently: [camunda-deployment-references#2939](https://github.com/camunda/camunda-deployment-references/pull/2939)

This documentation change is independent of that PR and describes behavior that is already true today.

## When should this change go live?

- [x] This is a bug fix, security concern, or something that needs **urgent release support**. (add `bug` or `support` label)
- [ ] This is already available but undocumented and should be released within a week. (add `available & undocumented` label)
- [ ] This is on a **specific schedule** and the assignee will coordinate a release with the Documentation team. (create draft PR and/or add `hold` label)
[...]

PR #2963, opened 2023-11-29
Added docs about failJob

jonathanlukas:
## Description

This change documents a new connector feature linked in the issue-to-be-closed.

Documents this change: https://github.com/camunda/connectors/pull/1445

closes #2962 

## When should this change go live?

- [x] This change is not yet live and should not be merged until {ADD_DATE} (apply `hold` label or convert to draft PR)?
- [ ] There is no urgency with this change.
- [ ] This change or page is part of a marketing blog, conference talk, or something else on a schedule.
- [ ] This functionality is already available but undocumented.
- [ ] This is a bug fix or security concern.

## PR Checklist

- [ ] I have added changes to the relevant `/versioned_docs` directory, or they are not for an **already released version**.
- [x] I have added changes to the main `/docs` directory (aka `/next/`), or they are not for **future versions**.
- [x] My changes require an [Engineering review](https://github.com/camunda/camunda-docs/blob/main/howtos/documentation-guidelines.md#review-process), and I've assigned an engineering manager or tech lead as a reviewer, or my changes do not require an Engineering review.
- [x] My changes require a [technical writer review](https://github.com/camunda/camunda-docs/blob/main/howtos/documentation-guidelines.md#review-process), and I've assigned @christinaausley as a reviewer, or my changes do not require a technical writer review.

Issue #2939, opened 2023-11-21
Document Zeebe's default gRPC service config


npepinpe:
**Description**

Document the new default gRPC service config policy found in Zeebe, and how it can be used to enable per-call default retry policies.

megglos:
ZDP-Triage:
- we merged the policy recently and it needs to be documented
- you have to opt-in to use it, every grpc client could make use of it
- we should doc it for 8.4 for people to make use of it

akeller:
@npepinpe can you please triage this? It's still labeled for 8.4

npepinpe:
Oh wow, I totally missed that. That's when I was an IC as well :smile: I'll have it done today

npepinpe:
![shame](https://external-content.duckduckgo.com/iu/?u=https%3A%2F%2Fmedia4.giphy.com%2Fmedia%2F67HhViM1DcinS%2Fgiphy.gif&f=1&nofb=1&ipt=d1dc03a20edeb306d413f718ebc4dd9c815cef778239d2f047836c610ef84454&ipo=images)

Will do

PR #9542, opened 2026-08-04
docs(self-managed): collapse the deployment readiness script into a details block

leiicamundi:
## Description

The `check-deployment-ready.sh` reference script grows from 14 to roughly 140 lines in [camunda-deployment-references#2939](https://github.com/camunda/camunda-deployment-references/pull/2939). Four pages embed that file verbatim with ` ```bash reference `, so the moment the reference PR merges, each of them renders a wall of defensive bash — argument validation, three `jq` filter functions, a diagnostics printer — in the middle of a "track the progress of the installation" step. The prose on three of those pages even says "using the following command", which stops being accurate.

This PR keeps the auto-synced reference but stops it from dominating the page.

- A short paragraph describing what the script does: it polls until every pod is ready, and on a stall it reports the unready containers with restart counts, `CrashLoopBackOff`, `OOMKilled` and exit codes, plus recent warning events.
- A table of the environment variables the script now accepts: `CAMUNDA_NAMESPACE`, `DEPLOYMENT_READY_TIMEOUT_SECONDS`, `DEPLOYMENT_READY_INTERVAL_SECONDS`.
- The full source moved into a `details` block.

That content lives in a single partial, `_partials/_deployment-readiness-check.md`, following the existing pattern of `_identity-secret.md`. The four guides import it rather than repeating it, so there is one copy per version instead of one per page. Partials sit next to the docs they serve, so each version keeps the reference URL pinned to its own branch, which a shared React component in `src` could not do.

On the Kind page the reader has already obtained a local copy of the reference architecture through `get-your-copy.sh`, so the invocation is shown directly:

```bash
export CAMUNDA_NAMESPACE=camunda
./generic/kubernetes/single-region/procedure/check-deployment-ready.sh
```

The EKS, AKS, and OpenShift pages never instruct a clone and treat embedded scripts as copy-paste material, so on those the source stays visible inside the `details` block rather than being replaced by a command the reader could not run.

### Scope

Only `docs` (Next) and `versioned_docs/version-8.9` are touched, eight files in total. The 8.8 and 8.7 pages pin their reference URLs to `stable/8.8` and `stable/8.7`, which keep the original 14-line script, so changing them would describe something that does not exist on those branches.

The `openshift/dual-region.md` pages reference a different script, `generic/openshift/dual-region/procedure/check-deployment-ready.sh`, which is untouched by the reference PR and therefore left alone here.

### Dependency

This PR is a draft on purpose. It describes behavior that only exists once [camunda-deployment-references#2939](https://github.com/camunda/camunda-deployment-references/pull/2939) and its 8.9 backport [#2963](https://github.com/camunda/camunda-deployment-references/pull/2963) are merged. Merging earlier would document environment variables the released scripts do not yet support.

Related: [camunda-docs#9499](https://github.com/camunda/camunda-docs/pull/9499) adds the Identity troubleshooting entry that motivated the reference change.

### Validation

`npm run build` completes successfully and `prettier` reports no formatting issues on the eight files. The broken-anchor warnings the build emits are pre-existing and unrelated, they appear identically on an unmodified tree.

## PR Checklist

### DIFF
--- a/docs/self-managed/deployment/helm/cloud-providers/_partials/_deployment-readiness-check.md
+++ b/docs/self-managed/deployment/helm/cloud-providers/_partials/_deployment-readiness-check.md
@@ -1,0 +1,16 @@
+The script polls the namespace until every pod is `Running` with all of its containers ready. If the deployment stalls, it reports the containers that are not ready, with their restart count, waiting reason, last termination reason, and exit code, together with recent warning events, so you can see what is blocking it.
+
+Configure it with the following environment variables:
+
+| Variable                            | Default   | Purpose                                                          |
+| ----------------------------------- | --------- | ---------------------------------------------------------------- |
+| `CAMUNDA_NAMESPACE`                 | `camunda` | Namespace to watch                                               |
+| `DEPLOYMENT_READY_TIMEOUT_SECONDS`  | `1800`    | Wall-clock budget in seconds. Set it to `0` to wait indefinitely |
+| `DEPLOYMENT_READY_INTERVAL_SECONDS` | `5`       | Delay in seconds between two polls                               |
+
+<details>
+<summary>See the check-deployment-ready.sh script</summary>
+```bash reference
+https://github.com/camunda/camunda-deployment-references/blob/main/generic/kubernetes/single-region/procedure/check-deployment-ready.sh
+```
+</details>
--- a/docs/self-managed/deployment/helm/cloud-providers/_partials/_deployment-readiness-check.md
+++ b/docs/self-managed/deployment/helm/cloud-providers/_partials/_deployment-readiness-check.md
@@ -1,6 +1,14 @@
 The script polls the namespace until every pod is `Running` with all of its containers ready. If the deployment stalls, it reports the containers that are not ready, with their restart count, waiting reason, last termination reason, and exit code, together with recent warning events, so you can see what is blocking it.
 
-Configure it with the following environment variables:
+If you don't have a local copy of the reference architecture, download the script and run it:
+
+```bash
+curl -fsSL https://raw.githubusercontent.com/camunda/camunda-deployment-references/main/generic/kubernetes/single-region/procedure/check-deployment-ready.sh -o check-deployment-ready.sh
+chmod +x check-deployment-ready.sh
+./check-deployment-ready.sh
+```
+
+Review the script before you run it. Configure it with the following environment variables:
 
 | Variable                            | Default   | Purpose                                                          |
 | ----------------------------------- | --------- | ---------------------------------------------------------------- |
--- a/docs/self-managed/deployment/helm/cloud-providers/_partials/_deployment-readiness-check.md
+++ b/docs/self-managed/deployment/helm/cloud-providers/_partials/_deployment-readiness-check.md
@@ -1,14 +1,10 @@
+import DeploymentReadinessDownload from './\_deployment-readiness-download.md'
+
 The script polls the namespace until every pod is `Running` with all of its containers ready. If the deployment stalls, it reports the containers that are not ready, with their restart count, waiting reason, last termination reason, and exit code, together with recent warning events, so you can see what is blocking it.
 
-If you don't have a local copy of the reference architecture, download the script and run it:
+{props.download && <DeploymentReadinessDownload />}
 
-```bash
-curl -fsSL https://raw.githubusercontent.com/camunda/camunda-deployment-references/main/generic/kubernetes/single-region/procedure/check-deployment-ready.sh -o check-deployment-ready.sh
-chmod +x check-deployment-ready.sh
-./check-deployment-ready.sh
-```
-
-Review the script before you run it. Configure it with the following environment variables:
+Configure it with the following environment variables:
 
 | Variable                            | Default   | Purpose                                                          |
 | ----------------------------------- | --------- | ---------------------------------------------------------------- |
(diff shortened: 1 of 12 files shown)


christinaausley (commented):
Looks good! Two minor questions:

1. On `kind.md` the direct invocation (`./generic/kubernetes/single-region/procedure/check-deployment-ready.sh`) is followed by the partial's "If you don't have a local copy... download the script and run it" paragraph with a different curl-based invocation. Since the Kind guide has already given the reader a local copy via `get-your-copy.sh`, this reads almost contradictory with the instructions back to back? Was this intentional, or should we reword for when importing the partial?

2. The versioned partial pins `stable/8.9` for the reference script, but the 8.9 backport (2963) is still open per your dependency note...is 2963 guaranteed to expose the same three env vars with the same names/defaults as main's 2939? or could the 8.9 backport diverge from what this table documents?

leiicamundi:
Thanks, both are fair and the first one was a real defect. Fixed in 69bfb4342.

## 1. The two invocations on the Kind guide

You're right, and "almost contradictory" is generous. The reader got two runnable commands back to back:

```
./generic/kubernetes/single-region/procedure/check-deployment-ready.sh   <- from the page
curl -fsSL ... -o check-deployment-ready.sh && ./check-deployment-ready.sh   <- from the partial
```

It was intentional only in the weak sense that I wrote the sentence conditionally ("If you don't have a local copy...") so it would not be *wrong* on Kind. But not-wrong is a low bar for a tutorial, and the proximity made it read as a correction of the line above it.

The reason it was in the shared partial at all: Kind is the only one of the four guides that hands the reader a local copy, via `get-your-copy.sh`. EKS, AKS, and OpenShift never instruct a clone and treat embedded scripts as copy-paste material, so for them the download is the *primary* action, not a fallback. Dropping it would have left them with no way to obtain the script.

So I took your suggestion to reword on import, and made it opt-in instead. The download moved to its own partial, rendered from a `download` prop, which is the pattern already used in the agentic AI configuration partials:

```
{props.download && <DeploymentReadinessDownload />}
```

The three cloud guides pass `<DeploymentReadinessCheck download />`. Kind passes `<DeploymentReadinessCheck />` and now goes straight from its own invocation to what the script does and how to configure it. I also moved "Review the script before you run it" into the download partial, since it only makes sense when you have just fetched a script off the internet.

Verified in the built HTML rather than by eye:

| Page | curl block | env var table |
| --- | --- | --- |
| `kind` | absent | present |
| `eks-helm` | present | present |
| `aks-helm` | present | present |
| `redhat-openshift` | present | present |

## 2. Divergence between 8.9 and main

Good instinct to check rather than trust the note. The answer is that they cannot diverge today, and I verified it rather than assuming:

```
$ diff <(git show origin/main:generic/.../check-deployment-ready.sh) \
       <(git show origin/backport-2939-to-stable/8.9:generic/.../check-deployment-ready.sh)
check-deployment-ready.sh      BYTE-IDENTICAL
wait-for-keycloak.sh           BYTE-IDENTICAL
```

And the three variables with their defaults, extracted from each branch independently:

| Branch | Defaults |
| --- | --- |
| `main` | `CAMUNDA_NAMESPACE:-camunda`, `DEPLOYMENT_READY_TIMEOUT_SECONDS:-1800`, `DEPLOYMENT_READY_INTERVAL_SECONDS:-5` |
| `stable/8.9` backport | `CAMUNDA_NAMESPACE:-camunda`, `DEPLOYMENT_READY_TIMEOUT_SECONDS:-1800`, `DEPLOYMENT_READY_INTERVAL_SECONDS:-5` |

camunda/camunda-deployment-references#2963 is a plain cherry-pick of the three commits from #2939, with no adaptation: the two scripts applied cleanly because they were already identical on both branches before the change. Every review fix, including the Copilot findings and a later manual pass, was cherry-picked across too, and I re-checked byte-identity after each one.

Where I'd temper the guarantee: nothing *enforces* it going forward. If someone amends one branch without the other, the table could drift. Two things limit the blast radius: the partial is versioned, so `versioned_docs/version-8.9` and `docs` each pin their own branch and would have to be updated separately anyway, and the values are documented per version rather than shared. If you'd rather not carry even that risk, I can hold this PR until #2963 is merged and re-verify against the merged `stable/8.9` instead of the branch head. Say the word.

One status update since you reviewed: #2939 is now merged into `main` as `32179d4e`, so the `docs/` side of this table describes shipped behaviour. Only the 8.9 side is still pending #2963.

christinaausley (approved):
Thanks for the thoughtful response! I really appreciate it and think this looks great!

The `download` prop pattern is a nice fix - reusing the same approach as the agentic AI partials keeps this consistent rather than inventing a new mechanism, I like you verified the built HTML output directly rather than trusting the source. Moving "review the script before you run it" into the download partial makes sense too, since yeah it only applies once you're fetching something off the internet.

On holding for #2963: let's go ahead and merge now rather than wait perhaps? The backport being an unmodified cherry-pick with every review fix re-applied and re-verified gives me enough confidence, and if something does drift before #2963 merges, that's a quick follow-up fix to the table, not a structural problem. Not worth blocking this PR on another PR's timeline, but happy with whatever your preference is.

marcel-dias (approved):

leiicamundi:
> [!IMPORTANT]
> **Correction.** This comment claimed camunda-deployment-references#2963 still needed an approving review and was gating this PR. That was already wrong when I posted it: #2963 merged at 09:29 UTC, roughly two hours earlier. I read `mergeStateStatus: UNKNOWN` and inferred "still open" instead of checking `state`/`mergedAt`.
>
> The ordering advice itself held and was followed — #2963 merged first, this PR second — so nothing was published in the wrong order. But the stated reason for waiting was stale, and anyone reading it would have been misled about what was blocking what.

Merging as the backport is now merged, thanks @christinaausley for the feedback on download, I'll try to apply it to others PRs

---

PR #9558, opened 2026-08-05
docs: correct document store isolation validation and containment rules

### LINKED ISSUE #9555: Document store isolation for physical tenants: correct validation behavior and document containment rules
## Problem

[camunda/camunda#59383](https://github.com/camunda/camunda/pull/59383) (`fix: document store detect document containment`, closes camunda/camunda#56448) changes how `DocumentStoreIsolationValidation` compares document store locations across physical tenants. Several statements in [Storage isolation for Physical Tenants](https://github.com/camunda/camunda-docs/blob/main/docs/self-managed/concepts/physical-tenants/storage-isolation.md) are now incorrect, and the new rules are undocumented.

Before the change, a location was the *configured* values, lowercased and stripped of trailing slashes, compared for equality. After the change, a location is what the store *resolves to at startup*, split into a `namespace` that must match exactly and a `keyPrefix` that may not enclose or be enclosed by another tenant's.

## Incorrect statements to fix

In `docs/self-managed/concepts/physical-tenants/storage-isolation.md`:

| Current text | Correct behavior |
| --- | --- |
| **Validation**: Startup warning logged if two tenants share the same storage location | `DocumentStoreIsolationValidation` throws `UnifiedConfigurationException`, so the cluster fails to start. |
| Storage configuration matrix: Document Store collision detection = Startup warning (logs) | Startup error, the same as RDBMS and Elasticsearch/OpenSearch. |
| Don't point two tenants to the same bucket without distinct subpaths | Subpaths must be siblings. Nesting is rejected unconditionally, so no tenant can use the bucket or container root while another uses a path inside it. |
| **Subpath structure**: `s3://bucket/{physicalTenantId}/documents/` | No such convention exists. The prefix is the configured `bucket-path`, coerced to end in `/`. |
| Hybrid example uses `bucket-prefix: "tenant-a"` | The property is `bucket-path`. `bucket-prefix` does not exist. |

## Missing content

Add a section describing how document store locations are compared, covering:

- A location is `(provider, namespace, keyPrefix)`. Two tenants collide when the namespace matches and one key prefix is a prefix of the other.
- Namespace per provider: AWS = bucket and endpoint; GCP = bucket; Azure = container and the resolved blob endpoint; local = the configured directory through `Path.of`.
- Why containment is rejected without qualification: a document id is caller-supplied and appended to the prefix as given, and `/` draws no boundary any object storage service enforces. `docs/` plus the id `archive/invoice` resolves into `docs/archive/`.
- Object store key prefixes are now case-sensitive, so `Tenant-A/` and `tenant-a/` are distinct. Local paths are case-folded on every platform.
- Azure: a connection string is resolved to its endpoint, so two accounts configured through connection strings are now accepted, while a connection string and an explicit `endpoint` for the same account are now rejected. `UseDevelopmentStorage=true` resolves to the emulator endpoint. A SAS token is excluded from the location and no longer printed in the error.
- GCP: an unset `prefix` resolves to `temp/`, so an unset prefix collides with an explicit `temp/`.
- In-memory stores are exempt, because they are ephemeral and process-local.

Add a limitations section covering the known gaps:

- Aliases are not resolved. Two endpoint URLs or DNS names fronting the same backend are treated as separate locations.
- Legacy `DOCUMENT_*` environment configuration is not compared, because those stores are resolved after the check runs.
- Local paths are case-folded regardless of platform, so on a case-sensitive filesystem two directories differing only in case are reported as a collision.
- The document id charset is unrestricted at the storage layer. Only the configuration that would let it matter is rejected, at startup.
- Nested local directories are not compared. `LocalStorageDocumentStore` rejects `/`, `\`, and `..` in a document id, so the parent cannot descend into the child.

## Upgrade impact

Configurations that started before the change now fail startup, so this belongs in `docs/reference/announcements-release-notes/8100/8100-announcements.md`:

- An unset AWS `bucket-path` alongside a tenant with a set `bucket-path`.
- An unset GCP `prefix` alongside a tenant using `temp/` or `temp`.
- Any nested prefix, including a container or bucket root paired with a path inside it.
- An Azure connection string and an explicit `endpoint` addressing the same account.
- A Windows path written with `\` separators matching another tenant's `/` path.

Remediation is the same in every case: give each tenant its own sibling prefix.

## Related gap

`docs/self-managed/concepts/document-handling/configuration/camunda-8-run.md`, `docker.md`, and `helm.md` describe the GCP `prefix` property as optional but never state its `temp/` default. That default is now load-bearing for physical tenant validation.

konstantinos-mylosis:

## PR Checklist

closes #9555

### DIFF
--- a/docs/self-managed/concepts/document-handling/configuration/camunda-8-run.md
+++ b/docs/self-managed/concepts/document-handling/configuration/camunda-8-run.md
@@ -134,11 +134,11 @@ camunda:
         prefix: documents/ # optional
 ```
 
-| Property                                | Required | Description                                                         |
-| --------------------------------------- | -------- | ------------------------------------------------------------------- |
-| `camunda.document.gcp.<id>.bucket-name` | Yes      | Name of the Google Cloud Storage bucket where documents are stored. |
-| `camunda.document.gcp.<id>.prefix`      | No       | Folder-like prefix within the GCS bucket.                           |
-| `camunda.document.default-store-id`     | Yes      | Instance ID of the store to use as the default.                     |
+| Property                                | Required | Description                                                                                                                                                                 |
+| --------------------------------------- | -------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `camunda.document.gcp.<id>.bucket-name` | Yes      | Name of the Google Cloud Storage bucket where documents are stored.                                                                                                         |
+| `camunda.document.gcp.<id>.prefix`      | No       | Folder-like prefix within the GCS bucket. Defaults to `temp/` when unset. The value is used as written, with no separator appended, so a prefix isn't necessarily a folder. |
+| `camunda.document.default-store-id`     | Yes      | Instance ID of the store to use as the default.                                                                                                                             |
 
 <details>
 <summary>Deprecated: legacy environment variable equivalents</summary>
--- a/docs/self-managed/concepts/document-handling/configuration/docker.md
+++ b/docs/self-managed/concepts/document-handling/configuration/docker.md
@@ -139,11 +139,11 @@ camunda:
         prefix: documents/ # optional
 ```
 
-| Property                                | Required | Description                                                         |
-| --------------------------------------- | -------- | ------------------------------------------------------------------- |
-| `camunda.document.gcp.<id>.bucket-name` | Yes      | Name of the Google Cloud Storage bucket where documents are stored. |
-| `camunda.document.gcp.<id>.prefix`      | No       | Folder-like prefix within the GCS bucket.                           |
-| `camunda.document.default-store-id`     | Yes      | Instance ID of the store to use as the default.                     |
+| Property                                | Required | Description                                                                                                                                                                 |
+| --------------------------------------- | -------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `camunda.document.gcp.<id>.bucket-name` | Yes      | Name of the Google Cloud Storage bucket where documents are stored.                                                                                                         |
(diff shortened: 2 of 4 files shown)


docs/self-managed/concepts/document-handling/configuration/camunda-8-run.md
```diff
@@ -134,11 +134,11 @@ camunda:
...
-| `camunda.document.gcp.<id>.prefix`      | No       | Folder-like prefix within the GCS bucket.                           |
-| `camunda.document.default-store-id`     | Yes      | Instance ID of the store to use as the default.                     |
+| Property                                | Required | Description                                                                                                                                                                 |
+| --------------------------------------- | -------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `camunda.document.gcp.<id>.bucket-name` | Yes      | Name of the Google Cloud Storage bucket where documents are stored.                                                                                                         |
+| `camunda.document.gcp.<id>.prefix`      | No       | Folder-like prefix within the GCS bucket. Defaults to `temp/` when unset. The value is used as written, with no separator appended, so a prefix isn't necessarily a folder. |
```
deepthidevaki, line 140: @konstantinos-mylosis This is not a breaking change compared to previous legacy configuaration, right?

deepthidevaki (approved):
Great documentation 🚀

docs/self-managed/concepts/document-handling/configuration/camunda-8-run.md
konstantinos-mylosis: Nope, it's working the same as before.
Just mentioned now

christinaausley (approved):
Looks great!

PR #9561, opened 2026-08-06
In process restore guide

### LINKED ISSUE #59357: New restore guide
Introduce a new standalone page for the _In process restore_ that wraps the new sequence of restoring a broker's partition data both for ES/OS and RDBMS

panagiotisgts:
## Description

Guide for the new in-process restore available in 8.10 for Zeebe's partition data (primary storage). Available under: `docs/next/self-managed/operational-guides/backup-restore/in-process-restore/`

closes: [#59357](https://github.com/camunda/camunda/issues/59357)

## PR Checklist

### DIFF
--- a/docs/self-managed/operational-guides/backup-restore/in-process-restore.md
+++ b/docs/self-managed/operational-guides/backup-restore/in-process-restore.md
@@ -184,7 +184,7 @@ The overall `status` reports the state of the cluster change that performs the r
 | `IN_PROGRESS` | The restore is running.                                              |
 | `COMPLETED`   | Every partition was restored and the brokers returned to processing. |
 | `FAILED`      | The restore change failed and did not complete.                      |
-| `CANCELLED`   | The restore change was cancelled.                                    |
+| `CANCELLED`   | The restore change was canceled.                                     |
 
 Each partition entry reports the progress of a single broker's copy of that partition:
 
@@ -238,10 +238,8 @@ Automatic retries can't help if the problem is the backup itself, for example if
 
    The restore status reports the change as `CANCELLED`, and the cluster stays in recovery mode.
 
-2. Send a new [restore request](#3-trigger-the-restore). Because each restore drops the local partition data before it writes the backup data, the new attempt does not build on the partial result of the cancelled one, and you can select a different backup target.
-
-Repeat this as often as needed. A restore stays acceptable while every broker is in recovery mode, so cancelling and retrying does not require another mode change.
+2. Send a new [restore request](#3-trigger-the-restore). Because each restore drops the local partition data before it writes the backup data, the new attempt does not build on the partial result of the canceled one, and you can select a different backup target.
 
 :::warning
-Don't leave a partially failed restore unfinished. Between cancelling a restore and completing a new one, Zeebe's internal data is a mix of restored and pre-restore state and cannot be trusted. Keep the cluster in recovery mode and retry until every partition reaches `RESTORED`. If you switch the cluster back to `PROCESSING` in that state, treat it as unrecoverable and restore again from a clean state.
+Don't leave a partially failed restore unfinished. Between canceling a restore and completing a new one, Zeebe's internal data is a mix of restored and pre-restore state and cannot be trusted. Keep the cluster in recovery mode and retry until every partition reaches `RESTORED`. If you switch the cluster back to `PROCESSING` in that state, treat it as unrecoverable and restore again from a clean state.
 :::
--- a/docs/self-managed/operational-guides/backup-restore/in-process-restore.md
+++ b/docs/self-managed/operational-guides/backup-restore/in-process-restore.md
@@ -11,22 +11,19 @@ keywords:
     "recovery mode",
     "in-process restore",
   ]
-description: "Restore Zeebe partition data on running brokers by switching the cluster into recovery mode and triggering a restore over the Orchestration Cluster REST API."
+description: "Restore Zeebe partition data in-process without requiring the restart of the brokers themselves."
 ---
 
-In-process restore restores Zeebe partition data on the brokers that are already running. In Camunda 8.10 and later, you switch the cluster into recovery mode, trigger a restore over the [Orchestration Cluster REST API](/apis-tools/orchestration-cluster-api-rest/orchestration-cluster-api-rest-overview.md), and track its progress per broker and partition.
-
-Compared to the [standalone restore application](./elasticsearch/restore.md#restore-zeebe-cluster), in-process restore requires no deployment changes. The brokers keep running, so you don't override the broker start command, set restore-only environment variables, or clear broker data directories yourself.
-
-The procedure uses two endpoints: [change cluster mode](/apis-tools/orchestration-cluster-api-rest/specifications/change-cluster-mode.api.mdx) and [restore from a backup](/apis-tools/orchestration-cluster-api-rest/specifications/restore.api.mdx).
+With Camunda 8.10 and later, you can restore Zeebe partition data in-process without requiring the restart of the brokers themselves. In-process restore is a downtime operation that runs while the cluster is in recovery mode, but it does not require any deployment changes or broker restarts, in contrast to the old [standalone restore application](./elasticsearch/restore.md#restore-zeebe-cluster). The brokers keep running, so you don't override the broker start command, set restore-only environment variables, or clear broker data directories yourself.
 
 ## How in-process restore works
 
 A restore runs in three phases, driven by two API requests:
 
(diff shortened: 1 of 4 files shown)


docs/self-managed/operational-guides/backup-restore/in-process-restore.md
```diff
@@ -0,0 +1,247 @@
...
+| Camunda version  | Camunda 8.10 or later, restored with the exact version the backup was created with.                                                                                                                                                        |
+| Backup store     | Every broker is configured with the same backup store that holds the backup, as described in the [Elasticsearch and OpenSearch](./elasticsearch/backup.md#prerequisites) or [RDBMS](./rdbms/backup.md#prerequisites) backup prerequisites. |
+| Completed backup | A completed backup exists for every partition. List the available backups with the [Zeebe backup management API](./zeebe-backup-and-restore.md#list-backups-api).                                                                          |
+| Partition count  | The partition count of the cluster matches the partition count of the backup. Brokers can be scaled between backup and restore as long as the partition count is unchanged.                                                                |
```
deepthidevaki, line 37: Is there a way to restore from an old backup before scaling? In theory it should be possible.

docs/self-managed/operational-guides/backup-restore/in-process-restore.md
```diff
@@ -0,0 +1,247 @@
...
+{
+  "changeId": "7",
+  "plannedChanges": [
+    { "operation": "ModeChangeOperation", "mode": "RECOVERING" },
+    { "operation": "AwaitModeChangeOperation", "mode": "RECOVERING" }
+  ]
+}
+```
+
+Wait until this change has completed before you trigger the restore. Query the [cluster monitoring API](/self-managed/components/orchestration-cluster/zeebe/operations/cluster-scaling.md#monitoring-api) and check that `lastChange.id` matches the returned `changeId` and that no `pendingChange` is reported:
```
deepthidevaki, line 72: 💭 It would be nice to monitor it via the same api for the mode change.

docs/self-managed/operational-guides/backup-restore/in-process-restore.md
```diff
@@ -0,0 +1,247 @@
...
+
+```bash
+curl "${ORCHESTRATION_CLUSTER_MANAGEMENT_API}/actuator/cluster"
+```
+
+A restore is only accepted while every broker of the cluster is in recovery mode. Requests sent earlier are rejected with `409`.
+
+### 2. Restore secondary storage
+
+With the cluster in recovery mode, nothing is exported to secondary storage, so restore it now. Skip this step if secondary storage was already restored another way, for example as part of a wider disaster recovery procedure.
```
deepthidevaki, line 82: Can we really skip this step? Secondary storage can be only restored after the partition switch to recovery mode. Otherwise, it would have unexpected state.

docs/self-managed/operational-guides/backup-restore/in-process-restore.md
```diff
@@ -0,0 +1,247 @@
...
+
+   ```bash
+   curl -X DELETE "${ORCHESTRATION_CLUSTER_MANAGEMENT_API}/actuator/cluster/changes/8"
+   ```
+
+   The restore status reports the change as `CANCELLED`, and the cluster stays in recovery mode.
+
+2. Send a new [restore request](#3-trigger-the-restore). Because each restore drops the local partition data before it writes the backup data, the new attempt does not build on the partial result of the cancelled one, and you can select a different backup target.
+
+Repeat this as often as needed. A restore stays acceptable while every broker is in recovery mode, so cancelling and retrying does not require another mode change.
```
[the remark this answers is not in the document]
deepthidevaki: >  A restore stays acceptable while every broker is in recovery mode,

I don't understand what this means.

docs/self-managed/operational-guides/backup-restore/in-process-restore.md
```diff
@@ -0,0 +1,247 @@
...
+    { "operation": "PartitionPreRestoreOperation", "mode": null },
+    { "operation": "PartitionRestoreOperation", "mode": null },
+    { "operation": "ModeChangeOperation", "mode": "PROCESSING" },
+    { "operation": "AwaitModeChangeOperation", "mode": "PROCESSING" },
+    { "operation": "UpdateIncarnationNumberOperation", "mode": null }
+  ]
+}
+```
+
+The partition operations repeat once per broker and partition, and the plan does not name the broker, the partition, or the resolved backup. Use the [restore status](#4-track-the-restore) to see which backups a partition is restored from.
```
deepthidevaki, line 118: Technically, you have the backupId already at this point because the validation is completed. So why not also include it in the response?

deepthidevaki (dismissed):
Looks good to me. My comments are mostly about the feature not for the documentation. 

I didn't review the writing style. So please wait for a Techwriter's review before merging.

docs/self-managed/operational-guides/backup-restore/in-process-restore.md
panagiotisgts: This was required in the old restore as well. I guess since we only scale-up right now it should be no issue as the _extra_ partition backup would simply be ignored. When scale down is implemented I think it won't be that clear. But i'd have to verify it a bit more thorough

docs/self-managed/operational-guides/backup-restore/in-process-restore.md
panagiotisgts: It could be skipped I guess if you first do a hard pose exporting and then perform the secondary storage restore. Then there's no need to do it again in this sequence.

docs/self-managed/operational-guides/backup-restore/in-process-restore.md
panagiotisgts: I was considering this as a follow up when done with the rest of the admin ops, as it would also be handy for RDBMS dry runs mostly. It is possible by extracting the information from the operations, yes :+1:

docs/self-managed/operational-guides/backup-restore/in-process-restore.md
deepthidevaki: What I meant is that when we detect the backup has lower number of partitions, then we only restore and bootstrap those partitions and stop all the remaining scaled up partitions. 

Ignore this for now. This could be a later followup.

docs/self-managed/operational-guides/backup-restore/in-process-restore.md
panagiotisgts: Right now I don't find the necessity for a dedicated API endpoint to track the mode change. We could consider this in the future if we see fit.

deepthidevaki: The monitoring is only exposed via actuators, then? So a user of a physical tenant cannot really monitor it.

christinaausley (approved):
Looks good 👍  The "Validate a restore without applying it" section says both `/mode` and `/restore` accept `dryRun`, with an example calling `POST /restore?dryRun=true`. The current `cluster.yaml` spec on `camunda/camunda@main` only defines `dryRun` as a parameter on `PATCH /mode` - the `/restore` POST operation has no `parameters` block, only a `requestBody`. Is `dryRun` on `/restore` still pending in an unmerged engine PR, or should this section be scoped to mode-change dry runs only?

panagiotisgts:
@christinaausley 

> Looks good 👍 The "Validate a restore without applying it" section says both /mode and /restore accept dryRun, with an example calling POST /restore?dryRun=true. The current cluster.yaml spec on camunda/camunda@main only defines dryRun as a parameter on PATCH /mode - the /restore POST operation has no parameters block, only a requestBody. Is dryRun on /restore still pending in an unmerged engine PR, or should this section be scoped to mode-change dry runs only?

Good catch, it was forgotten in the `cluster.yaml` spec, I'll address this, but it will have a `dryRun` parameter as described.

panagiotisgts:
Addressed the above in https://github.com/camunda/camunda/pull/59716

docs/self-managed/operational-guides/backup-restore/in-process-restore.md
panagiotisgts: The completed/planned changes should be visible in the `/cluster` REST API endpoint no? Or that would only include the cluster-wide topology?

christinaausley (dismissed):

christinaausley:
@panagiotisgts Don't want to mess with your PR too much so I will let you resolve the rebasing and then can approve and merge for you 👍

christinaausley (approved):

Issue #6813, opened 2025-09-22
Add setting env vars to deploying with Docker Compose


christinaausley:
A user complains the envs are not set:

WARN[0000] The "CAMUNDA_ZEEBE_VERSION" variable is not set. Defaulting to a blank string. WARN[0000] The "CAMUNDA_OPERATE_VERSION" variable is not set. Defaulting to a blank string. WARN[0000] The "CAMUNDA_TASKLIST_VERSION" variable is not set. Defaulting to a blank string. WARN[0000] The "CAMUNDA_CONNECTORS_VERSION" variable is not set. Defaulting to a blank string. WARN[0000] The "ELASTIC_VERSION" variable is not set. Defaulting to a blank string. unable to get image 'camunda/tasklist:': Error response from daemon: invalid reference format

Need to confirm with team and add relevant details to docs.

bkenez:
Addressed by #9504, which restores complete distribution archive installation and documents the required `.env` handling across 8.8, 8.9, and 8.10.

PR #9569, opened 2026-08-06
docs(self-managed): Remove Connectors from document store IRSA configuration

bkenez:
## Description

Updates the Amazon EKS document-store IRSA guide for [camunda-platform-helm#6813](https://github.com/camunda/camunda-platform-helm/pull/6813). Related work remains tracked in [camunda-platform-helm#6778](https://github.com/camunda/camunda-platform-helm/issues/6778).

- Removes the Connectors service account from the document-store IAM role trust policy and Helm values example because Connectors accesses documents through the Orchestration REST API.
- Adds migration guidance for connector tasks that use the AWS default credentials chain, static AWS credentials, or a mounted GCP service account file.
- Directs users to configure connector-task credentials through Connectors-owned settings and a connector-specific role or Secret.
- Applies the correction to the next documentation and the maintained 8.7, 8.8, and 8.9 versions.

Review requested from @camunda/tech-writers.

## PR Checklist

### DIFF
--- a/docs/self-managed/deployment/helm/cloud-providers/amazon/amazon-eks/irsa.md
+++ b/docs/self-managed/deployment/helm/cloud-providers/amazon/amazon-eks/irsa.md
@@ -268,8 +268,7 @@ The role's **trust policy** must allow the relevant Kubernetes service accounts 
       "Condition": {
         "StringEquals": {
           "oidc.eks.<region>.amazonaws.com/id/<oidc-id>:sub": [
-            "system:serviceaccount:<namespace>:<release-name>-orchestration",
-            "system:serviceaccount:<namespace>:<release-name>-connectors"
+            "system:serviceaccount:<namespace>:<release-name>-orchestration"
           ]
         }
       }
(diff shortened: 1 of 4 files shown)


christinaausley (commented):
Nice work!

I checked the companion Helm chart PR this docs change depends on (camunda-platform-helm#6813, "fix: stop propagating document store credentials to connectors") and it shows as not yet merged. I noticed this PR is being backported to the maintained 8.7, 8.8, and 8.9 versions, all stating that "Camunda no longer propagates document-store credentials to the Connectors pod."

If a reader is on an unpatched chart version (before #6813 ships in a patch release for their minor), this statement won't be true for them yet...their Connectors pod would still receive the ambient credentials, right? Should this PR wait until #6813 merges and ships in patch releases for 8.7/8.8/8.9, or is there a specific patch version per minor we should call out as the cutover point so readers on older patches know they're still exposed?

bkenez:
You're right, and thanks for catching the ordering — I had it backwards.

The statement is version-scoped: it only becomes true for a reader once #6813 is in a *released* patch for their minor. So this PR shouldn't merge ahead of that, and I'll keep it in draft until the fix has shipped for 8.7, 8.8, and 8.9. camunda-platform-helm#6813 is approved and covers all four versions (8.7–8.10), matching this PR's surface, so the sequence is: #6813 merges → patch releases go out → this goes ready.

One alternative if you'd rather not hold the docs that long: state the minimum patch version inline per version instead of an unqualified claim. Happy either way — let me know which you prefer.

christinaausley:
Agreed, holding until #6813 ships makes sense either way since we don't have the patch numbers yet.

When you come back to this I'd lean toward keeping the version-scoped phrasing even after it ships, rather than switching to an unqualified claim: something like "As of chart version 8.7.X/8.8.Y/8.9.Z/8.10.0, Camunda no longer propagates document-store credentials to the Connectors pod." That way a reader who's a few patches behind still gets an accurate answer, instead of the doc going stale again for anyone not fully caught up.

Ping me once the chart fix has patch versions and I'll take another pass 💪

marcel-dias (approved):

PR #9572, opened 2026-08-06
Tenant Issue 10: Java Client Multi-Tenant Support

### LINKED ISSUE #8896: Tenant Issue 10: Java Client Multi-Tenant Support
**Review Stakeholders:** @nicpuppa @lenaschoenburg @deepthidevaki

- [#3640 Physical Tenant - Multi-Tenant Connector Runtime and Java Client](https://github.com/camunda/product-hub/issues/3640)

**Deliverables**
- Java client multi-tenant overview page covering:
  - One client instance targets multiple Physical Tenants
  - Multi-tenant configuration patterns (all tenants or subset)
  - Comparison with logical tenant targeting (existing behavior)
- Configuration guide for ZeebeClient covering:
  - gRPC gateway endpoint configuration
  - Physical Tenant targeting via header injection
  - Interceptor patterns for automatic tenant header addition
  - REST client configuration (if applicable in 8.10)
  - Credentials and auth header setup
- REST client guide covering (if in scope):
  - Tenant-scoped endpoint targeting
  - Query parameter vs path-based tenant specification
  - Base URL configuration per tenant or dynamic routing
- Code examples covering:
  - Creating client and targeting specific tenant
  - Job worker registration per tenant
  - Process instance creation scoped to tenant
  - Variable update scoped to tenant
  - Multi-tenant job worker (handling jobs from multiple tenants)
- Best practices section covering:
  - When to use one client vs multiple clients
  - Connection pooling across tenants
  - Error handling for unauthorized tenant access
  - Monitoring and logging across tenants

**Context We Have**
- Java client will support multi-tenant targeting via gRPC header
- One client instance can register job workers for multiple tenants
- Logical tenant ID (`tenantId`) remains in API method signatures
- Physical Tenant selected via header/interceptor, not API parameter
- Client must handle authentication/authorization per tenant
- gRPC interceptor pattern recommended for header injection

**Engineering Questions**
1. Should Java client builder have method like `.forPhysicalTenant(String tenantId)` to scope all subsequent calls, or should each API call accept optional tenant parameter?
2. If job worker registered without specific physical tenant scope, does it receive jobs from all tenants client is authorized for, or must workers be explicitly scoped?
3. For error messages, should "unauthorized tenant access" be distinguishable from "invalid credentials"?
4. Should Java client provide built-in metrics or tracing for multi-tenant operations, or left to application layer?

christinaausley:
## Description

Closes #8896

- New page: `apis-tools/java-client/physical-tenants.md` - Physical Tenant vs. logical tenant comparison, targeting a single/multiple Physical Tenants, code examples, best practices.
- `camunda-spring-boot-starter/configuration.md` - new multi-client (`camunda.clients.<name>.*`) config section, `@JobWorker` fan-out behavior, multi-client lifecycle events.
- `camunda-spring-boot-starter/properties-reference.md` - fixed a stale `physical-tenant-id` description that only mentioned the gRPC header, not the REST path effect.
- One open TODO remains inline: whether unauthorized-Physical-Tenant errors are a distinct client exception type

## PR Checklist

### DIFF
--- a/docs/apis-tools/camunda-spring-boot-starter/configuration.md
+++ b/docs/apis-tools/camunda-spring-boot-starter/configuration.md
@@ -169,29 +169,35 @@ A single `camunda.client.*` configuration targets exactly one [Physical Tenant](
 ```yaml
 camunda:
   clients:
-    teamA:
-      physical-tenant-id: teamA
+    finance:
+      physical-tenant-id: finance
       primary: true
-    teamB:
-      physical-tenant-id: teamB
+    risk:
+      physical-tenant-id: risk
 ```
 
-Each named client is a sparse overlay on top of the base `camunda.client.*` configuration — any property you don't set per client (address, auth, and so on) falls back to the shared `camunda.client.*` value. Exactly one client should be marked `primary: true`; if you configure only one client, it's implicitly primary.
+The map key is a free-form client name; it does not have to match the Physical Tenant ID. A `physical-tenant-id` must be lowercase alphanumeric and at most 64 characters — an invalid value fails application startup.
 
-All clients must share the same `camunda.client.auth.method` — mixing authentication methods across clients in one application throws an `IllegalArgumentException` at startup. Credentials themselves (client ID, secret, and so on) can still differ per client.
+Each named client is a sparse overlay on top of the base `camunda.client.*` configuration — any property you don't set per client (address, auth, and so on) falls back to the shared `camunda.client.*` value.
+
+At most one client may be marked `primary: true` — configuring more than one throws `IllegalArgumentException` at startup. If you configure only one client, it's implicitly primary. If you configure more than one client and mark none of them `primary: true`, there is no primary client at all: looking one up throws `IllegalStateException`, and the `camundaClientConfiguration` bean isn't registered.
+
+Each client's resolved `auth.method` (its own `camunda.clients.<name>.auth.method`, overlaid on the base `camunda.client.auth.method`) must be the same across all clients — mixing resolved authentication methods in one application throws an `IllegalArgumentException` at startup. An unset method normalizes to `none`, so leaving it unset on one client and setting `none` explicitly on another is fine; unset and `basic` together is a conflict. Credentials themselves (client ID, secret, and so on) can still differ per client.
 
 If you don't configure any `camunda.clients.*` entries, your application has a single, implicitly-named `default` client — existing single-client configuration is unaffected.
 
 #### Access a named client
 
-Each named client is also available as a Spring bean, named `<name>CamundaClient` (for example, `teamACamundaClient`). To look up clients by name at runtime, inject `CamundaClientRegistry`:
+Each named client is also available as a Spring bean, named `<name>CamundaClient` (for example, `financeCamundaClient`). Kebab-case client names are camel-cased for the bean name (`camunda.clients.risk-eu` → `riskEuCamundaClient`). The primary client's bean is `@Primary`, so a plain `@Autowired CamundaClient` still resolves it, and the historical `camundaClient` bean name is preserved as an alias for the primary client — `@Qualifier("camundaClient")` and `getBean("camundaClient")` keep working.
+
+To look up clients by name at runtime, inject `CamundaClientRegistry`. Injecting the registry does not eagerly instantiate every client — beans are resolved lazily on first lookup:
 
 ```java
 @Autowired
 private CamundaClientRegistry clientRegistry;
 
-public void useTeamAClient() {
-  CamundaClient teamAClient = clientRegistry.get("teamA");
+public void useFinanceClient() {
+  CamundaClient financeClient = clientRegistry.get("finance");
   // ...
 }
 ```
@@ -970,11 +976,13 @@ camunda:
 
 ##### Physical Tenant fan-out for multi-client applications
 
-The `@JobWorker` annotation has no attribute to bind a worker to one named client. In a [multi-client](#multi-client-configuration-physical-tenants) application, every `@JobWorker` method registers against **all** configured clients — it fans out across the whole `CamundaClientRegistry`, not just the primary or default one. A worker is never silently dropped for the clients it isn't explicitly bound to, since no such binding exists.
+The `@JobWorker` annotation has no attribute to bind a worker to one named client. In a [multi-client](#multi-client-configuration-physical-tenants) application, every `@JobWorker` method registers against **all** configured clients — it fans out across the whole `CamundaClientRegistry`, not just the primary or default one.
 
 In an application with only one client, this has no visible effect — there's only one client to fan out to.
 
-If you need a worker to run against only some of your configured Physical Tenants, filter inside the handler using the tenant identifier available to you at runtime, rather than relying on annotation-level scoping.
+Per-client worker overrides don't restrict this fan-out either: `worker.*`/`worker.override.*` properties set under `camunda.clients.<name>` have no effect, since the customizer that applies those overrides is built from the single, global `camunda.client.*` properties bean. Setting `camunda.clients.risk.worker.override.shipOrder.enabled: false`, for example, silently does nothing.
+
+If you need a worker to run against only some of your configured Physical Tenants, filter inside the handler using the tenant identifier available to you at runtime, rather than relying on annotation-level scoping or per-client worker configuration.
 
 #### Define the job timeout
 
@@ -1050,6 +1058,10 @@ public class MyRandomBean {
   // make sure this bean is registered
 }
 ```
+
+:::note Multi-client applications
+In a [multi-client](#multi-client-configuration-physical-tenants) application, every `@Deployment` resource is deployed to **all** configured clients — one `CamundaPostDeploymentSpringEvent` is published per client. For a multi-tenant application, this means your BPMN lands in every configured Physical Tenant, which is often what you want, but shouldn't come as a surprise. The same applies to `@ClusterVariables` processing.
+:::
 
 ### Specify resources to deploy
 
@@ -1250,6 +1262,8 @@ public void onDeploymentCreated(CamundaPostDeploymentEvent event) {
 ```
 
 The event will grant you access to a list of deployments that have been created.
+
+In a [multi-client](#multi-client-configuration-physical-tenants) application, one event fires per configured client, since `@Deployment` resources are deployed to every configured client.
 
 ## Observe metrics
 
--- a/docs/apis-tools/camunda-spring-boot-starter/configuration.md
+++ b/docs/apis-tools/camunda-spring-boot-starter/configuration.md
@@ -162,6 +162,42 @@ camunda:
 
 This does also affect the default tenant being used by all job workers, however there are [more possibilities](#control-tenant-usage) to configure them.
 
+### Multi-client configuration (Physical Tenants)
+
+A single `camunda.client.*` configuration targets exactly one [Physical Tenant](/self-managed/concepts/physical-tenants/index.md) (the default tenant, unless `physical-tenant-id` is set). To target multiple Physical Tenants from one Spring application, configure multiple named clients under `camunda.clients.<name>.*`:
+
+```yaml
+camunda:
+  clients:
+    teamA:
+      physical-tenant-id: teamA
+      primary: true
+    teamB:
+      physical-tenant-id: teamB
+```
+
+Each named client is a sparse overlay on top of the base `camunda.client.*` configuration — any property you don't set per client (address, auth, and so on) falls back to the shared `camunda.client.*` value. Exactly one client should be marked `primary: true`; if you configure only one client, it's implicitly primary.
+
+All clients must share the same `camunda.client.auth.method` — mixing authentication methods across clients in one application throws an `IllegalArgumentException` at startup. Credentials themselves (client ID, secret, and so on) can still differ per client.
+
+If you don't configure any `camunda.clients.*` entries, your application has a single, implicitly-named `default` client — existing single-client configuration is unaffected.
+
+#### Access a named client
+
+Each named client is also available as a Spring bean, named `<name>CamundaClient` (for example, `teamACamundaClient`). To look up clients by name at runtime, inject `CamundaClientRegistry`:
+
+```java
+@Autowired
+private CamundaClientRegistry clientRegistry;
+
+public void useTeamAClient() {
+  CamundaClient teamAClient = clientRegistry.get("teamA");
+  // ...
+}
+```
+
+`CamundaClientRegistry` also provides `find(String)` (returns an `Optional`), `getPrimary()`, `clientNames()`, and `all()` (a `Map` of every configured client, keyed by name).
+
 ## Authentication
 
 The authentication method is determined by `camunda.client.auth.method`. If omitted, the client will try to detect the authentication method based on the provided properties.
@@ -932,6 +968,14 @@ camunda:
             - foo
 ```
 
+##### Physical Tenant fan-out for multi-client applications
+
+The `@JobWorker` annotation has no attribute to bind a worker to one named client. In a [multi-client](#multi-client-configuration-physical-tenants) application, every `@JobWorker` method registers against **all** configured clients — it fans out across the whole `CamundaClientRegistry`, not just the primary or default one. A worker is never silently dropped for the clients it isn't explicitly bound to, since no such binding exists.
+
+In an application with only one client, this has no visible effect — there's only one client to fan out to.
+
+If you need a worker to run against only some of your configured Physical Tenants, filter inside the handler using the tenant identifier available to you at runtime, rather than relying on annotation-level scoping.
+
 #### Define the job timeout
 
 To define the job timeout, you can set the annotation (`long` in milliseconds):
@@ -1191,6 +1235,8 @@ public class CamundaLifecycleListener implements CamundaClientLifecycleAware {
   }
 }
 ```
+
+In a [multi-client](#multi-client-configuration-physical-tenants) application, one `CamundaClientCreatedSpringEvent`/`CamundaClientClosingSpringEvent` fires per configured client, each carrying that client's name — listen for these instead if you need to tell clients apart.
 
 ### Post deployment event
 
--- a/docs/apis-tools/camunda-spring-boot-starter/properties-reference.md
+++ b/docs/apis-tools/camunda-spring-boot-starter/properties-reference.md
@@ -215,7 +215,7 @@ Type: <code>string</code>
 
 <td>
 
-The physical tenant ID applied to every outgoing call: sent as the `camunda-physical-tenant` gRPC header, and used to prefix the REST base path (see `camunda.client.prefix-physical-tenant-path`) unless disabled. When `null`, the header is omitted and REST calls target the default physical tenant.
+The physical tenant ID applied to every outgoing call: sent as the `camunda-physical-tenant` gRPC header, and used to prefix the REST base path for the per-tenant API (see `camunda.client.prefix-physical-tenant-path`) unless disabled — the cluster-scoped REST API (`/cluster/v2/...`) is never prefixed. Leave this unset (`null`) rather than an empty string to target the default physical tenant; the header is then omitted and REST calls target the default physical tenant.
 
 Type: <code>string</code>
 
--- a/docs/apis-tools/camunda-spring-boot-starter/properties-reference.md
+++ b/docs/apis-tools/camunda-spring-boot-starter/properties-reference.md
@@ -215,7 +215,7 @@ Type: <code>string</code>
 
 <td>
 
-The physical tenant ID sent as the `camunda-physical-tenant` gRPC header on every outgoing call. When `null` the header is omitted.
+The physical tenant ID applied to every outgoing call: sent as the `camunda-physical-tenant` gRPC header, and used to prefix the REST base path (see `camunda.client.prefix-physical-tenant-path`) unless disabled. When `null`, the header is omitted and REST calls target the default physical tenant.
 
 Type: <code>string</code>
 
--- a/docs/apis-tools/java-client/physical-tenants.md
+++ b/docs/apis-tools/java-client/physical-tenants.md
@@ -22,7 +22,7 @@ These two mechanisms are independent and can be combined:
 | **Isolation**           | Logical only — shared engine, shared storage    | Strong — separate primary/secondary storage, separate authorization  |
 | **Targeting mechanism** | API parameter, per call or per worker           | gRPC header and REST path prefix, per client instance                |
 
-A Physical Tenant can have its own set of logical tenants. For example, Physical Tenant `teamA` and Physical Tenant `teamB` can each have a logical tenant called `foo` — `teamA`'s `foo` and `teamB`'s `foo` are completely isolated from each other.
+A Physical Tenant can have its own set of logical tenants. For example, Physical Tenant `teama` and Physical Tenant `teamb` can each have a logical tenant called `foo` — `teama`'s `foo` and `teamb`'s `foo` are completely isolated from each other.
 
 ## Target a Physical Tenant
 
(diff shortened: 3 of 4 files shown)


docs/apis-tools/java-client/physical-tenants.md
```diff
@@ -0,0 +1,179 @@
...
+
+|                         | Logical tenant                                  | Physical Tenant                                                      |
+| ----------------------- | ----------------------------------------------- | -------------------------------------------------------------------- |
+| **Configured on**       | Job worker or API call (`tenantId`/`tenantIds`) | The client itself (`physicalTenantId`)                               |
+| **Scope**               | A subdivision within one Physical Tenant        | A separate, isolated execution unit within the Orchestration Cluster |
+| **Isolation**           | Logical only — shared engine, shared storage    | Strong — separate primary/secondary storage, separate authorization  |
+| **Targeting mechanism** | API parameter, per call or per worker           | gRPC header and REST path prefix, per client instance                |
+
+A Physical Tenant can have its own set of logical tenants. For example, Physical Tenant `teamA` and Physical Tenant `teamB` can each have a logical tenant called `foo` — `teamA`'s `foo` and `teamB`'s `foo` are completely isolated from each other.
```
nicpuppa: `teamA` / `teamB` are not valid physical tenant IDs, so none of the examples on this page work as written. The ID must be lowercase alphanumeric (`[a-z0-9]+`) and at most 64 characters — the Spring starter validates it and fails startup with an `IllegalArgumentException`, mirroring the server-side `PhysicalTenantResolver` rule:

https://github.com/camunda/camunda/blob/eda44750b6fe48a0ef82b5ce58ba2a74fa66af5d/clients/camunda-spring-boot-starter/src/main/java/io/camunda/client/spring/properties/MultiCamundaClientPropertiesResolver.java#L190-L204

Could we switch to something valid and realistic throughout — e.g. `finance` / `risk`, or `teama` / `teamb`? Worth documenting the constraint itself somewhere on the page too, since it's a fail-fast at startup rather than a runtime error.

_(This applies to the whole page, not just this line.)_

docs/apis-tools/java-client/physical-tenants.md
```diff
@@ -0,0 +1,179 @@
...
+
+## Target a Physical Tenant
+
+Set the Physical Tenant on the client builder:
```
nicpuppa, line 29: Could we mention the non-builder configuration paths here as well? Both settings are also reachable via plain client properties and environment variables, and the env-var names are **different** from the Spring starter's, which is an easy trap:

- bare Java client: `CAMUNDA_PHYSICAL_TENANT_ID`, `CAMUNDA_PREFIX_PHYSICAL_TENANT_PATH` — https://github.com/camunda/camunda/blob/eda44750b6fe48a0ef82b5ce58ba2a74fa66af5d/clients/java/src/main/java/io/camunda/client/impl/CamundaClientEnvironmentVariables.java#L29-L31
- Spring starter: `CAMUNDA_CLIENT_PHYSICALTENANTID` (as listed in the properties reference)

docs/apis-tools/java-client/physical-tenants.md
```diff
@@ -0,0 +1,179 @@
...
+```java
+CamundaClient client = CamundaClient.newClientBuilder()
+    .physicalTenantId("teamA")
+    .build();
+```
+
+This does two things:
+
+- Adds the `Camunda-Physical-Tenant` gRPC metadata header to every gRPC call.
+- Prefixes the REST base path with `/physical-tenants/teamA` (for example, `https://your-cluster/physical-tenants/teamA/v2/...`).
```
nicpuppa: This is accurate for the per-tenant API, but there's an important exception missing: the **cluster-scoped** API (`/cluster/v2/...`, e.g. `newStatusRequest()`) is never physical-tenant-prefixed. The client deliberately builds it next to, rather than below, the per-tenant root, and even strips a `/physical-tenants/<id>` segment that's already present in the configured address:

https://github.com/camunda/camunda/blob/eda44750b6fe48a0ef82b5ce58ba2a74fa66af5d/clients/java/src/main/java/io/camunda/client/impl/http/HttpClientFactory.java#L174-L190

Covered by `shouldNotPrefixClusterStatusPathWithPhysicalTenant`:

https://github.com/camunda/camunda/blob/eda44750b6fe48a0ef82b5ce58ba2a74fa66af5d/clients/java/src/test/java/io/camunda/client/PhysicalTenantRestPathTest.java#L88-L100

Suggest adding a third bullet, since a reader debugging an unexpected cluster-status URL will land on this page.

docs/apis-tools/java-client/physical-tenants.md
```diff
@@ -0,0 +1,179 @@
...
+
+If your REST address already points at a tenant-specific path — for example, behind a reverse proxy that performs the routing — disable the automatic path prefix so the configured address is used as-is:
+
+```java
+CamundaClient client = CamundaClient.newClientBuilder()
+    .physicalTenantId("teamA")
+    .restAddress(URI.create("https://your-proxy/teamA/v2"))
+    .prefixPhysicalTenantPath(false)
+    .build();
+```
```
nicpuppa: This example doesn't work: `/v2` is appended by the client unconditionally, independent of `prefixPhysicalTenantPath`, so the address as written resolves to `https://your-proxy/teamA/v2/v2/...`.

https://github.com/camunda/camunda/blob/eda44750b6fe48a0ef82b5ce58ba2a74fa66af5d/clients/java/src/main/java/io/camunda/client/impl/http/HttpClientFactory.java#L151-L166

`shouldUseVerbatimRestBasePathWhenAppendDisabled` asserts the outgoing path is `/v2/topology`:

https://github.com/camunda/camunda/blob/eda44750b6fe48a0ef82b5ce58ba2a74fa66af5d/clients/java/src/test/java/io/camunda/client/PhysicalTenantRestPathTest.java#L153-L163

"the configured address is used as-is" is also slightly off — the flag suppresses only the auto-inserted `/physical-tenants/<id>` segment; `/v2` is still appended, and a trailing `/physical-tenants/<id>` is still stripped when forming the cluster-scoped base.

````suggestion
If your REST address already points at a tenant-specific path — for example, behind a reverse proxy that performs the routing — disable the automatic path prefix so the client does not insert the tenant segment a second time:

```java
CamundaClient client = CamundaClient.newClientBuilder()
    .physicalTenantId("teama")
    .restAddress(URI.create("https://your-proxy/teama"))
    .prefixPhysicalTenantPath(false)
    .build();
```
````

The client still appends the `/v2` API path to whatever address you configure, so leave it off.

File: docs/apis-tools/java-client/physical-tenants.md [hunk 5]
```diff
@@ -0,0 +1,179 @@
...
+
+`prefixPhysicalTenantPath(false)` only affects the REST base path. The gRPC header is always sent when `physicalTenantId` is set, regardless of this setting.
```
nicpuppa, line 66: Worth one more clause here: with a **blank** (whitespace-only) `physicalTenantId` the two protocols diverge in the bare client — the REST prefix is skipped, but the gRPC interceptor is still installed and sends an empty header value, because it is added whenever the configured value is non-`null`:

https://github.com/camunda/camunda/blob/eda44750b6fe48a0ef82b5ce58ba2a74fa66af5d/clients/java/src/main/java/io/camunda/client/impl/CamundaClientImpl.java#L590-L599

The Spring starter normalizes blank to `null` before it reaches the client, so this only bites people wiring the builder from their own config. A short "set it to `null`, not `\"\"`, to target the default tenant" is probably enough.

docs/apis-tools/java-client/physical-tenants.md
```diff
@@ -0,0 +1,179 @@
...
+<TabItem value="single" label="One Physical Tenant">
+
+```java
+teamAClient.newWorker()
+    .jobType("shipOrder")
+    .handler(new ShipOrderHandler())
+    .open();
+```
+
+This worker only polls Physical Tenant `teamA` — it uses the gRPC channel configured with the `teamA` header.
```
nicpuppa: "it uses the gRPC channel configured with the `teamA` header" is true only for gRPC workers. With `preferRestOverGrpc` the worker polls over REST and the targeting comes from the path prefix instead.

Suggest something protocol-neutral, e.g. "this worker only polls Physical Tenant `teama` — the targeting comes from the client, on whichever protocol it uses".

docs/apis-tools/java-client/physical-tenants.md
```diff
@@ -0,0 +1,179 @@
...
+
+```java
+Stream.of(teamAClient, teamBClient)
+    .forEach(client -> client.newWorker()
+        .jobType("shipOrder")
+        .handler(new ShipOrderHandler())
+        .open());
+```
+
+Each registration opens an independent polling loop against its own client's Physical Tenant. Inside the handler, call `activatedJob.getPhysicalTenantId()` to find out which Physical Tenant the job belongs to — this is populated on every `ActivatedJob`, whether the job arrived over gRPC or REST, and returns the default Physical Tenant's ID when none was set at job creation:
```
nicpuppa: The semantics are slightly off: the value is the physical tenant that the **job-activation request was routed to**, not one that was "set at job creation" (nothing sets it at creation). From the protocol definition:

https://github.com/camunda/camunda/blob/eda44750b6fe48a0ef82b5ce58ba2a74fa66af5d/zeebe/gateway-protocol/src/main/proto/gateway.proto#L144-L146

Suggest: "…returns the ID of the Physical Tenant the activation request was routed to, which is the default Physical Tenant when the request did not target one." The "populated on every `ActivatedJob`, gRPC or REST" part is correct.

docs/apis-tools/java-client/physical-tenants.md
```diff
@@ -0,0 +1,179 @@
...
+
+### Error handling for unauthorized tenant access
+
+At the API level, an unrecognized Physical Tenant ID returns `404 Not Found`, and a request the caller isn't authorized for returns `401 Unauthorized` (missing or invalid credentials for the tenant) or `403 Forbidden` (valid credentials, insufficient permission) — see [How to determine who can access a tenant](/self-managed/concepts/physical-tenants/authorization-model.md#how-to-determine-who-can-access-a-tenant) for the full breakdown of these status codes at the API level.
+
+:::note
+Whether the Java client surfaces these as distinct, easily-distinguishable exception types (as opposed to a generic client exception carrying the HTTP/gRPC status) is not yet confirmed. Check the status code or gRPC error details on the thrown exception until this is documented more specifically.
+:::
+
+<!-- TODO: Confirm whether the Java client exposes unauthorized-Physical-Tenant errors as a distinct exception type, or only as a generic exception carrying the HTTP/gRPC status? -->
```
nicpuppa: Both the open question and the gRPC half can be resolved from the code, so I think this section can be stated positively and the TODO dropped.

There is **no** dedicated exception type for physical-tenant errors:

- REST → `ProblemException` (extends `ClientHttpException`; `code()` for the status, `details()` for the `ProblemDetail`)
- gRPC → `ClientStatusException` (`getStatus()` / `getStatusCode()`)

And the status codes differ by protocol, which matters more than the exception type. Over gRPC, an unknown tenant is `NOT_FOUND` only when the API is **unprotected**; with auth enabled it is deliberately `UNAUTHENTICATED`, with the tenant ID not echoed, so tenant existence isn't revealed to an unauthenticated caller:

https://github.com/camunda/camunda/blob/eda44750b6fe48a0ef82b5ce58ba2a74fa66af5d/zeebe/gateway-grpc/src/main/java/io/camunda/zeebe/gateway/interceptors/impl/AuthenticationInterceptor.java#L77-L96

The REST `404` in the current text is correct and comes from the pre-security filter leaving unknown tenants to the catch-all chain. Worth documenting the asymmetry explicitly — otherwise a user seeing `UNAUTHENTICATED` will go hunting for a credentials bug when the real problem is a typo'd tenant ID.

docs/apis-tools/java-client/physical-tenants.md
```diff
@@ -0,0 +1,179 @@
...
+
+### Monitoring and logging across tenants
+
+Built-in metrics are currently client-instance-scoped, not tenant-aware: job worker metrics are tagged by job type and action, not by client name or Physical Tenant. If you run multiple client instances and need to distinguish their metrics, tag your own application-level metrics using `ActivatedJob.getPhysicalTenantId()` inside job handlers, or the client/tenant identifier you already have at hand when issuing other commands. There is no built-in tracing for multi-tenant operations today — this is left entirely to the application layer.
```
nicpuppa: Slightly stronger than stated: worker metric tags are only `type` and `action`, so the same job type across several clients doesn't just lack a tenant dimension — the series **collapse into one**:

https://github.com/camunda/camunda/blob/eda44750b6fe48a0ef82b5ce58ba2a74fa66af5d/clients/camunda-spring-boot-starter/src/main/java/io/camunda/client/metrics/JobHandlerMetrics.java#L44-L50

Same story for the job-worker actuator endpoint: its DTO carries no client name, so a multi-client app sees indistinguishable duplicate entries. Worth a sentence, since the actuator is the first place people look.

docs/apis-tools/camunda-spring-boot-starter/configuration.md
```diff
@@ -162,6 +162,42 @@ camunda:
...
+
+```yaml
+camunda:
+  clients:
+    teamA:
+      physical-tenant-id: teamA
+      primary: true
+    teamB:
+      physical-tenant-id: teamB
+```
```
nicpuppa, line 177: `physical-tenant-id: teamA` fails at startup — the value must be lowercase alphanumeric (`[a-z0-9]+`), max 64 chars, validated per client:

https://github.com/camunda/camunda/blob/eda44750b6fe48a0ef82b5ce58ba2a74fa66af5d/clients/camunda-spring-boot-starter/src/main/java/io/camunda/client/spring/properties/MultiCamundaClientPropertiesResolver.java#L190-L204

Note the map key is a free-form client name and does **not** have to match the tenant ID, which is worth saying out loud since the example makes them look coupled.

````suggestion
```yaml
camunda:
  clients:
    finance:
      physical-tenant-id: finance
      primary: true
    risk:
      physical-tenant-id: risk
```

The map key is a free-form client name; it does not have to match the Physical Tenant ID. A `physical-tenant-id` must be lowercase alphanumeric and at most 64 characters — an invalid value fails application startup.
````

File: docs/apis-tools/camunda-spring-boot-starter/configuration.md [hunk 11]
```diff
@@ -162,6 +162,42 @@ camunda:
...
+
+Each named client is a sparse overlay on top of the base `camunda.client.*` configuration — any property you don't set per client (address, auth, and so on) falls back to the shared `camunda.client.*` value. Exactly one client should be marked `primary: true`; if you configure only one client, it's implicitly primary.
```
nicpuppa: "Exactly one client should be marked `primary: true`" reads as advisory, but the real rule is stricter in one direction and looser in the other:

- **at most one** may be flagged — two or more throws `IllegalArgumentException` at startup
- if none is flagged and more than one client is configured, there is **no primary at all**: `CamundaClientRegistry#getPrimary()` throws `IllegalStateException`, and the `camundaClientConfiguration` bean isn't registered (`OnResolvablePrimaryClientCondition`)

https://github.com/camunda/camunda/blob/eda44750b6fe48a0ef82b5ce58ba2a74fa66af5d/clients/camunda-spring-boot-starter/src/main/java/io/camunda/client/spring/bean/CamundaClientRegistry.java#L50-L57

Suggest stating both failure modes — silently having no primary is the confusing one.

docs/apis-tools/camunda-spring-boot-starter/configuration.md
```diff
@@ -162,6 +162,42 @@ camunda:
...
+
+All clients must share the same `camunda.client.auth.method` — mixing authentication methods across clients in one application throws an `IllegalArgumentException` at startup. Credentials themselves (client ID, secret, and so on) can still differ per client.
```
nicpuppa: Two small corrections:

- the constraint is on each client's **resolved** `auth.method`, i.e. `camunda.clients.<name>.auth.method` overlaid on the base — not on `camunda.client.auth.method` specifically
- an unset method normalizes to `none`, so leaving it unset on one client and setting `none` explicitly on another is fine, while unset + `basic` is a conflict

https://github.com/camunda/camunda/blob/eda44750b6fe48a0ef82b5ce58ba2a74fa66af5d/clients/camunda-spring-boot-starter/src/main/java/io/camunda/client/spring/properties/MultiCamundaClientPropertiesResolver.java#L109-L140

docs/apis-tools/camunda-spring-boot-starter/configuration.md
```diff
@@ -162,6 +162,42 @@ camunda:
...
+
+If you don't configure any `camunda.clients.*` entries, your application has a single, implicitly-named `default` client — existing single-client configuration is unaffected.
+
+#### Access a named client
+
+Each named client is also available as a Spring bean, named `<name>CamundaClient` (for example, `teamACamundaClient`). To look up clients by name at runtime, inject `CamundaClientRegistry`:
```
nicpuppa: A few bean-level facts worth adding here, all part of the public surface:

- the primary client bean is `@Primary`, so a plain `@Autowired CamundaClient` still resolves
- the historical `camundaClient` bean name is preserved as an **alias** for the primary client, so `@Qualifier("camundaClient")` and `getBean("camundaClient")` keep working
- kebab-case client names are camel-cased for the bean name (`camunda.clients.team-a` → `teamACamundaClient`)
- injecting `CamundaClientRegistry` does not eagerly instantiate every client — beans are resolved lazily on first lookup

docs/apis-tools/camunda-spring-boot-starter/configuration.md
```diff
@@ -932,6 +968,14 @@ camunda:
             - foo
 ```
 
+##### Physical Tenant fan-out for multi-client applications
+
+The `@JobWorker` annotation has no attribute to bind a worker to one named client. In a [multi-client](#multi-client-configuration-physical-tenants) application, every `@JobWorker` method registers against **all** configured clients — it fans out across the whole `CamundaClientRegistry`, not just the primary or default one. A worker is never silently dropped for the clients it isn't explicitly bound to, since no such binding exists.
```
nicpuppa: The last sentence ("A worker is never silently dropped for the clients it isn't explicitly bound to, since no such binding exists") is circular — suggest deleting it. The preceding two sentences already carry the point.

docs/apis-tools/camunda-spring-boot-starter/configuration.md
```diff
@@ -932,6 +968,14 @@ camunda:
...
+
+In an application with only one client, this has no visible effect — there's only one client to fan out to.
+
+If you need a worker to run against only some of your configured Physical Tenants, filter inside the handler using the tenant identifier available to you at runtime, rather than relying on annotation-level scoping.
```
nicpuppa: This is the right advice, but it will read as a preference unless we say why annotation-level scoping isn't the only thing missing: **per-client worker configuration is ignored too**. `worker.*` / `worker.override.*` set under `camunda.clients.<name>` has no effect, because the customizer that applies those overrides is built from the single global `camunda.client.*` properties bean:

https://github.com/camunda/camunda/blob/eda44750b6fe48a0ef82b5ce58ba2a74fa66af5d/clients/camunda-spring-boot-starter/src/main/java/io/camunda/client/spring/configuration/CamundaClientAllAutoConfiguration.java#L135-L141

Otherwise the first thing a reader tries is `camunda.clients.risk.worker.override.shipOrder.enabled: false`, which silently does nothing.

docs/apis-tools/camunda-spring-boot-starter/configuration.md
```diff
@@ -1192,6 +1236,8 @@ public class CamundaLifecycleListener implements CamundaClientLifecycleAware {
 }
 ```
 
+In a [multi-client](#multi-client-configuration-physical-tenants) application, one `CamundaClientCreatedSpringEvent`/`CamundaClientClosingSpringEvent` fires per configured client, each carrying that client's name — listen for these instead if you need to tell clients apart.
```
nicpuppa, line 1251: Correct — and it raises a gap: `@JobWorker` isn't the only thing that fans out. **Startup deployment does too**: every `@Deployment` resource is deployed to *every* configured client, with one `CamundaPostDeploymentSpringEvent` published per client, and `@ClusterVariables` processing likewise runs per client (all annotation processors are driven by the per-client lifecycle events).

https://github.com/camunda/camunda/blob/eda44750b6fe48a0ef82b5ce58ba2a74fa66af5d/clients/camunda-spring-boot-starter/src/main/java/io/camunda/client/spring/annotation/processor/DeploymentAnnotationProcessor.java#L98-L106

For a multi-tenant app that's a significant behavior — your BPMN lands in every Physical Tenant, which is often what you want but should never be a surprise. Probably best placed in the "Post deployment event" and "Deploying resources on start-up" sections rather than here.

docs/apis-tools/camunda-spring-boot-starter/properties-reference.md
```diff
@@ -215,7 +215,7 @@ Type: <code>string</code>
 
 <td>
 
-The physical tenant ID sent as the `camunda-physical-tenant` gRPC header on every outgoing call. When `null` the header is omitted.
+The physical tenant ID applied to every outgoing call: sent as the `camunda-physical-tenant` gRPC header, and used to prefix the REST base path (see `camunda.client.prefix-physical-tenant-path`) unless disabled. When `null`, the header is omitted and REST calls target the default physical tenant.
```
nicpuppa: Good catch on the stale description — the REST effect was genuinely missing. Two refinements:

- the prefix applies to the per-tenant API only; the cluster-scoped API (`/cluster/v2/...`) is never prefixed
- worth noting the value should be `null`/unset rather than an empty string to target the default tenant (the starter normalizes blank to `null`, so this is mostly a note for people reading the property in isolation)

nicpuppa (changes requested):
Thanks for putting this together

docs/apis-tools/camunda-spring-boot-starter/configuration.md
christinaausley: Updated to state both failure modes explicitly -- this was written more as guidance ("you should mark one primary") rather than as a description of enforced behavior -- didn't realize the no-primary-with-multiple-clients case fails silently at the bean level rather than just being suboptimal config 👍

docs/apis-tools/camunda-spring-boot-starter/configuration.md
christinaausley: Updated -- this was based on the property's declared location (`camunda.client.auth.method`) rather than tracing how the multi-client resolver actually reads and normalizes it per client before comparing.

docs/apis-tools/camunda-spring-boot-starter/configuration.md
christinaausley: Added all four points -- this section only documented the happy-path lookup via `CamundaClientRegistry` and didn't dig into what else is on the public bean surface. Good to have the full picture here since a reader relying on `@Autowired CamundaClient` or the legacy `camundaClient` qualifier needs to know those still work.

docs/apis-tools/camunda-spring-boot-starter/configuration.md
christinaausley: Added -- the filtering advice assumed a reader would only try annotation-level scoping, not realize per-client worker config looked like a viable alternative too. Good to close that off explicitly.

docs/apis-tools/camunda-spring-boot-starter/configuration.md
christinaausley: Added notes to both sections. This page's multi-client section focused on job workers specifically, since that's where the fan-out behavior was first noticed - didn't extend the same check to deployment and cluster-variables processing, which turn out to follow the identical per-client lifecycle pattern 👍

docs/apis-tools/java-client/physical-tenants.md
christinaausley: Renamed to `teama`/`teamb` throughout, and added the format constraint (lowercase alphanumeric, max 64 chars, fails at startup) where the tenant ID is first introduced -- same root cause as the Spring Boot Starter config example, camelCase read as a natural "team" naming convention without checking it against the actual validation regex.

docs/apis-tools/java-client/physical-tenants.md
christinaausley: Added, including both sets of environment variable names. This page was written builder-first since that's the primary Java client entry point, and didn't cover the property/env-var paths that bypass it - the differing names between the bare client and the Spring starter is an easy trap, good to flag explicitly.

docs/apis-tools/java-client/physical-tenants.md
christinaausley: Fixed - this assumed `prefixPhysicalTenantPath(false)` meant "use the configured address verbatim," without confirming `/v2` is appended unconditionally downstream of that flag.

docs/apis-tools/java-client/physical-tenants.md
christinaausley: Updated -- the gRPC-channel framing was written with the default transport in mind and didn't account for `preferRestOverGrpc` changing how targeting happens.

docs/apis-tools/java-client/physical-tenants.md
christinaausley: Updated to "routed to" rather than "set at job creation." this described the value from the reader's perspective (what they'd observe) rather than the actual protocol semantics -- good catch that nothing sets it at creation.

docs/apis-tools/java-client/physical-tenants.md
christinaausley: Updated -- "lacks a tenant dimension" undersold the actual reader impact, which is that same-type jobs across clients become indistinguishable, not just unlabeled. Good addition on the actuator endpoint too, since that's the first place most people check 👍

christinaausley:
Thanks @nicpuppa -- resolved all and responded to a handful -- appreciate the feedback!

nicpuppa (approved):
lgtm, thanks

Issue #1662, opened 2023-02-01
Clarify URL/address inconsistencies across SaaS, Self-Managed, and Modeler


christinaausley:
Based on [this Slack thread](https://camunda.slack.com/archives/CKZK2E7RP/p1675082724078539), there are several inconsistencies (or rather, a lack of clarification) surrounding the topics of cluster URL and address.

See [this Google Doc](https://docs.google.com/document/d/10wZ6eMszqJHEsVbprdjaC1A-vcqoxANwaMpnwGkjE8I/edit?usp=sharing) for ongoing research and feedback before we move forward with an action plan and PR.

christinaausley:
Note several comments in [this Google Doc](https://docs.google.com/document/d/10wZ6eMszqJHEsVbprdjaC1A-vcqoxANwaMpnwGkjE8I/edit?usp=sharing) for feedback and questions still to answer.

Notably, I met with @melanie-butcher this week to discuss. She is working with her team on sifting through current usage in the strings, so we should have more answers following this mini research project in the best way to move forward, all based on our current usage.

akeller:
What's the status here? Anything left to get this to the finish line?

christinaausley:
Followed up with @melanie-butcher today to see what her team has uncovered -- will update accordingly this week with any necessary action items to get this across the finish line 👍

christinaausley:
I haven't heard from @melanie-butcher yet, so I've launched the doc for additional feedback in Slack before moving forward on glossary updates and a PR here.

melanie-butcher:
@christinaausley - The doc is good. I'm not sure exactly what you are looking for from me?

christinaausley:
@melanie-butcher Do you have any additional opinions or thoughts based on how the existing strings use these terms?

christinaausley:
I've updated this document according to my own research and some feedback.

This has been launched across engineering and teams for final feedback, so I can officially launch a PR next week.

christinaausley:
Feedback noted in Google Doc. Will add to Glossary and PR fixes as appropriate by EOD 8/18.

christinaausley:
Pushed back as a result of other priorities this week, but will review feedback and make adjustments next week 💯

christinaausley:
ALRIGHT! After lots of back and forth review, I've decided on the following action items, and terms to add to the glossary for the next update:

[ ] Web address should instead be referred to as URL.
[...]

PR #9592, opened 2026-08-11
Restore C7 Best Practices pages (#1662)

yuriy-tereshchuk-c:
Restored Camunda C7 best practices pages.

related to #1662


igpetrov:
Hi @yuriy-tereshchuk-c, I think, we need to restore it in the different repository, likely here: https://github.com/camunda/camunda-docs-manual/tree/7.24/content/user-guide.

This repo is for the https://docs.camunda.io/. The C7 docs is at https://docs.camunda.org/manual/7.24/.

Also, we need to verify whether similar docs don't exist /not migrated already.

yuriy-tereshchuk-c:
Changes applied in PR https://github.com/camunda/camunda-docs-manual/pull/1912

Issue #9522, opened 2026-07-31
Authorization and identity configuration


christinaausley:
**Why this matters**

Highest-severity active Documentation item on the support board. "Configuring authorization of components is very confusing for many of our customers" - rated High, linked to camunda-docs 8123. OIDC setup has been marked resolved twice and tickets continue to come in. Console role changes in 8.8 generated two simultaneous tickets (SUPPORT-31441, SUPPORT-31590). Certificate and Keycloak setup recurring blocker for new installations.

**Support tickets to examine**
- SUPPORT-31441 — SaaS 8.8 Console role changes
- SUPPORT-31590 — SaaS 8.8 Console role changes
- Open tickets labeled `component/Identity`

https://app.asana.com/1/384886434863350/project/1216161833611915/list/1216163466683151

https://github.com/camunda/camunda-docs/issues/8123

christinaausley:
To close this issue:

## Already strong (from the Entra + Ping playbook drafts)

- Concrete IdP-side setup + a full self-contained Helm example (#8123 items 2–3) — both drafts are unusually well fact-checked, with explicit "confirmed `<date>` against `<version>`" annotations.
- Component-level role granting (part of item 7) — covered via the Management Identity mapping-rule UI steps.

## Not yet covered by anything drafted

- **Item 4** — ongoing claim-mapping example for regular users/groups. Drafts only show the one-time bootstrap admin mapping rule, not a group-claim → role example tied to the Helm config.
- **Items 5–6** — multi-tenancy setup and its mapping rules. Absent entirely.
- **Item 7 (second half)** — individual-resource permissions (process definitions, tasks) — only component roles are covered.
- **Item 8** — a structured diagnostic workflow (JWT inspection → mapping-rule check → direct IdP auth test). Currently just scattered per-error troubleshooting entries.
- **Item 1** — explicit 8.7→8.8 architectural-change framing. "How identity works" explains the two-subsystem model as a timeless concept, not "here's what changed and why your setup broke."
- **Certificate/truststore trust configuration** (Identity ↔ Keycloak/IdP TLS) — a distinct, currently undocumented pain point per SUPPORT-24738; the support engineer's own "todo: write confluence" was never done.

## Open decisions

- #8123 asks for "one page." The emerging plan is several linked pages (concept page + per-IdP playbooks + existing canonical reference pages) — worth deciding whether that satisfies the ask, since true single-page consolidation hits the same maintenance problem already flagged internally.
[...]

PR #9601, opened 2026-08-12
Per-IdP playbooks

christinaausley:
https://github.com/camunda/camunda-docs/issues/9522

## Summary

Adds a Ping Identity playbook and brings the existing Microsoft Entra and external Keycloak guides in line with it, closing PR 2 of #9522 (per-IdP playbooks). All Helm keys and Spring Boot properties referenced below were independently verified against the `camunda-platform-helm` chart source (14.5.0–14.8.2) and `camunda/camunda`, not just the source drafts.

## Changes

**New:**
- `ping-identity.md` — PingFederate/PingOne playbook: app registration walkthrough, dual signing-key (access token vs. ID token) detection, audience/Access Token Manager setup, Ping-specific troubleshooting. Scoped to only what's Ping-specific; defers to `generic-oidc-provider.md` for shared setup steps. Added to the sidebar and category index.

**Updated:**
- `microsoft-entra.md` — `preferred_username` corrected from "optional" to required (explains the silent-403 failure mode if omitted), added `userInfoEnabled`/HTTP header-size gotchas with Helm examples, new Entra-specific troubleshooting section, `oid` vs `preferred_username` guidance for `initialClaimValue`
- `external-keycloak.md` — added troubleshooting entries (admin API connectivity check, expected first-boot pod restart, realm-exists-but-clients-missing, demo user password behavior), noted the nested vs. deprecated-flat secret form
- `generic-oidc-provider.md` — new "Handle separate access token and ID token signing keys" section covering `additionalJwkSetUris`, since this isn't Ping-specific and other providers can hit it too
- `jwt-token-claims.md` — added a Ping row to the provider claim-patterns table
- `index.md` — added Ping to the reference list

## Bug fixes in already-published content

Found while cross-referencing the drafts against real chart source — unrelated to this PR's scope but fixed along the way:

- `troubleshooting-oidc.md`: `SERVER_MAXHTTPREQUESTHEADERSIZE` → `SERVER_MAX_HTTP_REQUEST_HEADER_SIZE` (missing underscores meant Spring Boot's relaxed binding would never have picked this up)
- `external-keycloak.md`: same bug class, `IDENTITY_CLIENTID` → `IDENTITY_CLIENT_ID`
- `microsoft-entra.md`: `identityPostgresql.auth.secretKeys.userPasswordKey` was copy-pasted to point at the *admin* password secret key instead of the user one (appeared twice)

## Deliberately not carried over from the source drafts

Verification showed these were incorrect, so they were left out rather than merged in:

- Keycloak 16/WildFly `/auth` context-path branch — dead content; only Keycloak 26.x is currently supported
- Claim that `KEYCLOAK_REALM`/`IDENTITY_CLIENTID` env vars are redundant — the chart templates indicate `KEYCLOAK_REALM` is still load-bearing for custom-realm setups
- `identityPostgresql` requirement for external Keycloak — not needed when Keycloak is the provider (Identity's state lives in Keycloak itself); confirmed via the chart's own values.yaml comment
- Entra's `accessTokenAcceptedVersion` top-level manifest step — that's the deprecated Azure AD Graph manifest schema; the existing page's `api.requestedAccessTokenVersion` is correct for the current Microsoft Graph manifest format

## Follow-ups (not blocking, but worth tracking)

- **Open architecture question from Steve**: whether a "generic OIDC provider" page can ever be more than a list of provider-specific exceptions, given how much each provider's implementation actually differs. This PR keeps `generic-oidc-provider.md` as a shared base with thin provider-specific delta pages (Entra, Ping) rather than fully self-contained playbooks per provider. Worth a decision before Okta (and whatever else comes out of Steve's `#ask-identity`-style thread) gets drafted, since it determines that page's shape too.
- Okta, OneLogin, and Google Cloud Identity/Workspace are explicitly out of scope for this PR, but could be included if we want.

### DIFF
--- a/docs/self-managed/deployment/helm/configure/authentication-and-authorization/external-keycloak.md
+++ b/docs/self-managed/deployment/helm/configure/authentication-and-authorization/external-keycloak.md
@@ -142,8 +142,9 @@ global:
       realm: /realms/<realm>
       auth:
         adminUser: <keycloak_admin>
-        existingSecret: "camunda-credentials"
-        existingSecretKey: "identity-keycloak-admin-password"
+        secret:
+          existingSecret: "camunda-credentials"
+          existingSecretKey: "identity-keycloak-admin-password"
     auth:
       identity:
         clientId: <identity_client_id>
@@ -157,11 +158,15 @@ identity:
   env:
     - name: KEYCLOAK_REALM
       value: <realm>
-    - name: IDENTITY_CLIENTID
+    - name: IDENTITY_CLIENT_ID
       value: <identity_client_id>
 ```
 
 Add the section under `global.identity` to the `global` configuration you created in the previous step.
+
+:::note
+Older chart versions used a flat `global.identity.keycloak.auth.existingSecret` / `auth.existingSecretKey` form instead of the nested `auth.secret.*` shown above. The flat form still works but logs a deprecation warning — use the nested form for new deployments.
+:::
 
 The `identity.firstUser` field defines the initial user that Management Identity creates in Keycloak with full access to all Camunda components.
 By default, this user is named `demo`. To use a different name, set `identity.firstUser.username`.
@@ -201,8 +206,9 @@ global:
       realm: /realms/<realm>
       auth:
         adminUser: <keycloak_admin>
-        existingSecret: "camunda-credentials"
-        existingSecretKey: "identity-keycloak-admin-password"
+        secret:
+          existingSecret: "camunda-credentials"
+          existingSecretKey: "identity-keycloak-admin-password"
   security:
     authentication:
       method: oidc
@@ -216,7 +222,7 @@ identity:
   env:
     - name: KEYCLOAK_REALM
       value: <realm>
-    - name: IDENTITY_CLIENTID
+    - name: IDENTITY_CLIENT_ID
       value: <identity_client_id>
 
 optimize:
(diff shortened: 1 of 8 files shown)


docs/self-managed/deployment/helm/configure/authentication-and-authorization/jwt-token-claims.md
```diff
@@ -68,12 +68,13 @@ Specifies the intended audience of the token.
...
-| Okta            | `email`                         | `client_id`          | Client ID                 |
+| Provider                    | User claim                      | Client claim         | Audience default                                       |
+| --------------------------- | ------------------------------- | -------------------- | ------------------------------------------------------ |
+| Microsoft Entra             | `preferred_username`            | `azp`                | Client ID                                              |
+| Keycloak                    | `email` or `preferred_username` | `azp` or `client_id` | May require configuration                              |
+| Auth0                       | `email`                         | `client_id`          | Client ID                                              |
+| Okta                        | `email`                         | `client_id`          | Client ID                                              |
+| Ping (PingFederate/PingOne) | `sub`                           | `client_id`          | Requires an Access Token Manager configured per client |
```
StevePascoe, line 77: We actually need to split Ping into two rows here, one for PingOne and one for PingFederate. 
For PingFederate:
user claim = sub
client claim = client_id
audience default = Shared Access Token Manager for all clients 

For PingOne
user claim = sub
client claim = client_id
audience default = Per-application, via resource/scope grants

docs/self-managed/deployment/helm/configure/authentication-and-authorization/external-keycloak.md
```diff
@@ -157,12 +158,16 @@ identity:
...
       value: <realm>
-    - name: IDENTITY_CLIENTID
+    - name: IDENTITY_CLIENT_ID
       value: <identity_client_id>
 ```
 
 Add the section under `global.identity` to the `global` configuration you created in the previous step.
 
+:::note
+Older chart versions used a flat `global.identity.keycloak.auth.existingSecret` / `auth.existingSecretKey` form instead of the nested `auth.secret.*` shown above. The flat form still works but logs a deprecation warning — use the nested form for new deployments.
```
StevePascoe, line 168: We might want a note here. I think the flat form works in 8.8 and 8.9, but fails in 8.10. Check PR 5399(https://github.com/camunda/camunda-platform-helm/pull/5399)

docs/self-managed/deployment/helm/configure/authentication-and-authorization/microsoft-entra.md
StevePascoe, line 283: I think with 8.10 identityPostgresql and webModelerPostgresql isn't supported anymore. I think it's replaced with identity.externalDatabase and webModeler.restapi.externalDatabase for 8.10.

StevePascoe (changes requested):

christinaausley:
Ok -- I think I've resolved your comments @StevePascoe 😄 Feel free to add any other reviewers if you'd like anyone else to take a look?

StevePascoe (dismissed):

StevePascoe:
@Ben-Sheppard , @bmudric , @daniel-ewing @gustavo-camunda  - Would like to get a couple of you to take a look at this PR as well.

bmudric:
Reads good, but found one issue when running a machine check against the changes combined with Helm repo main branch:

> The PR tells users to configure Web Modeler's external database under camundaHub.restapi.externalDatabase. That key doesn't work:
> - It's absent entirely in chart versions 8.6–8.9 (the 14.x line the PR says it verified against).
> - In chart 8.10 (15.x, still alpha) it exists only as an empty stub ({}) — no url/username/secret sub-keys.
> - The actual working key in all these chart versions is webModeler.restapi.externalDatabase.*.
> 
> Since Helm silently ignores unknown keys, a user following this doc would set values that are never applied — Web Modeler's DB connection stays unconfigured with no error. This same mistake is repeated in 4 places:
> - generic-oidc-provider.md lines 219-226 and 269-276
> - external-keycloak.md lines 89-96
> - microsoft-entra.md lines 439-446 and 503-510

Not sure if the charts are simply still behind/incomplete, i.e. a rename of `webModeler` to `camundaHub` is pending that will make these doc changes valid.

christinaausley:

PR #9598, opened 2026-08-12
docs(app-integrations): document Enable App Integrations cluster setting for Teams notifications

yhaskell:
## Description

Documents the new **Enable App Integrations** cluster setting and what it means for Camunda for Microsoft Teams notifications.

On Camunda 8 SaaS, clusters running generation `8.9+gen13` or later deliver user task notifications to Microsoft Teams only after this setting is turned on in the cluster settings. Until then, notification rules can be created and saved but nothing is delivered. In exchange, these clusters get more than the older behavior: alongside the "a task was created" notification, cards now refresh in place as a task is assigned, completed, or canceled, and tasks assigned to you later also notify you. Clusters on earlier generations need no configuration and continue to receive the task-creation notification only.

Changes:

- **Cluster settings** — a new `Enable App Integrations` section describing what the setting does, that it is off by default, which generations offer it, and who can change it.
- **Notification rules** — a new section explaining that delivery depends on the cluster generation, with a table comparing what each delivers.
- **Getting started** — a prerequisite row for the SaaS tab.
- **Chatbot** — the automatic card update behavior is now correctly qualified rather than stated unconditionally.
- **Troubleshooting** — a new entry for notifications that are not delivered, including the message Teams shows when the setting is off.

Scope is deliberately limited to SaaS and to Microsoft Teams. Self-Managed behavior is unchanged and untouched.

Applied to `docs/` (8.10) and `versioned_docs/version-8.9/`. 8.8 needs no change: its Teams documentation already describes notifications as new-task-only, which remains accurate for that generation.

Relates to the App Integrations SaaS exporter rollout.

## When should this change go live?

- [ ] This is a bug fix, security concern, or something that needs **urgent release support**. (add `bug` or `support` label)
- [ ] This is already available but undocumented and should be released within a week. (add `available & undocumented` label)
- [x] This is on a **specific schedule** and the assignee will coordinate a release with the Documentation team. (create draft PR and/or add `hold` label)
- [ ] This is part of a scheduled **alpha or minor**. (add alpha or minor label)
- [ ] There is **no urgency** with this change (add `low prio` label)

> Added `hold` because this documents behavior tied to the `8.9+gen13` generation rollout. If that generation is already live, drop the label and release.

## PR Checklist

- [x] The commit history of this PR is cleaned up, using [`{type}(scope): {description}` commit message(s)](https://github.com/camunda/camunda-docs/blob/main/CONTRIBUTING.MD#commit-message-header-formatting)

[...]

PR #9611, opened 2026-08-13
docs(app-integrations): document Physical Tenant support and fix the removed auth.audience setting

yhaskell:
## Description

Documents Physical Tenant support in App Integrations, and corrects two errors in the published Microsoft Teams installation guide.

A single App Integrations deployment can serve every Physical Tenant of a Self-Managed orchestration cluster in 8.10. Each tenant gets its own API endpoint, Operate and Tasklist links, and notification rules, while the backend, its database, and the Teams app registration stay shared. The new page covers how a tenant is identified on each path, how to configure one, how per-tenant isolation is achieved through audiences, and what changes for an existing cluster that adopts tenants — most importantly that notification rules do not carry over and must be recreated per tenant.

It also states plainly what is not supported: a separate identity provider per Physical Tenant. App Integrations holds one identity provider for the whole deployment and distinguishes tenants by audience, which is Model B in the identity deployment models.

Two fixes to already-released 8.9 content, both backported:

- The installation guide documents an `auth.audience` setting that no longer exists. It was replaced by `auth.audiences`, and the application now refuses to start if the old key is present — so an operator following the published page today produces a configuration that cannot boot.
- A broken in-page link to Step 2.

Stacked on #9598 and should merge after it.

The `auth.audience` correction is urgent for 8.9. The Physical Tenants page targets the 8.10 minor.

## PR Checklist

- [ ] I added a redirect for a renamed or deleted page to the .htaccess file. — not applicable, no page was renamed or removed.

### DIFF
--- a/docs/components/camunda-integrations/ms-teams/ms-teams-installation.md
+++ b/docs/components/camunda-integrations/ms-teams/ms-teams-installation.md
@@ -373,6 +373,10 @@ The `urls.tasklist` field supports two formats:
 
 :::
 
+:::note
+On Camunda 8.10, a cluster can host several [Physical Tenants](/self-managed/concepts/physical-tenants/index.md). To serve them from one App Integrations deployment, add a `physicalTenants` array to the cluster. See [App Integrations and Physical Tenants](/self-managed/concepts/physical-tenants/app-integrations.md).
+:::
+
 ### Stage values
 
 The `stage` field controls the environment label.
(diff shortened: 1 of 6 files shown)


christinaausley (dismissed):
Non-blocking comments 👍 

1. Why does the auth.audiences change land in versioned_docs/version-8.9 while the Physical Tenants page is docs/ only? I assume it's because the audiences migration is an App Integrations release (independent of Camunda's version) and Physical Tenant support is 8.10 - but if so, does 8.8 need it too?

2. The warning says auth.audience is rejected at startup. Are the example YAML blocks fully migrated? The diff is hard to read here, and a leftover `audience:` in an example would now be a config that fails to start.

3. On the exporter API key warning - "the cluster key is matched before the per-tenant keys." Is that ordering guaranteed by the implementation, or incidental? It's load-bearing for the silent-no-match failure mode described.

4. "exposeDefaultTenant ... No effect unless physicalTenants is set" - does setting it without physicalTenants warn, or is it silently ignored?

yhaskell:
> 1. Why does the auth.audiences change land in versioned_docs/version-8.9 while the Physical Tenants page is docs/ only? I assume it's because the audiences migration is an App Integrations release (independent of Camunda's version) and Physical Tenant support is 8.10 - but if so, does 8.8 need it too?

This is correct — audiences migration is an App Integrations Release and pertains both to 8.9 and 8.10 docs. 
It doesn't go to 8.8, since 8.8 was SaaS only, so users couldn't configure anything there.

> 2. The warning says auth.audience is rejected at startup. Are the example YAML blocks fully migrated? The diff is hard to read here, and a leftover `audience:` in an example would now be a config that fails to start.

Yes, all the examples are migrated too.

> 3. On the exporter API key warning - "the cluster key is matched before the per-tenant keys." Is that ordering guaranteed by the implementation, or incidental? It's load-bearing for the silent-no-match failure mode described.

It's guaranteed by the implementation. Users are supposed to either use the different keys for physical tenants, or the X-Physical-Tenant-Id header will be used automatically.

> 4. "exposeDefaultTenant ... No effect unless physicalTenants is set" - does setting it without physicalTenants warn, or is it silently ignored?

It is silently ignored when there are no physical tenants configuration provided.

christinaausley (approved):

PR #9614, opened 2026-08-14
docs(release-notes): document input mapping and FEEL temporal regressions in 8.7-8.9

remcowesterhoud:
## Description

Adds two known-regression entries to the **Key changes → Engine** section of the 8.7, 8.8, and 8.9 release announcements (`*-announcements.md`), matching the existing pattern used for [Output mapping behavior change for object variables](https://github.com/camunda/camunda-docs/blob/main/docs/reference/announcements-release-notes/870/870-announcements.md#output-mapping-behavior-change):

1. **Nested input mappings can silently drop sibling fields** — input mappings that copy multiple nested fields of a variable into a target sharing the variable's own root name populate only one of those fields, leaving the others `null`, with no warning or incident raised.
2. **Chained input mappings can silently drop FEEL temporal value types** — a FEEL temporal value (`duration`, `date`, `time`, `date-time`) produced by one input mapping loses its type before a later input mapping on the same element reads it, silently evaluating to `null`.

Both regressions:

- Were introduced in `8.7.36` / `8.8.34` / `8.9.15` and are reverted in `8.7.37` / `8.8.35` / `8.9.16` (these patch versions aren't released yet — this PR is intended to be released together with them).
- Follow the established "Regression" entry format for this section: a summary table row (`Patch release | Type | Key change`), then a `#### `/`### ` entry with `**Affected versions:**` / `**Action:**` prose — no `:::warning` admonition, consistent with how existing regressions (for example, the object-variable output mapping issue) are documented here.
- Are documented identically across all mirrored copies of the affected announcements files (`docs/` and the relevant `versioned_docs/version-8.x/` trees).

No GitHub issue links are included in the docs text — the regressions were fixed by reverting the introducing change rather than a tracked forward-fix, so there's no ongoing issue to point readers at.

The technical content (root cause, affected/introduced/reverted versions, and workarounds) was provided directly by the assignee of the underlying engineering work.

## PR Checklist


docs/reference/announcements-release-notes/870/870-announcements.md
```diff
@@ -111,6 +113,32 @@ Replace is the intended long-term behavior. The merge behavior in the affected p
...
+**Action:**
+
+- **Running 8.7.36:** replace the per-field mappings with a single mapping that rebuilds the object from the source in one expression — for example, use target `foo` with source `={bar: foo.bar, baz: foo.baz}` instead of mapping `foo.bar` and `foo.baz` separately.
+- **Upgrading to 8.7.37+:** the regression is reverted, so per-field mappings on a shared root name behave correctly again. The workaround above is no longer needed, but remains harmless to keep.
+
+### Chained input mappings can silently drop FEEL temporal value types {#chained-input-mapping-temporal-type-loss}
+
+**Affected versions:** 8.7.36. Reverted in 8.7.37.
+
+Camunda 8.7.36 introduced a regression in which a FEEL temporal value (`duration`, `date`, `time`, `date-time`, and their local variants) produced by one input mapping loses its type before a later input mapping on the same element reads it.
```
eppdot: 💭 are there other FEEL types that could be affected?

docs/reference/announcements-release-notes/870/870-announcements.md
```diff
@@ -111,6 +113,32 @@ Replace is the intended long-term behavior. The merge behavior in the affected p
...
+
+For example, mapping target `duration` from source `=@"P1D"` and target `days` from source `=duration.days` on the same element produces `days` as `null`, because the temporal type of `duration` isn't preserved between the two mappings.
```
eppdot: 💭 I liked it a bit better how its explained in your "Rules Doc", where you described the input mappings in order. 

Also it might help to understand the problem better, when we name `days` differently, I got shortly confused by `duration.days` vs `days`.

docs/reference/announcements-release-notes/870/870-announcements.md
```diff
@@ -111,6 +113,32 @@ Replace is the intended long-term behavior. The merge behavior in the affected p
...
+
+**Action:**
+
+- **Running 8.7.36:** combine the two mappings into a single expression so the temporal value never crosses a mapping boundary — for example, use target `days` with source `=@"P1D".days` instead of mapping `duration` and `days` separately.
```
eppdot: 💭 define what a mapping boundary is. I think here it would also help to have the mappings as a list in the examples etc.

eppdot (changes requested):
👍 Overall it looks fine to me, but I think we can make the descriptions a bit more understandable without having all the context. For both of us it's easy to follow the description you added, but for a customer it might be hard to understand what it means.

docs/reference/announcements-release-notes/870/870-announcements.md
remcowesterhoud: No, all other types are supported properly.

docs/reference/announcements-release-notes/870/870-announcements.md
remcowesterhoud: I removed the term entirely.

docs/reference/announcements-release-notes/870/870-announcements.md
remcowesterhoud: I made the examples a bit more obvious, so it's not just a paragraph of text.

docs/reference/announcements-release-notes/870/870-announcements.md
remcowesterhoud: Made this explicit in the doc text: "Only temporal types are affected — other FEEL types (strings, numbers, booleans, lists, and contexts) behave correctly."

docs/reference/announcements-release-notes/870/870-announcements.md
```diff
@@ -111,6 +113,42 @@ Replace is the intended long-term behavior. The merge behavior in the affected p
...
 - **Upgrading to 8.7.33+:** replace behavior is restored. The same processes identified above will behave differently after upgrading. If your process was relying on earlier tasks' values being kept, you need to fix it before upgrading: instead of assigning a whole object `result = {a: 1, b: 2}`, make sure it includes all the keys it needs explicitly — or write each key separately `result.a = 1, result.b = 2`.
 
+### Nested input mappings can silently drop sibling fields {#nested-input-mapping-sibling-fields}
+
+**Affected versions:** 8.7.36. Reverted in 8.7.37.
+
+Camunda 8.7.36 introduced a regression affecting elements with two or more input mappings that write to different nested fields of the same variable. Only one of the mapped fields ends up with the expected value — the others are silently set to `null`, with no warning or incident raised.
+
+For example, an element with these two input mappings, both writing into the same `foo` variable:
```
eppdot: 🔧 here the information is missing that there is a parent `foo` otherwise it's hard to understand imho.

apart from that I like this format much better 🎖️

docs/reference/announcements-release-notes/870/870-announcements.md
```diff
@@ -111,6 +113,42 @@ Replace is the intended long-term behavior. The merge behavior in the affected p
...
+
+1. Target `foo.bar` ← source `=foo.bar`
+2. Target `foo.baz` ← source `=foo.baz`
+
+produces a local `foo` where one of `bar` or `baz` is `null` — which field depends on the order the mappings are declared.
```
eppdot: 💭 we could say here, which one will become `null` right? it well be `baz` based on the ordering above, no?

docs/reference/announcements-release-notes/870/870-announcements.md
```diff
@@ -111,6 +113,42 @@ Replace is the intended long-term behavior. The merge behavior in the affected p
...
+**Action:**
+
+- **Running 8.7.36:** combine the mappings into a single mapping that rebuilds the whole object at once — for example, target `foo` with source `={bar: foo.bar, baz: foo.baz}` — instead of mapping `foo.bar` and `foo.baz` separately.
+- **Upgrading to 8.7.37+:** the regression is reverted, so mapping individual fields of the same variable behaves correctly again. The workaround above is no longer required, but is harmless to keep.
+
+### Chained input mappings can silently drop FEEL temporal value types {#chained-input-mapping-temporal-type-loss}
+
+**Affected versions:** 8.7.36. Reverted in 8.7.37.
+
+Camunda 8.7.36 introduced a regression affecting elements with two or more input mappings where one mapping produces a FEEL temporal value (`duration`, `date`, `time`, `date-time`, or their local variants) and a later mapping on the same element reads a property from it. The temporal value loses its type before the later mapping runs, so the property access silently evaluates to `null` instead of the expected value. Only temporal types are affected — other FEEL types (strings, numbers, booleans, lists, and contexts) behave correctly.
```
eppdot: 💭 here we could say it loses it's type and becomes a string (due to variable serialization between mappings)

docs/reference/announcements-release-notes/870/870-announcements.md
```diff
@@ -111,6 +113,42 @@ Replace is the intended long-term behavior. The merge behavior in the affected p
...
+
+For example, an element with these two input mappings:
+
+1. Target `age` ← source `=@"P1D"` (a duration)
+2. Target `ageDays` ← source `=age.days`
+
+produces `ageDays` as `null`, because `age`'s duration type isn't preserved between the two mappings.
```
eppdot, line 145: ❤️

eppdot (changes requested):
👍 I like this much better and becomes easier to reason about.

Some smaller nitpicks we could still do to improve understandability 🕵️

docs/reference/announcements-release-notes/870/870-announcements.md
remcowesterhoud: Added the parent-scope variable explicitly with concrete values: "given a parent-scope variable `foo: {bar: 1, baz: 2}`".

docs/reference/announcements-release-notes/870/870-announcements.md
remcowesterhoud: Right — stated it directly now: "the local `foo` becomes `{bar: 1, baz: null}` — `baz`, mapped by the later declaration, can no longer resolve against the parent scope."

docs/reference/announcements-release-notes/870/870-announcements.md
remcowesterhoud: Added: "...loses its type — becoming a plain string due to serialization between mapping evaluations — before the later mapping runs..."

eppdot (approved):
👍 LGTM

mesellings (approved):
Approving to unblock - I will make some changes in a separate PR around wording etc

mesellings:
Follow up TW review: https://github.com/camunda/camunda-docs/pull/9615

PR #9628, opened 2026-08-17
docs(style): discourage use of em dashes

afgambin:
## Description

Discourages the use of em dashes in Camunda documentation, favoring shorter, more concise sentences instead.

- `.github/instructions/content.instructions.md`: replaces "avoid excessive use of em dashes" with a direct "avoid use of em dashes."
- `howtos/technical-writing-styleguide.md` (source of truth): flips the Em dash table row so the "Avoid" column now shows em-dash usage and the "Use" column shows the same sentences rewritten without them.

## PR Checklist


christinaausley:
Thanks for putting this together @afgambin! Directionally I don't disagree, but before I approve I'd like to understand the intent a bit better, since "avoid excessive" already covers overuse, and jumping to a full ban is a bigger call.

A few questions:

1. What's driving the move from "excessive" to "none"? Is there a specific complaint behind this (recent PRs reading as too AI-generated, em dashes have become a pretty recognizable tell there), or is it more of a general preference? Knowing the actual motivation would help make the style guide's reasoning explicit rather than just asserting the rule.
2. Is a Vale rule planned as a follow-up? I checked, and nothing in `styles/camunda/all/` currently flags em-dash usage, and this PR doesn't add one either. Without automated enforcement, this becomes a manual nitpick that different reviewers will catch inconsistently, which risks reading as arbitrary, especially on the many docs PRs that come from engineers rather than writers, who are already navigating technical review on top of style. A simple `existence` rule would make "avoid" actually stick instead of being another paragraph people skim past. But also more notifications could annoy doc authors.
3. One example could use polishing, since it'll get copied as the reference pattern:

> "Camunda was founded in 2008, at a time when established industry players were advocating a low-code approach. We believed in a developer-first approach instead."

Splitting this into two sentences and tacking "instead" on the end reads choppier than the em-dash original and repeats "approach" awkwardly. Worth tightening so it holds up as the gold-standard rewrite -- maybe a good docs example?

Happy to approve once we've aligned on intent, just want to make sure this lands as a defensible policy rather than a preference that gets applied unevenly.

mesellings (dismissed):
I'm happy with the approach of removing em dashes if the current setup means they just get through anyway - I think most people now use them as a flag for AI-generated docs, and I would (contentiously) argue that they aren't needed in tech writing anyway. For example, as @christinaausley says I think the example sentence could be improved anyway, e.g.:

> Camunda was founded in 2008 in the true spirit of Berlin, as a diverse, distributed, and global organization with "Camundis" located around the world.<br/>At Camunda, our mission is to enable organizations to design, automate, and improve their processes, no matter where they are or what they entail.<br/>Founded at a time when established players advocated a low-code approach, Camunda instead believed in a developer-first approach.

(or something like that)

Also, isn't it "Camundi" rather than "Camundo"?

afgambin:
Thanks both for the feedback @christinaausley @mesellings! Replying to your questions/comments:

1. This was motivated by repeatedly seeing agents bypass that instruction and introduce em dashes. I don’t think they are necessary in most cases, and we often end up rewording those sentences manually, requiring human intervention. I should have included this context in the PR description. We could narrow the PR to update only the agent instructions, but I would prefer to avoid creating a drift between the agent instructions and the style guide (source of truth).

2. No, it is not. I have no objection to adding one, so let me know! However, AFAIK, we don’t have Vale checks for every style guide rule, which makes sense to me. With the adoption of AI, I think Vale checks should be reserved for critical rules, while “lighter” linting can be handled by AI agents. For example, it’s fine if an em dash appears on a page, but not if a product name is written incorrectly.

3. Polishing the example makes sense to me. Updated it based on your suggestions.

christinaausley:
> No, it is not. I have no objection to adding one, so let me know! However, AFAIK, we don’t have Vale checks for every style guide rule, which makes sense to me. With the adoption of AI, I think Vale checks should be reserved for critical rules, while “lighter” linting can be handled by AI agents. For example, it’s fine if an em dash appears on a page, but not if a product name is written incorrectly.

I agree with this -- just wanted to make sure it wasn't planned 👍

christinaausley (approved):

PR #9638, opened 2026-08-18
Add test for Hub API spec bundling with RefParser

kellervater:
## Description

This PR adds a test that guards against breaking changes to the `@apidevtools/json-schema-ref-parser` dependency used by the `sync-hub-api.js` script.

The `sync-hub-api.js` script runs only from the `update-postman` workflow on release-tag pushes, not on pull requests. Without this test, a breaking bump to the `json-schema-ref-parser` dependency could reach production undetected.

The test shells out to a real `node` process to exercise the exact runtime path that production uses, since the package ships ESM-only and Jest's module pipeline cannot directly require it in the test environment.

## PR Checklist

- [x] The commit history of this PR is cleaned up
- [x] My changes are for **an upcoming minor release** and are in the appropriate directory

## Test Plan

The added test (`sync-hub-api.spec.js`) validates that `RefParser.bundle()` can successfully process both the `hubsm` and `hubsaas` OpenAPI specifications without throwing errors. The test verifies that the bundled specs contain expected properties (OpenAPI version, title, and paths).

https://claude.ai/code/session_016UZGCRa9bMN8DwjMJGpY6S

### DIFF
--- a/hacks/postman/sync-hub-api.spec.js
+++ b/hacks/postman/sync-hub-api.spec.js
@@ -4,24 +4,68 @@
 // sync-hub-api.js. That script only runs from the `update-postman` workflow,
 // which triggers on release-tag pushes, not on pull requests — so a breaking
 // bump to this dependency would otherwise reach production undetected.
+//
+// The package ships ESM-only. Jest's own module pipeline can't `require()` it
+// (no Babel config transforms node_modules here), even though plain Node can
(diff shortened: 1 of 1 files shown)


clementnero (approved):

kellervater:
cc @alexronquillo 
introduced a small test suite to validate dependency upgrades in the future. This allows us to to make PRs like https://github.com/camunda/camunda-docs/pull/9621 automergable.

PR #9647, opened 2026-08-19
docs(self-managed): document removed Web Modeler file-download executor properties

### LINKED ISSUE #9646: Document removed Web Modeler file-download.executor.* properties in the 8.9 → 8.10 Self-Managed upgrade guide
A change in Web Modeler Self-Managed removes the file-download executor tuning properties and replaces them with a single property. Because startup now **fails fast** when a removed property is still set, operators need to be told to update their configuration *before* upgrading. This should be added to the 8.9 → 8.10 Self-Managed upgrade guide.

## What changed

The download endpoints no longer use a bounded executor; they stream synchronously on the request thread, bounded by a semaphore. The executor tuning properties were therefore removed:

**Removed** (no longer bound — and if any is still present, Web Modeler now fails to start):
- `camunda.modeler.file-download.executor.core-pool-size`
- `camunda.modeler.file-download.executor.max-pool-size`
- `camunda.modeler.file-download.executor.queue-capacity`

**Replacement:**
- `camunda.modeler.file-download.max-concurrent-downloads` — default `15` (equals the previous `max-pool-size` 5 + `queue-capacity` 10)

These are typically set via environment variables on Self-Managed.

## Behavior operators will see

On startup, if any removed key is still set, Web Modeler throws and refuses to boot, with a message along the lines of:

> The file-download executor properties [camunda.modeler.file-download.executor.max-pool-size] have been removed. File-download concurrency is now controlled by camunda.modeler.file-download.max-concurrent-downloads; set that instead and remove the executor properties.

This fail-fast is intentional (a warning log was considered too easy to miss), which is exactly why the migration needs to be documented so users can adjust config ahead of time.

Note the semantics differ slightly: the old executor ran at most 5 concurrent downloads and queued up to 10; the new limit lets all 15 stream concurrently. Downloads stream rather than buffer the whole file, so peak memory stays modest.

## Suggested location

8.9 → 8.10 Self-Managed upgrade guide, under the existing Camunda Hub / Web Modeler section:
`docs/self-managed/upgrade/components/890-to-8100.md`

## Source

- Implementation: camunda/camunda-hub#27598
- Reviewer request (ThorbenLindhauer): https://github.com/camunda/camunda-hub/pull/27598#discussion_r3811427731

pablocabrera85:
Documents the removal of the `camunda.modeler.file-download.executor.*` properties and their replacement with `camunda.modeler.file-download.max-concurrent-downloads` in the 8.9 → 8.10 Self-Managed upgrade guide, so operators can update config before upgrading and avoid the fail-fast startup error.

Closes #9646

### DIFF
--- a/docs/self-managed/upgrade/components/890-to-8100.md
+++ b/docs/self-managed/upgrade/components/890-to-8100.md
@@ -271,6 +271,24 @@ DELETE /api/v2/clusters/{clusterId}
 
 Read more about dynamic cluster management in the [Camunda Hub properties reference](/self-managed/components/hub/configuration/properties.md#dynamic-cluster-management).
 
+### File download concurrency configuration
+
+The file-download endpoints (used by Web Modeler features in Camunda Hub) no longer use a bounded executor. Downloads now stream synchronously on the request thread, bounded by a semaphore, which keeps peak memory usage modest. As a result, the following executor tuning properties have been **removed**:
+
+- `camunda.modeler.file-download.executor.core-pool-size`
+- `camunda.modeler.file-download.executor.max-pool-size`
+- `camunda.modeler.file-download.executor.queue-capacity`
+
+These have been replaced by a single property:
+
+- `camunda.modeler.file-download.max-concurrent-downloads` — default `15` (equal to the previous `max-pool-size` of `5` plus `queue-capacity` of `10`)
+
+:::warning
+If any of the removed properties are still set (for example, as environment variables), Camunda Hub **fails to start**. Remove the executor properties and, if you need to tune download concurrency, set `camunda.modeler.file-download.max-concurrent-downloads` instead before upgrading.
+:::
+
+Note the semantics differ slightly from 8.9: previously, at most 5 downloads ran concurrently with up to 10 queued; now, up to 15 downloads stream concurrently.
+
 ## Elasticsearch/OpenSearch Exporter
 
 ### Default replica count changed
--- a/docs/self-managed/upgrade/components/890-to-8100.md
+++ b/docs/self-managed/upgrade/components/890-to-8100.md
@@ -273,7 +273,7 @@ Read more about dynamic cluster management in the [Camunda Hub properties refere
 
 ### File download concurrency configuration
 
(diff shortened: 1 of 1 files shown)


pablocabrera85:
For reviewers: I searched the entire docs tree (current `docs/` and all `versioned_docs/`) for these properties. The removed `camunda.modeler.file-download.executor.*` keys and the new `camunda.modeler.file-download.max-concurrent-downloads` replacement are **not referenced anywhere else** in the docs — this upgrade guide section is the only place they appear.

The existing `file-download.md` pages (current + versioned 8.7–8.9) document the feature but never mention these tuning properties, and the other `executor`/`pool-size`/`queue-capacity` hits belong to the unrelated `camunda.api.rest.executor.*` settings. So no other pages need updating.

ThorbenLindhauer (dismissed):
Looks overall fine. We could add some details here and there (e.g. besides memory it's also the number of HTTP threads that get blocked with downloads; the memory argument is a bit unsharp since the files for download are loaded outside of the critical section guarded by the semaphore). 

But since this is a very niche configuration that we have never documented before, I don't think we have be perfect here.

docs/self-managed/upgrade/components/890-to-8100.md
```diff
@@ -271,6 +271,26 @@ DELETE /api/v2/clusters/{clusterId}
...
+- `camunda.modeler.file-download.executor.core-pool-size`
+- `camunda.modeler.file-download.executor.max-pool-size`
+- `camunda.modeler.file-download.executor.queue-capacity`
+
+These have been replaced by a single property:
+
+- `camunda.modeler.file-download.max-concurrent-downloads` — default `15` (equal to the previous defaults: `max-pool-size` `5` plus `queue-capacity` `10`)
+
+:::warning
+Camunda Hub fails to start if any of the removed properties are still set (for example, as environment variables).
```
wollefitz, line 287: @pablocabrera85 Why are the properties not just ignored?

christinaausley:
Hi @pablocabrera85 let me know if you need any assistance/final review to merge this PR?

Issue #9649, opened 2026-08-19
Update Algolia crawler and index


alexronquillo:
To "rotate" our Algolia API key in Vault, we need to perform the following operations:

- [x] Create a new crawler and index
- [x] Compare the new index to the old index
- [x] Deploy the docs pointed at the new index and verify everything works as expected
- [x] Update the secret in Vault to use the new crawler's API key
- [x] Test the new crawler
- [ ] Delete the old crawler and index (analytics remain even if the crawler is deleted)

Related to: https://github.com/camunda/camunda-docs/issues/9649#issue-5193073007

PR #9650, opened 2026-08-19
dx(search): Use new v3 index

alexronquillo:
## Description

Use new algolia index.

This is step 3 in the process of [swapping our search crawler/index](https://github.com/camunda/camunda-docs/issues/9649).

- [x] This is a bug fix, security concern, or something that needs **urgent release support**. (add `bug` or `support` label) -- No security concern label (used dx)

## PR Checklist

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -689,7 +689,7 @@ module.exports = {
       // These keys are for our new standalone algolia instance!
       apiKey: "68db7725a8410eace68419c29385ad1e",
       appId: "6KYF3VMCXZ",
-      indexName: "camunda-v2",
+      indexName: "camunda-v3",
       placeholder: "Search Camunda 8 docs",
     },
     languageTabs: [


alexronquillo:
After testing, the results look the same, but have different priority, so I will look into that inconsistency before merging.

mesellings (approved):
Approving to unblock you as you see fit - but do we also need to update the crawler.config.js file? (we might not, but I know that references the old crawler as well) I think this then aligns with the search config in the actual crawler?

alexronquillo:
@mesellings That file isn't used, so it's not high priority. It's only there so someone can see it without logging into Algolia and therefore may be worth reconsidering if we even need it (Maybe it's worth it to let AI see it? Not sure). But either way, this change is _only_ for the search side, not the crawler side. I'll be updating the crawler stuff in a [future step](https://github.com/camunda/camunda-docs/issues/9649).

alexronquillo:
I updated the index search configuration in Algolia, and now the results look identical. So I'll merge this and move to the next step. @camunda/tech-writers

PR #9632, opened 2026-08-17
Tenant: Cluster Admin Operations

christinaausley:
## Description

Answers Panos/Deepthi's question about whether there's a dedicated docs page for Cluster Admin operations.

Adds `docs/components/admin/cluster-admin.md`, defining the cluster-admin role and the four `/cluster/v2/...` operations it protects (status, topology, mode, restore), including how to configure access via Basic auth or OIDC. Linked from the Admin overview and cross-linked from the Physical Tenants concept docs (`authorization-model.md`, `api-routing.md`, `configuration-reference.md`), which previously described this content in scattered, partially stale terms ("still being wired") - verified against the actual `camunda/camunda` implementation, which is further along than the existing docs suggested.

A few details (exact OIDC claim schema, precise semantics of the mode-change endpoint, GA timeline) are still pending confirmation from engineering - flagged separately in Slack, not blocking this PR.

## PR Checklist

- [x] My changes are for **an upcoming minor release** and are in the `/docs` directory.
- [x] I included my new page in the sidebar file(s).
- [ ] I added a DRI, team, or delegate as a reviewer for technical accuracy and grammar/style

PR #9662, opened 2026-08-20
Tenant issue 12: Backup restore and cluster admin

### LINKED ISSUE #8898: Tenant Issue 12: Backup, Restore, and Scaling Operations
**Review Stakeholders:** @houssain-barouni @lenaschoenburg @deepthidevaki

- [#3646 Physical Tenant - Day-2 Operations: Management API, Backup/Restore, and Cluster Topology](https://github.com/camunda/product-hub/issues/3646)

**Deliverables**
- Backup and restore operations guide covering:
  - Per-tenant backup procedure
  - Cluster-wide backup procedure
  - Backup ID semantics and uniqueness
  - Backup artifact location and naming conventions
  - Restore from backup per tenant
  - Restore full cluster
  - Authorization: ClusterAdmin required for cluster backup/restore
  - Recovery scenarios (e.g., restore after data loss)
  - Storage backend specifics (RDBMS vs ES/OS vs document store)
  - Cross-tenant backup considerations (same backup store, separate artifacts)
- Scaling operations guide covering:
  - What can be scaled per tenant (storage, replica count, etc.)
  - Cluster-wide scaling (broker count, gateway count)
  - Scaling procedure (add broker, rebalance partitions, verify readiness)
  - Restart requirements (if any)
  - Performance expectations pre/post scaling
  - Monitoring and validation during scaling

**Context We Have**
- Backup/restore exposed via REST API with ClusterAdmin authorization
- Per-tenant and cluster-wide backup endpoints
- Same backup store across tenants with separate artifacts
- RDBMS, ES/OS, and document store all support per-tenant isolation
- Restore fallback for restoring other tenants if primary storage unavailable
- Scaling semantics still being finalized

**Engineering Questions**
1. For per-tenant backup, does backup include only that tenant's data, or does it include shared configuration/state?
2. If cluster backup restored, does every tenant's data get restored, or can individual tenants be selectively restored?
3. What happens if backup from Tenant A accidentally used to restore Tenant B? Is there validation to prevent this?
4. When scaling cluster, are partitions automatically rebalanced across tenants, or does rebalancing need manual trigger?

christinaausley:
## Description

Adds Day-2 operational documentation for Physical Tenants, covering backup, restore, scaling, and topology verification. Closes #8898.

Two new pages:

- `self-managed/concepts/physical-tenants/backup-restore-scaling.md` tenant-scoped and cluster-wide procedures for backing up, restoring, and scaling a Physical Tenant plus how to verify the result.
- `components/admin/cluster-admin.md` the cluster-admin role and how to configure access via Basic auth or OIDC.

**Supersedes #9632**, which is closed in favor of this PR. The two topics cross-link in both directions, and under `onBrokenLinks: "throw"` splitting them would fail the build on whichever merged first. #9632's six files are a strict subset of the ten here.

Both pages are cross-linked from the Physical Tenants concept docs (`authorization-model.md`, `api-routing.md`, `configuration-reference.md`, `index.md`, `storage-isolation.md`, `authentication-authorization.md`), which previously described this content in scattered, partially stale terms ("still being wired", "not yet available in the current alpha"). Verified against the actual `camunda/camunda` implementation, which is further along than the existing docs suggested.

### Review feedback from #9632

All three of @deepthidevaki lingering comments are addressed. Both "physical tenant" wording corrections are applied, and the hand-maintained endpoint table (method/path/auth) is replaced by area groupings that link to the generated spec pages, with runtime backup, history backup, and exporting control now included.

Her comment also caught an error: the draft claimed those endpoints weren't yet in the API reference. They are - `get-cluster-status`, `get-cluster-topology`, `restore-as-cluster-admin`, and `change-cluster-mode-as-cluster-admin` all have generated pages, so the table links them directly. That note is removed.

Full auto-generation isn't possible yet. The backup endpoints are real in the product (`ClusterRuntimeBackupController`, `ClusterHistoryBackupController`, defined in `cluster-admin.yaml`), but that file hasn't reached this repo's generated spec, so only `mode`, `restore`, `status`, and `topology` appear under `/cluster/v2/`. Once it does, the Backup row becomes spec links like the others and the page stops needing manual upkeep. I will open a separate issue for this.

## PR Checklist


docs/self-managed/concepts/physical-tenants/backup-restore-scaling.md
```diff
@@ -0,0 +1,247 @@
...
+| `POST`   | `/`                 | Create a runtime backup |
+| `GET`    | `/?prefix={prefix}` | List runtime backups    |
+| `GET`    | `/{backupId}`       | Check backup status     |
+| `DELETE` | `/{backupId}`       | Delete a runtime backup |
+
+A backup ID is unique within a Physical Tenant. Reusing an existing ID for the same tenant returns a conflict. The same ID can be used by another Physical Tenant because each tenant has its own backup namespace.
+
+When a tenant has scheduled backups enabled, backup IDs are generated and an explicit ID is rejected. Because backup configuration is per tenant, tenants in the same cluster can be in different modes. A cluster-wide request with an explicit ID fails if any tenant uses scheduled backups, and a request without an explicit ID fails if any tenant does not. Mixed configurations must use the tenant-scoped endpoints.
+
+<!-- TODO(physical-tenants-day-2): Add concrete per-tenant backup-store configuration properties and artifact examples for runtime and history backups. Owner/reviewer: Houssain Barouni. -->
```
christinaausley: CC @houssain-barouni

docs/self-managed/concepts/physical-tenants/backup-restore-scaling.md
```diff
@@ -0,0 +1,247 @@
...
+POST /cluster/v2/restore
+```
+
+A cluster-wide restore can target every configured Physical Tenant, or a named tenant. When no `physicalTenantId` is supplied, the operation targets all tenants. A named tenant targets only that tenant.
+
+Supply per-tenant restore arguments through the request's `overrides` field. Naming a single tenant and supplying overrides in the same request is rejected, because the two express conflicting targets.
+
+Both restore endpoints accept `?dryRun=true`, which validates the request and reports the resulting plan without changing any data. Use a dry run before every production restore.
+
+<!-- TODO(physical-tenants-day-2): Confirm and document the server-side validation and error response when a backup from one Physical Tenant is submitted for another tenant. Owner/reviewers: Lena Schönburg and Deepthi Devaki. -->
```
christinaausley: CC @lenaschoenburg @deepthidevaki

docs/self-managed/concepts/physical-tenants/backup-restore-scaling.md
```diff
@@ -0,0 +1,247 @@
...
+
+If the tenant's primary storage is unavailable, use the configured backup store and the restore procedure for the affected storage backend. A fallback restore does not change the scope of the operation: restore the affected tenant through its tenant-scoped endpoint unless you intentionally need a cluster-wide recovery.
+
+### Recover the full cluster
+
+Use cluster-wide restore when multiple tenants require recovery or when the cluster must be restored to a coordinated state. A cluster-wide request can restore every configured tenant or select a named tenant with tenant-specific restore arguments. Validate every tenant's terminal state and topology before resuming normal traffic.
+
+<!-- TODO(physical-tenants-day-2): Add backend-specific restore fallback steps for primary-storage loss, including the supported RDBMS, Elasticsearch/OpenSearch, and document-store boundaries. Owner/reviewer: Houssain Barouni. -->
```
christinaausley: CC @houssain-barouni

christinaausley:

docs/components/admin/cluster-admin.md
```diff
@@ -0,0 +1,101 @@
...
+| ------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Status and topology | [Cluster status](/apis-tools/orchestration-cluster-api-rest/specifications/get-cluster-status.api.mdx), [cluster topology](/apis-tools/orchestration-cluster-api-rest/specifications/get-cluster-topology.api.mdx)                           |
+| Recovery            | [Cluster restore](/apis-tools/orchestration-cluster-api-rest/specifications/restore-as-cluster-admin.api.mdx), [cluster mode change](/apis-tools/orchestration-cluster-api-rest/specifications/change-cluster-mode-as-cluster-admin.api.mdx) |
+
+Backup and exporting control are per Physical Tenant in 8.10 and are not served under `/cluster/v2/...`. See [backup, restore, and scaling](/self-managed/concepts/physical-tenants/backup-restore-scaling.md). Scaling and multi-region failover use the actuator surface rather than this API.
```
deepthidevaki: :x: This is not correct. Backups and exporting are also cluster wide.

docs/components/admin/cluster-admin.md
```diff
@@ -0,0 +1,101 @@
...
+
+## How cluster admin differs from tenant-scoped access
+
+- Cluster admin is authenticated and authorized separately from Orchestration Cluster user sessions. It does not share credentials, roles, or mapping rules with tenant-scoped users.
+- Cluster admin credentials are verified against an isolated user store bound to `/cluster/v2/**`. A cluster admin has no route to `/physical-tenants/{physicalTenantId}/v2/...` endpoints, and no existing tenant role grants cluster admin implicitly. Configure it explicitly, as described in [Configure cluster admin access](#configure-cluster-admin-access).
+- Authorization is coarse-grained by design. Cluster admin grants access to all cluster-level operations. There are no sub-roles or partial cluster admin grants.
+
+## Cluster-wide operations
+
+Cluster admin protects the operations served under the `/cluster/v2/...` path prefix. In 8.10 these cover cluster status and topology, cluster mode changes, and restore.
```
deepthidevaki: 🔧 Not sure if we have to mention explicitly "In 8.10". Is that the convention used across docs?

docs/self-managed/concepts/physical-tenants/backup-restore-scaling.md
```diff
@@ -0,0 +1,250 @@
...
+| Runtime backup                    | `/physical-tenants/{physicalTenantId}/v2/backups/runtime` | Per tenant only        |
+| History backup                    | `/physical-tenants/{physicalTenantId}/v2/backups/history` | Per tenant only        |
+| Exporting control                 | `/physical-tenants/{physicalTenantId}/v2/exporting`       | Per tenant only        |
+| Restore                           | `/physical-tenants/{physicalTenantId}/v2/restore`         | `/cluster/v2/restore`  |
+| Topology                          | `/physical-tenants/{physicalTenantId}/v2/topology`        | `/cluster/v2/topology` |
+| Cluster mode                      | `/physical-tenants/{physicalTenantId}/v2/mode`            | `/cluster/v2/mode`     |
+| Scaling and configuration changes | `/actuator/cluster?physicalTenant={physicalTenantId}`     | `/actuator/cluster`    |
+
+Tenant-scoped routes address the selected tenant's partition group and storage. In 8.10, backup and exporting control are per tenant only: there is no cluster-wide backup endpoint. Cluster-wide restore, topology, and mode operations use the `/cluster/v2/...` prefix and require cluster-admin access.
```
deepthidevaki: > In 8.10, backup and exporting control are per tenant only: there is no cluster-wide backup endpoint

This is not correct. Cluster wide backup/exporting are available now.

docs/self-managed/concepts/physical-tenants/backup-restore-scaling.md
```diff
@@ -0,0 +1,250 @@
...
+title: "Backup, restore, and scaling"
+sidebar_label: "Backup, restore, and scaling"
+description: "Back up, restore, scale, and inspect Physical Tenants in a Camunda 8 Self-Managed cluster."
+---
+
+Learn how to back up, restore, scale, and inspect Physical Tenants after deployment.
+
+## Choose the operation scope
+
+Use a tenant-scoped endpoint when you need to affect one Physical Tenant. Use a cluster-wide endpoint when you need to affect every Physical Tenant or inspect the complete cluster.
```
deepthidevaki: The existing documentation uses the actuator endpoints for the backup/exporting and the restore app for the restore process. Instead we should update the docs to use the new rest endpoints for the backup/exporting and the restore api for the restore. There we should provide two variants of the process - backup/restore per tenant or backup/restore cluster wide using the cluster-admin api. 

PS: A page for restore api might already exists. 

@christinaausley Can we look at this topic holistically and restructure the doc?

docs/self-managed/concepts/physical-tenants/backup-restore-scaling.md
```diff
@@ -0,0 +1,250 @@
+---
+id: backup-restore-scaling
+title: "Backup, restore, and scaling"
+sidebar_label: "Backup, restore, and scaling"
+description: "Back up, restore, scale, and inspect Physical Tenants in a Camunda 8 Self-Managed cluster."
+---
+
+Learn how to back up, restore, scale, and inspect Physical Tenants after deployment.
```
deepthidevaki: I would not put backup,restore and scale in one page.

docs/self-managed/concepts/physical-tenants/backup-restore-scaling.md
```diff
@@ -0,0 +1,250 @@
...
+
+| Operation                         | Tenant-scoped surface                                     | Cluster-wide surface   |
+| --------------------------------- | --------------------------------------------------------- | ---------------------- |
+| Runtime backup                    | `/physical-tenants/{physicalTenantId}/v2/backups/runtime` | Per tenant only        |
+| History backup                    | `/physical-tenants/{physicalTenantId}/v2/backups/history` | Per tenant only        |
+| Exporting control                 | `/physical-tenants/{physicalTenantId}/v2/exporting`       | Per tenant only        |
+| Restore                           | `/physical-tenants/{physicalTenantId}/v2/restore`         | `/cluster/v2/restore`  |
+| Topology                          | `/physical-tenants/{physicalTenantId}/v2/topology`        | `/cluster/v2/topology` |
+| Cluster mode                      | `/physical-tenants/{physicalTenantId}/v2/mode`            | `/cluster/v2/mode`     |
+| Scaling and configuration changes | `/actuator/cluster?physicalTenant={physicalTenantId}`     | `/actuator/cluster`    |
```
deepthidevaki: Here again, we can restructure the existing page for scaling and restructure it to include physical tenant specific operations and cluster wide.

docs/self-managed/concepts/physical-tenants/backup-restore-scaling.md
```diff
@@ -0,0 +1,250 @@
...
+| Resource   | Permissions                           | Used for                                                    |
+| ---------- | ------------------------------------- | ----------------------------------------------------------- |
+| `BACKUP`   | `CREATE`, `READ`, `DELETE`, `RESTORE` | Creating, viewing, deleting, and restoring backups          |
+| `EXPORTER` | `PAUSE`                               | Pausing, soft-pausing, and resuming exporting during backup |
+
+Cluster-admin access is coarse-grained. It is not restricted to a single backup ID or Physical Tenant.
+
+To configure cluster-admin credentials, see [cluster admin](/components/admin/cluster-admin.md). To obtain a token, see [Orchestration Cluster API authentication](/apis-tools/orchestration-cluster-api-rest/orchestration-cluster-api-rest-authentication.md).
+
+## Back up a Physical Tenant
```
deepthidevaki: :x: As said before, it is better we link to the existing backup process documentation and adapt it.

deepthidevaki (changes requested):
@christinaausley Happy to discuss this further.

docs/components/admin/cluster-admin.md
christinaausley: Fixed. This was written against alpha4, before the cluster-wide backup and exporting controllers went out and the OpenAPI spec vendored in this repo still reflects that state (`cluster-admin.yaml` only carries `/cluster/v2/{mode,restore,status,topology}`), which is what I built the table from. I've now rebuilt it from the `@ClusterScoped` controllers in `camunda/camunda` rather than the spec.

The table now includes backup (`/cluster/v2/backups/runtime`, `/backups/runtime/state`, `/backups/history`), exporting (`/cluster/v2/exporting`, `/pause`, `/resume`), and rebalance, plus a note that each accepts an optional `physicalTenantId` to narrow the call to one tenant.

Is a spec sync planned before GA? There are no generated API reference pages for these yet, so the table uses literal paths instead of links. I think spec adjustments are owned by eng and not by tech writers, though we do review and merge the updates once they come through on the docs side.

docs/components/admin/cluster-admin.md
christinaausley: No it isn't a consistent convention. "In 8.10" appears about 70 times across `docs`, but it's meant for scoping a limitation that will change, not as a blanket qualifier. I used it here to note that the endpoint list was a snapshot of a surface still under active development, but since these are versioned docs the qualifier is redundant, and it would go stale silently when 8.11 is cut.

Removed from this sentence and from the config warning below. I kept the one in the intro note ("Cluster admin was added in 8.10 alongside Physical Tenants")... that's availability info someone upgrading from 8.9 needs.

docs/self-managed/concepts/physical-tenants/backup-restore-scaling.md
christinaausley: Done...same root cause as the `cluster-admin.md` thread: I built the scope table from the OpenAPI spec vendored in this repo, which still only carries `/cluster/v2/{mode,restore,status,topology}` and predates the cluster-wide backup and exporting controllers. Rebuilt the endpoint list from the `@ClusterScoped` controllers in `camunda/camunda` instead.

This page is also gone - per your other comments, the content is now split into the existing operational guides rather than living in a parallel Physical Tenants copy. The corrected scope table is in [back up and restore](https://github.com/camunda/camunda-docs/blob/tenant-issue-12/docs/self-managed/operational-guides/backup-restore/backup-and-restore.md), and it now shows both variants for runtime backup, runtime state, history backup, and exporting control, plus the optional `physicalTenantId` query parameter for narrowing a cluster-wide call to one tenant.

docs/self-managed/concepts/physical-tenants/backup-restore-scaling.md
christinaausley: Agreed, and restructured. To explain the original shape: the Physical Tenants section was built as a self-contained concept area (isolation, storage, routing, auth, provisioning), and I kept day-2 operations inside it so tenant-specific behavior stayed in one reviewable place while the endpoints were still moving. Now that the surface is stable that's the wrong home, since it forks the backup story in two and operators land on the general guide first.

What changed:

- [backup-and-restore.md](https://github.com/camunda/camunda-docs/blob/tenant-issue-12/docs/self-managed/operational-guides/backup-restore/backup-and-restore.md) gains a "Back up a cluster with multiple Physical Tenants" section with both variants, the authorization split, backup ID semantics, and per-backend rules. [elasticsearch/backup.md](https://github.com/camunda/camunda-docs/blob/tenant-issue-12/docs/self-managed/operational-guides/backup-restore/elasticsearch/backup.md) gets a note that its actuator calls are cluster-wide by default, can be scoped with `?physicalTenant=`, and have REST equivalents.
- for restore, yes the page exists: `in-process-restore.md`. It now has a "Restore a cluster with multiple Physical Tenants" section mapping the tenant-scoped and `/cluster/v2/...` variants of the mode to trigger to track to confirm flow, plus `overrides` and cross-tenant safety.
- scaling moved into `cluster-scaling.md` per your other comment.
- `backup-restore-scaling.md` is deleted and out of the sidebar.

One thing I deliberately did not do was rewrite the ES backup procedure end to end from actuator to REST. That page is ~800 lines and applies to every 8.10 user, not just multi-tenant ones, so switching the primary surface is a behavior-affecting change that deserves its own PR. Should the REST API become the documented default in 8.10, or stay a documented alternative until the actuator path is deprecated? Happy to open a follow-up issue either way.

docs/self-managed/concepts/physical-tenants/backup-restore-scaling.md
christinaausley: See above

docs/self-managed/concepts/physical-tenants/backup-restore-scaling.md
christinaausley: Agreed. Scaling is now in [cluster-scaling.md](https://github.com/camunda/camunda-docs/blob/tenant-issue-12/docs/self-managed/components/orchestration-cluster/zeebe/operations/cluster-scaling.md) rather than duplicated here. The original placement was so tenant-specific behavior stayed in one reviewable place while the endpoints were still moving, but that's no longer a good reason to keep it separate.

The new "Scale a cluster with multiple Physical Tenants" section sits right after **Considerations**, so readers hit the scope question before the procedure. It covers which dimension is tenant-scoped vs. cluster-wide (partition count, broker count, replication factor, partition join/leave/priority, routing state, purge), a worked `PATCH /actuator/cluster?physicalTenant=` example, the `400`/`404` rejection cases, and the fact that partition ids restart at `1` per tenant.

One nuance you should maybe check to confirm: I documented that a partition-count change **without** `physicalTenant` targets the default tenant only, while `GET /actuator/cluster` without the parameter reports every tenant. That's from `shouldApplyPatchClusterToOnlyTheDefaultTenantWhenUnscoped` plus ADR 003 D3. Is that read/write asymmetry intentional and staying for GA? It seems like the kind of thing that gets an operator to scale the wrong tenant.

docs/self-managed/concepts/physical-tenants/backup-restore-scaling.md
christinaausley: Agreed...this is now the approach throughout. There's no parallel Physical Tenants backup procedure anymore. `backup-restore-scaling.md` is deleted, and the tenant-specific detail was folded into the existing guides: [backup-and-restore.md](https://github.com/camunda/camunda-docs/blob/tenant-issue-12/docs/self-managed/operational-guides/backup-restore/backup-and-restore.md) for scope, authorization, backup ID semantics, and per-backend rules, and a note in [elasticsearch/backup.md](https://github.com/camunda/camunda-docs/blob/tenant-issue-12/docs/self-managed/operational-guides/backup-restore/elasticsearch/backup.md) on scoping its actuator calls.

The cross-links in `storage-isolation.md`, `authorization-model.md`, and `physical-tenants/index.md` now point into those guides instead. I also dropped the duplicated authorization section as`authorization-model.md` already carries the `BACKUP`/`EXPORTER` permission table... I've extended that table to cover history backup and restore rather than restating it in a second place.

docs/self-managed/components/orchestration-cluster/zeebe/operations/cluster-scaling.md
```diff
@@ -28,6 +28,57 @@ How brokers are identified and scaled depends on whether the cluster is [zone-aw
...
+| ---------------------------------- | ------------ | --------------------------------------------------------------------------------------------------- |
+| Partition count                    | Per tenant   | `PATCH /actuator/cluster?physicalTenant={physicalTenantId}`                                         |
+| Broker count                       | Cluster-wide | `PATCH /actuator/cluster` or `POST /actuator/cluster/brokers`, without a `physicalTenant` parameter |
+| Replication factor                 | Cluster-wide | `PATCH /actuator/cluster`, without a `physicalTenant` parameter                                     |
+| Partition join, leave, or priority | Per tenant   | `POST` or `DELETE /actuator/cluster/brokers/{brokerId}/partitions/{partitionId}?physicalTenant=`    |
+| Routing state                      | Per tenant   | `PATCH /actuator/cluster/routing-state?physicalTenant={physicalTenantId}`                           |
+| Purge                              | Both         | `POST /actuator/cluster/purge`, optionally scoped with `?physicalTenant={physicalTenantId}`         |
```
deepthidevaki: :x: purge does not belong to scaling. Is there a different page where purge is documented?

docs/self-managed/concepts/physical-tenants/backup-restore-scaling.md
deepthidevaki: > One thing I deliberately did not do was rewrite the ES backup procedure end to end from actuator to REST. That page is ~800 lines and applies to every 8.10 user, not just multi-tenant ones, so switching the primary surface is a behavior-affecting change that deserves its own PR. Should the REST API become the documented default in 8.10, or stay a documented alternative until the actuator path is deprecated? Happy to open a follow-up issue either way.

Yes. I think we should rewrite the existing backup process with the new rest endpoints, and provide the actuator endpoints as an alternative and backward compatible option. A new customer should use the rest endpoints. An existing user can easily find the documentation with the actuator endpoints, and should be encouraged to migrate to the rest endpoint. The deprecation of the actuator endpoint is not decided, but it is likely that we move all similar endpoints to rest and deprecate actuator endpoints eventually.

A follow up issue + PR would be great. Thanks.

deepthidevaki (dismissed):
Thanks.

christinaausley:

christinaausley:
Thanks for the review @deepthidevaki. Updated...

- removed the `Purge` row from the scaling dimensions table and moved the tenant-scoping guidance to `data-purge.md` where purge is actually documented. That page previously called purge "a cluster-wide, asynchronous operation" with no tenant qualifier, which IMO reads as reassuring when it actually means every tenant's data - it now says so explicitly and documents `?physicalTenant={physicalTenantId}`.

One question still open from the scaling thread - not blocking this PR:

`PATCH /actuator/cluster` without `physicalTenant` targets the default tenant only, while `GET /actuator/cluster` without it reports every tenant. I documented the asymmetry in a note (from `shouldApplyPatchClusterToOnlyTheDefaultTenantWhenUnscoped` plus ADR 003 D3), but if it isn't deliberate it seems like a good way for an operator to scale the wrong tenant. Worth a product issue rather than a docs note? Whatever you think is best.

christinaausley:

deepthidevaki:
> One question still open from the scaling thread - not blocking this PR:
> 
> `PATCH /actuator/cluster` without `physicalTenant` targets the default tenant only, while `GET /actuator/cluster` without it reports every tenant. I documented the asymmetry in a note (from `shouldApplyPatchClusterToOnlyTheDefaultTenantWhenUnscoped` plus ADR 003 D3), but if it isn't deliberate it seems like a good way for an operator to scale the wrong tenant. Worth a product issue rather than a docs note? Whatever you think is best.

This is a deliberate decision for keeping backwards compatibility of the patch api.

deepthidevaki (approved):

Issue #9629, opened 2026-08-17
Document project/process application migration changes for SaaS


alexronquillo:
To support the upcoming Web Modeler data structure changes and migration, we need to make the following documentation changes:

## Web Modeler API v1

Process applications aren't explicitly referenced in any v1 endpoints, they're implicitly referenced as folders. With the upcoming changes:

- Process applications are no longer surfaced as folders anywhere in the v1 APIs.
- We're introducing new parameters/endpoints (implementation not yet finalized) to make process applications accessible.

> **Note:** The v1 reference docs are generated from the OpenAPI spec.

## Web Modeler UI

We're updating the SaaS UI to (among other things): 

- Prevent users from creating files and folders at the top level of a project. IDP applications can still be created at the project level.
- Prevent users from creating process applications as a child of anything other than a project.

8.10 docs will reference the new workspaces/projects UI for Camunda Hub, but the 8.9 docs will need to reflect the SaaS changes in some way (implementation not yet finalized) that accounts for the fact that 8.9 Self-Managed users will use the same documentation and not see the same UI.

> **Note:** The SaaS UI is temporary until 8.10 releases.

## Acceptance criteria

- [x] Create a new migration guide highlighting the v1 API changes and reasons behind them.
  - [x] Include in 8.9 docs targeting SaaS only.
  - [x] Include in 8.10 docs targeting SaaS and Self-Managed.
  - [x] Changes:
    - [x] All public API v1 changes listed [here](https://github.com/camunda/camunda-hub/issues/25831#issuecomment-5319049300). 🟡
    - [x] All changes listed [here](https://github.com/camunda/camunda-hub/issues/27564). 
- [x] Update the 8.9 documentation to account for UI changes **(SaaS-only, when necessary)**
  - [x] [Build forms with Modeler](https://docs.camunda.io/docs/components/modeler/forms/utilizing-forms/)
    - [x] Instruct users to create a form _within a process application_ rather than at the top level of a project.
    - [x] Update/remove screenshot that shows a form at the top level of a project.
  - [x] [Generate a BPMN element template](https://docs.camunda.io/docs/components/modeler/web-modeler/element-templates/element-template-generator/#generate-a-bpmn-element-template)
    - [x] Change "Camunda recommends storing element templates in root project folders, not process applications". No longer possible.
    - [x] Update instructions.
[...]

PR #9690, opened 2026-08-24
docs(hub): add references to new process application fields and endpo…

alexronquillo:
## Description

Process applications are no longer treated as folders in Web Modeler API v1. The changes to address this behavior change is broken into two parts:

- Remove implicit references to process applications in the existing endpoints
- Add new endpoints, so users can access process application data explicitly

This PR addresses the second part.

> **Note:** This is not ready for review, because it hasn't been implemented yet and is subject to change.

Related to: https://github.com/camunda/camunda-docs/issues/9629

## PR Checklist

### DIFF
--- a/docs/apis-tools/migration-manuals/web-modeler-v1-apis.md
+++ b/docs/apis-tools/migration-manuals/web-modeler-v1-apis.md
@@ -46,7 +46,7 @@ Affected endpoints:
 - `GET /api/v1/folders/{folderId}`
 - `PATCH /api/v1/folders/{folderId}`
 
-Instead of passing the process application ID to the folders endpoints, you can use the new process application endpoints:
+Instead of passing the process application ID to the folders endpoints, use the new process application endpoints:
 
 - `DELETE /api/v1/processApplications/{processApplicationId}`
 - `GET /api/v1/processApplications/{processApplicationId}`
@@ -60,7 +60,7 @@ GET /api/v1/processApplications/e005e49a-dce8-42ee-b0db-30b1d5555ebd  # must be 
 
 ### Parent IDs reference folders
 
-In folders API requests, you'll receive a `404 NOT FOUND` if you pass a process application ID as the `parentId`. Previously, you could pass a process application's ID when writing data. Instead of passing the process application ID as the `parentId`, you can use the new `processApplicationId` field to place the folder in a process application.
+In folders API requests, you'll receive a `404 NOT FOUND` if you pass a process application ID as the `parentId`. Previously, you could pass a process application's ID when writing data. Use the new `processApplicationId` field to specify the target process application and, optionally, the `parentId` field to specify a target folder.
 
 Example:
 
@@ -69,7 +69,7 @@ POST /api/v1/folders
 {
   "name": "Nested folder",
   "projectId": "fb928277-6268-44bb-b3e6-1925fa730ecf",
-  "parentId": "1ef492f5-7ddc-43a7-b5e2-f5ad5c14b676"  # must be a folder, not a process application
+  "parentId": "1ef492f5-7ddc-43a7-b5e2-f5ad5c14b676",  # must be a folder, not a process application
   "processApplicationId": "e005e49a-dce8-42ee-b0db-30b1d5555ebd"  # new process application field
 }
 ```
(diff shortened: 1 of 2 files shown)


christinaausley (dismissed):
Couple comments 😄 

1. POST /api/v1/files example is missing a comma after processApplicationId I think? That example sends both folderId and processApplicationId, but the wording says to use one instead of the other. Can both be sent? If so which wins?

2. "Parent IDs reference folders" lists DELETE and GET as affected, but parentId is a body field and neither has a body. Should that list be PATCH + POST only?

3. Only DELETE/GET/PATCH /api/v1/processApplications are shown. Is there a POST to create one? should these get their own section rather than living inside a folders subsection?

4. processApplications entries show only id and name, unlike the full folder/file objects in the same response. Accurate or just trimmed to keep things clean?

alexronquillo:
> POST /api/v1/files example is missing a comma after processApplicationId I think? That example sends both folderId and processApplicationId, but the wording says to use one instead of the other. Can both be sent? If so which wins?

The implementation isn't finalized, but I believe you can send both (this document won't be finalized until we get the full implementation and test). A file can be part of both a process application (must be) and a folder (might be). I can reword the sentence that seems to suggest you can't send both.

> "Parent IDs reference folders" lists DELETE and GET as affected, but parentId is a body field and neither has a body. Should that list be PATCH + POST only?

Only PATH and POST [are mentioned](https://github.com/camunda/camunda-docs/pull/9690/changes#diff-e85f42db69180bd87dfeed0723a1539c1d23fd5dac269ce3fda9987786d7363dR79).

> Only DELETE/GET/PATCH /api/v1/processApplications are shown. Is there a POST to create one? should these get their own section rather than living inside a folders subsection?

No, a create endpoint is out of scope, because there was no way to do this implicitly using the folders API before. I decided to embed them in the folders section because it supports the way we're thinking about this, namely: this functionality was there before, but it was implicitly available with the folders API. Now, we're removing the implicit options and making them explicit. Open to counterarguments, though.

> processApplications entries show only id and name, unlike the full folder/file objects in the same response. Accurate or just trimmed to keep things clean?

This hasn't been finalized yet, so I'll come back to these when I test and can use real response examples. In the meantime, I can at least add the timestamps, as those should be there.

docs/apis-tools/migration-manuals/web-modeler-v1-apis.md
```diff
@@ -188,6 +219,21 @@ Affected endpoints:
...
+```shell
+POST /api/v1/files/search
+{
+  "filter": {
+    "processApplicationId": "e005e49a-dce8-42ee-b0db-30b1d5555ebd"
+  }
+}
+```
+
 ## Projects API
```
ThorbenLindhauer, line 237: For the examples in these sections, the `content.files` and `content.folders` items are gaining a `processApplicationId` property (similar to the responses we have adjusted above).

And let's make it consistent then: they should reference a process application that is included in `content.processApplications`.

docs/apis-tools/migration-manuals/web-modeler-v1-apis.md
```diff
@@ -188,6 +219,21 @@ Affected endpoints:
...
 
 You can no longer read or write process application IDs in any projects API context.
```
ThorbenLindhauer: We should update that sentence, similar to what we did for the other API sections.

ThorbenLindhauer (changes requested):
Thanks, looks overall good. I have two findings below. And let's wait with the merge until https://github.com/camunda/camunda-hub/issues/27564 is completed (please let me know if you'd like me to merge then when you are finishing your day).

docs/apis-tools/migration-manuals/web-modeler-v1-apis.md
alexronquillo: Do you mean something like "You can no longer access process applications as folders in any projects API context"?

ThorbenLindhauer:
```suggestion
You can no longer access process applications as folders in any projects API context. Instead, process applications are now explicitly represented.
```

Maybe like this. I realize now that we don't have the second sentence for the other API introductions either. It think it would be a good addition since otherwise reading this sentence in isolation (e.g. because you have been linked to this subsection) suggests we have only removed things.

alexronquillo: good point

ThorbenLindhauer (approved):
Looks good, thanks!

PR #9703, opened 2026-08-26
docs(concepts): document broker secret resolution and its effect on job activation

### LINKED ISSUE #60332: [Secret Resolution] Docs: broker resolution path and job workers
Parent: #56591

### Description

Document the broker resolution path and its effect on job workers.

### Motivation

The activation semantics are surprising and operationally visible: a job can be skipped, dropped, or turned into an incident because of secret resolution, and none of that is currently documented. There is also an entire configuration namespace with no documentation at all.

### Acceptance criteria

- [ ] Documented that the broker resolves references asynchronously ahead of activation, via a scheduler, into the shared cache
- [ ] Configuration documented under `camunda.processing.engine.secrets.*`: `interval` (default `5s`), `retry-max-attempts` (default `3`), `retry-initial-delay` (default `1s`), `retry-max-delay` (default `30s`), `retry-backoff-factor` (default `2`), `batch-resolution-limit` (default `20`)
- [ ] Documented that during batch collection a job whose references are not yet cached is skipped without consuming a batch slot, so jobs behind it can still activate. It is not blocked and not failed
- [ ] Documented that a job whose injected values would push the response past the maximum message size is dropped from the activation
- [ ] Documented that a job whose injection fails is dropped and reported for an incident, and what the user sees
- [ ] Both long polling and job push covered, and stated that job push injects the same values
- [ ] Documented that state, records and logs retain the placeholder text and only the activation response carries values
- [ ] `docs/components/concepts/job-workers.md` and `docs/components/concepts/outbound-connectors-job-workers.md` updated or cross-linked

### Breakdown

- Write the job worker facing content
- Add the `camunda.processing.engine.secrets.*` properties to the configuration reference, coordinating with D5 so the two partials do not conflict

### Notes

Source of truth is `JobSecretInjector`, `JobSecretLookup`, `SecretResolutionScheduler` and `EngineSecrets.java` in `camunda/camunda`.

HeleneW-dot:
closes [#60332](https://github.com/camunda/camunda/issues/60332)

## Description

- Adds a draft concepts page covering the background resolution scheduler, the long polling and job push activation semantics, and where resolved values do and do not appear. 
- Adds the camunda.processing.engine.secrets.* properties to the processing partial

Additional changes:
Restores the `_processing.md` import that `properties.md` lost in a February merge, without which the new secret properties would render nowhere.

Note for review: the new page is intentionally **not** added to `sidebars.js`. Its title, file name, placement, and whether it needs an alpha marker are owned by [#60326](https://github.com/camunda/camunda/issues/60326), which is still open. The page builds and passes link validation, but is not reachable from navigation until that decision lands and the sidebar entry is added in a follow-up.

## PR Checklist


docs/components/concepts/secret-resolution-and-job-activation.md
```diff
@@ -0,0 +1,199 @@
...
+later variable merge overwrote it, or if the variables cannot be read at all, the job is not
+activated and an incident is raised. The incident also takes the job out of activation until the
+incident is resolved, so the same failing injection is not retried on every activation.
+
+For how to inspect and resolve either incident, see [incidents](incidents.md).
+
+<!-- The incident messages, the diagnosis steps, and the operator runbook are owned by
+camunda/camunda#60964. Link into that content from here once it lands, instead of restating it. -->
+
+## Tune the resolution scheduler
```
ajanoni-camunda, line 122: This section tunes `camunda.processing.engine.secrets.*` but never mentions the cache config that governs the cache lifecycle this whole page is built on: `camunda.secrets.cache.ttl` (default `20m`) and `camunda.secrets.cache.max-size` (default `1000`). Those are owned by camunda/camunda#60331, so they don't belong here in full, but the behavioral consequence does: a resolved value is cached only for the TTL (or until it's evicted at `max-size`), so a job re-activated after the entry expires (e.g. after a timeout/retry) finds the reference uncached again, re-parks, and re-resolves. Right now nothing on the page ties the re-park/re-resolve behavior to the cache lifetime.

Suggest either (a) a deferral comment like the ones already used for #60963/#60964, e.g. `<!-- Cache config (camunda.secrets.cache.ttl, max-size) is owned by camunda/camunda#60331. Link its cache section here once it lands. -->`, and/or (b) one sentence here noting a resolved value stays cached only for `camunda.secrets.cache.ttl`, cross-linking #60331 when it lands.

ajanoni-camunda (dismissed):
LGTM. Just one small comment.

docs/components/concepts/secret-resolution-and-job-activation.md
HeleneW-dot: Adjusted 👍

HeleneW-dot:
@giorgionaps This is ready for tech writer review now

ajanoni-camunda (dismissed):
✅ LGTM

giorgionaps (approved):
@HeleneW-dot thanks for making the changes! I've committed some wording corrections, otherwise, lgtm!

PR #9709, opened 2026-08-26
Document the job lease concept

### LINKED ISSUE #56097: Document job lease concept
## Description

Document the job lease concept in the user-facing docs — probably under https://docs.camunda.io/docs/next/components/concepts/job-workers/.

Points to cover:

- **What a lease is**: an opt-in, opaque token returned per activated job when a worker sets `withLease: true`. It identifies a specific activation of a job.
- **How to use it**: request with `withLease`, read `leaseToken` off the activated job, and pass the matching lease back on complete / fail / throw-error. Optionally include it on update-timeout (and other job updates) to verify you still own the job — a mismatch rejection is a definitive signal the activation was superseded.
- **Enforcement**: complete, fail, and throw-error on a leased job require the matching lease; a missing or mismatched lease is rejected with `INVALID_STATE`. Property updates (update-timeout, update-retries, priority) never require a lease — operators and batch operations are unaffected — but validate a lease if one is supplied (mismatch is rejected).
- **Rejection handling**: a lease-mismatch rejection means the worker's activation was superseded. It is **non-retryable** — drop the work; expected to be logged at debug level, not treated as an error.
- **Permanently lease-only (important caveat)**: once any worker leases a job, that job is **never again served to a non-leasing worker of the same type**. Non-leasing polls silently skip it (no error surface). If a job type is served by a mix of leasing and non-leasing workers, users must understand this consequence before opting in.
- **Fleet guidance**: recommend a homogeneous fleet per job type (all workers leasing, or none). Mixed fleets are supported but should be transitional (e.g. during rollout); before rolling a leasing worker fleet back to a non-leasing version, drain in-flight leased jobs first.
- **Streaming**: leases are supported on the job streaming (push) path; a leased job only matches streams opened with `withLease`. Leasing activations (poll or stream) match all jobs — an unleased job pushed to a leasing stream becomes leased; non-leasing activations match only unleased jobs.
- **Capability check (version skew)**: after activating with `withLease: true`, check `leaseToken != null` **per job**. Null means the activation is not fenced — the cluster does not support leases, or (during a rolling upgrade) this job's partition leader is not yet upgraded. Consumers that **require** the fence must not process the job: fail it with a retry backoff and a clear message, **preserving retries** (set retries to the job's current value so the fail does not decrement them — otherwise a long rolling upgrade burns the job into an incident); the retry converges once the partition is upgraded. Consumers that merely prefer fencing may proceed. (Over REST an old cluster rejects the unknown `withLease` field with a 400; over gRPC the field is silently ignored — hence the null check.) The same contract belongs in the Java client Javadoc for `withLease`. Against a cluster that permanently lacks lease support this manifests as a visible fail/backoff loop — intentionally observable, rather than silently running unfenced.
- **Worker rollback**: the permanently-lease-only consequence includes rollbacks — if a leasing worker deployment is rolled back to a non-leasing version, jobs leased in the interim are never served to the rolled-back workers, and those process instances appear stuck with no error surface (the lease-skip metric from #54845 is the operator signal). There is currently no un-lease operation (deferred — not needed yet). Rescue options: redeploy any worker for the job type with `withLease: true` to drain the leased jobs, or use process-instance modification (terminate the stuck element and activate it again, producing a fresh unleased job).

## Notes

- **Epic issue**: https://github.com/camunda/product-hub/issues/3462

korthout:
## Description

Documents the job lease concept, introduced in https://github.com/camunda/camunda/issues/54840.

- Adds a new **Job leasing** section to `docs/components/concepts/job-workers.md`, covering: what a lease is, how to request and pass one, enforcement and rejection handling, the permanent lease-only caveat (including the rollback scenario), fleet guidance, and how to check lease support during a rolling upgrade.
- Cross-links the new section from the existing **Job streaming** section.
- Syncs the proto snippets in `docs/apis-tools/zeebe-api/gateway-service.md` with the `withLease`/`leaseToken` fields already present in `gateway.proto` on `camunda/camunda@main`, across the 8 affected RPCs (`StreamActivatedJobs`, `ActivateJobs`, `CompleteJob`, `FailJob`, `ThrowError`, `UpdateJobRetries`, `UpdateJobTimeout`, `UpdateJobPriority`).

This PR is stacked on top of [#9710](https://github.com/camunda/camunda-docs/pull/9710), which documents the `UpdateJobPriority` RPC itself (previously undocumented, unrelated to leasing). Merge #9710 first.

Out of scope: the REST API spec pages already document `withLease`/`leaseToken` (auto-generated).

closes https://github.com/camunda/camunda/issues/56097

## PR Checklist

### DIFF
--- a/docs/apis-tools/zeebe-api/gateway-service.md
+++ b/docs/apis-tools/zeebe-api/gateway-service.md
@@ -61,6 +61,15 @@ message ActivateJobsRequest {
   int64 requestTimeout = 6;
   // a list of IDs of tenants for which to activate jobs
   repeated string tenantIds = 7;
+  // whether to activate the jobs with a lease; when true, each activated job is assigned a
+  // distinct, opaque lease token, returned as ActivatedJob.leaseToken. The lease fences the
+  // complete, fail, and throw-error commands against a superseded activation of the same job
+  // (e.g. after the job timed out or failed and was re-activated by another worker): a command
+  // carrying a stale lease token is rejected rather than racing with the newer activation. Once
+  // a job has been activated with a lease, it is served only to leasing workers of that job
+  // type; a homogeneous fleet per job type is recommended. Defaults to false, which activates
+  // jobs without a lease.
+  bool withLease = 9;
 }
 ```
 
@@ -134,6 +143,9 @@ message ActivatedJob {
   JobKind kind = 15;
   // the listener event type of the job.
   ListenerEventType listenerEventType = 16;
+  // the lease token identifying this activation; unset when the job was activated without a
+  // lease
+  optional string leaseToken = 21;
 }
 ```
 
@@ -247,6 +259,13 @@ message CompleteJobRequest {
   // The result of the completed job as determined by the worker.
   // This functionality is currently supported only by user task listeners
   optional JobResult result = 3;
+  // the token identifying a leased job's activation, obtained from ActivatedJob.leaseToken.
+  // For a leased job, the matching token must be supplied to prove the command comes from the
+  // worker that holds the current lease; a command with no token is rejected. A command carrying
+  // a stale token is likewise rejected, fencing the job against a superseded activation (e.g.
+  // after the job timed out or failed and was re-activated by another worker). A job that was
+  // activated without a lease requires no token.
+  optional string leaseToken = 4;
 }
 
 message JobResult{
@@ -933,6 +952,13 @@ message FailJobRequest {
   // "b" respectively, with their associated values. [{ "a": 1, "b": 2 }] would not be a
   // valid argument, as the root of the JSON document is an array and not an object.
   string variables = 5;
+  // the token identifying a leased job's activation, obtained from ActivatedJob.leaseToken.
+  // For a leased job, the matching token must be supplied to prove the command comes from the
+  // worker that holds the current lease; a command with no token is rejected. A command carrying
+  // a stale token is likewise rejected, fencing the job against a superseded activation (e.g.
+  // after the job timed out or failed and was re-activated by another worker). A job that was
+  // activated without a lease requires no token.
+  optional string leaseToken = 6;
 }
 ```
 
@@ -1279,6 +1305,13 @@ message StreamActivatedJobsRequest {
   repeated string fetchVariable = 5;
   // a list of identifiers of tenants for which to stream jobs
   repeated string tenantIds = 6;
+  // whether to stream jobs with a lease; when true, each job pushed on this stream is
+  // assigned a distinct, opaque lease token, returned as ActivatedJob.leaseToken. The lease
+  // fences the complete, fail, and throw-error commands against a superseded activation of
+  // the same job (e.g. after the job timed out or failed and was re-activated by another
+  // worker): a command carrying a stale lease token is rejected rather than racing with the
+  // newer activation. Defaults to false, which pushes jobs without a lease.
+  bool withLease = 8;
 }
 ```
 
@@ -1340,6 +1373,9 @@ message ActivatedJob {
   JobKind kind = 15;
   // the listener event type of the job.
   ListenerEventType listenerEventType = 16;
+  // the lease token identifying this activation; unset when the job was activated without a
+  // lease
+  optional string leaseToken = 21;
 }
 ```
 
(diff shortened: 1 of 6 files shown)


docs/apis-tools/zeebe-api/gateway-service.md
```diff
@@ -61,6 +61,15 @@ message ActivateJobsRequest {
   int64 requestTimeout = 6;
   // a list of IDs of tenants for which to activate jobs
   repeated string tenantIds = 7;
+  // whether to activate the jobs with a lease; when true, each activated job is assigned a
```
ce-dmelnych, line 64: 🔧 This is unrelated to the `jobLease` changes, but we could also add the missing field:
```proto
// the tenant filtering strategy - determines whether to use provided tenant IDs or assigned tenant IDs
TenantFilter tenantFilter = 8;
```

This would complete the documentation for ActivateJobsRequest and avoid the gap between fields 7 and 9.

ce-dmelnych: Or, we could create a dedicated documentation issue to review and update this page, as it looks like the documentation for other messages is also lagging behind the implementation.

For example, `ActivatedJob` is missing documentation for the following fields:
```proto
  // the listener event type of the job.
  ListenerEventType listenerEventType = 17;
  // the list of tags
  repeated string tags = 18;
  // the priority of the job; higher values indicate higher priority
  int32 priority = 19;
  // the business ID of the owning process instance; null when the instance has none
  string businessId = 20;
  //... and
  // the id of the physical tenant that the job-activation request was routed to;
  // the default physical tenant when the request did not specify one
  string physicalTenantId = 22;
```

ce-dmelnych (dismissed):
LGTM ✅

Thanks for documenting the `jobLease` concept so clearly in the public documentation! 🙌🏻 

I only left one minor nit, which does not need to block this PR and could also be addressed separately in a dedicated follow-up issue/PR.

docs/apis-tools/zeebe-api/gateway-service.md
korthout: @afgambin Is this something you could look into for us?

docs/self-managed/upgrade/components/890-to-8100.md
```diff
@@ -312,3 +312,46 @@ zeebe:
...
+
+### Job leasing during the rolling upgrade
+
+Camunda 8.10 introduces [job leasing](/components/concepts/job-workers.md#job-leasing).
+During a rolling upgrade from 8.9, jobs on partitions that are still running 8.9 return a `null` `leaseToken`, even when activated with `withLease` set to `true`, because 8.9 brokers do not support job leasing.
+
+If your worker requires the lease to process a job safely, treat a `null` `leaseToken` as not yet supported on that partition: fail the job with a retry backoff, and set retries back to the job's current value so the fail doesn't burn through them while the upgrade is in progress.
+The lease becomes available after that job's partition leader is upgraded to 8.10, and the retry converges.
+
+When using the REST API, a partition still running 8.9 rejects the unknown `withLease` field with a `400` error instead of returning `null`. This handling is no longer required after the entire cluster has been upgraded to 8.10.
```
afgambin: I got this from the AI-review, worth checking it :

""Over REST, a partition still on 8.9 rejects the unknown withLease field with a 400." The strict unknown-field rejection (FAIL_ON_UNKNOWN_PROPERTIES) lives in the gateway's REST Jackson config (zeebe/gateway-rest/.../JacksonConfig.java), not in a broker/partition - partitions don't serve HTTP. It's the version of the gateway instance handling the REST request that matters, not which partition leader ends up processing the command. Even the source issue says "an old cluster" (also loose), but the PR narrows it to "partition," which is more specifically wrong. Worth asking the engineering reviewer to confirm/reword (e.g., "an unupgraded gateway instance")."

docs/apis-tools/zeebe-api/gateway-service.md
afgambin: @korthout @ce-dmelnych  Filed it [here](https://github.com/camunda/camunda-docs/issues/9736). Please let me know whether this is part of the real-time agent visibility epic (to understand priorities). If so, let me know whether you want to take it on first or would prefer me to take a first pass and then ask for your review. Thanks!

afgambin (dismissed):
@korthout I did a review pass and lgtm now. I left one (I believe non-blocking) comment.

docs/apis-tools/zeebe-api/gateway-service.md
korthout: It's not related to the real-time agent visibility epic. But other features that were added to the gRPC API and not documented. I suggest we check alignment as well for older versions.

docs/self-managed/upgrade/components/890-to-8100.md
korthout: Confirmed — the 400 is a gateway-side REST rejection (FAIL_ON_UNKNOWN_PROPERTIES in the gateway's Jackson config), independent of which partition/broker would have handled the command. Reworded to attribute it to the gateway instance in 22aebd5e4.

korthout:
@afgambin looks like one of the links is broken

afgambin (approved):

Issue #9547, opened 2026-08-04
Document the three-region Kubernetes + RDBMS multi-region strategy


leiicamundi:
## Summary

The multi-region resilience section documents two strategies — [Cold Recovery](https://docs.camunda.io/docs/next/self-managed/concepts/multi-region/cold-recovery/) and [Dual-Region](https://docs.camunda.io/docs/next/self-managed/concepts/multi-region/dual-region/). A third strategy is being delivered for 8.10 and has no page: a three-region Kubernetes deployment backed by a globally replicated relational database, where **losing a region requires no operator intervention at all**.

That is the differentiator, and it is not expressible in the current comparison table. Both documented strategies have a manual, operator-initiated failover. This one does not have a failover procedure, because the cluster never loses quorum.

- Product epic: camunda/product-hub#3553
- Tier framework: camunda/product-hub#3586
- Implementation: camunda/camunda-deployment-references#2940 (`aws/kubernetes/eks-multi-region-rdbms`)

## Why the existing pages do not cover it

| | Cold Recovery | Dual-Region | This |
|---|---|---|---|
| Region loss | Restore from backup | Quorum lost, engine **stops** until brokers are force-removed | Quorum preserved, engine **keeps processing** |
| Failover mode | Manual | Manual, operator-initiated | **None required** |
| Failback | Manual restore | Eight steps including an Elasticsearch snapshot and cross-region restore | Redeploy the region |
| Secondary storage | Object storage backups | One Elasticsearch per region, two exporters | One database, one exporter, replication owned by the database |

## Proposed content

A page under `self-managed/concepts/multi-region/`, listed as a third option on the [resilience tiers](https://docs.camunda.io/docs/next/self-managed/concepts/multi-region/resilience-tiers/) page and added to its comparison table.

**1. Why three regions.** Zeebe distributes partitions round-robin over consecutive node IDs and the Helm chart computes `nodeId = ordinal * global.multiregion.regions + regionId`. Setting `replicationFactor = number of regions` places exactly one replica of every partition in every region. A partition keeps its majority while `N-1 > N/2`, which is true for every `N >= 3`:

| Regions | Replication factor | Region losses tolerated |
|---|---|---|
| 2 | 4 (dual-region) | **0** |
| 3 | 3 | 1 |
| 4 | 4 | 1 |
| 5 | 5 | 2 |

[...]

PR #9700, opened 2026-08-25
docs: add time lag replication documentation

stoerti:
## When should this change go live?

- [ ] This is a bug fix, security concern, or something that needs **urgent release support**. (add `bug` or `support` label)
- [ ] This is already available but undocumented and should be released within a week. (add `available & undocumented` label)
- [ ] This is on a **specific schedule** and the assignee will coordinate a release with the Documentation team. (create draft PR and/or add `hold` label)
- [x] This is part of a scheduled **alpha or minor**. (add alpha or minor label)
- [ ] There is **no urgency** with this change (add `low prio` label)

## PR Checklist

- [ ] The commit history of this PR is cleaned up, using [`{type}(scope): {description}` commit message(s)](https://github.com/camunda/camunda-docs/blob/main/CONTRIBUTING.MD#commit-message-header-formatting)

- [x] My changes are for **an upcoming minor release** and are in the `/docs` directory.
- [ ] My changes are for an **already released minor** and are in a `/versioned_docs` directory.

- [ ] I added a DRI, team, or delegate as a reviewer for technical accuracy and grammar/style:
  - [ ] [Engineering team review](https://github.com/camunda/camunda-docs/blob/main/howtos/documentation-guidelines.md#review-process)
  - [ ] [Technical writer review](https://github.com/camunda/camunda-docs/blob/main/howtos/documentation-guidelines.md#review-process) via `@camunda/tech-writers` unless working with an embedded writer.

PR #9712, opened 2026-08-26
docs: add three-region active-active (RDBMS) multi-region option

theburi:
## Description

Adds three-region active-active (RDBMS) as a third multi-region resilience option on the [multi-region resilience](https://docs.camunda.io/docs/next/self-managed/concepts/multi-region/resilience-tiers/) page, alongside the existing Cold Recovery and Dual-Region strategies.

This is a three-region Kubernetes deployment where the Orchestration Cluster runs active-active across all regions (partition quorum survives the loss of one region, with no operator-initiated failover) and a relational database (RDBMS) with cross-region replication handles secondary storage.

Per direction from `@andreybelik`, this update stays inline on `resilience-tiers.md` — no new dedicated page is added yet.

Related: camunda/camunda-docs#9547 (docs gap issue proposing this content), camunda/product-hub#3553, camunda/product-hub#3586, camunda/camunda-deployment-references#2940.

Note: the underlying feature targets the 8.10 release and its reference implementation is still in development, so the specifics here may need to change before general availability.

## PR Checklist


docs/self-managed/concepts/multi-region/resilience-tiers.md
```diff
@@ -18,35 +18,40 @@ Camunda provides a structured multi-region resilience framework for Self-Managed
...
+| **Recovery time (RTO)** | ~1–4 hours                                                                              | ~15 minutes                                                                                                                                                       | Primary storage: ~0 (no failover procedure). Secondary storage: depends on the database's own failover behavior.                                                               |
+| **Data loss (RPO)**     | 15 min – 4 hours (backup-interval dependent)                                            | 0 minutes                                                                                                                                                         | 0 minutes, under nominal database replication lag (see note below)                                                                                                             |
```
entangled90: 🔧 I would frame it differently as it's not just depending on the replication lag, but also on the database itself. so for example for Aurora it's 0.

docs/self-managed/concepts/multi-region/resilience-tiers.md
```diff
@@ -18,35 +18,40 @@ Camunda provides a structured multi-region resilience framework for Self-Managed
...
 
 :::important
 Cold Recovery RTO and RPO targets are bounded by data volume, backup frequency, and operator restore speed. Treat published ranges as planning targets, not contractual commitments.
 
 Dual-Region RTO is based on internal operational tests. Actual times may vary depending on your environment, level of automation and the specific manual steps performed during recovery. See [Dual-Region](./dual-region.md#recovery-objectives) for a phase-by-phase breakdown.
+
+Three-region active-active (RDBMS) reaches an RPO of 0 by keeping the Orchestration Cluster's exported log history until the database confirms replication, rather than by requiring the database itself to replicate synchronously. On databases that expose a replication-lag signal (for example, log sequence number tracking on Aurora PostgreSQL), the Orchestration Cluster monitors that lag and pauses new exports if it grows too large. This preserves consistency on recovery as long as the database's replication stays within nominal operating parameters; it doesn't protect against a database that has stopped replicating entirely.
```
entangled90: 💭  it's 0 for Aurora MySQL as well and for Self-hosted PostgreSQL/MicrosoftSQL server.
I think we should create another page to explain the support matrix for the database etc.
This can be fixed when we create such page.

docs/self-managed/concepts/multi-region/resilience-tiers.md
leiicamundi: Agreed with @entangled90 that it's database-dependent rather than lag-dependent. Worth making the precondition explicit, because it isn't the default: `async-replication.enabled` defaults to `false`, `min-sync-replicas` to `1`, and `pause-on-max-lag-exceeded` to `false` (see the property table in the RDBMS [Multi-region support](https://docs.camunda.io/docs/next/self-managed/concepts/databases/relational-db/configuration/#multi-region-support) section). RPO 0 needs a `type` selected, `min-sync-replicas: 2`, and `GRANT PG_MONITOR` on PostgreSQL.

Relevant to the reference implementation: camunda/camunda-deployment-references#2940 sets `CAMUNDA_DATA_SECONDARYSTORAGE_RDBMS_ASYNCREPLICATION_ENABLED=true` and nothing else — no `type`, so no strategy is selected. Either that PR configures the strategy, or this cell says "0 when replication monitoring is configured; see Multi-region support".

docs/self-managed/concepts/multi-region/resilience-tiers.md
```diff
@@ -18,35 +18,40 @@ Camunda provides a structured multi-region resilience framework for Self-Managed
...
+| **Data loss (RPO)**     | 15 min – 4 hours (backup-interval dependent)                                            | 0 minutes                                                                                                                                                         | 0 minutes, under nominal database replication lag (see note below)                                                                                                             |
+| **Failover mode**       | Manual, operator-initiated                                                              | Manual, operator-initiated                                                                                                                                        | None required for a single region loss; partition quorum is preserved                                                                                                          |
```
leiicamundi: "None required for a single region loss" conflicts with the RTO cell two rows up, which says secondary-storage recovery depends on the database's failover behavior.

Aurora Global Database has one writer region. camunda/camunda-deployment-references#2940 ships a `failover.sh` whose unplanned path runs `remove-from-global-cluster` to promote a survivor, and camunda/camunda-deployment-references#3290 exists because that leaves Terraform state drift.

Suggest: "None for the Orchestration Cluster; database writer failover required if the lost region held the writer."

Related, next row: "single RDBMS instance ... with replication owned by the database" reads as contradictory — a single instance has nothing to replicate. #2940 describes the database tier as active-standby, with one writer serving every region.

docs/self-managed/concepts/multi-region/resilience-tiers.md
leiicamundi: Rather than adding a heading here for the anchor, this looks like a good candidate for the database page @entangled90 suggests. The RDBMS [Multi-region support](https://docs.camunda.io/docs/next/self-managed/concepts/databases/relational-db/configuration/#multi-region-support) section already documents the mechanism with the property table, vendor support and required grants, and #9700 is extending it with `TIME_LAG`. Restating it here means two descriptions drifting apart.

Also missing from the paraphrase, and stated on the canonical page: deferring compaction "may drastically increase the disk space usage of the logstream". RPO 0 is bought with broker disk — relevant on a page that compares cost.

Suggest replacing the paragraph with a one-line pointer, and letting the vendor matrix live on the new page.

docs/self-managed/concepts/multi-region/resilience-tiers.md
```diff
@@ -18,35 +18,40 @@ Camunda provides a structured multi-region resilience framework for Self-Managed
...
-| **Compliance fit**      | Basic business continuity management (BCM) requirements                                 | Certified, auditable region-recovery posture with a published runbook                                                                                             |
-| **Relative cost**       | **$** (lower cost): Object storage only; no standing second region                      | **$$$** (higher cost): Orchestration Cluster running across both regions with extra capacity to sustain load in case of Region failure, plus cross-region traffic |
+| Consideration           | Cold Recovery                                                                           | Dual-Region                                                                                                                                                       | Three-region active-active (RDBMS)                                                                                                                                             |
```
leiicamundi: This is where the transposition would help most. Three columns already forces ~500-character source lines and reflows the whole table on any edit; a fourth strategy makes it worse. Per-strategy sections would also give Copilot's requested anchors without having to invent them.

Minor, whichever way this goes: the separator rows here and in the "Choose your strategy" table above are wider than their header rows. `prettier --write` runs on `*.md` via lint-staged, so this will reflow on the next touch — worth running `npm run format` to avoid a second churn commit.

leiicamundi (commented):
Three of us landed on the same structural point from different angles, so consolidating.

Copilot asks for stable anchors and notes the compliance-fit cell answers a different question than its row. @entangled90 notes RPO is database-dependent (0 on Aurora, Aurora MySQL, self-hosted PostgreSQL, MSSQL) and suggests a separate page for the vendor support matrix. Both point at the same shape:

**One section per strategy, one row per consideration.** `### Cold Recovery`, `### Dual-Region`, `### Three-region active-active`, each with a `| Consideration | Value |` table. Copilot's anchors come for free. Cells get room for answers that aren't a single scalar (this strategy's RTO genuinely has two parts). Adding a fourth strategy appends a section instead of reflowing every line of a 3-column table — the current diff is 19+/14− for what is semantically one column.

The vendor matrix then extends the existing [Multi-region support](https://docs.camunda.io/docs/next/self-managed/concepts/databases/relational-db/configuration/#multi-region-support) section rather than living here. The RPO cell links to it instead of restating it — worth noting #9700 is currently editing that same section to add `TIME_LAG`, so a paraphrase here will drift on merge.

Two factual items inline need resolving regardless of structure. One question: given the `multi-region-concept` branch argues that "active-active" collapses independent concerns, and that dual-region is already active-active at the data layer, is that the taxonomy we're landing on? If so this strategy's name may want to change with it.

leiicamundi:
similar pr: https://github.com/camunda/camunda-docs/pull/9739

docs/self-managed/concepts/multi-region/resilience-tiers.md
theburi: I don't want to provide too much details in this overview. 
So modified to say "0 minutes, database dependent"

docs/self-managed/concepts/multi-region/resilience-tiers.md
theburi: The usability of columns is much better, especially when we talk about tiers here.

docs/self-managed/concepts/multi-region/resilience-tiers.md
theburi: Updated to address this comment

theburi: Updated with new version

docs/self-managed/concepts/multi-region/img/multi-region-overview.png
entangled90, line 1: 💭  I noticed that the secondary storage is not showed for the Dual-Region active-active setup 
In that case we have two separate ES clusters (if using ES) or a single RDBMS with sync replication and potentially other setups. 

I think that right now it might be a bit misleading?

docs/self-managed/concepts/multi-region/resilience-tiers.md
```diff
@@ -18,35 +18,39 @@ Camunda provides a structured multi-region resilience framework for Self-Managed
...
+| **Typical use case**    | Low-criticality production; environments where hours-long recovery is acceptable | Enterprise production workloads that must survive a region failure                                                                                                | Enterprise workloads that require automatic resilience to a single region loss, without a manual failover step                          |
+| **Compliance fit**      | Basic business continuity management (BCM) requirements                          | Certified, auditable region-recovery posture with a published runbook                                                                                             | No manual failover runbook to certify for the Orchestration Cluster; database-level failover procedures are certified separately        |
```
entangled90, line 47: 💭 To maintain the same level of performance the operator should also move the raft leader to the RDBMS writer region, especially if the ping between the region is high. Note that this is potentially not required as throughput might be sufficient.

Probably not worth adding it here, but let's keep it in mind

entangled90 (dismissed):

docs/self-managed/concepts/multi-region/resilience-tiers.md
theburi: At the level this page targets, I would not go to this level ... of detail

docs/self-managed/concepts/multi-region/img/multi-region-overview.png
theburi: Good point, I will get it fixed

entangled90 (approved):

PR #9717, opened 2026-08-27
docs(concepts): document secret resolution failures, incidents, and the operator runbook

### LINKED ISSUE #60964: [Secret Resolution] Docs: secret resolution failures, incidents and troubleshooting
### 

### Description

Document what happens when secret resolution fails, and what an operator does about it.

D6 (#60332) covers the broker resolution path and mentions incidents in a single acceptance criterion. That is not enough to act on. This ticket owns the failure modes and the operator runbook.

Three outcomes are possible when a job's secrets cannot be delivered, and they are not interchangeable:

1. **Resolution failed** — an incident, error type `SECRET_RESOLUTION_ERROR`.
2. **Values too large** — the job is dropped from the activation. Usually silent and retried; an incident only in one specific case.
3. **Pointer no longer addresses a text leaf** — no incident at all, and the worker receives the unresolved placeholder.

### Motivation

Every failure here is operationally visible and none of it is self-explanatory. A job that does not activate, an incident naming a placeholder, or a worker that receives the literal `camunda.secrets.<name>` text all send the operator to the store, the configuration, or support — with nothing in the documentation to check against.

### Acceptance criteria

- [ ] `SECRET_RESOLUTION_ERROR` documented: when it is raised, where it appears, and the message text — currently "Failed to resolve secret '&lt;reference&gt;' from &lt;store&gt;. Ensure the secret exists and the store is available, then resolve the incident to retry." Verify the wording against `SecretReferenceBatchCreateIncidentsProcessor` before publishing, since #60404 and #60405 reworked it
- [ ] Documented that resolving the incident retries, and that jobs blocked on unresolved references are picked up again once resolution succeeds — no redeployment needed
- [ ] Size behaviour documented, distinguishing the two cases. A job whose injected values would push the activation response past the max message size is dropped from that batch **together with every job after it**, and those jobs stay activatable — the next activation picks them up. A `MESSAGE_SIZE_EXCEEDED` incident is raised **only** when the oversized job is first in the batch, i.e. its values can never fit any batch. Verify against `JobSecretInjector#injectSecretValues` and `dropJobsThatNoLongerFit`
- [ ] Documented that the effective limit is the max message size (4 MB by default), not a fixed secret-value cap. #60060 changed this — do not carry over the previous 8 KB figure, and check whether the incident message now names the correct limit
- [ ] Confirm whether the long-poll and job-push paths still differ on the size check. #60060 reported that push performed no size check at all, so a streaming worker received the same job without an incident. If that asymmetry still exists, document it
- [ ] Documented that secret resolution does not un-park a suspended job: once suspension overrides `WAITING_FOR_SECRET_RESOLUTION`, secret completion is a silent no-op (#59561)
- [ ] Documented that when a reference's JSON pointer no longer addresses a text leaf, injection is skipped, **no incident is raised**, and the worker is handed the literal `camunda.secrets.<name>` placeholder unresolved — activated once, not a reactivation loop (#58614). This is the most surprising behaviour in the set and the easiest to misread as a bug
- [ ] Operator runbook: diagnose (store reachable, secret present, name within the reference charset `[\p{Alnum}_-]+`, permission granted), fix, and what happens next
- [ ] Placed where a reader arriving from an incident will find it — the incident or troubleshooting reference, not only the secrets pages
- [ ] Cross-linked with #60332 so the two do not restate each other

### Links

Behaviour: https://github.com/camunda/camunda/issues/60060, https://github.com/camunda/camunda/issues/59561, https://github.com/camunda/camunda/issues/58614
Incident messages: https://github.com/camunda/camunda/issues/60404, https://github.com/camunda/camunda/issues/60405

### Breakdown

- Verify the current incident message wording and the current size limit in code — both changed recently
- Confirm the long-poll versus job-push size-check asymmetry
- Write the runbook and place it in the incident or troubleshooting reference

### Notes

Source of truth in `camunda/camunda`:

- `zeebe/engine/.../processing/secretreference/SecretReferenceBatchCreateIncidentsProcessor.java` — `SECRET_RESOLUTION_ERROR` and its message
- `zeebe/engine/.../processing/job/JobSecretInjector.java` — injection, size budget, drop behaviour
- `zeebe/engine/.../processing/secretreference/SecretReferenceBatchReactivateJobsProcessor.java` — reactivation
- `zeebe/docs/adr/0007-810-job-waiting-for-secret-resolution-state.md` — the `WAITING_FOR_SECRET_RESOLUTION` state

Depends on #60332 for the activation model this page references. Cross-link rather than restate.

HeleneW-dot:
closes [#60964](https://github.com/camunda/camunda/issues/60964)

Depends on https://github.com/camunda/camunda-docs/pull/9703

## Description

Adds the operator runbook for secret resolution failures ([camunda/camunda#60964](https://github.com/camunda/camunda/issues/60964)), part of the Centralized Secret Resolution doc set ([camunda/camunda#56591](https://github.com/camunda/camunda/issues/56591)).

Documents the three outcomes when a job's secret references can't be delivered:

SECRET_RESOLUTION_ERROR — store resolution failed, or injection failed
MESSAGE_SIZE_EXCEEDED — resolved values can never fit an activation
Silent outcomes that need no operator action (batch drop, no-placeholder pointer, job push's size-check asymmetry), plus suspension overriding secret-waiting

## PR Checklist

### DIFF
--- a/docs/components/concepts/secret-resolution-incidents.md
+++ b/docs/components/concepts/secret-resolution-incidents.md
@@ -1,0 +1,245 @@
+---
+id: secret-resolution-incidents
+title: "Troubleshoot secret resolution failures"
+description: "Diagnose the incidents raised when a job's secret references cannot be resolved or their values cannot be injected, fix the cause, and know what happens next."
+---
+
+<!-- Draft. The page title, file name, placement, and the name of the feature itself are owned by
+camunda/camunda#60326 and are not settled, so this page is deliberately not listed in sidebars.js
+yet and every term used for the feature here is provisional. -->
+
+When a job's [secret references](secret-resolution-and-job-activation.md) cannot be delivered, the
+cluster reacts in one of three ways: it raises a `SECRET_RESOLUTION_ERROR` incident, it raises a
+`MESSAGE_SIZE_EXCEEDED` incident, or it defers the job silently and retries it. The three are not
+interchangeable, and only two of them need you.
+
+This page is for the operator who already has a symptom. For how resolution and activation work in
+the first place, see [secret resolution and job activation](secret-resolution-and-job-activation.md).
+
+## Find your symptom
+
+| Symptom                                                                              | What happened                                                                                     | Section                                                                                                 |
+| :----------------------------------------------------------------------------------- | :------------------------------------------------------------------------------------------------ | :------------------------------------------------------------------------------------------------------ |
+| An incident whose message starts with `Failed to resolve secret`                     | The secret store could not return the value, permanently or after exhausting its retries.         | [A secret could not be resolved](#a-secret-could-not-be-resolved)                                       |
+| An incident whose message names a job key and a variable path                        | The value was available, but it could not be written into the job's variables.                    | [A secret value could not be injected](#a-secret-value-could-not-be-injected)                           |
+| An incident whose message names a growth in bytes and the configured message size    | The resolved values are too large to ever fit an activation.                                      | [Secret values are too large to activate](#secret-values-are-too-large-to-activate)                     |
+| A job of a type that references secrets is never activated, and there is no incident | The references have not resolved yet, or the store is backing off.                                | [Failures that raise no incident](#failures-that-raise-no-incident)                                     |
+| Jobs activate, but later and in smaller batches than usual, and there is no incident | Injected values did not fit the batch, so jobs were dropped from it and picked up by a later one. | [Failures that raise no incident](#failures-that-raise-no-incident)                                     |
+| A job of a suspended process instance is not activated after its secret resolves     | Suspension overrides secret waiting, and secret completion does not undo it.                      | [Secret resolution does not resume a suspended job](#secret-resolution-does-not-resume-a-suspended-job) |
+
+All of these incidents are job incidents. They appear wherever job incidents appear: in
+[Operate](/components/operate/userguide/resolve-incidents-update-variables.md), and in the
+[search incidents](/apis-tools/orchestration-cluster-api-rest/specifications/search-incidents.api.mdx)
+endpoint, where you can filter on `errorType` for `SECRET_RESOLUTION_ERROR` or
+`MESSAGE_SIZE_EXCEEDED`.
+
+No incident message ever contains a secret value. A message names the reference, the variable path,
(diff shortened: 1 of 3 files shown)


docs/components/concepts/secret-resolution-incidents.md
```diff
@@ -0,0 +1,245 @@
...
+
+- **The reference names the secret you think it does.** In an input mapping, a reference is detected
+  from the parsed FEEL expression, so FEEL's own rules decide where the name ends. A dash is
+  subtraction to FEEL, so `=camunda.secrets.db-password` is read as the reference `db` minus the
+  variable `password`. Escape any name that is not a bare FEEL identifier with backticks, as in
+  ``=camunda.secrets.`db-password` ``. An incident naming a shorter secret than you wrote is this
+  mistake.
+- **The name matches the secret in the store.** The reference name and the secret name must match
+  exactly, including case. Names that Camunda's own secret endpoints reject, such as a name holding
+  a dot, are still detected and resolved from a store when they are backtick-escaped in the
```
ajanoni-camunda: 🔧 The issue's acceptance criteria call out the runbook should check "name within the reference charset `[\p{Alnum}_-]+`" as its own diagnosis step. This bullet covers the behavior around it (backtick-escaping, dot-named secrets rejected by the API but resolvable via the store) but never states the charset itself as a checkable fact, so a name using a character outside `[\p{Alnum}_-]+` (e.g. a space or other punctuation) isn't explicitly called out as a cause to check for.

Suggest adding one clause naming the charset explicitly here, e.g.: "...a store-managed name can work while the API refuses to create it. The reference itself only ever matches `[\p{Alnum}_-]+` after the prefix, so a name using any other character is never detected as a reference at all."

ajanoni-camunda (dismissed):
👍 Looks good. Just one comment.

docs/components/concepts/secret-resolution-incidents.md
HeleneW-dot: Yes that makes sense, I adjusted to make it more explilcit

HeleneW-dot:
@giorgionaps This PR is also ready for tech writer review 🙏

lev-safro (approved):

PR #9720, opened 2026-08-27
docs(modeler): document Configuration element-template property type

nikku:
## Description

Documents the core configuration mechanism for element templates in the 8.10 (`/docs`) tree: the standard `type: "Configuration"` property type and its embedded `configurationTemplates`.

- `template-properties.md`: new **Configuration input type** section — the chooser, the `configurationTemplate` / `configurationTemplateVersion` keys, how a configuration is stored as a cluster variable with metadata, design-time selection vs. runtime resolution, secret references, and the `zeebe:input` / `zeebe:property` binding with modeler-only display metadata.
- `template-metadata.md`: new **Embedding configurations: `configurationTemplates`** section covering the top-level embedding key and its attributes (including `kind`).
- `template-example.md`: extends the shared example to demonstrate `Configuration` end-to-end.

Scope is the generic core mechanism; credentials are only referenced as the domain concept built on top. 

Two images added/updated:

* [x] `docs/components/modeler/element-templates/img/field-configuration.png` needs to be added
* [x] `docs/components/modeler/element-templates/img/overview.png` needs to be updated to showcase configuration chooser

Closes https://github.com/bpmn-io/internal-docs/issues/1365

## When should this change go live?

- [ ] This is a bug fix, security concern, or something that needs **urgent release support**. (add `bug` or `support` label)
- [ ] This is already available but undocumented and should be released within a week. (add `available & undocumented` label)
- [ ] This is on a **specific schedule** and the assignee will coordinate a release with the Documentation team. (create draft PR and/or add `hold` label)
- [x] This is part of a scheduled **alpha or minor**. (add alpha or minor label)
- [ ] There is **no urgency** with this change (add `low prio` label)

## PR Checklist

- [x] The commit history of this PR is cleaned up, using [`{type}(scope): {description}` commit message(s)](https://github.com/camunda/camunda-docs/blob/main/CONTRIBUTING.MD#commit-message-header-formatting)

- [x] My changes are for **an upcoming minor release** and are in the `/docs` directory.
- [ ] My changes are for an **already released minor** and are in a `/versioned_docs` directory.

- [ ] I added a DRI, team, or delegate as a reviewer for technical accuracy and grammar/style:
  - [ ] [Engineering team review](https://github.com/camunda/camunda-docs/blob/main/howtos/documentation-guidelines.md#review-process)
[...]

PR #9726, opened 2026-08-27
docs(connectors): add connector credentials pages for Hub, Desktop Modeler, and Web Modeler

crobbins215:
## Description

Documents the new **Credentials** feature landing in the 8.10 minor: reusable connector authentication and connection configuration that you create once and then select on a connector task, instead of re-entering the same settings in every diagram.

Epic: https://github.com/camunda/product-hub/issues/3396

### New pages

A new **Credentials** category under **Components > Connectors > How to use connectors**:

| Page | Covers |
| --- | --- |
| `credentials/index.md` | Concepts, terminology, credential types, the `camunda.secrets.` reference syntax, and the full Camunda Hub walkthrough (Managed and Clusters only tabs, three-step create wizard, credential states, edit, delete, permissions, limitations) |
| `credentials/desktop-modeler.md` | Selecting, creating, editing, and upgrading a credential from the properties panel, what the picker offers per connection/permission state, and the single-cluster scope of Desktop Modeler credentials |
| `credentials/web-modeler.md` | The same flow in Web Modeler, plus the two differences: no offline state, and credentials created here are Hub-managed immediately |

Three screenshots are included (Managed empty state, wizard step 1, Clusters only tab). Further screenshots (populated list, Configure step, Review step, the pickers, the inline create form) are still to be added, see the open questions below.

### Changes to existing pages

Cross-links only, no behavior changes:

- `connectors/use-connectors/index.md` — new short **Using credentials** section above **Using secrets**
- `hub/organization/manage-clusters/manage-secrets.md` — pointer to credentials, noting a credential's sensitive fields reference secrets created here
- `modeler/desktop-modeler/connect-to-camunda-8.md` and `hub/workspace/modeler/run-or-publish-your-process.md` — disambiguation pointers, because both pages use the word "credentials" for the client ID and secret that authenticate Modeler against a cluster, which is a different concept from connector credentials

### Validation

`npm run build` passes with no broken links or anchors introduced by these pages.

### Open questions for reviewers

1. **Hub navigation path** to the **Credentials** page is not stated anywhere in the docs, because I could not confirm which section of Hub it sits under. Please confirm so I can add it.
2. **Test credential** appears as a requirement in the design docs but I could not confirm it shipped in 8.10, so it is not documented. Confirm whether it should be.
3. The **Clusters only** screenshot shows a cluster version string of "8.10 Versioned SNAPSHOT". Happy to retake it against a release build if preferred.
4. Please confirm the **credential type list** (AWS Credential, REST Authentication, JDBC Connection) is complete for the minor. It is framed as a growing list rather than exhaustive.

/cc @camunda/tech-writers

## PR Checklist


docs/components/modeler/desktop-modeler/connect-to-camunda-8.md
```diff
@@ -6,6 +6,8 @@ description: "Set up a connection from Desktop Modeler to Camunda 8 Orchestratio
 
 To deploy diagrams, start process instances, or test tasks, you must first connect Desktop Modeler to a Camunda 8 Orchestration Cluster. Follow the steps below to connect to **Camunda 8 SaaS**. To connect to a local installation, visit the [Camunda 8 Self-Managed guide](../../../self-managed/components/modeler/desktop-modeler/connect-to-self-managed.md).
 
+The credentials on this page authenticate Desktop Modeler itself. To authenticate an individual connector task, see [configure credentials in Desktop Modeler](../../connectors/use-connectors/credentials/desktop-modeler.md).
```
nikku: I'm not sure we need this - what do we guard against? Where there is ambiguity, clearly tell it apart - call the thing we use here *cluster credentials* - while we tell the other ones (in ambiguous contexts) *connector credentials*.

docs/components/hub/organization/credentials/index.md
```diff
@@ -0,0 +1,163 @@
...
+title: Connector credentials
+description: "Create reusable connector credentials in Camunda Hub, then select them in your connector tasks instead of repeating authentication settings in every diagram."
+keywords: [credential, credentials, authentication, reusable, connection]
+---
+
+A credential stores the authentication information a connector needs to reach an external service, so you can reuse it across connector tasks instead of entering the same values in every diagram.
+
+## About credentials
+
+Without credentials, you configure a connector's authentication and connection settings directly on each connector task. If ten tasks call the same REST API, you configure the same authentication ten times, and you update all ten when something changes.
```
nikku, line 14: Credentials is a standalone concern that can be used *fully without connectors*. I'd love to see it documented as such.

Connectors is the *main* user today, but as infrastructure it can be used by *any* consumer. We risk falling into the connector template trap if we bucket them like this - make it harder for folks to understand how our platform composes.

nikku:
High-level feedback:

* Make `credentials` a standalone concern - it currently is in the HUB UI and it is in practice. Connectors are main users, but they work standalone.
* Consider disambiguation you add across the docs - remove it or simplify (call credentials *cluster* credentials where they are) or call out the difference explicitly (via a `::note` block) - don't disambiguate in inline prose.

alexronquillo:
I'll wait until after @nikku's comments are addressed before reviewing in case there are significant changes.

docs/components/modeler/desktop-modeler/connect-to-camunda-8.md
crobbins215: Removed the note. Renamed to *cluster credentials* here instead, so it reads unambiguously without an aside.

docs/components/hub/organization/credentials/index.md
crobbins215: Moved Credentials out of the Connectors tree entirely, to `components/hub/organization/credentials/` (sidebar: Camunda Hub > Organization > Credentials), and reworded the intro/About section to describe it as Hub infrastructure that any element template can use via the `Configuration` property type, with connectors called out as today's main consumer rather than the framing. Connectors now gets a single linking section (mirrors the existing *Using secrets* pattern) instead of owning any credentials pages.

crobbins215:
Addressed @nikku's review — moved Credentials to `components/hub/organization/credentials/` (sidebar: Camunda Hub > Organization > Credentials), reworded the intro to describe it as Hub infrastructure any element template can use rather than a connector-only concept, and replaced the Connectors-side Credentials category with a single linking section on the existing guide page (mirrors the *Using secrets* pattern). Removed the inline disambiguation sentences and renamed to *cluster credentials* at the point of use instead, per the review comment. Replies inline on the two original threads.

@chillleader — added a scaffolded page at [`docs/components/connectors/custom-built-connectors/credential-templates.md`](https://github.com/camunda/camunda-docs/blob/docs/connector-credentials/docs/components/connectors/custom-built-connectors/credential-templates.md) for **how to author a credential template for a custom connector** (declaring a `kind: CREDENTIAL` configuration template, wiring it to an element template's `Configuration` property and connector input, secret fields, versioning). Per the epic transition notes this is Connectors-team content — headings and a TODO comment are in place as scaffolding, needs your write-up before merge.

docs/components/hub/organization/credentials/desktop-modeler.md
```diff
@@ -0,0 +1,69 @@
+---
+id: credentials-desktop-modeler
+title: Configure credentials in Desktop Modeler
```
alexronquillo, line 3: We've decided to treat Desktop Modeler and Camunda Hub as completely separate products, so the desktop modeler documentation should be moved from the Hub space to the [Desktop Modeler](https://docs.camunda.io/docs/next/components/modeler/desktop-modeler/) space. Also this seems to be prioritizing Desktop Modeler for explanations and instructions rather than the modeling interface within Camunda Hub, even though these are described as part of Hub. I suggest we: 

1. Move the source explanations to the Hub docs (what you have as Web Modeler now)
2. Move the Desktop Modeler docs to the Desktop Modeler section
3. Reference the Hub docs from the Desktop Modeler section

docs/components/hub/organization/credentials/web-modeler.md
```diff
@@ -0,0 +1,39 @@
+---
+id: credentials-web-modeler
+title: Configure credentials in Web Modeler
```
alexronquillo, line 3: Web Modeler as a product is now Camunda Hub, so we should update this terminology. When referring to modeling withing Hub, we typically use the "Hub modeling interface/canvas"

docs/components/hub/organization/credentials/web-modeler.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Select a credential
+
+Connectors that support credentials show a credential field in the properties panel, such as **AWS Credential**. Select the field to open the credential picker, which lists the credentials on the cluster that match the credential type the connector needs.
+
+Selecting a credential stores only a reference to it in your diagram. The credential's values stay on the cluster.
+
+The picker behaves as described in [Configure credentials in Desktop Modeler](./desktop-modeler.md#what-you-can-do-in-the-picker), with the differences below.
+
+## Differences from Desktop Modeler
```
alexronquillo, line 22: I would put this in the Desktop Modeler space and reference Camunda Hub, rather than putting it in Camunda Hub and referencing Desktop Modeler.

docs/components/hub/organization/credentials/web-modeler.md
```diff
@@ -0,0 +1,39 @@
...
+
+- Web Modeler always works against an authenticated Camunda 8 cluster, so what you can do always comes from that cluster's authorization configuration. There is no offline state where the picker cannot reach a cluster.
+- A credential you create in Web Modeler is managed in Camunda Hub immediately. It appears on the **Managed** tab of the [**Credentials** page](./index.md#managed-credentials) without the **Clusters only** discovery step that a credential created from Desktop Modeler requires.
+
+## Create, edit, or upgrade a credential
+
+Creating, editing, and upgrading a credential works the same as in Desktop Modeler. See [Create a credential](./desktop-modeler.md#create-a-credential) for the fields and the steps.
```
alexronquillo, line 29: I think we would rather duplicate content here, rather than sending users to Desktop modeler docs, which are now treated as a completely separate product, even if there is overlap in the UI. Even if we don't, I would reference Camunda Hub from the Desktop Modeler space, rather than referencing Desktop Modeler from the Hub space.

alexronquillo (changes requested):
Thanks for the docs submission @crobbins215! Overall, I think this is great. I've left a few comments.

There's one large information architecture theme I think we should consider. Namely, I would re-prioritize these docs with Camunda Hub modeling interface as the first priority. This is a Hub feature, and that's where Credentials are stored and managed. As you mention in the overview doc "Credentials are Hub infrastructure".

Desktop Modeler is an alternative modeling interface that makes use of Camunda Hub's credentials. So Camunda Hub docs should feel self contained, and any explanation of differences or references between the two should live in the Desktop Modeler space.

What do you think?

docs/components/hub/organization/credentials/desktop-modeler.md
crobbins215: Done, all three steps. The Desktop Modeler page moved out of the Hub space to `components/modeler/desktop-modeler/credentials.md` and is now listed in the Desktop Modeler sidebar. The Hub modeling interface page holds the source explanations and instructions, and the Desktop Modeler page references Camunda Hub for the concept and central management.

docs/components/hub/organization/credentials/web-modeler.md
crobbins215: Renamed throughout. The page is now titled *Configure credentials in the modeling interface* at `hub/organization/credentials/modeling-interface.md`, and body copy says Camunda Hub or the Hub modeling interface. No remaining Web Modeler references in these pages.

docs/components/hub/organization/credentials/web-modeler.md
crobbins215: Done. Desktop Modeler instructions now live in the Desktop Modeler space and reference Camunda Hub; the Hub page no longer references Desktop Modeler.

docs/components/hub/organization/credentials/web-modeler.md
crobbins215: Duplicated as you suggested. The Hub modeling interface page is now self-contained: full picker table, create steps, and edit/upgrade steps, with no links into the Desktop Modeler docs. The Desktop Modeler page carries its own copy and links back to Camunda Hub.

crobbins215:
@alexronquillo actioned all four comments, replies inline. Summary:

- Desktop Modeler credentials page moved out of the Hub space to `components/modeler/desktop-modeler/credentials.md`, listed in the Desktop Modeler sidebar.
- The Hub modeling interface page (`hub/organization/credentials/modeling-interface.md`) is now the source for the shared explanations and instructions, self-contained, with no references into the Desktop Modeler docs. I also dropped the "offline, no cluster connection" row from its picker table, since that case only applies to Desktop Modeler.
- "Web Modeler" renamed to Camunda Hub / the Hub modeling interface throughout.
- Shared picker, create, and edit/upgrade instructions are duplicated on the Desktop Modeler page rather than cross-linked, and that page references Camunda Hub for the concept and central management.

One thing worth a check: the **Clusters only** tab on the Hub Credentials page exists specifically because credentials created outside Hub (Desktop Modeler, or the cluster API directly) are not tracked in Hub. I kept Desktop Modeler as a plain-text mention there with no link into the Desktop Modeler space, since the tab does not make sense without naming where those credentials come from. Say the word if you would rather it not be named at all.

Build passes with no new broken links.

crobbins215:
@chillleader this PR is blocked on your section before it can merge.

The page is scaffolded at [`docs/components/connectors/custom-built-connectors/credential-templates.md`](https://github.com/camunda/camunda-docs/blob/docs/connector-credentials/docs/components/connectors/custom-built-connectors/credential-templates.md) with headings and a TODO comment in place, but no content.

docs/components/hub/organization/credentials/index.md
```diff
@@ -0,0 +1,165 @@
...
+title: Credentials
+description: "Create a reusable credential in Camunda Hub, then select it wherever an element template needs authentication or connection configuration, instead of repeating the settings in every diagram."
+keywords: [credential, credentials, authentication, reusable, connection]
+---
+
+A credential stores authentication and connection configuration you create once and reuse, instead of entering the same values every time an element template asks for them.
+
+## About credentials
+
+Credentials are Hub infrastructure, not a connector-only feature. Any element template can declare a field that uses one, by using the `Configuration` element template property type. Connectors are the main consumer today, so this page and the credential types below are described in connector terms, but the same mechanism is available to any element template author. Element template authors can read how to declare this property type in [element template properties](/components/modeler/element-templates/template-properties.md).
```
alexronquillo: Do we need to update `template-properties.md` to include the `Configuration` type? I don't see a reference to it there.

alexronquillo (commented):
Great work on this! I pushed a small commit with the biggest impact change being the change from "Credentials" to "Manage credentials". This is mostly to:

- Be consistent with all the other sections
- Lead with an action verb

Otherwise, I left one question, and we'll wait on the `credential-templates.md` changes 👍

docs/components/hub/organization/credentials/index.md
nikku: Updates for `Configuration` type ship independently via https://github.com/camunda/camunda-docs/pull/9720.

nikku: (I'll get #9720 ready to review today).

nikku: > Credentials are Hub infrastructure, not a connector-only feature. Any element template can declare a field that uses one, by using the `Configuration` element template property type. Connectors are the main consumer today, so this page and the credential types below are described in connector terms, but the same mechanism is available to any element template author. Element template authors can read how to declare this property type in [element template properties](/components/modeler/element-templates/template-properties.md).

I wonder if we should re-phrase this section. It goes in length about why we write what we write later - a user skipping that explaination is left confused. What is the purpose of this page?

In my words:

* Credentials allow you to define and manage re-usable configuration for your job workers, connectors and other reusable assets (previously: re-usable building blocks).
* Managed via Hub, but also available in standalone Camunda distributions
* Defined alongside element templates through embedded configuration templates (link to other documentation)
* Stored as cluster variables and made available that way to job workers (and connectors)

Further context can:

* link to connectors ("how they use credentials")
* link to element templates ("how to define credentials through Configuration templates")
* link to modeler sections ("how to choose credentials from the modeler UI")

---

I don't think "credentials is a hub feature is the right framing, too - credentials are generally usable infrastructure - hub just adds a federated view for them - that is the "hub feature". :)

crobbins215: Rewritten to your four points, and the "Hub infrastructure" framing is gone. The section now reads:

- A credential lets you define and manage reusable configuration for job workers, connectors, and other element templates.
- Credentials are generally usable infrastructure, available in any Camunda 8 distribution, including standalone ones that run without Hub. Camunda Hub adds an organization-wide, federated view and central management across your clusters, which is what the rest of the page covers.
- A credential type is defined alongside an element template, through an embedded configuration template.
- A credential is stored as a cluster variable, which makes it available by name to any job worker or connector that references it.

Then a short paragraph on it being selected as a whole and resolved at runtime, and a "Learn more" list linking out the three ways you suggested: connectors (how a connector task uses one), the `Configuration` property type (how an element template declares the picker), and the modeler UI (how to choose one). The long "without credentials you configure it ten times" preamble is dropped.

I merged main so #9720 is in, and linked its real anchors: `template-properties.md#configuration-input-type` and `template-metadata.md#embedding-configurations-configurationtemplates`.

@alexronquillo to your question on the same thread: no change needed to `template-properties.md` from this PR, #9720 covers it and has now merged.

crobbins215:
Pushed. Two changes:

**1. About section reframed** per @nikku's feedback, replies inline on that thread. Also merged `main` so #9720 is in, and linked its real anchors instead of a bare page link.

**2. @chillleader — the credential template page is no longer a stub.** I drafted it out so it needs a technical review from you rather than authorship: [`credential-templates.md`](https://github.com/camunda/camunda-docs/blob/docs/connector-credentials/docs/components/connectors/custom-built-connectors/credential-templates.md).

It now covers:

- What a credential template is (a configuration template with `kind: CREDENTIAL`), and its top-level fields
- A worked AWS Credential example with a conditional auth-type dropdown, plus the stored value it produces
- Embedding under `configurationTemplates`, including the canonical-definition rule for repeated embedding and why you embed two versions to support in-place upgrades
- Declaring the `Configuration` chooser property, the binding to a dedicated connector input, and `configurationTemplateVersion` as a floor rather than an exact match
- Keeping an inline fallback behind an `isEmpty` condition, and when to make the chooser required instead
- `secret: true` fields and `camunda.secrets.<KEY>` resolution
- Forward-compatible-only versioning, and a note that testing a credential is the real safety step rather than the version number

Sources I worked from, all cited in a comment at the top of the file: `connections-design` (`CONFIGURATION_TEMPLATE.md`, `CONFIGURATION_ELEMENT_TEMPLATE.md`, `CONFIGURATION_VERSIONING.md`, `CONFIGURATION_FIELD_REFERENCES.md`) and `camunda/connectors` (`RestAuthenticationConfiguration.java`, `http-json-connector.json`, and your team's `docs/credentials-in-element-templates.md`).

Worth a specific check from you:

- Whether the JSON examples match what the generator actually emits today. I wrote them by hand from the design docs, not by regenerating a template.
- Whether `secret: true` is the shipped field name, and whether the editor treats it as a hint or enforces it.
- I deliberately left out the Java-side authoring path (`@Configuration`, `@TemplateProperty(type = Configuration)`, `@ElementTemplate(configurations = ...)`) since this page sits under Build custom connectors and the audience is hand-authoring JSON. Tell me if the Java path belongs here too.

The four open questions from the PR description are still unanswered, if anyone can pick them up.

Build passes.

docs/components/connectors/custom-built-connectors/credential-templates.md
```diff
@@ -0,0 +1,202 @@
...
+
+| Field         | Required | Description                                                                                                      |
+| ------------- | -------- | ---------------------------------------------------------------------------------------------------------------- |
+| `id`          | Yes      | Uniquely identifies the credential template. Referenced by a `Configuration` property's `configurationTemplate`. |
+| `name`        | Yes      | Display name shown in the credential editor, and the default label for the picker.                               |
+| `version`     | Yes      | An integer. Bump it when the credential's shape changes (see [Versioning](#versioning-a-credential-template)).   |
+| `kind`        | Yes      | The kind of configuration. Set this to `CREDENTIAL`.                                                             |
+| `description` | No       | Short description shown in the credential editor.                                                                |
+| `properties`  | Yes      | The fields that make up the credential's stored value (see below).                                               |
+
```
chillleader, line 28: nit: this is documented already in the modeler / bpmn.io docs, we could just link it here instead of repeating

docs/components/connectors/custom-built-connectors/credential-templates.md
```diff
@@ -0,0 +1,202 @@
...
+      "version": 1,
+      "kind": "CREDENTIAL",
+      "properties": [ ... ]
+    }
+  ],
+  "properties": [ ... ]
+}
+```
+
+If more than one connector uses the same credential type, embed the identical `configurationTemplates` entry (same `id` and `version`) in each. The first copy Modeler loads becomes the canonical definition; later copies are only accepted if they match it exactly. If your connector needs to support in-place upgrades of a bound credential to a newer template version, embed both versions. Modeler needs the source version to read the existing value, and the target version to render the upgraded form.
```
chillleader: We can mention here that our [element template generator](https://github.com/camunda/connectors/tree/main/element-template-generator) is able to manage those embedded schemas automatically, generate them from the Java code and insert into every template where they are needed

docs/components/connectors/custom-built-connectors/credential-templates.md
```diff
@@ -0,0 +1,202 @@
...
+
+If your connector already has inline authentication fields, keep them. Hide them once a credential is selected, using an `isEmpty` condition on the chooser property:
+
+```json
+{
+  "id": "accessKey",
+  "type": "String",
+  "condition": { "property": "awsCredential", "isEmpty": true }
+}
+```
```
chillleader: Worth mentioning as a fallback pattern but IMO our recommendation should be to use credentials only, especially when using the new connector. This also requires explicit runtime support as mentioned below so it's not something that always works out of the box - more of an ad-hoc way to ensure backward compatibility.

docs/components/connectors/custom-built-connectors/credential-templates.md
```diff
@@ -0,0 +1,202 @@
...
+DRAFT: @chillleader please verify against current 8.10 behavior before this merges, per the
+PDP-3396 epic (https://github.com/camunda/product-hub/issues/3396) transition-to-implement
+notes assigning custom-credential guides to the Connectors team. Content below is grounded in
+camunda/connections-design (design/CONFIGURATION_TEMPLATE.md, CONFIGURATION_ELEMENT_TEMPLATE.md,
+CONFIGURATION_VERSIONING.md, CONFIGURATION_FIELD_REFERENCES.md) and the shipped REST connector
+(camunda/connectors: RestAuthenticationConfiguration.java, http-json-connector.json,
+docs/credentials-in-element-templates.md). Check it still matches what ships.
+-->
+
+Define a credential type for your custom connector, so users can create and select a credential for it in Camunda Hub or Modeler, instead of configuring authentication fields directly on the connector task. See [Credentials](/components/hub/organization/credentials/index.md) for the concept this page builds on, and [Configuration input type](/components/modeler/element-templates/template-properties.md#configuration-input-type) for the generic `Configuration` property type it uses.
```
alexronquillo:
```suggestion
Define a credential type for your custom connector, so users can create and select a credential for it in Camunda Hub or Modeler, instead of configuring authentication fields directly on the connector task. See [credentials](/components/hub/organization/credentials/index.md) for the concept this page builds on, and [`Configuration` input type](/components/modeler/element-templates/template-properties.md#configuration-input-type) for the generic `Configuration` property type it uses.
```

docs/components/connectors/custom-built-connectors/credential-templates.md
```diff
@@ -0,0 +1,202 @@
...
+
+## About credential templates
+
+A **credential template** is a [configuration template](/components/modeler/element-templates/template-metadata.md#embedding-configurations-configurationtemplates) whose `kind` is `CREDENTIAL`. It defines the fields a credential of that type has, and how they render in the credential editor in Camunda Hub and Modeler.
```
alexronquillo:
```suggestion
A **credential template** is a [configuration template](/components/modeler/element-templates/template-metadata.md#embedding-configurations-configurationtemplates) whose `kind` is `CREDENTIAL`. It defines the fields a credential of that type has, and how they render in the credential editor in Camunda Hub and Desktop Modeler.
```

docs/components/connectors/custom-built-connectors/credential-templates.md
```diff
@@ -0,0 +1,202 @@
...
+
+A credential template is not an element template property. It's a separate, self-contained schema, embedded in your connector's element template under the top-level `configurationTemplates` key. Your connector's element template then declares a `Configuration`-type property that locks to it, which renders as the credential picker in the properties panel.
```
alexronquillo:
```suggestion
A credential template is not an element template property. It's a separate, self-contained schema, embedded in your connector's element template under the top-level `configurationTemplates` key. Your connector's element template then declares a `Configuration`-type property that locks to it, which renders as the credential chooser in the properties panel.
```

We used "chooser" in the template doc.

docs/components/connectors/custom-built-connectors/credential-templates.md
alexronquillo:
```suggestion
If more than one connector uses the same credential type, embed the identical `configurationTemplates` entry (same `id` and `version`) in each. The first copy the modeler loads becomes the canonical definition; later copies are only accepted if they match it exactly. If your connector needs to support in-place upgrades of a bound credential to a newer template version, embed both versions. The modeler needs the source version to read the existing value, and the target version to render the upgraded form.
```

As this can be Desktop Modeler or the modeling interface in Camunda Hub, I think we should use lowercase modeler to not sound like a particular product.

docs/components/connectors/custom-built-connectors/credential-templates.md
```diff
@@ -0,0 +1,202 @@
...
+  ],
+  "properties": [ ... ]
+}
+```
+
+If more than one connector uses the same credential type, embed the identical `configurationTemplates` entry (same `id` and `version`) in each. The first copy Modeler loads becomes the canonical definition; later copies are only accepted if they match it exactly. If your connector needs to support in-place upgrades of a bound credential to a newer template version, embed both versions. Modeler needs the source version to read the existing value, and the target version to render the upgraded form.
+
+## Add a credential field to your element template
+
+Declare a `Configuration`-type property that locks to your credential template, and bind it to your connector's dedicated configuration input:
```
alexronquillo:
```suggestion
Declare a `Configuration` property that locks to your credential template, and bind it to your connector's dedicated configuration input:
```

Can we drop the "-type"?

docs/components/connectors/custom-built-connectors/credential-templates.md
```diff
@@ -0,0 +1,202 @@
...
+{
+  "id": "accessKey",
+  "label": "Access key",
+  "type": "String",
+  "secret": true,
+  "binding": { "type": "property", "name": "authentication.accessKey" }
+}
+```
+
+This is a rendering hint for the credential editor in Hub and Modeler. It doesn't restrict what the field can hold, but it tells the editor to treat entered values as secret references rather than literals. A user enters an existing secret's key, and the editor stores it as `camunda.secrets.<KEY>`. The engine resolves this reference when the job worker or connector task activates; your connector never sees the marker itself, only the resolved value.
```
alexronquillo:
```suggestion
This is a rendering hint for the credential editor in Hub and Desktop Modeler. It doesn't restrict what the field can hold, but it tells the editor to treat entered values as secret references rather than literals. A user enters an existing secret's key, and the editor stores it as `camunda.secrets.<KEY>`. The engine resolves this reference when the job worker or connector task activates; your connector never sees the marker itself, only the resolved value.
```

alexronquillo (changes requested):
Thanks for the draft! I gave some quick feedback. I didn't want to push changes that might collide with changes you might be making to address the technical review.

nikku:
https://github.com/camunda/camunda-docs/pull/9720 is merged in case you want to sensibly cross-link.

docs/components/connectors/custom-built-connectors/credential-templates.md
crobbins215: Good call. Dropped the repetition, now just links to [element template properties](https://docs.camunda.io/docs/next/components/modeler/element-templates/template-properties/) and states only the three credential-specific restrictions: `property` binding only, no nested `Configuration`, no `optional`.

docs/components/connectors/custom-built-connectors/credential-templates.md
crobbins215: Added as a tip under that paragraph, pointing at the [element template generator](https://github.com/camunda/connectors/tree/main/element-template-generator): if you build in Java you don't maintain the embedded schemas by hand, it generates them from the Java code and inserts them into every template that needs them.

docs/components/connectors/custom-built-connectors/credential-templates.md
crobbins215: Reframed. The section is now **Supporting inline fields as a fallback** and leads with the recommendation: for a new connector, use credentials only, mark the `Configuration` property `notEmpty`, and don't ship inline authentication fields at all. Inline fallback is then described as a backward-compatibility pattern for a connector that already shipped with inline auth, and I called out explicitly that it isn't automatic, your connector runtime has to resolve which of the two sources to use.

docs/components/connectors/custom-built-connectors/credential-templates.md
crobbins215: Applied.

docs/components/connectors/custom-built-connectors/credential-templates.md
crobbins215: Applied.

docs/components/connectors/custom-built-connectors/credential-templates.md
crobbins215: Applied, and I made "chooser" consistent everywhere else too: the credentials landing page, the Hub modeling interface page, and the Desktop Modeler page all said "picker" and now say "chooser", including the **What you can do in the chooser** headings.

docs/components/connectors/custom-built-connectors/credential-templates.md
crobbins215: Applied. Agreed on lowercase where it could be either product, so this paragraph now says "the modeler" throughout, and the bullets below it too. I kept "Desktop Modeler" capitalized where the sentence names that specific product, per your other two suggestions.

docs/components/connectors/custom-built-connectors/credential-templates.md
crobbins215: Applied.

docs/components/connectors/custom-built-connectors/credential-templates.md
crobbins215: Applied.

crobbins215:
Thanks both. All review comments applied, replies inline, threads resolved.

@nikku on cross-linking now that #9720 is merged: the credentials pages already link forward to [`Configuration` input type](https://docs.camunda.io/docs/next/components/modeler/element-templates/template-properties/#configuration-input-type), [`configurationTemplates`](https://docs.camunda.io/docs/next/components/modeler/element-templates/template-metadata/#embedding-configurations-configurationtemplates), and the worked example template. I've now added the return links too, so it goes both ways:

- The `CREDENTIAL` note in `template-properties.md` links to the credentials page, and to the credential template guide for defining your own type.
- The `kind` bullet in `template-metadata.md` points at the worked `CREDENTIAL` example.

@chillleader the three substantive points are in: credentials-only is now the stated recommendation for new connectors with inline fields reframed as a backward-compatibility pattern needing explicit runtime support; the property shape rules link out instead of repeating; and the element template generator is called out as handling the embedded schemas from Java.

@alexronquillo all six suggestions applied verbatim, and I propagated "picker" to "chooser" across the other three pages so it's consistent with the element template reference.

crobbins215:
I think this looks good, and we can go with it and create follow-up PRs rather than keep adjusting now @alexronquillo.

alexronquillo (approved):

nikku (commented):
Great stuff, great work. Thanks for cross-linking the concerns, too ⭐

PR #9770, opened 2026-08-31
docs(configuration): document camunda.secrets.* file store and cache

### LINKED ISSUE #60331: [Secret Resolution] Docs: configure the file secret store and cache (Self-Managed)
Parent: #56591

### Description

Document the file secret store, the secret cache, and the per-physical-tenant configuration surface, for Self-Managed.

Scope is deliberately the file store and everything shared by all stores. The AWS and GCP stores are split out into their own ticket, because they carry their own authentication and IAM detail and need different reviewers.

### Motivation

Nothing works until an operator has configured a store, and none of these properties are documented anywhere today.

### Acceptance criteria

- [ ] New property reference partial created under `docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/` and imported into `properties.md`
- [ ] File store documented: `camunda.secrets.stores.file.<id>.path`, default `/etc/camunda/secrets`
- [ ] Cache documented: `camunda.secrets.cache.ttl` (default `20m`, minimum `1m`, whole minutes only) and `camunda.secrets.cache.max-size` (default `1000`). One cache per store, so `max-size` bounds each store separately and the worst-case footprint is stores multiplied by `max-size`. TTL is what lets a rotated secret be picked up without a restart
- [ ] Documented that exactly one store per physical tenant is supported and its id must be `default`; any other id is rejected at startup
- [ ] Per-physical-tenant overrides documented via `camunda.physical-tenants.<id>.secrets.*`, and `docs/self-managed/concepts/physical-tenants/configuration-reference.md` updated
- [ ] Startup validation failures documented, since they are operator-visible error messages
- [ ] Documented that a store entry whose name cannot form a valid `camunda.secrets.<name>` reference is silently omitted from list results and cannot be resolved through `/v2/secrets`. The reference charset is `[\p{Alnum}_-]+`, so a dash is valid; a dot is the common case that is not
- [ ] The partial is structured so the cloud store ticket adds to it without reworking it

### Breakdown

- Write the partial and the import
- Update the physical tenants configuration reference

### Notes

Source of truth is `configuration/src/main/java/io/camunda/configuration/Secrets.java` in `camunda/camunda`.

The configuration property partials are hand-maintained — no generator exists in either repository, and the partials' git history is ordinary manual commits. The "harvested into the operator-facing configuration reference" javadoc on `Secrets.java` describes an intent, not a pipeline. Write the partial by hand.

SaaS availability is now tracked in its own ticket and is out of scope here.

HeleneW-dot:
## Description

Adds the property reference for the file secret store and the secret cache, plus the per-physical-tenant override surface and startup validation messages.

based off PR for [docs(concepts): document broker secret resolution and its effect on job activation ](https://github.com/camunda/camunda-docs/pull/9703), merge that one first.
related to [#60331](https://github.com/camunda/camunda/issues/60331)

## PR Checklist

### DIFF
--- a/docs/components/concepts/secret-resolution-and-job-activation.md
+++ b/docs/components/concepts/secret-resolution-and-job-activation.md
@@ -30,11 +30,7 @@ The scheduler then works through the references that are still pending:
 
 Resolution records carry no secret values. Only the store's cache holds a value, and only for as long as its cache entry lives.
 
-A cached value expires a fixed time after it is written, regardless of when it was last read. The store can also evict the value earlier if its cache is full. If the value is no longer cached when the broker tries to activate a job, the broker parks the job and resolves the reference again. This re-resolution also makes rotated secrets available to workers without a restart. The cache lifetime and size are configured per store under `camunda.secrets.cache`.
-
-<!-- The cache configuration (camunda.secrets.cache.ttl, camunda.secrets.cache.max-size) is owned
-by camunda/camunda#60331. Link its cache section here once it lands, instead of restating the
-defaults. -->
+A cached value expires a fixed time after it is written, regardless of when it was last read. The store can also evict the value earlier if its cache is full. If the value is no longer cached when the broker tries to activate a job, the broker parks the job and resolves the reference again. This re-resolution also makes rotated secrets available to workers without a restart. The cache lifetime and size are configured per store under [`camunda.secrets.cache`](/self-managed/components/orchestration-cluster/core-settings/configuration/properties.md#camundasecretscache).
 
 Two kinds of failure are treated differently:
 
--- a/docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/_secrets.md
+++ b/docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/_secrets.md
@@ -5,24 +5,28 @@ import TabItem from '@theme/TabItem';
 
 Configures the secret stores and cache used to resolve `camunda.secrets.<name>` references in process variables.
 
+This configuration is part of an [alpha feature](/components/early-access/alpha/alpha-features.md) and may be subject to change in future releases.
+
 <!-- TODO(camunda/camunda#60326): link the secret resolution concept page here once its title, placement, and the feature's terminology are finalized. -->
 
-`camunda.secrets.*` sets the root-level defaults, applied to the `default` physical tenant. Override those defaults per physical tenant under `camunda.physical-tenants.<tenant-key>.secrets.*`. See [Validation and constraints](/self-managed/concepts/physical-tenants/configuration-reference.md#validation-and-constraints) in the Physical Tenants configuration reference.
+`camunda.secrets.*` sets the defaults inherited by every physical tenant. Override them per physical tenant under `camunda.physical-tenants.<tenant-key>.secrets.*`. See [Validation and constraints](/self-managed/concepts/physical-tenants/configuration-reference.md#validation-and-constraints) in the Physical Tenants configuration reference.
+
+`<id>` is the store's identifier. Exactly one secret store is supported per physical tenant, and its `<id>` must be `default`. This check runs against the merged, per-tenant configuration: a tenant that already inherits a store cannot add a second one under a different id to override it, since that is two stores and fails startup the same way any other unsupported id does. A tenant overrides an inherited store only by reusing the id `default`.
+
+A secret name must match `[\p{Alnum}_-]+` and be at most 240 characters to be listed and resolved through `/v2/secrets` (the reference `camunda.secrets.<name>` is capped at 256 characters, and the `camunda.secrets.` prefix takes 16 of those). A dot is the common case that falls outside the charset, for example a file extension. A name that fails either check is stored, but it is silently omitted from list results and cannot be resolved by reference.
+
+<!-- AWS and GCP secret stores (camunda.secrets.stores.aws.<id>.* and camunda.secrets.stores.gcp.<id>.*) are tracked in camunda/camunda#60966 and documented separately. -->
 
 <Tabs>
   <TabItem value="application.yaml" label="Application properties">
 
 ### `camunda.secrets.stores.file`
 
-`<id>` is the store's identifier. Exactly one secret store is supported per physical tenant, and its `<id>` must be `default`. A store configured under any other id is rejected at startup.
+The `path` default below only takes effect once a `stores.file.<id>` entry exists somewhere in the merged configuration. With no file store declared, the physical tenant gets a no-op store instead: nothing is resolvable, and no directory is read.
 
-| Property                                | Description                                                                                                                                                                 | Default value          |
-| :-------------------------------------- | :-------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | :--------------------- |
-| `camunda.secrets.stores.file.<id>.path` | Path to the directory backing this file-based secret store. Each file in the directory is one secret: the file name is the secret name and the file contents are the value. | `/etc/camunda/secrets` |
-
-A secret name must match `[\p{Alnum}_-]+` to be listed and resolved through `/v2/secrets`. A dot is the common case that falls outside this charset, for example a file extension. Such a name is stored, but it is silently omitted from list results and cannot be resolved by reference.
-
-<!-- AWS and GCP secret stores (camunda.secrets.stores.aws.<id>.* and camunda.secrets.stores.gcp.<id>.*) are tracked in camunda/camunda#60966 and documented separately. -->
+| Property                                | Description                                                                                                                                                                                             | Default value          |
+| :-------------------------------------- | :------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ | :--------------------- |
+| `camunda.secrets.stores.file.<id>.path` | Path to the directory backing this file-based secret store, once the store is declared. Each file in the directory is one secret: the file name is the secret name and the file contents are the value. | `/etc/camunda/secrets` |
 
 ### `camunda.secrets.cache`
 
@@ -38,9 +42,9 @@ One cache is created per configured store.
 
 ### `CAMUNDA_SECRETS_STORES_FILE`
 
-| Property                                | Description                                                                                                                                                                 | Default value          |
-| :-------------------------------------- | :-------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | :--------------------- |
-| `CAMUNDA_SECRETS_STORES_FILE_<id>_PATH` | Path to the directory backing this file-based secret store. Each file in the directory is one secret: the file name is the secret name and the file contents are the value. | `/etc/camunda/secrets` |
+| Property                                | Description                                                                                                                                                                                             | Default value          |
+| :-------------------------------------- | :------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ | :--------------------- |
+| `CAMUNDA_SECRETS_STORES_FILE_<id>_PATH` | Path to the directory backing this file-based secret store, once the store is declared. Each file in the directory is one secret: the file name is the secret name and the file contents are the value. | `/etc/camunda/secrets` |
 
 ### `CAMUNDA_SECRETS_CACHE`
 
--- a/docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/_secrets.md
+++ b/docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/_secrets.md
@@ -7,7 +7,7 @@ Configures the secret stores and cache used to resolve `camunda.secrets.<name>` 
 
 <!-- TODO(camunda/camunda#60326): link the secret resolution concept page here once its title, placement, and the feature's terminology are finalized. -->
 
-`camunda.secrets.*` sets the root-level defaults, applied to the `default` physical tenant. Override those defaults per physical tenant under `camunda.physical-tenants.<tenant-key>.secrets.*`. See the [Physical Tenants configuration reference](/self-managed/concepts/physical-tenants/configuration-reference.md).
+`camunda.secrets.*` sets the root-level defaults, applied to the `default` physical tenant. Override those defaults per physical tenant under `camunda.physical-tenants.<tenant-key>.secrets.*`. See [Validation and constraints](/self-managed/concepts/physical-tenants/configuration-reference.md#validation-and-constraints) in the Physical Tenants configuration reference.
 
 <Tabs>
   <TabItem value="application.yaml" label="Application properties">
--- a/docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/_secrets.md
+++ b/docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/_secrets.md
@@ -1,0 +1,53 @@
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+## Secrets
+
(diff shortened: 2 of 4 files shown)


docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/_secrets.md
```diff
@@ -0,0 +1,53 @@
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+## Secrets
```
ajanoni-camunda, line 4: 🔧 The sibling reference pages for this same feature are marked as alpha: `list-secrets.api.mdx` and `resolve-secrets.api.mdx` are both titled "... (alpha)", carry `<MarkerAddedInVersion version="8.10" />`, and state "This endpoint is an [alpha feature](/components/early-access/alpha/alpha-features.md)".

This section documents the configuration side of that same alpha feature with no alpha marker and no version marker, on a page readers treat as GA reference. Either add a marker here or make an explicit decision that configuration is not alpha even though the API is.

Related: `components/early-access/alpha/alpha-features.md` currently has no secrets entry at all, even though both API pages link to it as an alpha feature. That predates this PR, and the concepts page in the base branch already flags it as owned by camunda/camunda#60326, so this is a pointer rather than a request to fix it here.

docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/_secrets.md
```diff
@@ -0,0 +1,53 @@
...
+
+Configures the secret stores and cache used to resolve `camunda.secrets.<name>` references in process variables.
+
+<!-- TODO(camunda/camunda#60326): link the secret resolution concept page here once its title, placement, and the feature's terminology are finalized. -->
+
+`camunda.secrets.*` sets the root-level defaults, applied to the `default` physical tenant. Override those defaults per physical tenant under `camunda.physical-tenants.<tenant-key>.secrets.*`. See [Validation and constraints](/self-managed/concepts/physical-tenants/configuration-reference.md#validation-and-constraints) in the Physical Tenants configuration reference.
```
ajanoni-camunda: ❌ This is not what the code does, and it contradicts two other places in the docs.

Root-level `camunda.secrets.*` is inherited by **every** physical tenant, per property, not only by `default`:

- `PhysicalTenantResolver.java:96-99` binds the root prefix into a fresh `Camunda` per tenant, then binds the per-tenant prefix on top of the same instance. The loop runs for every discovered tenant id.
- `PhysicalTenantMapOverlay.java:107-117` re-binds each root map key onto the tenant instance, so a partial tenant block layers only its delta over the fully populated root POJO.
- `PhysicalTenantSecretsOverlayTest.java:101` (`shouldInheritRootFileStoreWhenTenantHasNoSecretsOverride`) and `:187` assert exactly this for the non-default tenant `tenanta`. `:408-421` sets only `camunda.secrets.cache.ttl=10m` / `max-size=500` at root level and asserts `tenanta` gets 10m/500 rather than the built-in 20m/1000.

It also contradicts:

- This page's own next sentence, "Override those defaults per physical tenant", which only makes sense if root is the default for all tenants.
- `physical-tenants/configuration-reference.md`, "Use root-level `camunda.*` for shared defaults across all tenants."

Suggested: "`camunda.secrets.*` sets the defaults inherited by all physical tenants. Override them per physical tenant under `camunda.physical-tenants.<tenant-key>.secrets.*`."

docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/_secrets.md
```diff
@@ -0,0 +1,53 @@
...
+
+<Tabs>
+  <TabItem value="application.yaml" label="Application properties">
+
+### `camunda.secrets.stores.file`
+
+`<id>` is the store's identifier. Exactly one secret store is supported per physical tenant, and its `<id>` must be `default`. A store configured under any other id is rejected at startup.
```
ajanoni-camunda: 🔧 Worth one more clause here, because "override the defaults per tenant" plus "exactly one store, id must be `default`" invites a specific mistake: declaring a root store and then a tenant store under a different id, expecting the tenant one to win.

The store cap is applied to the **merged** per-tenant config, and the id check is applied to whatever survives the merge (`SecretStoreConfiguration.java:121-130` and `:234-254`). So a root `stores.file.default` plus a tenant `stores.file.primary` is a startup failure, not an override. A tenant can only override the store it inherits, by reusing the id `default`.

docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/_secrets.md
```diff
@@ -0,0 +1,53 @@
...
+
+| Property                                | Description                                                                                                                                                                 | Default value          |
+| :-------------------------------------- | :-------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | :--------------------- |
+| `camunda.secrets.stores.file.<id>.path` | Path to the directory backing this file-based secret store. Each file in the directory is one secret: the file name is the secret name and the file contents are the value. | `/etc/camunda/secrets` |
```
ajanoni-camunda: 🔧 The default value reads as though a file store always exists at this path. It does not.

`/etc/camunda/secrets` is a field default on `FileStore` (`Secrets.java:235`) and only materializes once a `stores.file.<id>` entry exists with `path` unset. With no `stores.file` entry anywhere in the merged config, the tenant gets a NOOP store: `SecretStoreConfiguration.java:155-161` registers it and logs "No secret stores configured for physical tenant".

Suggest saying that the path default applies once the store is declared, so nobody reads this table as "there is a file store at `/etc/camunda/secrets` unless you change it".

docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/_secrets.md
```diff
@@ -0,0 +1,53 @@
...
+
+A secret name must match `[\p{Alnum}_-]+` to be listed and resolved through `/v2/secrets`. A dot is the common case that falls outside this charset, for example a file extension. Such a name is stored, but it is silently omitted from list results and cannot be resolved by reference.
```
ajanoni-camunda: ❌ The charset is right, but it contradicts a page that is already published.

`REFERENCE_NAME_PATTERN = Pattern.compile("[\\p{Alnum}_-]+")` (`SecretServices.java:97`) allows the dash, so this line is correct. But `apis-tools/orchestration-cluster-api-rest/specifications/list-secrets.api.mdx:46-48` says names "containing a dot **or a dash**" are omitted. Those two statements cannot both be true, and the API page is the wrong one; the code comment at `SecretServices.java:388` also uses only the dot as its example.

The source of the error is upstream, in the OpenAPI description (`config/openapi/v2/secrets.yaml:79` in `camunda/camunda`), so the fix is a spec change plus a re-sync rather than an edit here. Flagging it so this PR does not ship a partial that contradicts the API reference next to it.

🔧 Second point on the same sentence: the charset is not the only condition. `isResolvableName` (`SecretServices.java:458-462`) also enforces `MAX_REFERENCE_LENGTH = 256` on prefix plus name, and with the 16-character `camunda.secrets.` prefix that caps the name at 240 characters. An over-long name is silently omitted from listings in exactly the same way a dotted name is, so it belongs in the same sentence.

docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/_secrets.md
```diff
@@ -0,0 +1,53 @@
...
+| `camunda.secrets.cache.ttl`      | How long a resolved secret is served from the cache before it is fetched from the store again, so a secret rotated in the store is picked up without a restart. Must be at least `1m` and a whole number of minutes; a shorter or fractional value is rejected at startup.                                                                              | `20m`         |
+| `camunda.secrets.cache.max-size` | Maximum number of secrets held in each store's cache. Once reached, caching another secret evicts one already held; which one is the cache implementation's choice. This is a per-cache limit, not a budget shared across stores, so the worst-case memory footprint is the number of configured stores multiplied by this value. Must be at least `1`. | `1000`        |
+
+  </TabItem>
+  <TabItem value="env" label="Environment variables">
+
+### `CAMUNDA_SECRETS_STORES_FILE`
```
ajanoni-camunda, line 43: 🙃 Nitpick, take or leave: everything a reader needs in order to configure this correctly, the `<id>` must be `default` rule, the name charset, and the AWS/GCP note, lives only in the Application properties tab. The Environment variables tab is bare tables, so a reader who lands on it never sees any of the constraints.

Not a house convention either way, since the sibling partials (`_cluster.md`, `_data.md`) carry no prose at all to duplicate. Either repeat the two constraint sentences in the env tab, or move them above `<Tabs>` so they apply to both.

docs/components/concepts/secret-resolution-and-job-activation.md
```diff
@@ -0,0 +1,199 @@
...
+
+For how to inspect and resolve either incident, see [incidents](incidents.md).
+
+<!-- The incident messages, the diagnosis steps, and the operator runbook are owned by
+camunda/camunda#60964. Link into that content from here once it lands, instead of restating it. -->
+
+## Tune the resolution scheduler
+
+The scheduler is configured under `camunda.processing.engine.secrets`. The defaults suit a store
+that responds in well under a second.
```
ajanoni-camunda: 🔧 Removing the `camunda.secrets.cache.ttl` sentence here is the right call now that this PR documents the cache properly, but nothing replaced the pointer.

After this change the concepts page mentions the cache four times, including "only for as long as its cache entry lives" (line 42) and "distinguishes a cache that is simply cold from a store that is not answering", but never says where cache lifetime is configured or links to the new section. The base branch at least had "The cache lifetime and size are configured per store under `camunda.secrets.cache`" plus an HTML comment saying to link this ticket's content once it landed. It has landed, in this PR, so the link should be added here or in Related resources.

The same HTML comment in the base branch also says to link "instead of restating the defaults", so a link rather than a duplicated `20m` is the intent.

docs/self-managed/concepts/physical-tenants/configuration-reference.md
```diff
@@ -102,7 +102,7 @@ For backward compatibility:
 
 ## Validation and constraints
 
-At startup, configuration validation enforces tenant-level constraints. All validation failures throw a `UnifiedConfigurationException` and prevent the cluster from starting. There is no separate error code, and the message is reported at startup rather than logged as a warning. For the exact error message when a tenant is missing `providers.assigned`, see [IdP provider assignment](./authentication-authorization.md#idp-provider-assignment).
+At startup, configuration validation enforces tenant-level constraints. Most validation failures throw a `UnifiedConfigurationException` and prevent the cluster from starting. Secret store and cache validation (see **Secrets** below) is the exception: it throws an `IllegalStateException` or an `IllegalArgumentException` directly instead. There is no separate error code, and the message is reported at startup rather than logged as a warning. For the exact error message when a tenant is missing `providers.assigned`, see [IdP provider assignment](./authentication-authorization.md#idp-provider-assignment).
```
ajanoni-camunda: ❓ Should this also say that secret misconfiguration prevents the cluster from starting?

As reworded, "prevent the cluster from starting" attaches only to the "Most validation failures" clause, and the secrets sentence that follows only says which exception types are thrown instead. The bullet further down, "Validation failures are startup failures, not runtime warnings", covers it globally, but a reader working through this paragraph could reasonably conclude the secret exceptions are handled differently in that respect too.

The exception types themselves are confirmed correct: `Secrets.java` and `SecretStoreConfiguration.java:124,244,260` throw `IllegalArgumentException` / `IllegalStateException`, while the sibling validators under `configuration/physicaltenants/` use `UnifiedConfigurationException`. Good catch documenting a real inconsistency rather than papering over it.

docs/self-managed/concepts/physical-tenants/configuration-reference.md
```diff
@@ -143,6 +144,41 @@ store location [provider=aws, namespace=[company-docs-bucket], keyPrefix='tenant
 
 If one tenant's path is nested inside another's rather than identical, the message instead reads: `tenant <enclosing> 's document store location [...] encloses tenant <enclosed> 's [...]`.
 
+**Secret store or cache misconfiguration:**
```
ajanoni-camunda, line 147: 👍 Checked all six of these against code and they match verbatim, including the rename guidance built from the `camunda.secrets.stores.` / `camunda.physical-tenants.<tenant>.secrets.stores.` prefixes: `SecretStoreConfiguration.java:124-129`, `:244-253`, `:256-267`, and `Secrets.java:130-227` for the three cache messages.

The closing note about the cache messages always reporting the canonical `camunda.secrets.cache.*` path is also correct and is exactly the kind of thing an operator grepping for a property name needs.

docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/_secrets.md
HeleneW-dot: Youre right, adjusted this to fix the wording to say its inherited by every tenant.

docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/_secrets.md
HeleneW-dot: Added a sentence explaining the store cap runs against the merged config, so overriding only works by reusing the `default` ID

docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/_secrets.md
HeleneW-dot: Clarified the default only applies once the store is declared

docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/_secrets.md
HeleneW-dot: Added the 240-character cap. 
Re the dash contradiction: on camunda/camunda: it's already fixed there so the spec's correct now but like you said it needs a resync, which should automatically happen every thursday so should be addressed tomorrow

docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/_secrets.md
HeleneW-dot: Moved the constraint sentences above the tabs so both app-properties and env-var readers see them

docs/components/concepts/secret-resolution-and-job-activation.md
HeleneW-dot: Found the actual TODO for this (owned by [camunda/camunda#60331](https://github.com/camunda/camunda/issues/60331)) a bit earlier in the file and resolved it there with a real link to the new cache section

docs/self-managed/concepts/physical-tenants/configuration-reference.md
HeleneW-dot: Yes I think so, reworded so "prevents the cluster from starting: covers both exception paths, not just the UnifiedConfigurationException

docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/_secrets.md
HeleneW-dot: Adjusted for now with #60326 owning the final alignment

ajanoni-camunda (dismissed):
🟢 LGTM

HeleneW-dot:
@giorgionaps This one is now also ready for tech writer review

lev-safro (approved):

PR #9781, opened 2026-09-01
docs: document secret resolution and cache metrics

### LINKED ISSUE #60963: [Secret Resolution] Docs: secret resolution and cache metrics
### 

### Description

Document the metrics emitted by secret resolution and the secret caches.

Six meters ship in 8.10 and none appear in the metrics reference:

| Meter | Source |
|---|---|
| `camunda.secret.resolution.duration` | `SecretResolutionMetricsDoc` |
| `camunda.secret.resolution.outcome` | `SecretResolutionMetricsDoc` |
| `camunda.secret.resolution.cycle.error` | `SecretResolutionMetricsDoc` |
| `camunda.secret.cache.result` | `SecretCacheMetricsDoc` |
| `camunda.secret.cache.evictions` | `SecretCacheMetricsDoc` |
| `camunda.secret.cache.size` | `SecretCacheMetricsDoc` |

### Motivation

Secret resolution sits between job activation and an external store, so it is exactly the kind of dependency an operator needs to watch. Without documented meters, a slow or failing store shows up only as jobs that do not activate, with no signal pointing at the cause.

### Acceptance criteria

- [ ] All six meters added to the Self-Managed metrics reference, with their type, tags and what a reading means
- [ ] Documented that no meter is tagged by secret name, and why (cardinality, and secret names are customer data)
- [ ] The relationship between `camunda.secret.cache.result` and `camunda.secret.resolution.outcome` documented, so a reader interpreting a miss rate knows which to read first
- [ ] `camunda.secret.cache.size` related to the `camunda.secrets.cache.max-size` setting documented in D5, including that the bound is per store
- [ ] Cross-linked from the configuration content in D5 and the broker path content in D6

### Links

Implementation: https://github.com/camunda/camunda/issues/59315

### Breakdown

- Read the meter documentation classes for the canonical descriptions and tags
- Add the entries to the metrics reference
- Cross-link from D5 and D6

### Notes

Source of truth is `SecretResolutionMetricsDoc` in `zeebe/engine` and `SecretCacheMetricsDoc` in `secret-store/secret-store-api`. Both carry per-meter descriptions written for this purpose; take the wording from there rather than inventing it.

HeleneW-dot:
## Description

Add the six meters emitted for secret reference resolution and the per-store secret cache to the Self-Managed metrics reference, using the descriptions from SecretResolutionMetricsDoc and SecretCacheMetricsDoc as the source of truth.

closes camunda/camunda#60963
base off https://github.com/camunda/camunda-docs/pull/9770, so merge that one first and rebase this one

## PR Checklist

### DIFF
--- a/docs/self-managed/operational-guides/monitoring/metrics.md
+++ b/docs/self-managed/operational-guides/monitoring/metrics.md
@@ -281,6 +281,100 @@ The `REPORT_NAME` and `REPORT_ID` labels identify the evaluated report or dashbo
 
 Report latency metrics are controlled by the `optimize.metrics.report-latency.enabled=true` configuration property. Set the property to `false` to disable report latency metrics.
 
+## Secret resolution and cache metrics
+
+Camunda emits meters for resolving secret references against a secret store, and for the in-memory
+cache that sits in front of each configured store. Together, they distinguish a secret store that
+is slow or unavailable from a cache that is simply cold, in a case that otherwise shows up only as
+jobs that fail to activate.
+
+No meter listed here is tagged by secret name: the cardinality is unbounded, and secret names are
+customer data.
+
+### Secret resolution metrics
+
+These meters cover resolving secret references against a secret store.
+
+| Metric name                             | Type    | Description                                                                                                                                                                                                                                                                                                                                                                  | Labels                        |
+| --------------------------------------- | ------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ----------------------------- |
+| `camunda.secret.resolution.duration`    | Timer   | Latency of one batch resolution call against a secret store, covering the call itself and not the follow-up commands the engine writes for its results. Split by how the call ended, so a store timing out does not distort the latency of the calls that came back.                                                                                                         | `store`, `result` (see below) |
+| `camunda.secret.resolution.outcome`     | Counter | Number of secret reference resolutions that produced an outcome, per store. Every result value is terminal for the reference it counts, so the values can be summed or divided by one another to derive rates. A reference whose store is unavailable but still has retry attempts left is not counted at all, since it has not reached a terminal outcome.                  | `store`, `result` (see below) |
+| `camunda.secret.resolution.cycle.error` | Counter | Number of resolution cycles in which a store failed in a way the engine does not model: an unexpected exception rather than a per-secret failure or an unreachable store. Counted per store. Always indicates a bug, either in the store implementation or in the engine. Counts cycles, not references, so it is a separate meter from `camunda.secret.resolution.outcome`. | `store`                       |
+
+The `store` label carries the ID of the secret store a reference belongs to.
+
+The `result` label carries different value domains depending on the meter, because
+`camunda.secret.resolution.duration` measures a whole batch call while
+`camunda.secret.resolution.outcome` measures one reference at a time:
+
+`result` values on `camunda.secret.resolution.duration`:
+
+| Value               | Description                                                                  |
+| ------------------- | ---------------------------------------------------------------------------- |
+| `RETURNED`          | The store returned, whatever the per-reference results were.                 |
+| `STORE_UNAVAILABLE` | The store could not be reached for this call.                                |
+| `ERROR`             | The store threw something the engine does not model. Always indicates a bug. |
+
+`result` values on `camunda.secret.resolution.outcome`:
+
+| Value               | Description                                                                                                                           |
+| ------------------- | ------------------------------------------------------------------------------------------------------------------------------------- |
+| `RESOLVED`          | The store returned a value for the reference.                                                                                         |
+| `NOT_FOUND`         | The store does not hold the reference.                                                                                                |
+| `ACCESS_DENIED`     | The store refused to read the reference.                                                                                              |
+| `INVALID_REF`       | The reference is not valid for the store.                                                                                             |
+| `UNREADABLE`        | The store holds the reference but could not read a value from it.                                                                     |
+| `STORE_UNAVAILABLE` | The store could not serve the reference at all: either it is not configured, or it could not be reached and no retry attempt is left. |
+
+### Secret cache metrics
+
+Each configured secret store resolves through an in-memory cache. These meters say how well that
+cache is doing its job.
+
+| Metric name                      | Type    | Description                                                                                                                                                                                                                                                                                                                          | Labels                        |
+| -------------------------------- | ------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ | ----------------------------- |
+| `camunda.secret.cache.result`    | Counter | Number of secret cache lookups, per store and per result, so the hit rate is `HIT / (HIT + MISS)`. Every lookup is counted exactly once. A name the store answers permanently (deleted, denied, or an invalid reference) is never cached, so it misses on every lookup for as long as it is referenced. That is not a cache to tune. | `store`, `result` (see below) |
+| `camunda.secret.cache.evictions` | Counter | Number of entries removed from a secret cache, per store and per cause.                                                                                                                                                                                                                                                              | `store`, `cause` (see below)  |
+| `camunda.secret.cache.size`      | Gauge   | Estimated number of entries a secret cache currently holds, per store. Estimated because eviction is asynchronous, so the value can briefly sit above the configured maximum: read it as a level to compare against that maximum, not as an exact count.                                                                             | `store`                       |
+
+The `store` label carries the ID of the secret store whose cache this is.
+
+`result` values on `camunda.secret.cache.result`:
+
+| Value  | Description                                                                            |
+| ------ | -------------------------------------------------------------------------------------- |
+| `HIT`  | The cache held a value for the name.                                                   |
+| `MISS` | The cache held no value for the name, so the caller had to reach the store or give up. |
+
+`cause` values on `camunda.secret.cache.evictions`:
+
+| Value       | Description                                                                                                                                |
+| ----------- | ------------------------------------------------------------------------------------------------------------------------------------------ |
+| `SIZE`      | The cache was full, so it dropped an entry to make room for another.                                                                       |
+| `EXPIRED`   | The entry's time-to-live elapsed.                                                                                                          |
+| `EXPLICIT`  | Something removed the entry by name. In practice, this is a store answering a name permanently (deleted, denied, or an invalid reference). |
+| `COLLECTED` | The entry's key or value was garbage collected.                                                                                            |
+
+### Read cache and resolution metrics together
+
+`camunda.secret.cache.result` and `camunda.secret.resolution.outcome` answer different questions.
+Reading them in the wrong order can make a healthy cache look broken. A falling cache hit rate only
+means the cache is not holding what callers ask for. It does not mean the value was cacheable in the
+first place. A reference that a store answers permanently as not found, denied, or invalid is never
+cached, so it registers a `MISS` on every lookup for as long as it is referenced. Read a low hit rate
+against `camunda.secret.resolution.outcome` first: if the misses concentrate on references that never
+resolve, the fix is not in the cache.
+
+### Cache size and the configured maximum
+
+`camunda.secret.cache.size` is bounded per store by the
+[`camunda.secrets.cache.max-size`](/self-managed/components/orchestration-cluster/core-settings/configuration/properties.md#camunda-secrets-cache)
+property. The bound applies per store, not as a shared budget, so the worst-case memory footprint
+across a deployment is the number of configured stores multiplied by that maximum.
+
+For how the broker resolves secret references before job activation, see
+[Secret resolution and job activation](/components/concepts/secret-resolution-and-job-activation.md).
+
 ## Grafana
 
 ### Zeebe
--- a/docs/self-managed/operational-guides/monitoring/metrics.md
+++ b/docs/self-managed/operational-guides/monitoring/metrics.md
@@ -285,8 +285,8 @@ Report latency metrics are controlled by the `optimize.metrics.report-latency.en
 
 Camunda emits meters for resolving secret references against a secret store, and for the in-memory
 cache that sits in front of each configured store. Together, they distinguish a secret store that
-is slow or unavailable from a cache that is simply cold, in a case that otherwise shows up only as
-jobs that fail to activate.
+is slow or unavailable from a cache that is simply cold, in cases that otherwise show up only as
+jobs that don't activate.
 
 No meter listed here is tagged by secret name: the cardinality is unbounded, and secret names are
 customer data.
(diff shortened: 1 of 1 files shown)


docs/self-managed/operational-guides/monitoring/metrics.md
```diff
@@ -281,6 +281,100 @@ The `REPORT_NAME` and `REPORT_ID` labels identify the evaluated report or dashbo
...
+| `camunda.secret.resolution.outcome`     | Counter | Number of secret reference resolutions that produced an outcome, per store. Every result value is terminal for the reference it counts, so the values can be summed or divided by one another to derive rates. A reference whose store is unavailable but still has retry attempts left is not counted at all, since it has not reached a terminal outcome.                  | `store`, `result` (see below) |
+| `camunda.secret.resolution.cycle.error` | Counter | Number of resolution cycles in which a store failed in a way the engine does not model: an unexpected exception rather than a per-secret failure or an unreachable store. Counted per store. Always indicates a bug, either in the store implementation or in the engine. Counts cycles, not references, so it is a separate meter from `camunda.secret.resolution.outcome`. | `store`                       |
```
ajanoni-camunda: ❌ A fourth resolution meter is missing: `camunda.secret.resolution.cycle.delay` (`SecretResolutionMetricsDoc.java:170`).

It is a Timer, tagged `result` with the values `DRAINING`, `WAKE`, `IDLE_BACKOFF`, and `RETRY_COOLDOWN`, plus the same `physicalTenant` / `partition` tags as the rest (`:189-190`).

So there are four resolution meters and three cache meters, seven in total. The PR description, this section, and the TODO comment on the concepts page all say six.

This is the one I would least want left out. The section opener promises to explain cases that "show up only as jobs that don't activate", and cycle delay is precisely the meter that separates "the scheduler is idle-backing-off" from "the scheduler is draining a backlog" from "the store is in retry cooldown". Without it, a reader diagnosing slow-to-start jobs has the outcome counters but nothing about scheduler cadence.

docs/self-managed/operational-guides/monitoring/metrics.md
```diff
@@ -281,6 +281,100 @@ The `REPORT_NAME` and `REPORT_ID` labels identify the evaluated report or dashbo
...
+
+The `store` label carries the ID of the secret store a reference belongs to.
```
ajanoni-camunda: ❌ The Labels column is incomplete for all three meters, and the same applies to the cache table below.

Every resolution meter returns `PartitionKeyNames.values()` from `getAdditionalKeyNames()` (`SecretResolutionMetricsDoc.java:57,111,150`), which is `physicalTenant` and `partition` (`PartitionKeyNames.java:20,30`). Those are applied as common tags when the registry is wrapped, at `zeebe/broker/.../startup/steps/MetricsStep.java:35`.

The cache meters carry `physicalTenant` as well, since the registry is wrapped per tenant at `SecretStoreConfiguration.java:164-166`. They have no `partition` tag, because the cache lives outside partitions, which is itself worth stating.

Why this matters more than a missing column entry: an operator who writes `sum(camunda_secret_cache_result_total)` from this page silently aggregates across every partition and every physical tenant without knowing it, and there is nothing here to tell them the tenant dimension exists. Given the sibling PR makes per-tenant secret stores a first-class concept, `physicalTenant` is the label most likely to be needed and the one hardest to guess.

Suggest listing all of them in the Labels column, and replacing "The `store` label carries the ID of the secret store a reference belongs to" with a short paragraph covering `store`, `physicalTenant`, and `partition`, noting that the cache meters have no `partition`.

docs/self-managed/operational-guides/monitoring/metrics.md
```diff
@@ -281,6 +281,100 @@ The `REPORT_NAME` and `REPORT_ID` labels identify the evaluated report or dashbo
...
+| --------------------------------------- | ------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ----------------------------- |
+| `camunda.secret.resolution.duration`    | Timer   | Latency of one batch resolution call against a secret store, covering the call itself and not the follow-up commands the engine writes for its results. Split by how the call ended, so a store timing out does not distort the latency of the calls that came back.                                                                                                         | `store`, `result` (see below) |
```
ajanoni-camunda: ❌ These dotted names will not match anything in a Prometheus query, and they break the convention of every other metric table on this page.

The tables above use exported names: `zeebe_incident_events_total`, `zeebe_stream_processor_records_total`, and, two sections up, `optimize_error_total`, which even shows sample scrape output.

I checked for a custom naming convention and there is none: no `NamingConvention` implementation and no renaming `MeterFilter` anywhere in the repo, and `PrometheusConfiguration.java` only calls `throwExceptionOnRegistrationFailure()`. So Micrometer's Prometheus defaults apply and the exported names are:

```
camunda_secret_resolution_duration_seconds  (plus _bucket / _count / _sum)
camunda_secret_resolution_cycle_delay_seconds
camunda_secret_resolution_outcome_total
camunda_secret_resolution_cycle_error_total
camunda_secret_cache_result_total
camunda_secret_cache_evictions_total
camunda_secret_cache_size
```

Note `camunda.secret.cache.size` is the one exception with no unit suffix, deliberately, per `SecretCacheMetricsDoc.java:141-142`.

To be clear, the dotted form is not wrong everywhere: it is the correct form for the `management.metrics.enable.*` filtering property documented earlier on this page. So the fix is to give both forms rather than swap one for the other, the way the Optimize sections give the exported name and this page's Filtering section gives the meter name.

Also worth stating explicitly that the hit rate formula in the `camunda.secret.cache.result` description becomes `camunda_secret_cache_result_total{result="HIT"} / sum without(result) (camunda_secret_cache_result_total)` in practice, since `HIT / (HIT + MISS)` reads as pseudocode today.

docs/self-managed/operational-guides/monitoring/metrics.md
```diff
@@ -281,6 +281,100 @@ The `REPORT_NAME` and `REPORT_ID` labels identify the evaluated report or dashbo
...
+| `MISS` | The cache held no value for the name, so the caller had to reach the store or give up. |
+
+`cause` values on `camunda.secret.cache.evictions`:
+
+| Value       | Description                                                                                                                                |
+| ----------- | ------------------------------------------------------------------------------------------------------------------------------------------ |
+| `SIZE`      | The cache was full, so it dropped an entry to make room for another.                                                                       |
+| `EXPIRED`   | The entry's time-to-live elapsed.                                                                                                          |
+| `EXPLICIT`  | Something removed the entry by name. In practice, this is a store answering a name permanently (deleted, denied, or an invalid reference). |
+| `COLLECTED` | The entry's key or value was garbage collected.                                                                                            |
```
ajanoni-camunda: 🙃 Worth a "not emitted in the current configuration" note on this row.

`CaffeineSecretCache.java:82-90` configures neither weak keys nor soft values, so `COLLECTED` is currently unreachable and the series sits permanently at zero. Harmless to list for completeness, but as written it invites someone to go looking for garbage-collection pressure that this cache cannot report.

docs/self-managed/operational-guides/monitoring/metrics.md
```diff
@@ -281,6 +281,100 @@ The `REPORT_NAME` and `REPORT_ID` labels identify the evaluated report or dashbo
...
+`result` values on `camunda.secret.cache.result`:
+
+| Value  | Description                                                                            |
+| ------ | -------------------------------------------------------------------------------------- |
+| `HIT`  | The cache held a value for the name.                                                   |
+| `MISS` | The cache held no value for the name, so the caller had to reach the store or give up. |
+
+`cause` values on `camunda.secret.cache.evictions`:
+
+| Value       | Description                                                                                                                                |
```
ajanoni-camunda: 👍 Verified all four values against `SecretCacheEvictionCause` (`SecretCacheMetricsDoc.java:218-234`) and the list is exactly right, including the non-obvious decision to leave out Caffeine's fifth cause, `REPLACED`.

That omission is correct for a reason worth recording somewhere: `from(RemovalCause)` throws `IllegalArgumentException` on `REPLACED` (`:254-257`), and Caffeine only calls `recordEviction` when `cause.wasEvicted()`, which excludes it. So even though `CachingSecretStore.resolveFromStore` does overwrite entries via `cache.put` (`CachingSecretStore.java:104`), no `REPLACED` ever reaches the counter.

The `EXPLICIT` description is also precisely right, and matches `invalidateOnPermanentFailure` (`CachingSecretStore.java:118-127`) including the detail that `UNREADABLE` is deliberately not invalidated.

docs/self-managed/operational-guides/monitoring/metrics.md
```diff
@@ -281,6 +281,100 @@ The `REPORT_NAME` and `REPORT_ID` labels identify the evaluated report or dashbo
...
+Reading them in the wrong order can make a healthy cache look broken. A falling cache hit rate only
+means the cache is not holding what callers ask for. It does not mean the value was cacheable in the
+first place. A reference that a store answers permanently as not found, denied, or invalid is never
+cached, so it registers a `MISS` on every lookup for as long as it is referenced. Read a low hit rate
+against `camunda.secret.resolution.outcome` first: if the misses concentrate on references that never
+resolve, the fix is not in the cache.
+
+### Cache size and the configured maximum
+
+`camunda.secret.cache.size` is bounded per store by the
```
ajanoni-camunda, line 378: 💭 Flagging an upstream contradiction rather than anything to change here, because I think this PR got it right.

The PR description names `SecretCacheMetricsDoc` as the source of truth, but that class's javadoc for `CACHE_EVICTIONS` and `CACHE_SIZE` (`:82-97,120-128`) states that the cache max size and TTL are "not configurable today" and quotes `CaffeineSecretCache.DEFAULT_MAX_SIZE` / `DEFAULT_TTL` as fixed values. That is stale: the wiring passes `config.getMaxSize()` and `config.getTtl()` from `camunda.secrets.cache` (`SecretStoreConfiguration.java:191-192`).

This section correctly ignores that and links the `max-size` property instead, which is the accurate behaviour. Worth an upstream javadoc fix so the stated source of truth stops contradicting the docs built from it, otherwise the next person to regenerate this section from those descriptions will reintroduce the error.

docs/self-managed/operational-guides/monitoring/metrics.md
```diff
@@ -281,6 +281,100 @@ The `REPORT_NAME` and `REPORT_ID` labels identify the evaluated report or dashbo
...
+[`camunda.secrets.cache.max-size`](/self-managed/components/orchestration-cluster/core-settings/configuration/properties.md#camundasecretscache)
+property. The bound applies per store, not as a shared budget, so the worst-case memory footprint
+across a deployment is the number of configured stores multiplied by that maximum.
+
+For how the broker resolves secret references before job activation, see
+[Secret resolution and job activation](/components/concepts/secret-resolution-and-job-activation.md).
```
ajanoni-camunda, line 408: 🔧 This link goes to the concepts page, but the concepts page does not link back here, and its TODO for this exact ticket is still open.

`components/concepts/secret-resolution-and-job-activation.md:193` still carries:

> The six secret resolution and secret cache meters, their tags, and how to read them are owned by camunda/camunda#60963. Name and link them here once they are in the metrics reference.

This PR is that content, so the instruction is now satisfiable, but the concepts page keeps both the stale comment and its generic "see the metrics reference" pointer instead of naming the meters and linking this section's anchor. Its "Monitor secret resolution" section is the place a reader arrives from, so it is the one that most needs the specific link.

The base PR has the mirror-image version of this gap, where a pointer was removed and nothing replaced it, so it may be worth fixing both together once the stack is rebased. It also means the "six meters" wording needs updating in two places, not one, once `cycle.delay` is added.

docs/self-managed/operational-guides/monitoring/metrics.md
HeleneW-dot: Yup missed that one 😬  Added it as a fourth row

docs/self-managed/operational-guides/monitoring/metrics.md
HeleneW-dot: Added both tags to every resolution meter, added physicalTenant to tje cache meters and rewrote the label paragraphs to explain the difference between them

docs/self-managed/operational-guides/monitoring/metrics.md
HeleneW-dot: Yes fair point. Added a new `Metric names in Prometheus` table mapping all meters, plus the real PromQL hit-rate expression

docs/self-managed/operational-guides/monitoring/metrics.md
HeleneW-dot: Yes true , added `not emitted in the current configuration ` to that row

docs/self-managed/operational-guides/monitoring/metrics.md
HeleneW-dot: Yes this file is being edited by [the other oopen PR ](https://github.com/camunda/camunda-docs/pull/9717)and touching it here would create a conflict, lets address it there and then It should get resolved when that PR is merged and this one is rebased

docs/self-managed/operational-guides/monitoring/metrics.md
HeleneW-dot: Yes agreed I think the doc is right and the source javadoc is stale, opened [small follow up ](https://github.com/camunda/camunda/issues/61694)to fix that

docs/self-managed/operational-guides/monitoring/metrics.md
HeleneW-dot: Actually I realise these are different stacks (60331 and 60332). So cant address it right now, the steps should be:

1. Merge [#9703](https://github.com/camunda/camunda-docs/issues/9703) (60332-secret-resolution-broker-and-job-workers) into main.
2. Rebase [#9770](https://github.com/camunda/camunda-docs/issues/9770) (60331-secret-store-and-cache-config) onto the updated main, push, merge it.
3. Rebase my #9781 (60963-secret-metrics-docs) onto the updated main (now includes step 2), push, merge it. This is the point where the "Secret resolution and cache metrics" section actually lands on main.
4. Rebase [#9717](https://github.com/camunda/camunda-docs/issues/9717) (60964-secret-resolution-failures) onto main (now includes step 3). Only now does #secret-resolution-and-cache-metrics actually exist in 9717's tree.
5. Then edit secret-resolution-and-job-activation.md in 9717: replace the TODO comment and the generic metrics-reference link with the real pointer.
6. Push, let CI run, merge [#9717](https://github.com/camunda/camunda-docs/issues/9717).

HeleneW-dot: even though I can fix it in the already open PR once the stack merges resolve I [opened a small follow up ticket ](https://github.com/camunda/camunda/issues/61695)just to make sure I dont forget, all of the PRs still need tech writer review so it will be a bit longer 😬

ajanoni-camunda (dismissed):
🟢 LGTM

HeleneW-dot:
@giorgionaps This one is ready for tech writer review

lev-safro (approved):

PR #9782, opened 2026-09-01
docs(configuration): document camunda.secrets.stores.aws and .gcp

### LINKED ISSUE #60966: [Secret Resolution] Docs: configure AWS and GCP secret stores (Self-Managed)
### 

### Description

Document the AWS Secrets Manager and GCP Secret Manager secret stores for Self-Managed.

Split out of #60331, which covers the file store, the cache and the per-physical-tenant surface. The cloud stores carry their own authentication and IAM detail and are reviewed by different people, so they are tracked separately.

### Motivation

These are the stores a production Self-Managed installation will actually use, and none of their properties are documented anywhere today. The authentication model in particular is not guessable: no static credentials are accepted at all.

### Acceptance criteria

- [ ] AWS Secrets Manager store documented: `region`, `path-prefix`, `batch-enabled`, `batch-size` (1 to 20, default 20), `container-secret-id`
- [ ] Documented that AWS authentication is always identity-based via the AWS SDK default credentials provider chain, and that no static credentials are accepted
- [ ] Documented that `batch-enabled` requires the `secretsmanager:BatchGetSecretValue` IAM action, and that it is mutually exclusive with `container-secret-id`
- [ ] GCP Secret Manager store documented: `project-id`, `path-prefix`, `endpoint`, `container-secret-id`
- [ ] Documented that GCP authentication uses the Application Default Credentials chain
- [ ] Documented that GCP secret ids allow only `[a-zA-Z0-9_-]` and are capped at 255 characters, so an AWS-style `camunda/` prefix produces invalid ids
- [ ] Container-secret mode documented for both stores: every reference becomes a key in one flat JSON object; nested objects and arrays are not supported
- [ ] Startup validation failures for both stores documented, since they are operator-visible error messages
- [ ] Added to the property reference partial created in #60331, without reworking it

### Links

Split from: https://github.com/camunda/camunda/issues/60331

### Breakdown

- Write the AWS section
- Write the GCP section
- Fold both into the partial from #60331

### Notes

Source of truth is `configuration/src/main/java/io/camunda/configuration/Secrets.java`, plus `AwsSecretsManagerStoreConfig` and `GcpSecretManagerStoreConfig` under `secret-store/`.

Depends on #60331 having created the partial, and on the D0 answer about whether partials are generated or hand-maintained.

HeleneW-dot:
## Description

Add the AWS Secrets Manager and GCP Secret Manager secret store sections to the secrets configuration partial

closes https://github.com/camunda/camunda/issues/60966

## PR Checklist


docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/_secrets.md
```diff
@@ -22,7 +22,34 @@ Configures the secret stores and cache used to resolve `camunda.secrets.<name>`
 
 A secret name must match `[\p{Alnum}_-]+` to be listed and resolved through `/v2/secrets`. A dot is the common case that falls outside this charset, for example a file extension. Such a name is stored, but it is silently omitted from list results and cannot be resolved by reference.
 
-<!-- AWS and GCP secret stores (camunda.secrets.stores.aws.<id>.* and camunda.secrets.stores.gcp.<id>.*) are tracked in camunda/camunda#60966 and documented separately. -->
+### `camunda.secrets.stores.aws`
```
ajanoni-camunda, line 29: 🔧 The rules that gate this section are up under `### camunda.secrets.stores.file`, where a reader arriving at AWS will not see them.

Three sentences at lines 17 and 23 apply to all store types, not just the file store:

- the definition of `<id>`, which this section's five rows use without introducing
- "Exactly one secret store is supported per physical tenant, and its `<id>` must be `default`"
- the secret name charset rule

The middle one matters most, because the cap counts across **all** store types combined, not per type: `STORE_BINDINGS.stream()...sum()` at `SecretStoreConfiguration.java:120-130`, with the id rule applied per binding at `:135` and `:234-254`. So adding an AWS store next to an existing file store is a startup failure ("only one is supported at this time"), and nothing in this section or the GCP one says so. Given the whole point of this PR is to present two more store types, that is the reading most likely to bite someone.

Suggest hoisting those sentences above `<Tabs>`, so they scope to the whole Secrets section rather than to the file store. That also fixes the undefined `<id>` in both new tables.

docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/_secrets.md
```diff
@@ -22,7 +22,34 @@ Configures the secret stores and cache used to resolve `camunda.secrets.<name>`
...
+| `camunda.secrets.stores.aws.<id>.batch-enabled`       | Resolve secrets via AWS's `BatchGetSecretValue` (fewer round-trips) instead of one `GetSecretValue` call per reference. Requires the `secretsmanager:BatchGetSecretValue` IAM action in addition to `GetSecretValue`. Mutually exclusive with `container-secret-id`: configuring both is rejected at startup.                                                                                                 | `false`       |
+| `camunda.secrets.stores.aws.<id>.batch-size`          | Maximum number of secret ids per `BatchGetSecretValue` call when `batch-enabled` is set. Must be between `1` and `20` (AWS's cap); a value outside that range is rejected at startup.                                                                                                                                                                                                                         | `20`          |
```
ajanoni-camunda: ❌ The range is enforced unconditionally, not only when `batch-enabled` is set.

`validate()` checks the bound before it ever looks at `batchEnabled` (`Secrets.java:347-353`):

```java
if (batchSize < 1 || batchSize > 20) {
  throw new IllegalArgumentException(
      "camunda.secrets.stores.aws." + storeId + ".batch-size must be between 1 and 20, but was " + batchSize);
}
```

As written, "when `batch-enabled` is set. Must be between `1` and `20`" reads as though the constraint is conditional on batching, so `batch-size: 50` with batching off looks harmless. It fails startup.

Suggest splitting the two ideas: the property only has an effect when `batch-enabled` is set, but the range is validated whether or not it is.

docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/_secrets.md
```diff
@@ -22,7 +22,34 @@ Configures the secret stores and cache used to resolve `camunda.secrets.<name>`
 
 A secret name must match `[\p{Alnum}_-]+` to be listed and resolved through `/v2/secrets`. A dot is the common case that falls outside this charset, for example a file extension. Such a name is stored, but it is silently omitted from list results and cannot be resolved by reference.
 
-<!-- AWS and GCP secret stores (camunda.secrets.stores.aws.<id>.* and camunda.secrets.stores.gcp.<id>.*) are tracked in camunda/camunda#60966 and documented separately. -->
+### `camunda.secrets.stores.aws`
+
+Configuration for a secret store backed by AWS Secrets Manager. Authentication is always identity-based, via the AWS SDK's default credentials provider chain. No static credentials are accepted at all.
```
ajanoni-camunda: 🔧 The IAM guidance is incomplete, and there are a few store behaviors nearby that an operator has to know and cannot find anywhere else.

**IAM:** the doc names `GetSecretValue` and, for batching, `BatchGetSecretValue`. Flat mode's startup probe also calls `ListSecrets(maxResults=1)` (`FlatSecretResolver.java:250-252`), so flat mode wants `secretsmanager:ListSecrets` too. Container mode probes with `GetSecretValue` on the container instead (`ContainerSecretResolver.java:151-155`), so the required actions differ by mode.

**Startup validation is best-effort.** A bad region, bad credentials, or a wrong prefix logs a WARN and lets startup continue; the failure surfaces on the first resolve (`AwsSecretsManagerSecretStore.java:91-98,145-155`). Only client-build failure and GCP's unresolvable project id actually stop the node. Since every other validation in this section is described as "rejected at startup", the contrast is worth one sentence.

**Secret version and stage are not configurable.** AWS reads the `AWSCURRENT` stage, GCP always reads version `latest` (`AwsSecretsManagerSecretStore.java:28-30`; `OneByOneSecretResolver.java:41`, `ContainerSecretResolver.java:70,164`). Anyone pinning versions in their store needs to know that.

**Binary secrets are unsupported.** An AWS secret with only `secretBinary` fails as `INVALID_REF` with "has no string value (binary secret)" (`FlatSecretResolver.java:199-207`).

**`path-prefix` also filters `list()`.** Only secrets starting with the prefix are listed, and the prefix is stripped from the logical name returned (`FlatSecretResolver.java:233-234`; GCP `OneByOneSecretResolver.java:83-84`). That changes what `/v2/secrets` returns, so it is more than an id-mapping detail and belongs in the `path-prefix` description.

docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/_secrets.md
```diff
@@ -22,7 +22,34 @@ Configures the secret stores and cache used to resolve `camunda.secrets.<name>`
...
+| `camunda.secrets.stores.aws.<id>.batch-size`          | Maximum number of secret ids per `BatchGetSecretValue` call when `batch-enabled` is set. Must be between `1` and `20` (AWS's cap); a value outside that range is rejected at startup.                                                                                                                                                                                                                         | `20`          |
+| `camunda.secrets.stores.aws.<id>.container-secret-id` | Instead of one AWS secret per reference, treat every reference as a JSON key inside this one named secret. Mutually exclusive with `batch-enabled`: batching fetches several distinct secrets in one call, while a container secret id means only one secret is ever fetched.                                                                                                                                 | `-`           |
+
+In container-secret mode, the named secret's value must be a flat JSON object mapping each reference name to a JSON string value, for example `{"db-password": "s3cr3t"}`. Nested objects and arrays are not supported.
```
ajanoni-camunda: 🔧 "Nested objects and arrays are not supported" does not say what happens when they appear, and the code is specific enough to state. This applies equally to the GCP paragraph at line 52.

The outcomes map directly onto the `result` values that sibling PR #9781 documents for `camunda.secret.resolution.outcome`, so naming them here connects the two pages:

- The container value is not JSON, or is JSON but not an object: **every** requested reference fails with `INVALID_REF`, and `list()` throws `STORE_UNAVAILABLE` rather than returning empty (AWS `ContainerSecretResolver.java:95-113,142-147`; GCP `:98-116,131-148`).
- A key is absent, or present with JSON `null`: `NOT_FOUND`, for that reference only (`:170-173`).
- A key is present but holds a number, object, or array: `INVALID_REF`, for that reference only (`:174-179`).

The first case is the one worth calling out, because a single malformed container secret takes down every reference in the store rather than the one that is wrong, and it surfaces differently through resolve than through list.

docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/_secrets.md
```diff
@@ -22,7 +22,34 @@ Configures the secret stores and cache used to resolve `camunda.secrets.<name>`
...
+| :---------------------------------------------------- | :------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ | :------------ |
+| `camunda.secrets.stores.gcp.<id>.project-id`          | ID of the GCP project that owns the secrets. Optional: when omitted, the client resolves it from the environment (`GOOGLE_CLOUD_PROJECT`) or the compute metadata server, via the Application Default Credentials chain.                                                                                                                                                                                                                                                                                | `-`           |
```
ajanoni-camunda: 🔧 "Optional" is true only where a default project actually exists. Worth naming the failure.

`GcpSecretManagerSecretStore.java:115-125` falls back to `ServiceOptions.getDefaultProjectId()` and, if that yields nothing, throws `SecretStoreUnavailableException` with an explicit message rather than deferring to a later NPE:

> GCP Secret Manager projectId is not configured and no default project could be resolved from the environment; set projectId explicitly or provide GOOGLE_CLOUD_PROJECT / a gcloud default project / a compute metadata server

So off-GCP, or anywhere the metadata server and `GOOGLE_CLOUD_PROJECT` are both absent, omitting this is fatal rather than merely defaulted. That is also one of the few things in these two stores that genuinely stops startup, which makes it worth stating.

docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/_secrets.md
```diff
@@ -22,7 +22,34 @@ Configures the secret stores and cache used to resolve `camunda.secrets.<name>`
...
+| `camunda.secrets.stores.gcp.<id>.path-prefix`         | Prefix prepended, with no separator inserted, to every reference name to form the GCP secret id (for example, prefix `camunda-` plus reference `db-password` resolves to secret id `camunda-db-password`; include the trailing separator here if one is wanted). When omitted, references map to bare secret ids. Also applies to `container-secret-id`, since that is itself resolved as a GCP secret id. Must contain only `[a-zA-Z0-9_-]`; a prefix with any other character is rejected at startup. | `-`           |
+| `camunda.secrets.stores.gcp.<id>.endpoint`            | Secret Manager endpoint override, for example a regional endpoint such as `secretmanager.europe-west1.rep.googleapis.com:443`, or a Private Service Connect endpoint. Unlike the AWS SDK, the GCP client libraries do not honor an endpoint override via environment variable. When omitted, the default global endpoint is used.                                                                                                                                                                       | `-`           |
```
ajanoni-camunda: 🙃 This row reads as though an AWS user configures an endpoint somewhere in Camunda. They cannot.

The comparison itself is accurate, and it is quoted almost verbatim from `Secrets.java:406-412`, so no complaint about the substance. But `endpoint` and `maxRetries` do exist on `AwsSecretsManagerStoreConfig.java:41-42` and are simply hardcoded to `null` and `DEFAULT_MAX_RETRIES` by `SecretStoreConfiguration.java:277-278`, i.e. unreachable from configuration. So for AWS the answer is "there is no endpoint property at all, use the SDK's own environment variable", and AWS retries are fixed at 3+1 attempts while GCP uses client-library defaults.

Rephrasing along the lines of "the AWS SDK picks this up from its own environment variables, so no property is offered; the GCP client libraries do not, hence this property" would make the asymmetry explicit rather than implied.

docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/_secrets.md
```diff
@@ -22,7 +22,34 @@ Configures the secret stores and cache used to resolve `camunda.secrets.<name>`
...
+| `camunda.secrets.stores.gcp.<id>.container-secret-id` | Instead of one GCP secret per reference, treat every reference as a JSON key inside this one named secret. Must contain only `[a-zA-Z0-9_-]`, the same as `path-prefix`. When set, the combined length of `path-prefix` plus `container-secret-id` must not exceed 255 characters; exceeding it is rejected at startup.                                                                                                                                                                                 | `-`           |
+
+GCP secret ids allow only `[a-zA-Z0-9_-]` and are capped at 255 characters. Both `path-prefix` and `container-secret-id` feed into secret ids, so an AWS-style `camunda/` prefix (which contains a slash) produces invalid ids; use, for example, `camunda-` instead.
```
ajanoni-camunda: 👍 This paragraph is the most useful thing in the PR and it is exactly right.

Verified: `SECRET_ID_PATTERN = [a-zA-Z0-9_-]*` and `MAX_SECRET_ID_LENGTH = 255` (`Secrets.java:381,383`), checked for `path-prefix` at `:489-497` and `container-secret-id` at `:498-506`, with the combined-length check at `:507-520`.

Calling out that an AWS-style `camunda/` prefix produces invalid GCP ids is a genuinely non-obvious trap, and it is the kind of cross-backend note a reader copying config between two stores actually needs.

One small precision if you want it: the combined-length check only runs when `container-secret-id` is set. Per-reference ids in flat mode are not length-checked at startup (`:468-470`), so an over-long prefix plus a long reference name fails at resolve time instead.

docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/_secrets.md
HeleneW-dot: Yes fair point,  I think this is already addressed with the rebase now because the base branch already moved this content above the tabs during its own rewrite, and we added the line spelling out that the one-store cap counts across all store types

docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/_secrets.md
HeleneW-dot: Yes true, removed the misleading clause and also reworded the neighboring batch enabled row

docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/_secrets.md
HeleneW-dot: Adjusted this

docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/_secrets.md
HeleneW-dot: Adjusted so it says it's optional only when a default project resolves from env/gcloud/metadata server, and startup fails outright if none do.

docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/_secrets.md
HeleneW-dot: Reworded

docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/_secrets.md
HeleneW-dot: Yes makes sense, added the 255-char combined check only runs when container-secret-id is set

docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/_secrets.md
HeleneW-dot: Added the outcomes for this one to aws and gcp sections

ajanoni-camunda (dismissed):
🟢 LGTM

HeleneW-dot:
@giorgionaps Ready for tech writer review 🙏

lev-safro (approved):

PR #9787, opened 2026-09-01
Update web modeler references 9

alexronquillo:
## Description

**Note:** https://github.com/camunda/camunda-docs/pull/9786 should be reviewed and merged first.

More Camunda Hub terms, instructions, and screenshot updates.

This time, the focus is on the modeling section. In an effort to stay organized, this tackles the files in the **Modeling** directory. The ones in the sub-directories are coming next.

Related to: #9427

## When should this change go live?

- [ ] This is a bug fix, security concern, or something that needs **urgent release support**. (add `bug` or `support` label)
- [ ] This is already available but undocumented and should be released within a week. (add `available & undocumented` label)
- [ ] This is on a **specific schedule** and the assignee will coordinate a release with the Documentation team. (create draft PR and/or add `hold` label)
- [x] This is part of a scheduled **alpha or minor**. (add alpha or minor label)
- [ ] There is **no urgency** with this change (add `low prio` label)

## PR Checklist

- [x] The commit history of this PR is cleaned up, using [`{type}(scope): {description}` commit message(s)](https://github.com/camunda/camunda-docs/blob/main/CONTRIBUTING.MD#commit-message-header-formatting)

- [x] My changes are for **an upcoming minor release** and are in the `/docs` directory.
- [ ] My changes are for an **already released minor** and are in a `/versioned_docs` directory.

- [x] I added a DRI, team, or delegate as a reviewer for technical accuracy and grammar/style:
  - [ ] [Engineering team review](https://github.com/camunda/camunda-docs/blob/main/howtos/documentation-guidelines.md#review-process)
  - [x] [Technical writer review](https://github.com/camunda/camunda-docs/blob/main/howtos/documentation-guidelines.md#review-process) via `@camunda/tech-writers` unless working with an embedded writer.

Issue #9427, opened 2026-07-20
Update references to Web Modeler and Console throughout the documentation


alexronquillo:
We reference Web Modeler and Console many times throughout the documentation. These need to be updated to reflect Camunda Hub terminology, instructions, and concepts.

Additional criteria:

- [x] [Mention Hub backups](https://camunda.slack.com/archives/C098M5EHRTK/p1787669130317799)
- [x] Web Modeler -> Camunda Hub
    - [x] Trivial changes
    - [x] The following files need a closer look:
        - [x] 8.9 SM migration guides (need to be removed)
        - [x] docs/components/concepts/user-task-listeners-guide.md (guide instructions)
        - [x] docs/components/modeler/forms/utilizing-forms.md (guide instructions)
        - [x] docs/components/rpa/getting-started.md (guide instructions)
        - [x] docs/reference/licenses.md (I don't want to change legal documents without specific approval; related: https://github.com/camunda/camunda-hub/issues/27928)
        - [x] docs/reference/notices.md (more research needed)
        - [x] all camunda hub (Using Camunda and SM) guides (guide instructions)
        - [x] docs/self-managed/concepts/databases/overview.md (technical approval needed)
        - [x] docs/self-managed/concepts/databases/relational-db/rdbms-setup-guide.md (guide instructions)
        - [x] self-managed deployment and operational guides (tech review)
        - [x] docs/reference/glossary.md (untangle "version" definitions)
        - [x] docs/reference/supported-environments.md (technical approval needed)
        - [x] docs/guides/getting-started-agentic-orchestration.md (getting started tutorial, so it could use a more thorough pass)
        - [x] docs/components/modeler/element-templates/template-metadata.md (`camundaWebModeler` engine - [not renaming](https://camunda.slack.com/archives/C0AGDE80US0/p1788855563171999))
- [x] Console -> Camunda Hub
    - [x] Using Camunda
    - [x] Self-Managed
- [x] Collaborator -> Member
- [x] (Org) member -> user
- [x] Project -> Workspace
- [x] Process application -> Project
- [x] Permissions (Project Admin -> Workspace Admin, Modeler -> Member, etc)
- [x] Web Modeler API -> Hub API (Management identity API permissions)
    - [x] Trivial changes
    - [x] The following files need a closer look:
[...]

PR #9789, opened 2026-09-01
docs(hub): update hub terms, instructions, and screenshots

alexronquillo:
## Description

**Note:** #9787 should be reviewed and merged first.

More Camunda Hub terms, instructions, and screenshot updates.

This time, the focus is on the advanced modeling section.

Related to: #9427

## PR Checklist


mesellings (approved):
Skim approval to unblock 👍 

Non-blocking: Without looking, is it a worse experience removing this many screenshots?

alexronquillo:
One could argue it's a worse experience because we removed the visual aid. One could also argue it's a better experience because unnecessary images break the flow of reading the instructions. 

In my opinion, the important question is: are the tradeoffs worth it?

Here's an example of the one of the old docs:

image

Here's an example of the same section in the new docs:

image

Here's the new UI for comparison:

image

The main problem is: images add a lot of maintenance cost. So we need to either update these images (because they're not correct anymore, as shown in the image above) or delete them. Retaking all the screenshots is, of course, a lot more work. As discussed previously, I think it's worth just removing the screenshots now and updating the instructions to match the new UI (for example, "A link icon appears at the bottom right." becomes "A link icon appears in the floating menu.") rather than retaking all these screenshots. We can always add them back later if we need to.

mesellings:
> One could argue it's a worse experience because we removed the visual aid. One could also argue it's a better experience because unnecessary images break the flow of reading the instructions.
> 
> In my opinion, the important question is: are the tradeoffs worth it?
> 
> Here's an example of the one of the old docs:
> 
> image
> Here's an example of the same section in the new docs:
> 
> image
> Here's the new UI for comparison:
> 
> image
> The main problem is: images add a lot of maintenance cost. So we need to either update these images (because they're not correct anymore, as shown in the image above) or delete them. Retaking all the screenshots is, of course, a lot more work. As discussed previously, I think it's worth just removing the screenshots now and updating the instructions to match the new UI (for example, "A link icon appears at the bottom right." becomes "A link icon appears in the floating menu.") rather than retaking all these screenshots. We can always add them back later if we need to.

No, I'm happy with the approach as we agreed, and that example does show the benefits very clearly.

It's more for any diagrams that could be useful - I noticed in the (unreviewed) IDP PR the diagram is removed rather than being redone.

Carry on as you are doing! 🚀 

P.s. I love that your example is in dark mode - once a developer always a developer 😊

alexronquillo:
Diagrams are trickier, for sure, but I actually have tried to keep the diagrams when possible. For example, in the IDP PR, I updated [this diagram](https://github.com/camunda/camunda-docs/pull/9797/changes#diff-17520b6409f631a6988728853ee8f1334d41245f4bde5fe49e86ed7f32d20e35). This has been a pain in some cases, because the diagrams are sometimes hard to find. When I update them, I'm including links to the source for future reference.

PR #9792, opened 2026-09-02
docs(cluster-variables): document the SECRET_REFERENCE cluster variable kind

### LINKED ISSUE #60330: [Secret Resolution] Docs: cluster variables of kind SECRET_REFERENCE
Parent: #56591

### Description

Document cluster variables of kind `SECRET_REFERENCE`.

### Motivation

`ClusterVariableKind.SECRET_REFERENCE` has shipped and already appears in the v2 REST API reference and the Java client, but no hand-written page explains it. A reader currently meets it only as an undocumented enum value in generated API output.

### Acceptance criteria

- [ ] The `SECRET_REFERENCE` kind documented in the cluster variable doc set (`docs/components/modeler/feel/cluster-variable/`) and on `docs/components/admin/cluster-variables.md`
- [ ] Documented that `JSON` is the default kind and `SECRET_REFERENCE` allows the value to contain `camunda.secrets.X` references
- [ ] Documented that the references are resolved at job activation time
- [ ] Interaction with scope, priority and namespace collisions covered, or explicitly stated as unchanged
- [ ] Required permissions stated
- [ ] Cross-linked to the concept page from D1

### Breakdown

- Audit the eight pages under `docs/components/modeler/feel/cluster-variable/` and decide which need the addition
- Update `docs/components/admin/cluster-variables.md`

### Notes

The relevant engine code is `ClusterVariableSecretReferenceScanner` and `ClusterVariableJobSecretResolver`.

HeleneW-dot:
## Description
Add the variable kinds section to the cluster variable data types page as the home for this content: JSON is the default kind, SECRET_REFERENCE marks a variable whose value may contain camunda.secrets.<name> references, and only a SECRET_REFERENCE-kind variable has its references resolved.

Also state where resolution happens on the usage page, note in the scope resolution page that priority is unchanged for every kind, add the kind collision case to the namespace collisions page, and cover creating, editing, and reading these variables plus the required permissions in Admin.

Relates to camunda/camunda#60330

## PR Checklist

### DIFF
--- a/docs/components/admin/cluster-variables.md
+++ b/docs/components/admin/cluster-variables.md
@@ -92,7 +92,7 @@ If a global variable with the same name exists, the tenant-level variable takes 
 
 ## Manage cluster variables of kind `SECRET_REFERENCE`
 
-A cluster variable of kind `SECRET_REFERENCE` can contain `camunda.secrets.<name>` references in its value. Camunda resolves them at job activation for a process that reads the variable in an input mapping. A `JSON`-kind variable whose value contains the same text is treated as ordinary text.
+A cluster variable of kind `SECRET_REFERENCE` can contain `camunda.secrets.<name>` references in its value. When a process reads the variable in an input mapping, Camunda resolves these references in the background ahead of activation, and injects the resolved values into the job only once it's handed to a worker. A `JSON`-kind variable whose value contains the same text is treated as ordinary text.
 
 The Admin UI create form has no kind field, so every variable you create there is a `JSON`-kind variable. To create a `SECRET_REFERENCE`-kind variable, use the Orchestration Cluster API with `"kind": "SECRET_REFERENCE"` in the request body, either [globally](/apis-tools/orchestration-cluster-api-rest/specifications/create-global-cluster-variable.api.mdx) or [for a tenant](/apis-tools/orchestration-cluster-api-rest/specifications/create-tenant-cluster-variable.api.mdx).
 
--- a/docs/components/admin/cluster-variables.md
+++ b/docs/components/admin/cluster-variables.md
@@ -11,6 +11,8 @@ Use Admin to manage cluster variables, which store configuration values centrall
 
 Cluster variables allow you to maintain environment-specific configurations, API endpoints, feature flags, and other shared values without hardcoding them into individual process definitions. Variables can be defined at the global (cluster-wide) or tenant level.
 
+Each variable also has a kind, either `JSON` or `SECRET_REFERENCE`, which determines how Camunda reads its value. `JSON` is the default. See [variable kinds](/components/modeler/feel/cluster-variable/data-types.md#variable-kinds).
+
 :::tip
 To learn more about cluster variables, including scope resolution, data types, and FEEL expression usage, see the [cluster variables overview](/components/modeler/feel/cluster-variable/overview.md).
 :::
@@ -88,8 +90,36 @@ If a global variable with the same name exists, the tenant-level variable takes 
 3. Click **Delete** next to the variable you want to delete.
 4. Confirm the deletion by clicking **Delete** in the confirmation dialog.
 
+## Manage cluster variables of kind `SECRET_REFERENCE`
+
+A cluster variable of kind `SECRET_REFERENCE` can contain `camunda.secrets.<name>` references in its value. Camunda resolves them at job activation for a process that reads the variable in an input mapping. A `JSON`-kind variable whose value contains the same text is treated as ordinary text.
+
+The Admin UI create form has no kind field, so every variable you create there is a `JSON`-kind variable. To create a `SECRET_REFERENCE`-kind variable, use the Orchestration Cluster API with `"kind": "SECRET_REFERENCE"` in the request body, either [globally](/apis-tools/orchestration-cluster-api-rest/specifications/create-global-cluster-variable.api.mdx) or [for a tenant](/apis-tools/orchestration-cluster-api-rest/specifications/create-tenant-cluster-variable.api.mdx).
+
+You can manage an existing `SECRET_REFERENCE`-kind variable in the Admin UI:
+
+- Updating its value keeps its kind, and Camunda scans the new value for references. A variable's kind is fixed at creation and cannot be changed.
+- The value shown in the variable list and in the variable details is the stored value, so you see the reference text rather than a resolved value.
+- Deleting it works the same as deleting any other cluster variable.
+
+For where resolved values appear and where they do not, see [secret resolution and job activation](/components/concepts/secret-resolution-and-job-activation.md).
+
+## Required permissions
+
+Managing cluster variables requires permissions on the `CLUSTER_VARIABLE` resource type. These permissions are the same for every kind, and there is no additional permission for a `SECRET_REFERENCE`-kind variable.
+
+| Action                    | Required permission |
+| ------------------------- | ------------------- |
+| Create a cluster variable | `CREATE`            |
+| View a cluster variable   | `READ`              |
+| Update a cluster variable | `UPDATE`            |
+| Delete a cluster variable | `DELETE`            |
+
+The resource identifier is the variable name, or `*` for all cluster variables. See [authorizations](/components/concepts/access-control/authorizations.md#available-resources) for how to grant these permissions.
+
 ## See also
 
 - [Cluster variables overview](/components/modeler/feel/cluster-variable/overview.md) — learn about scopes, data types, and FEEL expression usage.
 - [Get started with cluster variables](/components/modeler/feel/cluster-variable/get-started.md) — tutorial for creating and using your first cluster variable.
+- [Secret resolution and job activation](/components/concepts/secret-resolution-and-job-activation.md) — understand when a job that references secrets reaches a worker.
 - [Orchestration Cluster API: Create global cluster variable](/apis-tools/orchestration-cluster-api-rest/specifications/create-global-cluster-variable.api.mdx) — manage cluster variables via API.
--- a/docs/components/admin/cluster-variables.md
+++ b/docs/components/admin/cluster-variables.md
@@ -94,6 +94,8 @@ If a global variable with the same name exists, the tenant-level variable takes 
 
 A cluster variable of kind `SECRET_REFERENCE` can contain `camunda.secrets.<name>` references in its value. When a process reads the variable in an input mapping, Camunda resolves these references in the background ahead of activation, and injects the resolved values into the job only once it's handed to a worker. A `JSON`-kind variable whose value contains the same text is treated as ordinary text.
 
+Resolving these references requires a secret store to be configured. Without one, references are never resolved, and any job that depends on them stays unactivated with no error pointing at the cause. See [`camunda.secrets.stores.file`](/self-managed/components/orchestration-cluster/core-settings/configuration/properties.md#camundasecretsstoresfile).
+
 The Admin UI create form has no kind field, so every variable you create there is a `JSON`-kind variable. To create a `SECRET_REFERENCE`-kind variable, use the Orchestration Cluster API with `"kind": "SECRET_REFERENCE"` in the request body, either [globally](/apis-tools/orchestration-cluster-api-rest/specifications/create-global-cluster-variable.api.mdx) or [for a tenant](/apis-tools/orchestration-cluster-api-rest/specifications/create-tenant-cluster-variable.api.mdx).
 
 You can manage an existing `SECRET_REFERENCE`-kind variable in the Admin UI:
--- a/docs/components/modeler/feel/cluster-variable/data-types.md
+++ b/docs/components/modeler/feel/cluster-variable/data-types.md
@@ -21,3 +21,71 @@ Understand the data types supported by cluster variables for different configura
 :::note
 Access patterns may vary depending on how the array is used.
 :::
+
+## Variable kinds
+
+Every cluster variable has a kind, which determines how Camunda reads its value.
+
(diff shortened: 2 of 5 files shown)


docs/components/admin/cluster-variables.md
```diff
@@ -88,8 +90,36 @@ If a global variable with the same name exists, the tenant-level variable takes
...
+
+- Updating its value keeps its kind, and Camunda scans the new value for references. A variable's kind is fixed at creation and cannot be changed.
+- The value shown in the variable list and in the variable details is the stored value, so you see the reference text rather than a resolved value.
+- Deleting it works the same as deleting any other cluster variable.
+
+For where resolved values appear and where they do not, see [secret resolution and job activation](/components/concepts/secret-resolution-and-job-activation.md).
+
+## Required permissions
+
+Managing cluster variables requires permissions on the `CLUSTER_VARIABLE` resource type. These permissions are the same for every kind, and there is no additional permission for a `SECRET_REFERENCE`-kind variable.
```
ajanoni-camunda, line 111: ❌ **This sentence is true, but I think it reads as more reassuring than it should.**

Here is what the code actually does. `ClusterVariableCreateProcessor` checks exactly one thing: `CLUSTER_VARIABLE` / `CREATE`. It never checks a `SECRET` permission. Meanwhile `/v2/secrets/resolve` does require `SECRET` / `REVEAL`.

Put those together and you get a real gap:

> Someone holding only `CLUSTER_VARIABLE` / `CREATE` can create a `SECRET_REFERENCE` variable naming any secret in the store. Any process that reads it through an input mapping then hands the resolved plaintext to the worker. At no point did that person need `REVEAL`.

So `CLUSTER_VARIABLE` / `CREATE` is, in practice, secret-read capability. An operator reading this page today would reasonably conclude the two permission sets are independent, and size the grant accordingly.

Could we add a warning that says so directly? Something like:

> :::warning
> Granting `CREATE` or `UPDATE` on `CLUSTER_VARIABLE` effectively grants the ability to read any secret in the store, because the holder can create a `SECRET_REFERENCE` variable and have a process resolve it. This bypasses the `REVEAL` permission that `POST /v2/secrets/resolve` requires.
> :::

❓ One caveat before you act on this: I could not tell from the code whether this permission model is deliberate or an oversight. If it is intended, the warning above is the right fix. If it isn't, this belongs in a `camunda/camunda` issue and the docs should wait. Worth a quick check with the engineering DRI either way.

docs/components/modeler/feel/cluster-variable/data-types.md
```diff
@@ -21,3 +21,71 @@ Understand the data types supported by cluster variables for different configura
...
+
+### Read a variable of kind `SECRET_REFERENCE`
+
+Get and search responses return the stored value, so you see the reference text rather than a resolved value. References are resolved only when a process reads the variable in an input mapping, as described in [resolve secret references in a cluster variable](./usage-guide.md#resolve-secret-references-in-a-cluster-variable). For where resolved values appear and where they do not, see [secret resolution and job activation](/components/concepts/secret-resolution-and-job-activation.md).
+
+To find variables of a given kind, use the `kind` filter in [search cluster variables](/apis-tools/orchestration-cluster-api-rest/specifications/search-cluster-variables.api.mdx).
+
+### Required permissions
+
+A `SECRET_REFERENCE`-kind variable needs the same permissions as any other cluster variable. There is no additional permission for this kind.
```
ajanoni-camunda, line 84: ❌ Same issue as on the Admin page's permissions table — flagging it here so the fix lands wherever the table appears.

As written this says "nothing extra to worry about." What actually happens is that `CLUSTER_VARIABLE` / `CREATE` opens a route to resolved secret values that skips the `SECRET` / `REVEAL` check enforced on `/v2/secrets/resolve`.

Suggest mirroring whatever warning we settle on in `cluster-variables.md`. Same caveat applies: confirm with the engineering DRI that the permission model is intended before documenting it as such.

docs/components/modeler/feel/cluster-variable/data-types.md
```diff
@@ -21,3 +21,71 @@ Understand the data types supported by cluster variables for different configura
 :::note
 Access patterns may vary depending on how the array is used.
 :::
+
+## Variable kinds
```
ajanoni-camunda, line 25: ❌ **Nothing here tells the reader this is new, or that it rests on alpha infrastructure.**

Two separate gaps:

**1. It's new in 8.10.** The `kind` property is annotated `addedInVersion: "8.10"` on both the create and search request schemas in the v2 spec. These cluster-variable pages already exist in `versioned_docs/version-8.9/`, so this section is new content dropped into a page that isn't. Without a marker, a reader has no way to separate the two.

**2. It depends on an alpha feature.** `SECRET_REFERENCE` resolution is backed entirely by the secret store and cache subsystem. PR #9770 labels that subsystem an [alpha feature](/components/early-access/alpha/alpha-features.md), and the `/v2/secrets/*` endpoints are titled "(alpha)" throughout the API reference. This section inherits that dependency but none of the labeling.

Adding a version marker plus the standard alpha disclaimer would bring this in line with how #9770 and the secrets endpoints already handle the same feature.

ajanoni-camunda (commented):
Verified this branch against the implementation in `camunda/camunda` (`main`, `b24bcaeea37`) rather than reviewing it on prose alone.

👍 Most of it holds up exactly as written. The kind semantics, the immutable-kind-on-update behavior, the field-path narrowing, the listener exclusion, the input-mapping-only scope, and the collision outcome all match the code. The array warning even matches a dedicated regression test (`shouldDropJobWhenClusterVariableSecretIsNestedInAnArrayElement`). This is unusually precise reference documentation and it was a pleasure to check.

- ❌ **4 comments** — one security boundary that I think is understated, missing 8.10/alpha signposting, and one factual correction about array scanning.
- 🔧 **3 comments** — optional suggestions on completeness. None of these should block.

docs/components/modeler/feel/cluster-variable/data-types.md
```diff
@@ -21,3 +21,71 @@ Understand the data types supported by cluster variables for different configura
...
+| Kind               | Description                                                                        |
+| ------------------ | ---------------------------------------------------------------------------------- |
+| `JSON`             | The default. Your value is data, and Camunda reads it exactly as you stored it.    |
+| `SECRET_REFERENCE` | The value can contain `camunda.secrets.<name>` references, which Camunda resolves. |
+
+Only a `SECRET_REFERENCE`-kind variable has its references resolved. A `JSON`-kind variable whose value contains the same text is treated as ordinary text, and that text reaches your process unchanged.
+
+### Where references can appear in a value
+
+Camunda scans every string in a `SECRET_REFERENCE`-kind variable's value, including strings nested inside objects. Object keys are not scanned. A reference has the form `camunda.secrets.<name>`, where `<name>` can contain ASCII letters, digits, underscores, and dashes.
```
ajanoni-camunda: Two things on this sentence, one of which I think is a genuine error.

❌ **Arrays are scanned too.** `ClusterVariableSecretReferenceScanner.collect()` handles `node.isArray()` explicitly and emits index pointers like `/items/0`. Scoping the sentence to objects makes it contradict the array paragraph further down, which correctly says a reference inside an array *is* detected at create time. As it stands the two passages disagree.

Suggested fix — just widen the sentence:

> Camunda scans every string in a `SECRET_REFERENCE`-kind variable's value, including strings nested inside objects and arrays.

That way the later "do not place a reference inside an array" warning reads as the natural consequence (detected, but not resolvable) instead of a contradiction.

🔧 **No length limit is mentioned.** `REFERENCE_PATTERN` doesn't cap name length, but per #9770 a secret name has to be 240 characters or fewer to be listed and resolved through `/v2/secrets`. A longer name gets detected in a cluster variable and then can never resolve. Worth a clause next to the charset description, though this is a rare enough case that I'd leave it to your judgement.

docs/components/modeler/feel/cluster-variable/usage-guide.md
```diff
@@ -78,6 +78,19 @@ For example:
 camunda.vars.env.API_BASE_URL + "/api/v" + camunda.vars.env.API_VERSION + "/resource"
 ```
 
+### Resolve secret references in a cluster variable
+
+References in a [`SECRET_REFERENCE`-kind](./data-types.md#variable-kinds) cluster variable are resolved only when the variable is read by an input mapping on an element that creates a job for a job worker, such as a service task. In the other contexts on this page, including gateway conditions, script tasks, output mappings, and call activity input, the variable resolves to its stored value, so the reference text reaches your process unchanged.
```
ajanoni-camunda: 🔧 Small completeness point, take it or leave it.

Ad hoc sub-process jobs also carry resolved values. In `BpmnJobBehavior`, both `createNewJob` and `createNewAdHocSubProcessJob` pass `mergedSecretReferences(...)`, while the two listener paths pass `Map.of()` — which is exactly what makes your listener bullet below correct.

"such as a service task" isn't wrong, it's an example. But the rest of this paragraph draws a precise boundary, so a reader may well take the example as the full list. Either name the ad hoc sub-process case alongside it, or drop the example entirely and let "an element that creates a job for a job worker" stand on its own.

docs/components/modeler/feel/cluster-variable/namespace-collisions.md
```diff
@@ -72,6 +72,35 @@ camunda.vars.env.CONFIG_KEY.nested → null (trying to access property on string
...
+camunda.vars.env.API_CREDENTIALS
+→ the literal text "camunda.secrets.PAYMENT_API_KEY" (TENANT wins, kind JSON)
+
+camunda.vars.cluster.API_CREDENTIALS
+→ the resolved secret value (GLOBAL, kind SECRET_REFERENCE)
+```
+
+#### Why this happens
+
+Tenant scope has higher priority, so `camunda.vars.env` selects the tenant variable, and only that variable's kind is considered. A `JSON`-kind value is ordinary text, so its references are not resolved. The global variable's references are reachable only through `camunda.vars.cluster`, which bypasses tenant scope.
```
ajanoni-camunda, line 100: 🔧 This explanation matches `ClusterVariableJobSecretResolver.resolveInstance()` exactly — `env` resolves tenant-first, and the `kind == SECRET_REFERENCE` filter that follows is what makes the `JSON` tenant variable contribute nothing. Good.

One edge case isn't covered though. For `env`, the tenant hit passes through an extra filter: `getValueBuffer().capacity() > 0`. A tenant variable with an **empty value** is skipped, and resolution falls back to global.

In that case the global `SECRET_REFERENCE` variable wins through `camunda.vars.env` — the opposite of the outcome described here. Since the takeaway you're giving is "define the same kind at every scope," a reader hitting the empty-value case would see behavior that contradicts the rule. Probably worth a sentence.

docs/components/admin/cluster-variables.md
```diff
@@ -88,8 +90,36 @@ If a global variable with the same name exists, the tenant-level variable takes
 3. Click **Delete** next to the variable you want to delete.
 4. Confirm the deletion by clicking **Delete** in the confirmation dialog.
 
+## Manage cluster variables of kind `SECRET_REFERENCE`
+
+A cluster variable of kind `SECRET_REFERENCE` can contain `camunda.secrets.<name>` references in its value. When a process reads the variable in an input mapping, Camunda resolves these references in the background ahead of activation, and injects the resolved values into the job only once it's handed to a worker. A `JSON`-kind variable whose value contains the same text is treated as ordinary text.
```
ajanoni-camunda, line 95: 🔧 Missing prerequisite: none of this works unless a secret store is actually configured.

Per #9770, if no `stores.file.<id>` entry exists in the merged configuration, the physical tenant gets a **no-op store** — nothing resolves, and no directory is read.

The failure mode is quiet. An admin who creates a `SECRET_REFERENCE` variable on a cluster with no store configured just sees jobs that never activate, and nothing on this page points at the cause. A one-line pointer to the secret store configuration as a prerequisite would close that loop.

docs/components/admin/cluster-variables.md
HeleneW-dot: So ClusterVariableCreateProcessor checks only CLUSTER_VARIABLE/CREATE, /v2/secrets/resolve requires SECRET/REVEAL and job activation goes through neither, so I would say the read is accurate.
However, I also do not know if this is deliberate so I would like to hold off from documenting it until thats decided. I will ping in the project channel

HeleneW-dot: Asking [here](https://camunda.slack.com/archives/C099MRHA8U8/p1788511474826169)

docs/components/modeler/feel/cluster-variable/data-types.md
HeleneW-dot: Asking [here](https://camunda.slack.com/archives/C099MRHA8U8/p1788511474826169)

docs/components/admin/cluster-variables.md
HeleneW-dot: Answer [here](https://camunda.slack.com/archives/C099MRHA8U8/p1788512983719059?thread_ts=1788511474.826169&cid=C099MRHA8U8): "no additional permission for this kind" is accurate as written and already reflects the intended behaviour

docs/components/modeler/feel/cluster-variable/data-types.md
HeleneW-dot: https://github.com/camunda/camunda-docs/pull/9792#discussion_r3932878274

HeleneW-dot:
@ajanoni-camunda ready for re review 🙏

lev-safro (approved):

PR #9819, opened 2026-09-04
feat: add OC/operations page to change region priorities

entangled90:
## Description
In 8.10 it's now possible to easily change the priorities of RAFT for a zone-aware cluster.
The API themselves are already documented in `Management API`, but this guides provides some more context, helpful commands to the operator.

- [X] This is already available but undocumented and should be released within a week. (add `available & undocumented` label)

## PR Checklist

- [X] My changes are for **an upcoming minor release** and are in the `/docs` directory.

### DIFF
--- a/docs/self-managed/components/orchestration-cluster/zeebe/operations/migrate-raft-leadership.md
+++ b/docs/self-managed/components/orchestration-cluster/zeebe/operations/migrate-raft-leadership.md
@@ -1,0 +1,94 @@
+---
+id: migrate-raft-leadership
+title: "Move Raft leadership between zones"
+description: "Change zone priorities and rebalance a multi-region Zeebe cluster to move Raft partition leadership between zones."
+---
+
+You can move Raft partition leadership from one zone to another by reordering zone priorities and then rebalancing the cluster.
+
+This procedure changes only the priorities used for leader election. It doesn't change the partition distribution or move replicas between zones.
+
+Use this procedure to:
+
+- Perform a planned zone switchover.
+- Restore the preferred leadership placement after a zone failover.
+- Follow the sun by moving leadership to the preferred active region.
+
+## Align leadership with secondary storage
+
+Raft leadership placement is especially important when the Amazon Aurora secondary storage writer and Raft leaders are in different regions. Cross-region communication with the secondary storage writer adds network round-trip latency. After the secondary storage writer moves to another region, use this procedure to move Raft leadership to the same region when possible.
+
+The placement of Elasticsearch or OpenSearch secondary storage has less impact on this decision. In a multi-region setup, records are exported to secondary storage in both regions concurrently, so exporting has less dependency on cross-region network round-trip latency. Consider the location of the Aurora secondary storage writer first when choosing the preferred leader zone.
+
+## Prerequisites
+
+Before moving leadership, confirm the following conditions:
+
+- The cluster is fully zone-aware.
+- All brokers and partitions are healthy.
+- The target zone's replicas are caught up with the current leaders.
+- You can access the Zeebe Gateway [Management API](management-api.md) on its management port. The default port is `9600`.
+- You have identified a low-load maintenance window. Rebalancing can cause temporary unavailability while partition leaders are re-elected.
+
+## Check the current zone priorities
+
+Use the Management API to retrieve the current cluster topology and partition distribution:
+
+```bash
+curl \
+  'http://{zeebe-gateway}:9600/actuator/cluster' \
+  -H 'accept: application/json' \
+  | jq '.partitionDistribution.zones[] | {name, priority}'
+```
+
+The command returns each configured zone's name and priority. If `jq` isn't installed, omit the pipe to `jq` and manually inspect `partitionDistribution.zones` in the JSON response. You can also inspect `brokers[].partitions[]` in the full response to see the priority assigned to each partition replica.
+
+A higher priority makes a replica the preferred leader during an election. The zone with the highest configured priority is therefore the preferred zone for Raft partition leaders. Recording the current order also ensures that you preserve the relative priorities of any zones you aren't swapping.
+
+## Reorder the zone priorities
+
+The [Partition distribution API](management-api.md#partition-distribution-api) accepts a `zonePriorities` list. The first zone in the list receives the highest existing priority, the second zone receives the next highest priority, and so on.
+
+The request must include every currently configured zone. To exchange the priorities of two zones, reverse their positions and leave the other zones in their current order.
+
+For example, if `zone-a` currently has the highest priority and `zone-b` has the next highest priority, use the following request to make `zone-b` the preferred leader zone:
+
+```bash
+curl -X PUT \
+  'http://{zeebe-gateway}:9600/actuator/cluster/partition-distribution?dryRun=true' \
+  -H 'accept: application/json' \
+  -H 'Content-Type: application/json' \
+  -d '{
+    "zonePriorities": ["zone-b", "zone-a"]
+  }'
+```
+
+Review the `plannedChanges` and `expectedTopology` fields in the dry-run response. When the result matches the intended priority order, submit the same request without the `dryRun` parameter:
+
+```bash
+curl -X PUT \
+  'http://{zeebe-gateway}:9600/actuator/cluster/partition-distribution' \
+  -H 'accept: application/json' \
+  -H 'Content-Type: application/json' \
+  -d '{
+    "zonePriorities": ["zone-b", "zone-a"]
+  }'
+```
+
+The priority change is asynchronous. Monitor it by polling the cluster topology every five seconds:
+
+```bash
+watch -n 5 'curl -s http://{zeebe-gateway}:9600/actuator/cluster | jq "{pending: (.pendingChange.pending // null), status: .lastChange.status}"'
+```
+
+Wait until `pending` is `null` and `status` is `COMPLETED` before rebalancing.
+
+## Rebalance the cluster
+
+Changing zone priorities doesn't trigger a leader election. After the priority change completes, [manually rebalance the cluster](rebalancing.md#manual-rebalancing) to move partition leadership to the newly preferred zone.
+
+Before rebalancing, review the guide's [limitations](rebalancing.md#limitations), [impact](rebalancing.md#rebalancing-impact), and [readiness checks](rebalancing.md#when-to-rebalance). During rebalancing, partitions can be temporarily unavailable while new leaders are elected.
+
+After rebalancing completes, use `GET /actuator/cluster` and your cluster metrics to verify that the expected brokers in the target zone lead the partitions.
+
+You can also use the Orchestration Cluster REST API [`GET /v2/topology`](/apis-tools/orchestration-cluster-api-rest/specifications/get-topology.api.mdx) endpoint to check the partition leaders. This endpoint uses the v2 API rather than the Management API and requires different access permissions. Ensure your credentials are authorized for the v2 API before using it.
--- a/docs/self-managed/components/orchestration-cluster/zeebe/operations/migrate-raft-leadership.md
+++ b/docs/self-managed/components/orchestration-cluster/zeebe/operations/migrate-raft-leadership.md
@@ -75,13 +75,13 @@ curl -X PUT \
   }'
 ```
 
-The priority change is asynchronous. Monitor it by polling the cluster topology every five seconds:
+The priority change is asynchronous. Use the `changeId` from the response to poll that change every five seconds:
 
 ```bash
-watch -n 5 'curl -s http://{zeebe-gateway}:9600/actuator/cluster | jq "{pending: (.pendingChange.pending // null), status: .lastChange.status}"'
+watch -n 5 'curl -s http://{zeebe-gateway}:9600/actuator/cluster/changes/{changeId} | jq "{id, status, pending}"'
 ```
 
-Wait until `pending` is `null` and `status` is `COMPLETED` before rebalancing.
+Wait until `status` is `COMPLETED` before rebalancing. `GET /actuator/cluster` can also report a pending change for a cluster with only the default Physical Tenant, but use `GET /actuator/cluster/changes/{changeId}` for clusters with multiple Physical Tenants.
 
 ## Rebalance the cluster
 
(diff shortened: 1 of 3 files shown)


docs/self-managed/components/orchestration-cluster/zeebe/operations/migrate-raft-leadership.md
```diff
@@ -0,0 +1,94 @@
...
+  -d '{
+    "zonePriorities": ["zone-b", "zone-a"]
+  }'
+```
+
+The priority change is asynchronous. Monitor it by polling the cluster topology every five seconds:
+
+```bash
+watch -n 5 'curl -s http://{zeebe-gateway}:9600/actuator/cluster | jq "{pending: (.pendingChange.pending // null), status: .lastChange.status}"'
+```
```
deepthidevaki: 🔧 We have introduced a new GET /actuator/cluster/change endpoint to monitor pending changes. The other one still works if there is only the physical tenant.

deepthidevaki (dismissed):
Thanks. This should be linked from the dual-region operational procedure page as well. 

Please also get a review from a tech-writer.

docs/self-managed/components/orchestration-cluster/zeebe/operations/migrate-raft-leadership.md
entangled90: Done, will test it in a real cluster next week.

docs/self-managed/components/orchestration-cluster/zeebe/operations/migrate-raft-leadership.md
```diff
@@ -0,0 +1,94 @@
...
+  -H 'Content-Type: application/json' \
+  -d '{
+    "zonePriorities": ["zone-b", "zone-a"]
+  }'
+```
+
+The priority change is asynchronous. Use the `changeId` from the response to poll that change every five seconds:
+
+```bash
+watch -n 5 'curl -s http://{zeebe-gateway}:9600/actuator/cluster/changes/{changeId} | jq "{id, status, pending}"'
```
entangled90: 🔧 I think we can avoid filtering the json with those three keys and avoid using `jq` altogether, the payload is quite small and readable.

leiicamundi:
### Chart-side context: the priorities this page swaps are also declared in Helm values

This page documents moving Raft leadership by reordering zone priorities through the Partition Distribution API. The other half of that lifecycle is in camunda/camunda-platform-helm#6949, which adds `orchestration.multiregion.zones[].priority` to the 8.10 chart — the value a zoned cluster boots with. Two things worth reflecting here.

**A runtime priority swap is not written back to the values file.** The chart renders `priority` into `application.yaml` from `orchestration.multiregion.zones`. After an operator follows this procedure, the live ordering and the values file disagree, and the values file is what the next `helm upgrade` re-renders. A sentence in the procedure saying to mirror the new order back into `orchestration.multiregion.zones[].priority` would close the loop — otherwise the swap looks permanent until an unrelated upgrade quietly reverts the preferred leader zone. This bites hardest where leadership placement is load-bearing: camunda/camunda-deployment-references#2940 skews leaders to the zone hosting the Aurora writer specifically to avoid an inter-region round trip on every export flush.

**Priority `0` is no longer expressible from the chart.** #6949 raised the schema constraint on `zones[].priority` from `minimum: 0` to `minimum: 1`. That does not restrict this procedure — the API reassigns the *existing* priority values by rank, as the page says, so no reachable ordering is lost — but if the page ever gains an example that sets an absolute priority rather than reordering, `0` is not a value a Helm-deployed cluster can start from.

Neither is a correction to what the page currently says; the API behaviour described here is accurate. Flagging it so the runtime procedure and the declarative side stay in step.

entangled90:
> A  runtime priority swap is not written back to the values file

not a problem. the configuration is only used at first boot, then what matters is what is modified through the API. That's saved in the filesystem of the nodes.

christinaausley (dismissed):
This looks great to me!

leiicamundi:
Quick review at `52e1e6e3a`. The procedure is accurate — I checked the `zonePriorities` semantics against the Management API reference and they match exactly (existing priority values reused and reassigned by request order, must list exactly the configured zones, idempotent, does not itself move leaders). All four cross-page anchors resolve: `rebalancing.md#manual-rebalancing`, `#limitations`, `#rebalancing-impact`, `#when-to-rebalance`, and `management-api.md#partition-distribution-api`.

Six things, roughly in order of how much they'd cost an operator.

**1. The `id` and filename say "migrate", and the page opens by saying it doesn't.** `migrate-raft-leadership` versus line 2: "This procedure changes only the priorities used for leader election. It doesn't change the partition distribution or move replicas between zones." Worse, "migrate" is about to mean something specific and different in this area — camunda/camunda-platform-helm#7049 is adding numbered-to-zoned broker *migration* with coexisting StatefulSets. Someone searching "migrate" will land here and find the wrong procedure. The title is already right; `move-raft-leadership` or `change-zone-priorities` would match it.

**2. Nothing warns that port 9600 should not be reachable from where you are running these commands.** Every example is `http://{zeebe-gateway}:9600` and the only access note is about the v2 API at the very end. The management port is unauthenticated, and the reference architectures deliberately keep it off any public listener — the ECS one reaches it through `aws ecs execute-command` into a broker task for exactly this reason (camunda/camunda-deployment-references#3019). A line saying to reach it via `kubectl port-forward` or an exec, not by exposing it, would stop someone opening it up to run a one-off procedure.

**3. "The cluster is fully zone-aware" is a prerequisite the reader cannot check.** The Management API reference links `configuration/zone-aware-clusters.md` for this; the how-to should too, and it is worth saying what a non-zone-aware cluster looks like here — `partitionDistribution.zones` is absent, so the `jq` in the first step returns nothing and the reason is not obvious.

**4. Omitting a zone: say what happens.** Line 58 says the request must include every configured zone, but not that it is rejected otherwise. The reference page confirms it must list exactly the configured zones. Stating that it fails rather than silently dropping a zone is what makes an operator comfortable running it.

**5. Aurora is named in an otherwise vendor-neutral page.** Line 25 says "the Amazon Aurora secondary storage writer" while the neighbouring paragraph stays generic with "Elasticsearch or OpenSearch". "The RDBMS secondary storage writer (for example, Amazon Aurora)" keeps the guidance applicable to any RDBMS backend.

**6. `watch` is not installed by default on macOS,** and the `jq` fallback offered for the first command is not offered for the polling one. A plain `curl` loop, or repeating the "if `jq` isn't installed" note, covers it.

Separately, and not a correction to this page: for Helm-deployed clusters this procedure leaves `orchestration.multiregion.zones[].priority` stale, so the next `helm upgrade` re-renders the original order and undoes the swap. I raised that in an earlier comment on this PR — it matters most where leadership placement is load-bearing, which is exactly the Aurora case in section 1.

docs/self-managed/components/orchestration-cluster/zeebe/operations/migrate-raft-leadership.md
```diff
@@ -0,0 +1,94 @@
+---
+id: migrate-raft-leadership
```
leiicamundi: The `id` says migrate, and line 9 says this procedure does not migrate anything: it reorders priorities and leaves the partition distribution alone. The title already gets this right with "Move".

The collision is the bigger problem. "Migration" is about to mean something specific and different in this area: camunda/camunda-platform-helm#7049 adds numbered-to-zoned broker migration, where two StatefulSets run side by side and brokers really do change identity. Someone who searches for the migration procedure will land on this page and follow the wrong one.

`move-raft-leadership` or `change-zone-priorities` would match the title and stay clear of that.

The sidebar entry and any inbound links need the same rename.

docs/self-managed/components/orchestration-cluster/zeebe/operations/migrate-raft-leadership.md
```diff
@@ -0,0 +1,94 @@
...
+
+Raft leadership placement is especially important when the Amazon Aurora secondary storage writer and Raft leaders are in different regions. Cross-region communication with the secondary storage writer adds network round-trip latency. After the secondary storage writer moves to another region, use this procedure to move Raft leadership to the same region when possible.
+
+The placement of Elasticsearch or OpenSearch secondary storage has less impact on this decision. In a multi-region setup, records are exported to secondary storage in both regions concurrently, so exporting has less dependency on cross-region network round-trip latency. Consider the location of the Aurora secondary storage writer first when choosing the preferred leader zone.
+
+## Prerequisites
+
+Before moving leadership, confirm the following conditions:
+
+- The cluster is fully zone-aware.
```
leiicamundi: This is the one prerequisite a reader cannot check. The other four are all verifiable with something they already have.

On a cluster that is not zone-aware, `partitionDistribution.zones` is absent, so the `jq` on line 41 prints nothing and the reason is not obvious. It looks like a broken command rather than an unmet prerequisite.

The Management API reference links `configuration/zone-aware-clusters.md` at this exact point. Linking it here too, and saying that an empty or missing `partitionDistribution.zones` means the cluster is not zone-aware, turns this into something the reader can act on.

docs/self-managed/components/orchestration-cluster/zeebe/operations/move-raft-leadership.md
```diff
@@ -0,0 +1,94 @@
...
+The placement of Elasticsearch or OpenSearch secondary storage has less impact on this decision. In a multi-region setup, records are exported to secondary storage in both regions concurrently, so exporting has less dependency on cross-region network round-trip latency. Consider the location of the Aurora secondary storage writer first when choosing the preferred leader zone.
+
+## Prerequisites
+
+Before moving leadership, confirm the following conditions:
+
+- The cluster is fully zone-aware.
+- All brokers and partitions are healthy.
+- The target zone's replicas are caught up with the current leaders.
+- You can access the Zeebe Gateway [Management API](management-api.md) on its management port. The default port is `9600`.
```
leiicamundi, line 30: Worth saying how to reach the management port, not just that you need to.

Port 9600 is unauthenticated, and the reference architectures keep it off any public listener on purpose. The ECS dual-region one reaches it with `aws ecs execute-command` into a running broker task, because there is deliberately no ALB listener rule for it (camunda/camunda-deployment-references#3019). Every example on this page is a bare `http://{zeebe-gateway}:9600`, so a reader whose gateway is not reachable from their laptop has to guess, and the obvious guess is to expose the port.

A sentence pointing at `kubectl port-forward` or an exec into a broker would prevent that. The `http://` also assumes TLS is off, which is worth a word if the management endpoint can be secured.

leiicamundi (commented):
Detailed follow-up to my earlier comment, now inline. The procedure itself checks out: I compared the `zonePriorities` semantics against the Management API reference and they agree on all four points, that existing priority values are reused and reassigned by request order, that the request must list exactly the configured zones, that it is idempotent, and that it does not move leaders by itself. All five cross-page anchors resolve as well: `manual-rebalancing`, `limitations`, `rebalancing-impact`, `when-to-rebalance`, and `partition-distribution-api`.

The comments below are about the page around that procedure, not the API calls in it.

One item does not belong on any line, so I am leaving it here. For clusters deployed with Helm, this procedure leaves `orchestration.multiregion.zones[].priority` stale in the values file, and the next `helm upgrade` re-renders the original order and undoes the swap. That matters most in the Aurora case this page opens with, where leadership placement is the whole point.

docs/self-managed/components/orchestration-cluster/zeebe/operations/migrate-raft-leadership.md
```diff
@@ -0,0 +1,94 @@
...
+
+Use this procedure to:
+
+- Perform a planned zone switchover.
+- Restore the preferred leadership placement after a zone failover.
+- Follow the sun by moving leadership to the preferred active region.
+
+## Align leadership with secondary storage
+
+Raft leadership placement is especially important when the Amazon Aurora secondary storage writer and Raft leaders are in different regions. Cross-region communication with the secondary storage writer adds network round-trip latency. After the secondary storage writer moves to another region, use this procedure to move Raft leadership to the same region when possible.
```
leiicamundi: Aurora is named directly here while the next paragraph stays generic with "Elasticsearch or OpenSearch". The latency argument holds for any RDBMS secondary storage with a single writer, not just Aurora.

"The RDBMS secondary storage writer (for example, Amazon Aurora)" keeps the guidance usable for the other RDBMS backends and still names the case most readers will recognise. Same for the second mention on line 21.

docs/self-managed/components/orchestration-cluster/zeebe/operations/migrate-raft-leadership.md
```diff
@@ -0,0 +1,94 @@
...
+
+The command returns each configured zone's name and priority. If `jq` isn't installed, omit the pipe to `jq` and manually inspect `partitionDistribution.zones` in the JSON response. You can also inspect `brokers[].partitions[]` in the full response to see the priority assigned to each partition replica.
+
+A higher priority makes a replica the preferred leader during an election. The zone with the highest configured priority is therefore the preferred zone for Raft partition leaders. Recording the current order also ensures that you preserve the relative priorities of any zones you aren't swapping.
+
+## Reorder the zone priorities
+
+The [Partition distribution API](management-api.md#partition-distribution-api) accepts a `zonePriorities` list. The first zone in the list receives the highest existing priority, the second zone receives the next highest priority, and so on.
+
+The request must include every currently configured zone. To exchange the priorities of two zones, reverse their positions and leave the other zones in their current order.
```
leiicamundi: Say what happens when the list is incomplete. Right now a reader knows the rule but not the consequence of breaking it, which is the difference between running the command confidently and not running it at all.

The Management API reference is explicit that the request must list exactly the currently configured zones and that it is idempotent, so the answer is that it is rejected rather than silently dropping a zone. Stating that here is a short sentence and removes the main reason to hesitate.

docs/self-managed/components/orchestration-cluster/zeebe/operations/migrate-raft-leadership.md
leiicamundi: Two small things on this one line.

`watch` is not installed by default on macOS, and this is the only command on the page that needs it. Line 44 already offers a fallback for readers without `jq`, so the polling step is the one place where a missing tool is not covered. A plain `curl` in a loop, or a note, would match what line 44 already does.

The `jq` filter is also inside double quotes within the single-quoted `watch` argument, which works, but repeating the line 44 fallback here would help anyone who has neither tool.

entangled90:
All comments from @leiicamundi are addressed.

leiicamundi (approved):
lgtm, ty!

PR #9826, opened 2026-09-04
Enhance overview with Early access release details

toco-cam:
Added sections for Early access release and What to expect in the overview.

## PR Checklist

### DIFF
--- a/versioned_docs/version-8.9/components/early-access/overview.md
+++ b/versioned_docs/version-8.9/components/early-access/overview.md
@@ -46,7 +46,7 @@ An Early access release is the release of a new product. By testing this release
 
 ## What to expect
 
-|                                    | Alpha features                                                                                   |
+|                                    |                                                                                    |
 | ---------------------------------- | ------------------------------------------------------------------------------------------------ |
 | <b>Purpose</b>                     | Test maturity of a new product before GA                                                         |
 | <b>Suitable for production use</b> | No                                                                                               |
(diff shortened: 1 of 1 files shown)


tasso94 (changes requested):
Technical feedback: you are adding this to the 8.9 docs only.
Content feedback: let me think about it more. Looks good superfically.

toco-cam:
@tasso94 
> Technical feedback: you are adding this to the 8.9 docs only.

Should we change it in all maintained once + upcoming 8.10?

tasso94:
@toco-cam, do we want to add ProcessOS docs for all minor versions in standard support? If the answer is yes, I'd backport this disclaimer as well. If we only add docs to 8.10, I'm not sure if we need this disclaimer on all still supported minor versions.

toco-cam:
This PR is only for the definition of EA. This discussion also impacts the new ProcessOS Harness Docs currently available as preview.

Question: To which versions should we add the ProcessOS Harness docs?

Options:
1. Only 8.10 -> Safe bet - users visiting the docs are routed by default to the current release.
2. All maintained once -> ProcessOS works with all supported versions -> No reason not to do it.
3. All version -> No value as no user will read them and the Camunda Versions might not be compatible.

We should go for option 2.

WDYT @jan-ladleif

mesellings:
> This PR is only for the definition of EA. This discussion also impacts the new ProcessOS Harness Docs currently available as preview.
> 
> Question: To which versions should we add the ProcessOS Harness docs?
> 
> Options:
> 
> 1. Only 8.10 -> Safe bet - users visiting the docs are routed by default to the current release.
> 2. All maintained once -> ProcessOS works with all supported versions -> No reason not to do it.
> 3. All version -> No value as no user will read them and the Camunda Versions might not be compatible.
> 
> We should go for option 2.
> 
> WDYT @jan-ladleif

Option 2 would be good, as in the future we might need differences in the docs versions depending on how ProcessOS evolves (e.g. new features etc just like a normal product), and also existing users might not necessarily visit the current docs, and only stick with their versioned docs.

PR #9830, opened 2026-09-04
docs(concepts): document secret references in input mappings and connector coexistence

### LINKED ISSUE #60329: [Secret Resolution] Docs: using secret references in process models
Do not start working on this until [this slack thread](https://camunda.slack.com/archives/C099MRHA8U8/p1788161063844789?thread_ts=1787556425.767989&cid=C099MRHA8U8) is resolved.

Parent: #56591

### Description

Document how to use `camunda.secrets.<name>` references in process models.

### Motivation

This is the primary user-facing surface of the feature, and the rules are specific enough that a modeller will not guess them. A reference in the wrong position fails at deployment rather than at runtime.

### Acceptance criteria

- [ ] Documented that references live in input mappings (variable mappings) on job worker elements
- [ ] Documented that a reference used as a string literal (for example `="camunda.secrets.token"`) is rejected at deployment by the model validator, with the error a user will actually see
- [ ] Documented that concatenation works (for example `="Bearer " + camunda.secrets.token`)
- [ ] Documented that a trailing path access (for example `camunda.secrets.token.length`) is not treated as a reference
- [ ] Documented that a name containing a dash is valid, but must be backtick-escaped in FEEL (``=camunda.secrets.`db-password` ``), because a bare dash is FEEL's minus operator
- [ ] Documented as a limitation that backtick escaping also admits names the `/v2/secrets` endpoints reject — for example ``=camunda.secrets.`tls.crt` `` resolves in a process model, but the same secret cannot be listed or resolved through the API, because the endpoints accept only `[\p{Alnum}_-]+`
- [ ] Documented that secrets inside of an array are rejected at deployment by the model validator
- [ ] Documented that secret references across input mappings won't be resolved. Eg. `x <- camunda.secrets.x` and `y <- x` will ony resolve the secret at `x`, not at `y`
- [ ] Documented what a job worker or outbound connector receives after resolution
- [ ] Legacy `{{secrets.X}}` compatibility explained: it remains supported for existing models, and the connector runtime keeps receiving the old format as plain strings
- [ ] Hybrid connector support confirmed in code before anything is written, then documented accordingly. The carried-over claim is that centralised resolution via the Orchestration Cluster is not supported for hybrid connectors in 8.10; it is unverified and must not be published unverified
- [ ] Documented that legacy secrets are physical-tenant unaware, and that physical tenants apply to new secret usage only

### Breakdown

- Write or extend the modelling page agreed in D0
- Update `docs/components/connectors/use-connectors/index.md` ("Using secrets") with the coexistence rule

### Notes

Do not edit the roughly 60 out-of-the-box connector pages individually. The coexistence rule belongs on the hub page.

The hybrid connector limitation is carried over from the feature brief and has not been confirmed in code. Confirming it is a blocking acceptance criterion above, not an optional check.

ajanoni-camunda:
## Description

Documents the new `camunda.secrets.<name>` input-mapping secret references (alpha) and how they coexist with the legacy `{{secrets.*}}` connector syntax. This complements the already-merged broker-side resolution page (`secret-resolution-and-job-activation.md`) by covering the modelling and connector-authoring side.

**`docs/components/concepts/variables.md`** — new "Secret references in input mappings" section on the modelling/variables page:

- References live in input mappings on job-worker elements (service tasks, business rule tasks, ad hoc sub-processes); links to the resolution page for what the worker receives.
- A reference must be exactly `=camunda.secrets.<name>`. Documents deployment-time rejections with the real validator errors: string literal (`="camunda.secrets.API_TOKEN"`) and reference inside a list/`if`-context (`=[camunda.secrets.API_TOKEN]`).
- Concatenation works (`="Bearer " + camunda.secrets.API_TOKEN`).
- A trailing path (`camunda.secrets.API_TOKEN.length`) is treated as ordinary FEEL path access, not a reference.
- Cross-mapping references are not resolved (`x <- camunda.secrets.X`, then `y <- x` leaves `y` as the placeholder).
- "Names with special characters": a dash-containing name is valid but must be backtick-escaped (`` =camunda.secrets.`db-password` ``) because a bare dash is FEEL's minus operator. Documents the limitation that backtick escaping admits names the `/v2/secrets` endpoints reject (they accept only `[\p{Alnum}_-]+`).

**`docs/components/connectors/use-connectors/index.md`** — new "Using `camunda.secrets.*` references" subsection under "Using secrets" with the coexistence rule:

- `{{secrets.*}}` remains fully supported and is still resolved by the connector runtime at execution time as plain placeholder text.
- `camunda.secrets.<name>` is resolved by the orchestration cluster before the job reaches any worker; the connector receives the value already in place, whether co-located or run separately.
- Physical-tenant scoping difference between the two forms.
- Migration via `camunda.connector.secret-resolver.legacy.mode=FALLBACK`.

Technical claims (hybrid resolution, legacy `ON`/`FALLBACK` behavior, physical-tenant scoping) were verified against the `camunda/camunda` engine (`ClusterVariableJobSecretResolver`, `BpmnJobBehavior`) and `camunda/connectors` (`LegacySecretMode`, `CentralStoreSecretProvider`, `ConnectorsAutoConfiguration`) source.

## PR Checklist

Closes https://github.com/camunda/camunda/issues/60329

### DIFF
--- a/docs/components/concepts/variables.md
+++ b/docs/components/concepts/variables.md
@@ -206,6 +206,63 @@ Examples:
 | -                                      | **source:** `"Peter"`<br/>**target:** `sender`                                                               | `sender: "Peter"`                           |
 | `customer:{"name": "John"}`            | **source:** (not provided)<br/>**target:** `customer`                                                        | `customer: null`                            |
 
+### Secret references in input mappings
+
+An input mapping's `source` can reference a secret directly, without first storing it in a process variable. Write the reference as `camunda.secrets.<name>` in a FEEL expression.
+
+This is part of an [alpha feature](/components/early-access/alpha/alpha-features.md) and may be subject to change in future releases.
+
+| Process variables | Input mappings                                                                       | New variables                                                    |
+| ----------------- | ------------------------------------------------------------------------------------ | ---------------------------------------------------------------- |
+| -                 | **source:** `=camunda.secrets.API_TOKEN`<br/>**target:** `token`                     | `token` holds the secret's value when the job reaches the worker |
+| -                 | **source:** `="Bearer " + camunda.secrets.API_TOKEN`<br/>**target:** `authorization` | `authorization` holds `"Bearer "` followed by the secret's value |
+
+Secret references are only resolved in input mappings defined on elements that create a job for a job worker (for example, service tasks, business rule tasks, and ad hoc sub-processes). See [secret resolution and job activation](secret-resolution-and-job-activation.md) for how a reference is resolved and what the worker receives once the job is handed out; until then, the variable holds the placeholder text `camunda.secrets.<name>`.
+
+A reference must be an expression, and it must be exactly `camunda.secrets.<name>` with nothing else attached:
+
+- Writing the reference as a plain string, or quoting it inside an expression, is rejected at deployment rather than passed through as literal text:
+
+  ```feel
+  ="camunda.secrets.API_TOKEN"
+  ```
+
+  ```
+  Secret reference(s) 'camunda.secrets.API_TOKEN' must be used as an expression (e.g. '=camunda.secrets.<name>'), not as a string literal, in input mapping source '="camunda.secrets.API_TOKEN"'.
+  ```
+
+- A trailing path after the name, such as `camunda.secrets.API_TOKEN.length`, is not treated as a reference. It's evaluated as an ordinary FEEL path access instead.
+
+- A reference placed inside a list, or inside a context built by an `if` branch, is rejected at deployment, because the reference is no longer a value the mapping assigns directly:
+
+  ```feel
+  =[camunda.secrets.API_TOKEN]
+  ```
+
+  ```
+  Input mapping source '=[camunda.secrets.API_TOKEN]' puts a secret reference inside a list, or inside a context built by an 'if' branch. Camunda can only replace a secret where the mapping assigns it directly to a value, so this secret would never be filled in. Assign each secret reference to its own input mapping instead.
+  ```
+
+Give each secret its own input mapping. A later mapping that reads the _variable_ created by an earlier one, rather than writing `camunda.secrets.<name>` itself, does not get the secret resolved:
+
+| Input mappings                                                                                        | New variables                                                                                     |
+| ----------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------- |
+| **source:** `=camunda.secrets.API_TOKEN`<br/>**target:** `x`<br/>**source:** `=x`<br/>**target:** `y` | `x` holds the resolved secret value; `y` holds the literal placeholder text, not the secret value |
+
+#### Names with special characters
+
+A secret name containing a character FEEL doesn't allow in a bare identifier — most commonly a dash — must be backtick-escaped, the same way any other FEEL name with special characters is:
+
+```feel
+=camunda.secrets.`db-password`
+```
+
+Without the backticks, `camunda.secrets.db-password` parses as subtraction (`db` minus `password`), not as a reference.
+
+:::note
+Backtick escaping accepts any name, including a name your secret store accepts but the [`/v2/secrets`](/apis-tools/orchestration-cluster-api-rest/specifications/list-secrets.api.mdx) endpoints do not. Those endpoints list and resolve only names matching `[\p{Alnum}_-]+`. A name outside that set, such as `tls.crt`, can be backtick-escaped and resolved from a process model (``=camunda.secrets.`tls.crt` ``) if your secret store holds it under that name, but the same secret cannot be listed or resolved through `/v2/secrets`.
+:::
+
 ### Output mappings
 
 Output mappings can be used for several purposes:
--- a/docs/components/concepts/variables.md
+++ b/docs/components/concepts/variables.md
@@ -212,12 +212,14 @@ An input mapping's `source` can reference a secret directly, without first stori
 
 This is part of an [alpha feature](/components/early-access/alpha/alpha-features.md) and may be subject to change in future releases.
 
+Using secret references requires a configured [secret store](/self-managed/components/orchestration-cluster/core-settings/configuration/properties.md#secrets) that holds the secret. Without a configured store, the reference cannot be resolved.
+
 | Process variables | Input mappings                                                                       | New variables                                                    |
 | ----------------- | ------------------------------------------------------------------------------------ | ---------------------------------------------------------------- |
 | -                 | **source:** `=camunda.secrets.API_TOKEN`<br/>**target:** `token`                     | `token` holds the secret's value when the job reaches the worker |
 | -                 | **source:** `="Bearer " + camunda.secrets.API_TOKEN`<br/>**target:** `authorization` | `authorization` holds `"Bearer "` followed by the secret's value |
 
-Secret references are only resolved in input mappings defined on elements that create a job for a job worker (for example, service tasks, business rule tasks, and ad hoc sub-processes). See [secret resolution and job activation](secret-resolution-and-job-activation.md) for how a reference is resolved and what the worker receives once the job is handed out; until then, the variable holds the placeholder text `camunda.secrets.<name>`.
+Secret references are only resolved in input mappings defined on elements that create a job for a job worker (for example, service tasks, business rule tasks, and ad hoc sub-processes). See [secret resolution and job activation](secret-resolution-and-job-activation.md) for how a reference is resolved and what the worker receives once the job is handed out. The stored process variable always holds the placeholder text `camunda.secrets.<name>`; resolution replaces it only in the payload handed to the worker, not in the variable kept in the process instance's state. Any other consumer of that variable sees the placeholder.
 
 A reference must be an expression, and the reference itself must be exactly the three-segment path `camunda.secrets.<name>`. It can still take part in a supported expression, such as the concatenation shown above, but the following rejections apply:
 
@@ -249,6 +251,14 @@ Give each secret its own input mapping. A later mapping that reads the _variable
 | ----------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------- |
 | **source:** `=camunda.secrets.API_TOKEN`<br/>**target:** `x`<br/>**source:** `=x`<br/>**target:** `y` | `x` exposes the resolved secret value to the worker; `y` holds the literal placeholder text, not the secret value |
 
+:::note
+A secret reference can also come from a [cluster variable](/components/modeler/feel/cluster-variable/data-types.md) of kind `SECRET_REFERENCE`. An input mapping that selects such a variable, for example `=camunda.vars.env.MY_CONFIG`, resolves the `camunda.secrets.<name>` references embedded in its value the same way. See [resolve secret references in a cluster variable](/components/modeler/feel/cluster-variable/usage-guide.md#resolve-secret-references-in-a-cluster-variable).
+:::
+
+:::note
+Avoid using `camunda` as a process variable name. A process variable literally named `camunda` takes precedence over the secret namespace, so `camunda.secrets.<name>` resolves against that variable and no secret is injected.
+:::
+
 #### Escape secret names with special characters
 
 A secret name containing a character FEEL doesn't allow in a bare identifier (most commonly a dash) must be backtick-escaped, the same way any other FEEL name with special characters is:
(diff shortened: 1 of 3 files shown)


docs/components/concepts/variables.md
```diff
@@ -206,6 +206,63 @@ Examples:
...
+| ----------------- | ------------------------------------------------------------------------------------ | ---------------------------------------------------------------- |
+| -                 | **source:** `=camunda.secrets.API_TOKEN`<br/>**target:** `token`                     | `token` holds the secret's value when the job reaches the worker |
+| -                 | **source:** `="Bearer " + camunda.secrets.API_TOKEN`<br/>**target:** `authorization` | `authorization` holds `"Bearer "` followed by the secret's value |
+
+Secret references are only resolved in input mappings defined on elements that create a job for a job worker (for example, service tasks, business rule tasks, and ad hoc sub-processes). See [secret resolution and job activation](secret-resolution-and-job-activation.md) for how a reference is resolved and what the worker receives once the job is handed out; until then, the variable holds the placeholder text `camunda.secrets.<name>`.
+
+A reference must be an expression, and it must be exactly `camunda.secrets.<name>` with nothing else attached:
```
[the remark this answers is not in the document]
ajanoni-camunda: Good catch, fixed. Reworded to: the reference itself must be the three-segment `camunda.secrets.<name>` path, but it can still take part in a supported expression such as the concatenation shown above.

docs/components/concepts/variables.md
```diff
@@ -206,6 +206,63 @@ Examples:
...
+
+  ```
+  Input mapping source '=[camunda.secrets.API_TOKEN]' puts a secret reference inside a list, or inside a context built by an 'if' branch. Camunda can only replace a secret where the mapping assigns it directly to a value, so this secret would never be filled in. Assign each secret reference to its own input mapping instead.
+  ```
+
+Give each secret its own input mapping. A later mapping that reads the _variable_ created by an earlier one, rather than writing `camunda.secrets.<name>` itself, does not get the secret resolved:
+
+| Input mappings                                                                                        | New variables                                                                                     |
+| ----------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------- |
+| **source:** `=camunda.secrets.API_TOKEN`<br/>**target:** `x`<br/>**source:** `=x`<br/>**target:** `y` | `x` holds the resolved secret value; `y` holds the literal placeholder text, not the secret value |
```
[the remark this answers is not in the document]
ajanoni-camunda: Fixed. Row now reads "`x` exposes the resolved secret value to the worker" to avoid implying the secret is persisted in process variables.

docs/components/connectors/use-connectors/index.md
```diff
@@ -57,6 +57,19 @@ our [Connector SDK documentation](/components/connectors/custom-built-connectors
...
+- `camunda.secrets.<name>` is resolved before the job reaches any worker, including a connector runtime. The connector receives the value already in place, the same way whether the connector runtime is co-located with the cluster or run separately (for example, a self-managed runtime connecting to a SaaS cluster).
+
+`{{secrets.*}}` values are not scoped per [physical tenant](/self-managed/concepts/physical-tenants/connectors-runtime.md#per-tenant-secret-access) unless you opt in to `physicaltenantaware` in the connector runtime's own configuration. Physical tenant scoping of `camunda.secrets.<name>` is separate from that setting: each physical tenant resolves its own configured secret store.
+
+To migrate a field from `{{secrets.NAME}}` to `camunda.secrets.NAME` without moving the value out of its current provider first, set `camunda.connector.secret-resolver.legacy.mode` to `FALLBACK` on the connector runtime: a legacy-style reference whose name isn't found in a configured secret provider is then looked up in the same store `camunda.secrets.<name>` uses. The default, `ON`, only resolves legacy references from the configured providers.
```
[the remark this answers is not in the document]
ajanoni-camunda: Right, the framing was reversed. Rewrote it: `FALLBACK` keeps the legacy `{{secrets.NAME}}` syntax working after the value is moved into the store `camunda.secrets.<name>` uses (legacy lookup falls back to that store when not found in a provider). It does not make a new `camunda.secrets.*` reference read from a runtime provider.

docs/components/concepts/variables.md
```diff
@@ -206,6 +206,63 @@ Examples:
...
+
+A reference must be an expression, and it must be exactly `camunda.secrets.<name>` with nothing else attached:
+
+- Writing the reference as a plain string, or quoting it inside an expression, is rejected at deployment rather than passed through as literal text:
+
+  ```feel
+  ="camunda.secrets.API_TOKEN"
+  ```
+
+  ```
```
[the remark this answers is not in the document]
ajanoni-camunda: Fixed. Added the `text` language identifier to both error-output blocks.

docs/components/concepts/variables.md
```diff
@@ -206,6 +206,63 @@ Examples:
...
+
+Give each secret its own input mapping. A later mapping that reads the _variable_ created by an earlier one, rather than writing `camunda.secrets.<name>` itself, does not get the secret resolved:
+
+| Input mappings                                                                                        | New variables                                                                                     |
+| ----------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------- |
+| **source:** `=camunda.secrets.API_TOKEN`<br/>**target:** `x`<br/>**source:** `=x`<br/>**target:** `y` | `x` holds the resolved secret value; `y` holds the literal placeholder text, not the secret value |
+
+#### Names with special characters
+
+A secret name containing a character FEEL doesn't allow in a bare identifier — most commonly a dash — must be backtick-escaped, the same way any other FEEL name with special characters is:
```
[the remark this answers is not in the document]
ajanoni-camunda: Fixed. Heading is now "Escape secret names with special characters" and the em dashes were removed.

docs/components/concepts/variables.md
```diff
@@ -206,6 +206,63 @@ Examples:
 | -                                      | **source:** `"Peter"`<br/>**target:** `sender`                                                               | `sender: "Peter"`                           |
 | `customer:{"name": "John"}`            | **source:** (not provided)<br/>**target:** `customer`                                                        | `customer: null`                            |
 
+### Secret references in input mappings
```
HeleneW-dot, line 209: Can we also add a short mention that in order to use secret references, it is a pre requisite to configure a secret store first?

docs/components/concepts/variables.md
```diff
@@ -206,6 +206,63 @@ Examples:
...
+
+| Process variables | Input mappings                                                                       | New variables                                                    |
+| ----------------- | ------------------------------------------------------------------------------------ | ---------------------------------------------------------------- |
+| -                 | **source:** `=camunda.secrets.API_TOKEN`<br/>**target:** `token`                     | `token` holds the secret's value when the job reaches the worker |
+| -                 | **source:** `="Bearer " + camunda.secrets.API_TOKEN`<br/>**target:** `authorization` | `authorization` holds `"Bearer "` followed by the secret's value |
+
+Secret references are only resolved in input mappings defined on elements that create a job for a job worker (for example, service tasks, business rule tasks, and ad hoc sub-processes). See [secret resolution and job activation](secret-resolution-and-job-activation.md) for how a reference is resolved and what the worker receives once the job is handed out; until then, the variable holds the placeholder text `camunda.secrets.<name>`.
```
HeleneW-dot: This part: `until then, the variable holds the placeholder text `camunda.secrets.<name>`.`

So my understanding is technically, the variable _always_ holds the placeholder text. right? So every consumer of that variable would also be getting the placeholder. This reads as if the actual variable value will be replaced with the real secret

docs/components/connectors/use-connectors/index.md
```diff
@@ -57,6 +57,19 @@ our [Connector SDK documentation](/components/connectors/custom-built-connectors
 Using this in other areas can lead to unexpected results and incidents.
 :::
 
+### Using `camunda.secrets.*` references
+
+You can also reference a secret from a [secret store](/self-managed/components/orchestration-cluster/core-settings/configuration/properties.md#secrets) directly in a connector's input mapping, using `camunda.secrets.<name>` in a FEEL expression. This is part of an [alpha feature](/components/early-access/alpha/alpha-features.md). See [secret references in input mappings](/components/concepts/variables.md#secret-references-in-input-mappings) for the syntax and its rules.
+
+The two forms coexist and are handled differently:
```
HeleneW-dot: I think there's a third form: a `SECRET_REFERENCE`-kind cluster variable read by an input
mapping. Its references are recorded on the job and resolved the same way
(`ClusterVariableJobSecretResolver`, merged in `BpmnJobBehavior.mergedSecretReferences`),
so a connector field written `=camunda.vars.env.MY_CONFIG` can also arrive with values
already in place. This is also documented in `components/modeler/feel/cluster-variable/data-types.md`.

docs/components/concepts/variables.md
```diff
@@ -206,6 +206,63 @@ Examples:
...
+
+A reference must be an expression, and the reference itself must be exactly the three-segment path `camunda.secrets.<name>`. It can still take part in a supported expression, such as the concatenation shown above, but the following rejections apply:
```
HeleneW-dot, line 224: For the cluster-variable route, I think `=camunda.vars.env.MY_VAR`
does reach a secret when the variable is `SECRET_REFERENCE`-kind (see
`components/modeler/feel/cluster-variable/usage-guide.md#resolve-secret-references-in-a-cluster-variable`)
Maybe we could also link the cluster-variable page from here?

docs/components/concepts/variables.md
```diff
@@ -206,6 +206,63 @@ Examples:
...
+A secret name containing a character FEEL doesn't allow in a bare identifier (most commonly a dash) must be backtick-escaped, the same way any other FEEL name with special characters is:
+
+```feel
+=camunda.secrets.`db-password`
+```
+
+Without the backticks, `camunda.secrets.db-password` parses as subtraction (`db` minus `password`), not as a reference.
+
+:::note
+Backtick escaping accepts any name, including a name your secret store accepts but the [`/v2/secrets`](/apis-tools/orchestration-cluster-api-rest/specifications/list-secrets.api.mdx) endpoints do not. Those endpoints list and resolve only names matching `[\p{Alnum}_-]+`. A name outside that set, such as `tls.crt`, can be backtick-escaped and resolved from a process model (``=camunda.secrets.`tls.crt` ``) if your secret store holds it under that name, but the same secret cannot be listed or resolved through `/v2/secrets`.
```
HeleneW-dot: So this note's "backtick escaping accepts any name" applies only to a reference written directly in an input mapping source, right? Not to a reference inside a SECRET_REFERENCE-kind cluster variable value. 
We're already explicitly in the input mapping section, but just to avoid confusion can we scope it more explicitly? That way someone just searching for secret char escaping in the docs wont be mislead
Maybe we could add a similar explicit scope to cluster-variable/data-types.md, which also states the restricted charset but in that case for cluster variables

docs/components/concepts/variables.md
```diff
@@ -206,6 +206,63 @@ Examples:
...
+  ```
+
+Give each secret its own input mapping. A later mapping that reads the _variable_ created by an earlier one, rather than writing `camunda.secrets.<name>` itself, does not get the secret resolved:
+
+| Input mappings                                                                                        | New variables                                                                                                     |
+| ----------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------- |
+| **source:** `=camunda.secrets.API_TOKEN`<br/>**target:** `x`<br/>**source:** `=x`<br/>**target:** `y` | `x` exposes the resolved secret value to the worker; `y` holds the literal placeholder text, not the secret value |
+
+#### Escape secret names with special characters
```
HeleneW-dot, line 262: Should we also add a note about reserving the variable name `camunda`?
A process [variable literally named `camunda`](https://github.com/camunda/camunda/blob/2cad3f17822f1de7d83284390c5134d7c126dea0/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/expression/SecretReferenceEvaluationContext.java#L59) keeps precedence, so camunda.secrets.* resolves against it and no secret is injected.

docs/components/concepts/variables.md
```diff
@@ -206,6 +206,63 @@ Examples:
...
+
+A secret name containing a character FEEL doesn't allow in a bare identifier (most commonly a dash) must be backtick-escaped, the same way any other FEEL name with special characters is:
+
+```feel
+=camunda.secrets.`db-password`
+```
+
+Without the backticks, `camunda.secrets.db-password` parses as subtraction (`db` minus `password`), not as a reference.
```
HeleneW-dot: [Here](https://github.com/camunda/camunda/blob/2cad3f17822f1de7d83284390c5134d7c126dea0/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/model/element/SecretReference.java#L43) we document that the result of the subtraction does get read as a reference and then fails at evaluation

HeleneW-dot (changes requested):
Looks mostly good to me, I just have some comments on parts that we could make more clear 🙏

docs/components/concepts/variables.md
ajanoni-camunda: Good point. Added a prerequisite line: "Using secret references requires a configured secret store that holds the secret. Without a configured store, the reference cannot be resolved."

docs/components/concepts/variables.md
ajanoni-camunda: You're right, the old wording implied the stored value gets replaced. Reworded to: "The stored process variable always holds the placeholder text  camunda.secrets.<name> ; resolution replaces it only in the payload handed to the worker, not in the variable kept in the process instance's state. Any other consumer of that variable sees the placeholder."

docs/components/connectors/use-connectors/index.md
ajanoni-camunda: Added as a third bullet: a cluster variable of kind  SECRET_REFERENCE  can hold  camunda.secrets.<name>  references, and a connector field reading it (e.g.  =camunda.vars.env.MY_CONFIG ) receives the resolved value the same way, with a link to resolve secret references in a cluster variable.

docs/components/concepts/variables.md
ajanoni-camunda: Added a note that a reference can also come from a  SECRET_REFERENCE -kind cluster variable selected via  =camunda.vars.env.<name> , linking both the cluster-variable data types and usage guide pages.

docs/components/concepts/variables.md
ajanoni-camunda: Scoped it explicitly to references written directly in an input mapping source, and clarified it does not apply to references embedded in a  SECRET_REFERENCE -kind cluster variable value (which  follow the restricted cluster-variable character set). Also added the matching scope note to  cluster-variable/data-types.md .

docs/components/concepts/variables.md
ajanoni-camunda: Confirmed against  SecretReferenceEvaluationContext  (a real  camunda  variable is kept intact rather than shadowed). Added a note: "Avoid using  camunda  as a process variable name. A process variable literally named  camunda  takes precedence over the secret namespace, so  camunda.secrets.<name>  resolves against that variable and no secret is injected."

docs/components/concepts/variables.md
ajanoni-camunda: Corrected. Confirmed against  SecretReference  javadoc. Now reads: " camunda.secrets.db-password  parses as a subtraction ( db  minus  password ). FEEL reads the left operand as a reference named  db , which is not a valid secret reference, so the deployment fails at evaluation."

HeleneW-dot (approved):

PR #9847, opened 2026-09-07
docs(self-managed): fix Helm values that do not exist in the chart in the security reference

p-wunderlich:
## Description

The **Helm values** tab of the Orchestration Cluster security property reference documents 22 `orchestration.security.*` keys that no Camunda Helm chart declares or reads. Helm ignores unknown values silently, so an operator who copies them into `values.yaml` gets no error and no warning, and the setting simply never takes effect.

Partially addresses camunda/camunda-security-library#623 — see the "Version coverage" section below for why the issue stays open. No chart change is requested or needed; the chart is behaving correctly.

**Verified against `charts/camunda-platform-8.10` in `camunda/camunda-platform-helm`:** `csrf`, `httpHeaders`, and `http-headers` have zero matches in `values.yaml`, `templates/`, and `values.schema.json`. Every `orchestration.*` key in the changed tab was checked back against the chart's `values.yaml`, in both directions — every documented key exists, and every chart key under `orchestration.security.authentication.oidc` is now documented (17 of 17).

> All `charts/camunda-platform-8.10` `file:line` references below are against `camunda/camunda-platform-helm@f061fe4bfb96` at the time of writing. They are pointers for review, not durable citations.

### Group A — `csrf` and `httpHeaders` (17 rows)

Replaced the two tables with a single `orchestration.extraConfiguration` example carrying the real `camunda.security.csrf.*` and `camunda.security.http-headers.*` application properties, which are already documented correctly in the **Application properties** tab of the same page.

Deleting the tables outright would have left operators with no documented way to set CSP, HSTS, or CSRF under Helm at all. These keys will never be added to the chart: [ADR 0091](https://github.com/camunda/camunda-platform-helm/blob/main/docs/adr/0091-adopt-component-extraconfiguration-as-the-standard-application-configuration-mechanism.md) classifies Spring Boot application properties as Tier 1 / blocked, with `<component>.extraConfiguration` as the sanctioned mechanism.

The example deliberately sets only `csrf` and `hsts`. `content-security-policy.policy-directives` replaces the shipped policy wholesale — the default additionally permits `cdn.jsdelivr.net`, `*.chargebee.com`, `*.mixpanel.com`, `'unsafe-inline'` styles, `data:` images, and `blob:` workers — so an operator pasting a minimal `default-src 'self'` from the reference page would hard-break the web UIs. The surrounding text says this and points at the default policy.

The duplicated `#### Default Content Security Policy` block that sat under the removed `httpHeaders` table was dropped as well; the new text **names** the corresponding block in the **Application properties** tab rather than linking to it. A link would not have worked: `<Tabs>` here has no `groupId` and no `lazy`, so Docusaurus renders inactive `TabItem` panels with the `hidden` attribute instead of unmounting them. A same-page anchor into another tab therefore resolves in the built HTML — `onBrokenLinks: "throw"` and the link checkers all pass — but clicking it neither switches the tab nor scrolls, because the target sits inside a hidden subtree. On `main` this same reference was intra-tab and worked; deleting the triplicate heading is what turned it cross-tab. The only reference to its `#default-content-security-policy-2` anchor was inside the deleted table.

### Group B — 5 keys documented under the wrong name

| Documented | Actual chart key |
| :--- | :--- |
| `orchestration.security.authentication.oidc.clientSecret` | `...oidc.secret.inlineSecret`, `...secret.existingSecret`, `...secret.existingSecretKey` |
| `orchestration.security.authentication.oidc.userNameClaim` | `...oidc.usernameClaim` (lowercase `n`) |
| `orchestration.security.authentication.oidc.audiences` | `...oidc.audience` (singular) |
| `orchestration.security.multiTenancy.checksEnabled` | `orchestration.multitenancy.checks.enabled` (not under `security`) |
| `orchestration.security.multiTenancy.apiEnabled` | `orchestration.multitenancy.api.enabled` (not under `security`) |

The two multi-tenancy rows moved out from under the `orchestration.security.multiTenancy` heading, which is now `orchestration.multitenancy`. The chart 15.x deprecation note above the tab was widened from `orchestration.security.*` to `orchestration.*` to keep covering them.

### Group C — 5 wrong default values

Corrected against `charts/camunda-platform-8.10/values.yaml`: `usernameClaim` → `preferred_username`, `audience` → `orchestration-api`, `clientIdClaim` → `client_id`, `clientId` → `orchestration`, `redirectUrl` → `http://localhost:8080`. The first two were documented with application-side rather than chart defaults; the next two were plainly stale.

`redirectUrl` is the interesting one: the table showed `http://localhost:8080/sso-callback`, the effective redirect URI. The chart value is `http://localhost:8080` (`values.yaml:2140`) and `/sso-callback` is appended by the chart itself when it renders `redirect-uri` (`_helpers.tpl:549-552` passes the value through unchanged; `files/_application.yaml:192` does the `printf`). Pasting the documented value into `values.yaml` therefore produced a doubled suffix. The cell now shows the chart value and the description names the suffix.

### Group D — wrong description, and 5 missing rows

**`issuer` described the wrong property.** It read "Sets the OIDC provider's authorization URI explicitly", but `issuer` maps to `issuer-uri` (`_helpers.tpl:185-187`). The authorization URI comes from a separate value, `oidc.authUrl` (`_helpers.tpl:193-195` → `authorization-uri`) — which was not in the table at all, so there was no way to notice. Both rows now name the endpoint they set, and `authUrl`/`tokenUrl`/`jwksUrl` carry the chart's condition from `values.yaml:2115-2120`: they only apply when neither `oidc.issuer` nor `global.identity.auth.issuer` is set, and they support templating.

**Five values existed in the chart but not in the table:** `type`, `authUrl`, `preferUsernameClaim`, `scope`, `backwardsCompatibleAudiences`. The table now has exactly 17 rows — the 14 scalar keys under `orchestration.security.authentication.oidc` plus the three `secret.*` keys — matching `values.yaml:2110-2148` one-for-one.

One caveat worth stating rather than burying: **`type` is a chart-only switch.** Unlike the other four it emits no `camunda.security.*` property. Its single consumer is `orchestration.authIssuerType` (`_helpers.tpl:179-183`), which drives Keycloak-vs-generic endpoint derivation and hub-ping authorization gating. It belongs in the table because it is a real key under that prefix, but its description says what it actually does, so the row does not imply an application property that does not exist.

**`type` is also case-sensitive in a way nothing catches.** `camundaPlatform.authIssuerType` applies `upper()` to `global.identity.auth.type` (`common/_helpers.tpl:273-275`), and `constraints.tpl:388` validates only that global against `KEYCLOAK`/`MICROSOFT`/`GENERIC`. But `orchestration.authIssuerType` (`orchestration/_helpers.tpl:179-183`) passes `oidc.type` through raw. So `oidc.type: keycloak` renders without an error and silently takes the non-Keycloak branch, dropping the derived authorization, token, and JWKS endpoint URLs. The row now says the value must be uppercase and what happens if it is not.

Of the five added rows, `preferUsernameClaim`, `scope`, and `backwardsCompatibleAudiences` are deprecated in chart 15.x (`constraints.tpl:1029`, `:1089`, `:1092`); `type` and `authUrl` are not, same as `oidc.secret.*` and `audience`. `redirectUrl`, which was already in the table, is deprecated too (`constraints.tpl:1108`) — it is now marked as such, see below.

### Per-row deprecation markers

The tab-level note said "Some of the `orchestration.*` Helm values listed in this tab are deprecated", which left the reader to guess which. **14 of the 27 documented rows are deprecated** — a majority — so each one now carries a `**Deprecated in chart 15.x.**` lead-in in its Description cell, and the note was rewritten to point at those markers instead of saying "some". The lead-in follows the convention this same file already uses for `SPRING_PROFILES_ACTIVE` in the Environment variables tab.

The 14 were derived by attributing every entry in `templates/common/constraints.tpl` to its enclosing helper, to separate `camundaPlatform.keyDeprecated` (still works, logs a `DEPRECATION` warning) from `camundaPlatform.keyRemoved` (hard-rejected). All 14 marked rows are genuinely `keyDeprecated`. The only two `keyRemoved` entries in that prefix are the flat `oidc.existingSecret` / `oidc.existingSecretKey`, neither of which is a documented row — the table documents `oidc.secret.*`, a different key, which is not deprecated. Cross-checked 14/14 against `/self-managed/upgrade/helm/890-to-8100.md`, where five of them appear collapsed into one line in brace notation.

Marked: `authentication.authenticationRefreshInterval`, `authentication.unprotectedApi`, `oidc.redirectUrl`, `oidc.usernameClaim`, `oidc.preferUsernameClaim`, `oidc.clientIdClaim`, `oidc.groupsClaim`, `oidc.scope`, `oidc.backwardsCompatibleAudiences`, `authorizations.enabled`, `initialization.mappingRules`, `initialization.authorizations`, `multitenancy.checks.enabled`, `multitenancy.api.enabled`.

Because the note tells the reader to migrate each marked row "using the application property names from the **Application properties** tab", every marked row was checked against that tab. Twelve resolve in-page. Two did not, and now carry a cross-page link instead:

- **`oidc.redirectUrl`** sets three application properties, not one. Only `camunda.security.authentication.oidc.redirect-uri` appears on this page; `camunda.operate.identity.redirectRootUrl` and `camunda.tasklist.identity.redirectRootUrl` appear nowhere on it. The row now says so and links to the upgrade guide for the full mapping.
- **`oidc.backwardsCompatibleAudiences`** has its target on the page, but the semantics differ: the Helm key *appended* to the audience list, while `camunda.security.authentication.oidc.audiences` *replaces* it. Migrating verbatim rejects the component tokens with `UNAUTHENTICATED: Expected a valid token`. The row now warns and links to the troubleshooting section.

### `initialization.users` and `initialization.defaultRoles` showed blank defaults

Both Default cells were empty, but the chart ships non-empty, security-relevant defaults (`values.yaml:2158` ff. and `:2190`): a `demo` user with password `demo`, a `connectors` user with password `connector`, `demo` assigned to the `admin` role, and the `connectors` user and client assigned to `connectors`. A blank cell reads as "nothing is created by default", which is the opposite of what happens. The descriptions now name the shipped credentials and say to replace or remove them before exposing a cluster.

Same defect class as Group C, but outside the 22 keys in camunda/camunda-security-library#623 — found while reviewing the rest of the tab.

**On the Default column.** Two cells show an inherited default rather than the literal chart value, and both descriptions say so. `authentication.method` keeps `basic`: the chart value is `""` (`values.yaml:2103`) but it falls back to `global.security.authentication.method` (`values.yaml:132`), and `basic` is what an operator actually gets. `oidc.type` is the same shape — chart value `""`, fallback `global.identity.auth.type` = `KEYCLOAK` (`values.yaml:599`), via `camundaPlatform.authIssuerType` (`common/_helpers.tpl:273-275`), and `constraints.tpl:388` fails the render unless that global is one of the three flavors, so the effective value is never empty. Two further cells — the `initialization.users` and `initialization.defaultRoles` rows above — are **summaries** rather than literal values, because the real defaults are structured (a list of maps and a map of maps); they are marked as such in the cell so the column is not read as a list of strings. Every remaining cell is the literal `values.yaml` default, and the blank `issuer`/`authUrl`/`tokenUrl`/`jwksUrl` rows are blank because their fallbacks are empty too — named in the description either way.

### Version coverage

**This PR fixes 8.10 (`docs/`) only.** The Camunda Security Library is in place from 8.10 onward, and that is the scope agreed for this change.

The same tables are still wrong in `versioned_docs/version-8.9/` (the current default version) and `versioned_docs/version-8.8/` — the keys do not exist in those charts either. That is a deliberate scoping decision, not an oversight, which is why camunda/camunda-security-library#623 stays open and tracks 8.9/8.8 coverage. 8.7 does not carry these tables.

Two notes for whoever picks that up. 8.8 has no `partials/_security.md` at all — its copy of the tables lives inline in `configuration/properties.md`. And the 8.8 chart consumes `extraConfiguration` as a **map**, not the 8.9+ ordered list, and has no `spring.config.import`, so the snippet used here cannot be copied to 8.8 as-is — `orchestration.env` with `CAMUNDA_SECURITY_*` is the working mechanism there.

## PR Checklist


docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/_security.md
```diff
@@ -573,104 +573,86 @@ script-src-attr 'none'.
...
+| `orchestration.security.authentication.method`                        | The authentication method to use. Options: `basic`, `oidc`. The chart value is empty by default and inherits from `global.security.authentication.method`, which is `basic`.                                                                                                                                                                           | `basic`       |
+| `orchestration.security.authentication.authenticationRefreshInterval` | **Deprecated in chart 15.x.** The interval at which the memberships (groups, roles, tenants, component authorizations) are refreshed for logged in users. Find more details in [webserver and security](/self-managed/components/orchestration-cluster/core-settings/configuration/webserver.md#propagation-of-membership-changes-to-active-sessions). | `PT30S`       |
```
Ben-Sheppard, line 586: 🔧 NABD — 14 of these description cells now carry an identical literal `**Deprecated in chart 15.x.**` lead-in with no single place to update when chart 16.x changes status for a subset of them. Mirrors the existing `SPRING_PROFILES_ACTIVE` convention in this file, so not a blocker — just flagging the maintenance cost if this pattern keeps growing. A dedicated status column might scale better long-term.

docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/_security.md
```diff
@@ -573,104 +573,86 @@ script-src-attr 'none'.
...
+              enabled: true
+            http-headers:
+              hsts:
+                enabled: true
+                max-age-in-seconds: 31536000
+                include-subdomains: true
+```
 
-This is default value of the Content Security Policy when enabled:
+Leaving `content-security-policy` out, as above, keeps the shipped policy in place — see **Default Content Security Policy** in the **Application properties** tab. Setting `content-security-policy.policy-directives` replaces that policy wholesale rather than adding to it, so a minimal value such as `default-src 'self'` drops the sources the web applications need and breaks the UIs. Start from the default policy and adjust it.
```
Ben-Sheppard, line 637: 🔧 NABD — This is a same-page reference by name only, no link, because `<Tabs>` here has no `groupId`/`lazy` so an anchor into a hidden `TabItem` wouldn't actually navigate for the reader (correctly diagnosed in the PR description). Worth a short `<!-- comment -->` here noting *why* there's no link, so the next editor doesn't "fix" it by adding a broken one.

Ben-Sheppard (dismissed):
Docs-only fix for 22 fictitious Helm keys in the security reference, verified against the chart source (spot-checked directly against camunda-platform-helm@f061fe4bf) with an unusually thorough self-review. No correctness or security issues found — only two minor maintainability nits, neither blocking.

docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/_security.md
p-wunderlich: Added — a comment noting why there's no link, so a future editor doesn't "fix" it by adding a broken one.

docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/_security.md
p-wunderlich: Agreed this mirrors the `SPRING_PROFILES_ACTIVE` pattern already in the file. Not doing a status column here — it'd mean widening two already-wide tables on an approved PR. If the deprecated-lead-in pattern keeps growing (the trigger you named), that's the point to revisit it as a follow-up.

Ben-Sheppard (approved):
Re-approving after the follow-up commit — it addresses the unlinked cross-tab reference nit exactly as suggested (adds an explanatory HTML comment). The other nit (repeated deprecation-marker literal) was reasonably declined rather than changed; no blockers. Nothing else in the diff changed since the prior review, and the chart-side facts were separately verified against camunda-platform-helm@f061fe4bf.

PR #9879, opened 2026-09-09
docs(zeebe): align documentation for add/remove zone with latest changes

entangled90:
## Description

Extend/align the guide/reference in `management-api` with the latest changes in the application. 
This means:
- remove a zone defaults to `force=false` now
- `partition-distribution` -> `partitioning`

- camunda/camunda#63258
- camunda/camunda#63247

## PR Checklist


docs/self-managed/components/orchestration-cluster/zeebe/operations/add-remove-zone.md
```diff
@@ -0,0 +1,151 @@
+---
+id: add-remove-zone
+title: "Add or remove a zone"
+description: "Grow or shrink a zone-aware cluster by adding or gracefully removing a zone."
+---
+
+This guide walks through adding a new zone to a running [zone-aware](/self-managed/components/orchestration-cluster/zeebe/configuration/zone-aware-clusters.md) cluster, and gracefully removing an existing, healthy zone.
+
+The cluster must already be zone-aware before following this guide. If your cluster is still bare or only partially zoned, first migrate it using [`PUT /cluster/partition-distribution`](/self-managed/components/orchestration-cluster/zeebe/operations/management-api.md#partition-distribution-api) and [`PUT /cluster/zones`](/self-managed/components/orchestration-cluster/zeebe/operations/management-api.md#migrate-a-zone-to-a-zone-aware-topology).
```
entangled90, line 9: TODO: we need a dedicated how-to guide for migrating a bare/partially-zoned cluster to zone-aware (currently only documented at the API-reference level in `management-api.md`). This line should eventually link to that guide instead of the raw API calls.

leiicamundi:
Consolidated the zone-aware migration procedure into this PR, since the TODO on `add-remove-zone.md:9` and the documentation handoff from the chart are the same gap.

**Where it comes from.** [camunda/camunda-platform-helm#7049](https://github.com/camunda/camunda-platform-helm/pull/7049) carried the procedure as `docs/zone-aware-migration-procedure.md` and removed it in `4cda5a22e3` at maintainer request, so the operational guide can live here. This PR is its new home.

**What was added.**

- `migrate-to-zone-aware-brokers.md`: replacing numbered brokers with zoned ones one zone at a time, the coexistence phase, the management API steps (`PUT /cluster/partition-distribution` then `PUT /cluster/zones`), the checks before removing the old generation, and the PDB caveat during coexistence. The engine-level steps are deployment-agnostic; the chart specifics sit in their own `### Helm` subsections.
- `add-remove-zone.md:9` now links that guide instead of the two raw API endpoints, which resolves the TODO in https://github.com/camunda/camunda-docs/pull/9879#discussion_r3970458552.
- `zone-removal-notes.md` was removed from the repository root: it is working material (a TODO, source references, and an open question the notes answer themselves), not a documentation page.

**One thing to check before this leaves draft.** The `### Helm` subsections describe `orchestration.multiregion.mode`, `zone`, `zones[]`, and `keepUnzonedBrokers`, which only exist once camunda-platform-helm#7049 merges. That PR is still open with changes requested, so the value names can still move. The engine-level sections do not depend on it.

Two commits, split so each message matches its content: the guide, then the notes removal.

docs/self-managed/components/orchestration-cluster/zeebe/operations/add-remove-zone.md
leiicamundi: Addressed in a466ac7ba: this line now links [Migrate a cluster to zone-aware brokers](https://github.com/camunda/camunda-docs/blob/cs/add-zone-to-zoned-cluster/docs/self-managed/components/orchestration-cluster/zeebe/operations/migrate-to-zone-aware-brokers.md) instead of the two raw API endpoints. The guide is the procedure that was removed from camunda-platform-helm#7049 in 4cda5a22e3 for the handoff, rewritten with the engine-level steps first and the chart specifics in their own subsections.

leiicamundi:
Review pass on the consolidated diff. GitHub Copilot has not reviewed this PR: it is still a draft, and a review request does not stick until the PR is marked ready for review. So this is a manual pass instead, with the chart claims checked against the current head of [camunda-platform-helm#7049](https://github.com/camunda/camunda-platform-helm/pull/7049) rather than against the procedure file that was removed from it, because several commits landed in that area afterwards.

**Fixed in this branch**

- `24a95e985`: the guide explained what `regions`, `regionId`, and `clusterSize` mean during retention, but not that the chart enforces them. `constraints.tpl` fails the render when `regions` or `regionId` is missing, when `regionId` is not less than `regions`, or when `clusterSize` is not a positive integer divisible by `regions`. Those rules are now a table in step 1.
- `317173e1f`: the dual-region ordering rule (secondary zone first) only appeared in the introduction, while the reader acts on it in the last step. It is now restated where it applies.

**Re-verified against 7049 head, still accurate**

- The shared ServiceAccount stays unsuffixed across both generations (`8be82c0ab0`, "keep shared broker resources generation-neutral").
- The remote contact-point override through `orchestration.env` is preserved (`f65ec3762c`), so the guidance for brokers owned by another release still holds.
- The zoned `clusterSize` and `replicationFactor` checks stand down while retention is on and re-arm once it is off, which is what step 3 tells the operator to act on.

**Open, structural, needs a decision**

The page carries two layers: engine steps that work for any deployment, and chart steps that only exist once #7049 merges. Every other page under `zeebe/operations/` stays deployment-agnostic, and the canonical home for chart runbooks is `self-managed/deployment/helm/operational-tasks/`. Splitting the `### Helm` subsections out would let the engine half ship independently of a chart PR that still has changes requested. Leaving it as one page is defensible for a single reading flow, but then this PR cannot leave draft before the chart does.

**Open, low**

The `while true; do curl .../changes/{changeId}; sleep 5; done` block now appears three times across the docs (twice in `add-remove-zone.md`, once in the new guide, plus `in-process-restore.md`). It is an existing pattern rather than something this PR introduced, but it is a candidate for a shared partial.

entangled90:
as we're renaming the API/fields from `partition-distribution` to `partitioning` in camunda/camunda#63258 I'm modifying it directly in this PR

entangled90:
I rebased on main and decided to just include more details into the existing management API section instead of creating a new page. 
It includes fixes after the rename of the APIs in 
- camunda/camunda#63258
- camunda/camunda#63247

docs/self-managed/components/orchestration-cluster/zeebe/operations/management-api.md
```diff
@@ -18,15 +18,19 @@ The API is a custom endpoint available via [Spring Boot Actuator](https://docs.s
 For additional configurations such as security, refer to the official [Spring Boot documentation](https://spring.io/guides).
 :::
 
+The management port is typically not publicly exposed. If the gateway is not reachable from the machine where you run these commands, use a private connection such as `kubectl port-forward svc/camunda-zeebe-gateway 9600:9600`, then use `localhost` as the gateway host. The examples use `http://` for a management endpoint without TLS. If your endpoint uses TLS, use `https://` and the appropriate `curl` TLS options.
```
panagiotisgts: > If the gateway is not reachable from the machine where you run these commands,

I'd rephrase this a bit better

docs/self-managed/components/orchestration-cluster/zeebe/operations/management-api.md
```diff
@@ -315,20 +319,20 @@ Use this endpoint to update the [zone-aware](/self-managed/components/orchestrat
...
 ```
 curl -X 'PUT' \
-   'http://localhost:9600/actuator/cluster/partition-distribution' \
+   'http://localhost:9600/actuator/cluster/partitioning' \
    -H 'accept: application/json' \
    -H 'Content-Type: application/json' \
    -d '{
         "config": {
-          "type": "ZONE_AWARE",
+          "scheme": "ZONE_AWARE",
```
panagiotisgts, line 335: Really like the _scheme_ wording 💯

docs/self-managed/components/orchestration-cluster/zeebe/operations/management-api.md
```diff
@@ -432,23 +406,19 @@ POST actuator/cluster/zones/{zoneId}
...
-`numberOfBrokers` is the number of brokers deployed in the zone, from which the broker IDs
+Use `numberOfBrokers` when the zone's broker IDs are contiguous. The value is the number of brokers deployed in the zone, from which the broker IDs
 `<zoneId>_0` through `<zoneId>_<numberOfBrokers - 1>` are derived. These are the IDs the
 brokers of a zone-aware cluster assign themselves, so a zone whose brokers are numbered
 from zero without gaps needs nothing else. `numberOfBrokers` must be at least `1`; a lower
 value is rejected with HTTP `400`.
 
-Use `brokers` when the IDs are not contiguous, which is what a zone coming back with only
-some of its brokers looks like: only the explicit list can express that.
+Use `brokers` when the broker IDs are not contiguous, such as when a zone is re-added with only some of its brokers. List each broker ID explicitly in this array. Setting both `numberOfBrokers` and `brokers`, or omitting both, is rejected with HTTP `400`.
```
panagiotisgts: 💭  That might require some elevated visibility focusing on the _when a zone is re-added with only some of its brokers_ for clarity

docs/self-managed/components/orchestration-cluster/zeebe/operations/management-api.md
```diff
@@ -479,7 +449,38 @@ curl -X 'POST' \
...
+
+###### Dry run
+
+You can do a dry run without executing the change by setting the `dryRun` request parameter to `true`. By default, `dryRun` is set to `false`.
+
+##### Response
+
+The response is a JSON object with the same shape as the [cluster configuration response](#response). The `changeId` identifies the asynchronous operation. Poll the [Monitoring API](#monitoring-api) and wait until the operation is `COMPLETED` before shutting down brokers or taking further action.
+
+After the operation completes, verify that the zone appears in `partitioning` and that its brokers host their assigned partitions in the `brokers` array. You can also query the [Orchestration Cluster REST API specification for `GET /v2/topology`](/apis-tools/orchestration-cluster-api-rest/specifications/get-topology.api.mdx) to verify the broker and partition assignments.
```
panagiotisgts:
```suggestion
After the operation completes, verify that the zone is present under `partitioning` and that its brokers host their assigned partitions in the `brokers` array. You can also query the [Orchestration Cluster REST API specification for `GET /v2/topology`](/apis-tools/orchestration-cluster-api-rest/specifications/get-topology.api.mdx) to verify the broker and partition assignments.
```

docs/self-managed/components/orchestration-cluster/zeebe/operations/management-api.md
```diff
@@ -489,17 +490,19 @@ You can do a dry run without executing the change by setting the `dryRun` reques
 
 ##### Response
 
-The response is a JSON object with the same shape as the [partition distribution response](#response).
+The response is a JSON object with the same shape as the [cluster configuration response](#response). The `changeId` identifies the asynchronous operation. Poll the [Monitoring API](#monitoring-api) and wait until the operation is `COMPLETED` before shutting down brokers or taking further action.
+
+After the operation completes, verify that the removed zone no longer appears in `partitioning` and that its brokers no longer host partitions. Only then shut down the removed zone's brokers or scale down its StatefulSet.
```
panagiotisgts:
```suggestion
After the operation completes, verify that the removed zone is no longer present under `partitioning` and that its brokers no longer host partitions. Only then shut down the removed zone's brokers or scale down its StatefulSet.
```

docs/self-managed/components/orchestration-cluster/zeebe/operations/management-api.md
```diff
@@ -479,7 +449,38 @@ curl -X 'POST' \
...
+
+##### Response
+
+The response is a JSON object with the same shape as the [cluster configuration response](#response). The `changeId` identifies the asynchronous operation. Poll the [Monitoring API](#monitoring-api) and wait until the operation is `COMPLETED` before shutting down brokers or taking further action.
+
+After the operation completes, verify that the zone appears in `partitioning` and that its brokers host their assigned partitions in the `brokers` array. You can also query the [Orchestration Cluster REST API specification for `GET /v2/topology`](/apis-tools/orchestration-cluster-api-rest/specifications/get-topology.api.mdx) to verify the broker and partition assignments.
+
+#### Remove a zone
+
+By default, this operation gracefully drains the zone's partitions to the remaining zones before removing its brokers from cluster membership. Set `force=true` only if the zone is down or its brokers are unreachable. Forced removal evicts the zone's brokers without handing off leadership gracefully.
```
panagiotisgts: > Forced removal evicts the zone's brokers without handing off leadership gracefully.

Maybe that deserves a warning label

panagiotisgts (dismissed):

docs/self-managed/components/orchestration-cluster/zeebe/operations/management-api.md
entangled90: I actually removed it in the end :laughing: as it was just repeating what was done already above. I'll check again.

leiicamundi (commented):
Reviewed `18a1568f7` against the merged upstream contract in `camunda/camunda@main` (`dist/src/main/resources/api/cluster/cluster-api.yaml`, `components.yaml`).

Correct in this PR: `PUT /cluster/partitioning`, `scheme` required in `PartitioningConfig`, `force` defaulting to `false`. All four operations carry `x-added-in-version: "8.10.0"` and `versioned_docs/version-8.9` has no zone-aware content, so the versioning bot's "duplicate into version-8.9" comment is a false positive.

Three blocking issues. One is inline at line 466. The other two sit outside the diff, so they go here.

### 1. This PR breaks a cross-page anchor

`move-raft-leadership.md:52` links `management-api.md#partition-distribution-api`. Renaming the heading to `### Partitioning API` moves that anchor to `#partitioning-api`, so the link dies on merge.

CI will not catch it: `docusaurus.config.js` sets `onBrokenLinks: "throw"` and `onBrokenMarkdownLinks: "throw"`, but has no `onBrokenAnchors` entry, so broken anchors only warn. The green `validate-links` run says nothing about this case.

### 2. The rename covers one file out of four

Still on the old endpoint, all under `docs/`, which is the 8.10 version the rename ships in:

| Location | Content |
|---|---|
| `move-raft-leadership.md:43` | `jq '.partitionDistribution.zones[] ...'` |
| `move-raft-leadership.md:46` | prose, `partitionDistribution.zones` twice |
| `move-raft-leadership.md:60` | `curl ... /actuator/cluster/partition-distribution?dryRun=true` |
| `move-raft-leadership.md:72` | `curl ... /actuator/cluster/partition-distribution` |
| `cluster-scaling.md:187` | `You need to use PUT /actuator/cluster/partition-distribution/ instead` |

A reader who copies those `curl` commands gets a 404, and has no way to tell a stale page from a second endpoint. Please sweep all four in this PR.

### Also outside the diff

`management-api.md:537`, the last line of the page, still says "same shape as the [partition distribution response](#response)" while the two sibling `Response` blocks now say "cluster configuration response". The duplication behind that drift is flagged inline at 454.

### On the older threads

The 2026-09-17 rebase dropped `add-remove-zone.md` and `migrate-to-zone-aware-brokers.md`. Every thread on those two files now renders as outdated, which looks like resolved but is not. Replies added on each.

docs/self-managed/components/orchestration-cluster/zeebe/operations/management-api.md
leiicamundi: Blocking: the `force` condition is incomplete, and the missing case is the ordinary one.

This says `force=true` is needed only when the zone is down or unreachable. The spec gives a second, independent trigger:

> The `force` flag needs to be set when the region to be removed **contains the coordinator** or it's not currently reachable from the surviving regions.
> (`cluster-api.yaml`, `DELETE /zones/{zoneId}`)

camunda/camunda#63247 says why:

> Routes forced removal to a broker outside the zone being removed, which is necessary when the zone is not reachable from the surviving zones. A non-forced removal is routed to the elected coordinator, which can gracefully hand off coordination before leaving.
> (`ClusterConfigurationManagementRequestSender#removeZone`)

So a planned decommission of a healthy zone that happens to hold the coordinator follows this paragraph, sends `force=false`, and the request cannot be served. That is the case the new default was introduced for, and the page currently rules it out.

Suggested replacement for the second sentence:

> Set `force=true` if the zone is unreachable from the surviving zones, or if it contains the cluster coordinator. A non-forced removal is handled by the elected coordinator, so it cannot remove the zone the coordinator is in.

One more gap in the same paragraph: removing the only remaining zone is rejected, and the page does not say so. Two paths return it, `RemoveZoneTransformer` with "Cannot remove zone '<id>' because it is the last remaining zone in the partition distribution config.", and `ClusterConfigurationManagementRequestSender#removeZone` with "Cannot remove zone '%s' because it is the last remaining zone." when `force=true` leaves no coordinator outside the zone.

docs/self-managed/components/orchestration-cluster/zeebe/operations/management-api.md
```diff
@@ -432,23 +406,19 @@ POST actuator/cluster/zones/{zoneId}
 {
   "numberOfReplicas": <integer>,
   "priority": <integer>,
-  "numberOfBrokers": <integer>,
-  "brokers": [<brokerId1>, <brokerId2>, ...]
+  "numberOfBrokers": <integer>
```
leiicamundi: The schema block contradicts the three paragraphs under it.

`brokers` was dropped here, but line 413 requires "exactly one of `numberOfBrokers` or `brokers`", line 421 explains when to use `brokers`, and the second example posts a `brokers` array. Anyone reading the schema alone cannot discover the field; anyone reading the prose finds it undeclared.

`AddZoneRequest` keeps both as siblings (required `numberOfReplicas` and `priority`, then exactly one of `brokers` or `numberOfBrokers`), so the block should show both:

```
POST actuator/cluster/zones/{zoneId}
{
  "numberOfReplicas": <integer>,
  "priority": <integer>,
  "numberOfBrokers": <integer>,
  "brokers": [<brokerId1>, <brokerId2>, ...]
}
```

docs/self-managed/components/orchestration-cluster/zeebe/operations/management-api.md
```diff
@@ -386,44 +390,14 @@ The response is a JSON object. See the [OpenAPI spec](https://github.com/camunda
...
-
-You can do a dry run without executing the change by setting the `dryRun` request parameter to `true`. By default, `dryRun` is set to `false`.
-
-##### Response
-
-The response is a JSON object with the same shape as the [partition distribution response](#response).
+#### Add or re-add a zone
 
-#### Add back a previously force-removed zone
+To add a zone, first deploy its brokers and connect them to the existing cluster. Configure the brokers to use the zone you want to add.
```
leiicamundi, line 397: This claims a capability the linked spec does not describe.

"To add a zone, first deploy its brokers and connect them to the existing cluster" describes adding a brand-new zone. The contract only covers the re-add path: the `POST /zones/{zoneId}` summary is "Add back a previously removed zone", and `AddZoneRequest` repeats it.

`AddZoneToClusterIT` exists in the engine, so the page is probably right and the spec text is stale. The page points readers at that spec as the source of truth, so the two cannot disagree. Please confirm which one is correct and fix the other, with a spec follow-up in `camunda/camunda` if the page is right.

If adding a new zone is supported, the two flows have different prerequisites (new zone: deploy and join the brokers first; re-add: restart the previously removed brokers) and this section alternates between them paragraph by paragraph. Two subsections, or a two-row prerequisites table, would make that readable.

docs/self-managed/components/orchestration-cluster/zeebe/operations/management-api.md
```diff
@@ -489,17 +490,19 @@ You can do a dry run without executing the change by setting the `dryRun` reques
...
+The response is a JSON object with the same shape as the [cluster configuration response](#response). The `changeId` identifies the asynchronous operation. Poll the [Monitoring API](#monitoring-api) and wait until the operation is `COMPLETED` before shutting down brokers or taking further action.
+
+After the operation completes, verify that the removed zone no longer appears in `partitioning` and that its brokers no longer host partitions. Only then shut down the removed zone's brokers or scale down its StatefulSet.
 
 #### Migrate a zone to a zone-aware topology
 
-Migrates one zone of a bare or partially zoned cluster to a zone-aware topology. The request contains only the zone name. Before migrating a zone, update the persisted partition distribution with [`PUT /cluster/partition-distribution`](#partition-distribution-api), using a zone-aware partition distribution.
+Migrates one zone of a bare or partially zoned cluster to a zone-aware topology. The request contains only the zone name. Before migrating a zone, update the persisted partition distribution with [`PUT /cluster/partitioning`](#partitioning-api), using a zone-aware partition distribution.
```
leiicamundi, line 504: ~~The ordering rule that silently misplaces nodes is missing from the page.~~

Correction, this finding was wrong: the rule is documented at line 316, under the Partitioning API, which is where the spec puts it too.

> When migrating a bare or partially zoned cluster to zone-aware, list zones in `config.zones` in the order they should receive the existing (bare) nodes: the first zone receives node `0`, the second node `1`, and so on, wrapping around by zone count.

What is left is a cross-reference nit, not a gap. This step says to update the persisted distribution with `PUT /cluster/partitioning` first, and the reader has to already know that the list order in that call is load-bearing for this one migration. A link from here to the Partitioning API section would close it. Not blocking.

docs/self-managed/components/orchestration-cluster/zeebe/operations/management-api.md
```diff
@@ -479,7 +449,38 @@ curl -X 'POST' \
       }'
 ```
 
-Note that the IDs in `brokers` must be zone-aware, meaning they contain the zone in the name.
+</details>
+
+###### Dry run
```
leiicamundi, line 455: The duplication in this section already produced a drift bug, in this same diff.

The page now carries the "Dry run" paragraph verbatim four times (369, 454, 487, 531) and the "same shape as the ... response / `changeId` / poll the Monitoring API until `COMPLETED`" paragraph verbatim twice (460, 493). Of the three "same shape as the ... response" lines, two were updated to "cluster configuration response" and line 537 was missed, so it still says "partition distribution response".

One `#### Common parameters` subsection under `## Cluster API` would state the `dryRun` contract and the async contract (`202`, `changeId`, poll `GET /cluster/changes/{changeId}` until `COMPLETED`, `409` while another change is running) once, with each operation linking to it. That drops about 30 lines, removes the drift, and means the next endpoint added here inherits the async contract instead of restating it.

Related: the `409 ConcurrentChangeError` that all of these operations can return is not documented on the page at all. That was the substance of the Copilot finding on the now-deleted `migrate-to-zone-aware-brokers.md:122`.

docs/self-managed/components/orchestration-cluster/zeebe/operations/management-api.md
```diff
@@ -386,44 +390,14 @@ The response is a JSON object. See the [OpenAPI spec](https://github.com/camunda
...
-  <summary>Example request</summary>
-
-```
-curl -X 'DELETE' \
-   'http://localhost:9600/actuator/cluster/zones/zone-b' \
-   -H 'accept: application/json'
-```
-
-</details>
+Use the Zones API to add, remove, or migrate zones in a [zone-aware](/self-managed/components/orchestration-cluster/zeebe/configuration/zone-aware-clusters.md) cluster. The operations update the persisted partition distribution and run asynchronously. Use the [Monitoring API](#monitoring-api) to follow each change until its status is `COMPLETED`.
```
leiicamundi, line 393: Structure: this change moves operator procedure into an endpoint reference by deleting the pages that held it.

What this section now instructs, none of which is what the endpoint accepts:

- "first deploy its brokers and connect them to the existing cluster" (397)
- "start the ... brokers before sending the request" (400)
- "Only then shut down the removed zone's brokers or scale down its StatefulSet" (495)

The branch head on 2026-09-16 (`24a95e985`) had `add-remove-zone.md` and `migrate-to-zone-aware-brokers.md` as separate pages with sidebar entries, and the rebase dropped both. One of them existed to close https://github.com/camunda/camunda-docs/pull/9879#discussion_r3970458552: "we need a dedicated how-to guide for migrating a bare/partially-zoned cluster to zone-aware (currently only documented at the API-reference level in `management-api.md`)". That TODO is open again, and the page it complained about is the page now absorbing the procedure.

The split that existed is the right one: schemas, parameters, and status codes here; ordering, prerequisites, and the checks between steps in a how-to that links back. It also fixes the nesting, since the deepest operation on this page is now six heading levels down (`###### Dry run`).

The API-alignment part of this PR (`partitioning`, `scheme`, `force` default) is correct and worth landing on its own. The page deletions are the part worth reconsidering.

docs/self-managed/components/orchestration-cluster/zeebe/operations/management-api.md
leiicamundi: It is not quite a repeat of the paragraph above. The paragraph above documents the default; a caution documents the consequence of overriding it, and that consequence is not stated anywhere on the page.

camunda/camunda#63247 is explicit about the risk:

> Previously, removing a zone would automatically remove it via force which can potentially be a dangerous operation.

and `RemoveZoneRequest` on `main` spells out what force skips:

> Without `force`, the request is handled by the elected coordinator and the replicas on the departing brokers are removed before those brokers leave. Set `force` when the zone is not reachable from the surviving zones. Forced removal evicts the brokers without removing replicas first, at the cost of the replicas they held.
> `@param force` whether to evict the zone's brokers without first moving their partitions

"without handing off leadership gracefully" undersells "at the cost of the replicas they held". Suggested:

```
:::caution
Forced removal evicts the zone's brokers without first moving their partitions. The replicas those brokers held are lost. Use `force=true` only when a non-forced removal cannot be served.
:::
```

docs/self-managed/components/orchestration-cluster/zeebe/operations/add-remove-zone.md
leiicamundi: Reopening this one. The 2026-09-17 rebase removed `add-remove-zone.md` and `migrate-to-zone-aware-brokers.md` along with their sidebar entries, so the dedicated guide this TODO asked for is gone again.

The procedure moved into `management-api.md` instead, which is the page this TODO was filed against ("currently only documented at the API-reference level in `management-api.md`"). Details in the review comment on `management-api.md:393`.

docs/self-managed/components/orchestration-cluster/zeebe/operations/migrate-to-zone-aware-brokers.md
```diff
@@ -0,0 +1,187 @@
...
+
+## 1. Start the replacement brokers in the local zone
+
+Deploy the zoned brokers next to the existing numbered brokers, using zone-aware broker IDs of the form `{zone}_{index}` (see the [broker ID naming scheme](./cluster-scaling.md#broker-id-naming-scheme)). Point them at the existing cluster's contact points, and set `cluster.size` to the total broker count across all zones.
+
+At this point the zoned brokers join the cluster's gossip-based membership only. They are not part of the persisted partition distribution yet, so they hold no partitions and join no Raft groups. The numbered brokers keep serving traffic.
+
+### Helm
+
+The Camunda 8.10 chart (`camunda-platform` 15.x) renders both generations from a single release, so Helm owns every resource during the migration.
```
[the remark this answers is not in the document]
leiicamundi: Obsolete as filed: `migrate-to-zone-aware-brokers.md` was removed by the 2026-09-17 rebase, and no Helm content remains in this PR.

The finding still holds if the page comes back. The 8.10 chart does not define `keepUnzonedBrokers` and does not render a second broker StatefulSet, so the coexistence procedure needs camunda/camunda-platform-helm#7049 to land first.

docs/self-managed/components/orchestration-cluster/zeebe/operations/add-remove-zone.md
```diff
@@ -0,0 +1,176 @@
...
+:::note
+`GET /actuator/cluster` also reports a pending change, but only for the default [Physical Tenant](/self-managed/concepts/physical-tenants/index.md). On a cluster with multiple Physical Tenants, always poll `GET /actuator/cluster/changes/{changeId}` instead.
+:::
+
+### 4. Verify
+
+Confirm the new zone's brokers now host partitions:
+
+```bash
+curl -s 'http://{zeebe-gateway}:9600/actuator/cluster' | jq '.partitionDistribution.zones[] | {name, priority}'
```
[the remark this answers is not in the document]
leiicamundi: Addressed, in a different place than expected. `add-remove-zone.md` was removed by the 2026-09-17 rebase, and `management-api.md:462` now tells the reader to check that the zone's brokers host their assigned partitions in the `brokers` array, with `GET /v2/topology` as an alternative.

docs/self-managed/components/orchestration-cluster/zeebe/operations/migrate-to-zone-aware-brokers.md
```diff
@@ -0,0 +1,187 @@
...
+    "type": "ZONE_AWARE",
+    "zones": [
+      { "name": "zone-a", "numberOfReplicas": 3, "priority": 100 },
+      { "name": "zone-b", "numberOfReplicas": 3, "priority": 90 }
+    ]
+  }
+}
+```
+
+Then migrate the local zone with [`PUT actuator/cluster/zones`](./management-api.md#migrate-a-zone-to-a-zone-aware-topology):
```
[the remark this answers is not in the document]
leiicamundi: Partly addressed. `migrate-to-zone-aware-brokers.md` was removed by the 2026-09-17 rebase, and `management-api.md` now tells the reader to poll the Monitoring API until the change is `COMPLETED` before taking further action.

The rejection itself is still undocumented: none of these operations mention the `409 ConcurrentChangeError` returned while another change is running. Raised in the review comment on `management-api.md:454`.

docs/self-managed/components/orchestration-cluster/zeebe/operations/management-api.md
entangled90: Regarding ` The `force` flag needs to be set when the region to be removed contains the coordinator`, that needs to be fixed in the spec.

docs/self-managed/components/orchestration-cluster/zeebe/operations/management-api.md
entangled90: Yeah I'll reinstate it

docs/self-managed/components/orchestration-cluster/zeebe/operations/management-api.md
entangled90: From the engine perspective there's no different if it's add or re-add

docs/self-managed/components/orchestration-cluster/zeebe/operations/add-remove-zone.md
entangled90: I don't think it added that much value the separate page. We will likely need one for each reference architecture so it contains all the exact steps (including helm/terraform)

entangled90:
Fixes for `move-raft-leadership` in this PR #9960

docs/self-managed/components/orchestration-cluster/zeebe/operations/management-api.md
leiicamundi: Agreed, and the evidence I quoted in this thread supports you rather than me: `ClusterConfigurationManagementRequestSender#removeZone` says a non-forced removal "is routed to the elected coordinator, **which can gracefully hand off coordination before leaving**". The coordinator therefore does not require `force`, and `cluster-api.yaml` is the stale side. Withdrawn as a finding on this page.

entangled90: camunda/camunda#63387

docs/self-managed/components/orchestration-cluster/zeebe/operations/management-api.md
leiicamundi: The stale line this thread points at is fixed in e2684f48c: `management-api.md:538` now says "cluster configuration response" like its two siblings at 460 and 493.

The duplication that produced the drift is still there, so I'm leaving this thread open rather than resolving it. Not blocking, and your call whether a shared `#### Common parameters` subsection is worth it here or belongs in a follow-up.

leiicamundi (dismissed):
Just a last suggestion for the open thread:

```
:::warning
Forced removal evicts the zone's brokers without first moving their partitions. The replicas those brokers held are lost. Use `force=true` only when a non-forced removal cannot be served.
:::
```

Anyway, leaving this choice to you, the PR LGTM

entangled90:
Reworded with
```
:::warning
Forced removal of nodes that are running/reachable may cause data loss in extreme circumstances
:::
```

marcel-dias (dismissed):

christinaausley (approved):
Looks good! Reviewed and fixed:

**`management-api.md`** the "Add/re-add," "Remove," and "Migrate a zone" sections all link to `#response` for their response shape, but with 6 "Response" headings on the page, that bare anchor resolves to the first one (`GET actuator/cluster`), which has no `changeId`... yet each section's next sentence describes `changeId`, which only matches the Partitioning API's response. Added a `{#partitioning-response}` anchor and repointed all three, plus filled in the same missing explanation in Migrate for consistency.

**`move-raft-leadership.md`** not in your file list, but uses the same renamed endpoint. Updated `.partitionDistribution.zones` to `.partitioning.zones`, fixed a link to the now-dead `#partition-distribution-api` anchor, and updated both `curl` examples to `/actuator/cluster/partitioning`.

entangled90:
Since you mentioned move-raft-leadership I opened another PR (#9960) for that page, maybe it's worth combining them/closing the other one?

entangled90:
Actually nevermind, I can rebase that one on this one.

PR #9883, opened 2026-09-10
docs(feature): updates for centralized secret resolution

giorgionaps:
## Description

A PR that covers:

- [#60327](https://github.com/camunda/camunda/issues/60327)
- [#60328](https://github.com/camunda/camunda/issues/60328)
- [#60333](https://github.com/camunda/camunda/issues/60333)
- [#60336](https://github.com/camunda/camunda/issues/60336) (sidebar update)
- [#60334](https://github.com/camunda/camunda/issues/60334)
- [#60335](https://github.com/camunda/camunda/issues/60335)

## When should this change go live?

- [ ] This is a bug fix, security concern, or something that needs **urgent release support**. (add `bug` or `support` label)
- [ ] This is already available but undocumented and should be released within a week. (add `available & undocumented` label)
- [ ] This is on a **specific schedule** and the assignee will coordinate a release with the Documentation team. (create draft PR and/or add `hold` label)
- [ ] This is part of a scheduled **alpha or minor**. (add alpha or minor label)
- [ ] There is **no urgency** with this change (add `low prio` label)

## PR Checklist

- [ ] The commit history of this PR is cleaned up, using [`{type}(scope): {description}` commit message(s)](https://github.com/camunda/camunda-docs/blob/main/CONTRIBUTING.MD#commit-message-header-formatting)

- [ ] My changes are for **an upcoming minor release** and are in the `/docs` directory.
- [ ] My changes are for an **already released minor** and are in a `/versioned_docs` directory.

- [ ] I added a DRI, team, or delegate as a reviewer for technical accuracy and grammar/style:
  - [ ] [Engineering team review](https://github.com/camunda/camunda-docs/blob/main/howtos/documentation-guidelines.md#review-process)
  - [ ] [Technical writer review](https://github.com/camunda/camunda-docs/blob/main/howtos/documentation-guidelines.md#review-process) via `@camunda/tech-writers` unless working with an embedded writer.

PR #9887, opened 2026-09-10
docs(concepts): correct secret resolution documentation errors and gaps

### LINKED ISSUE #62709: [Secret Resolution] Docs: correct factual errors and fill gaps found in the merged documentation
### 

### Description

A DRI review of every **merged** documentation PR under #56591 found a set of statements that
contradict the implemented behavior, plus several crucial parts of the flow that are not documented
at all. This issue tracks the subset we decided to act on.

All code references below point at `camunda/camunda@f391a62` (`origin/main` at the time of review).
All documentation changes land in `camunda/camunda-docs`.

#### Corrections to the concept page (from camunda/camunda-docs#9703)

`docs/components/concepts/secret-resolution-and-job-activation.md`

**1. `WAITING_FOR_SECRET_RESOLUTION` is presented as an observable job state.**
The page says "While it waits, the job is in the state `WAITING_FOR_SECRET_RESOLUTION`". That is an
engine-internal enum with zero occurrences outside `zeebe/engine`: not in the protocol, not
exported, not in any v2 API response, not in Operate. Readers will try to filter on it and find
nothing.
Evidence: [JobState.java#L81](https://github.com/camunda/camunda/blob/f391a6211805ba9c8b684c6d9b82719ea0937536/zeebe/engine/src/main/java/io/camunda/zeebe/engine/state/immutable/JobState.java#L81)
Fix: describe the parking internally, then name what is actually observable (job not returned by
activation, no `JOB ACTIVATED` record, a `RESOLUTION_REQUESTED` record, and the
`zeebe_job_events_total{action="skipped uncached secret"}` meter).

**2. The page is written for multiple secret stores, but only one is supported.**
Phrases such as "groups them by store", "requests each store's group of references in one call",
"Configure each store's cache lifetime and size" and "tracks retry state per store" describe a
multi-store design. In reality each physical tenant supports exactly one store and its id must be
`default`; any other id fails startup. A reader who follows this page and configures
`camunda.secrets.stores.aws.vault` gets a cluster that does not start.
Evidence: [SecretStoreRegistry.java#L36](https://github.com/camunda/camunda/blob/f391a6211805ba9c8b684c6d9b82719ea0937536/secret-store/secret-store-api/src/main/java/io/camunda/secretstore/SecretStoreRegistry.java#L36),
[SecretStoreConfiguration.java#L298](https://github.com/camunda/camunda/blob/f391a6211805ba9c8b684c6d9b82719ea0937536/dist/src/main/java/io/camunda/application/commons/secrets/SecretStoreConfiguration.java#L298)
Fix: rewrite in the singular, and state the one-store-per-physical-tenant rule with the required
`default` id explicitly.

**3. "even if you don't configure a secret store yourself" is misleading, and the scope of the
effect needs clarifying.**
With no store configured, a `NoopSecretStore` is registered under `default` and fails every
reference permanently as `NOT_FOUND` with the message "No secret store configured", so the job gets
an incident rather than being delayed. A cluster whose models contain no secret references is not
affected at all. The current sentence alarms the wrong readers and hides the real failure mode.
Evidence: [NoopSecretStore.java#L29](https://github.com/camunda/camunda/blob/f391a6211805ba9c8b684c6d9b82719ea0937536/secret-store/secret-store-api/src/main/java/io/camunda/secretstore/NoopSecretStore.java#L29),
[SecretStoreConfiguration.java#L179](https://github.com/camunda/camunda/blob/f391a6211805ba9c8b684c6d9b82719ea0937536/dist/src/main/java/io/camunda/application/commons/secrets/SecretStoreConfiguration.java#L179)
Fix: state precisely who is affected. A cluster with no secret references in any model is unaffected;
a model that uses a reference on a cluster without a configured store gets a `SECRET_RESOLUTION_ERROR`
incident, not a delay.

**4. `camunda.processing.engine.secrets.wake-delay` is documented nowhere.**
Default `50ms`, and it is what governs activation latency in practice. It is missing from the concept
page tuning table, from `_processing.md`, and from the env-var table
(`CAMUNDA_PROCESSING_ENGINE_SECRETS_WAKEDELAY`). Every sibling property in that class is documented.
Because of this omission, the tuning advice "a shorter interval reduces that delay" points operators
at the wrong knob: `interval` is a ceiling that is rarely reached.
Evidence: [EngineSecrets.java#L129](https://github.com/camunda/camunda/blob/f391a6211805ba9c8b684c6d9b82719ea0937536/configuration/src/main/java/io/camunda/configuration/EngineSecrets.java#L129)
Fix: document the property in both tables, and correct the tuning guidance to describe the actual
cadence (fast `wake-delay` cadence while work is arriving, geometric idle backoff up to `interval`).

**5. Security: nothing warns that a worker can write the resolved value back.**
The page warns only about a worker logging its input variables. Nothing prevents a worker from
persisting the value through `completeJob` variables, an output mapping, a connector result variable,
or an error message. Any of those writes the plaintext into process variables, runtime state, and
every exported record, which defeats the guarantee the "Secret values location" table appears to
make.
Fix: add an explicit warning next to that table and qualify the table itself.

**6. Multi-tenancy is not mentioned at all.**
The store registry is per physical tenant, and stores and cache are overridable through
`camunda.physical-tenants.<id>.secrets.*`. A multi-tenant operator currently has no model of which
store serves which tenant.
Fix: add a short section, and cross-link the physical tenants configuration reference.

#### Correction to the troubleshooting runbook (from camunda/camunda-docs#9717)

`docs/components/concepts/secret-resolution-incidents.md`

**7. Drop row 1 of the "Diagnose the cause" table and add the row that matches reality.**
Row 1 maps the log line `Secret store '<id>' is not configured` to the cause "No secret store
configured". That log line does not fire in that situation: with nothing configured the noop store is
registered under `default`, the registry lookup succeeds, and the operator actually sees
`Secret '<ref>' in secret store 'default' failed permanently: NOT_FOUND - No secret store configured`.
That signature currently matches row 2, whose fix reads "Create the secret or correct the reference
name in the process", sending the operator to create a secret in a store that does not exist. Row 1's
real trigger cannot occur in a normal deployment.
Fix: remove row 1, and add a row keyed on the `NOT_FOUND - No secret store configured` message whose
fix is "configure a secret store".

#### Corrections to the 8.10 release notes (from camunda/camunda-docs#9751)

`docs/reference/announcements-release-notes/8100/8100-release-notes.md`

**8. The cache TTL default is wrong by a factor of 60.**
The entry says "Cache entries expire after the configured TTL, which is 20 seconds by default". The
default is 20 **minutes**, and a TTL below `1m` or one that is not a whole number of minutes is
rejected at startup, so 20 seconds is not even configurable. Both `_secrets.md` and the Camunda 8 Run
page state `20m`, so the release notes contradict the rest of the documentation.
Evidence: [Secrets.java#L212](https://github.com/camunda/camunda/blob/f391a6211805ba9c8b684c6d9b82719ea0937536/configuration/src/main/java/io/camunda/configuration/Secrets.java#L212)

**9. Output mappings are listed as a supported location. They are not.**
"Reference secrets as `camunda.secrets.NAME` in input mappings, expressions, and output mappings" is
wrong on the last item: secret detection runs only for input mappings, and the deploy-time validators
register on `zeebe:input` and `zeebe:property` only. A reference in an output mapping is never
detected, never rejected, and never resolved: it is written to process variables as the literal
string and exported.
Evidence: [VariableMappingTransformer.java#L84](https://github.com/camunda/camunda/blob/f391a6211805ba9c8b684c6d9b82719ea0937536/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/model/transformer/VariableMappingTransformer.java#L84)
(reached only from `transformInputMappings`, never from `transformOutputMappings` at
[#L170](https://github.com/camunda/camunda/blob/f391a6211805ba9c8b684c6d9b82719ea0937536/zeebe/engine/src/main/java/io/camunda/zeebe/engine/processing/deployment/model/transformer/VariableMappingTransformer.java#L170))
Fix: remove output mappings from the list.

**10. The file store is framed as development and testing only.**
"A file-based provider is available for development and testing" understates it. It is the production
mechanism on SaaS, where the operator projects Kubernetes Secrets into the store directory, and it is
a first-class Kubernetes option in the Helm chart through
`orchestration.secretStore.file.default.secret.existingSecret`. Readers on Kubernetes will skip the
option that actually fits them.
Fix: describe the file store as a supported production option on Kubernetes, and keep the dev and
test framing for the local Camunda 8 Run flow only.

#### Missing configuration property (from camunda/camunda-docs#9770)

`docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/_secrets.md`

**11. `camunda.secrets.max-concurrency` is documented nowhere.**
Canonical property, default `max(8, 2 x availableProcessors)`, minimum 1, validated at startup. It
bounds how many backend calls a node has in flight per physical tenant for stores whose cost scales
with call count, which makes it the knob for staying within an AWS or GCP rate quota. That is one of
the most likely reasons an operator touches this configuration at all.
Evidence: [Secrets.java#L109](https://github.com/camunda/camunda/blob/f391a6211805ba9c8b684c6d9b82719ea0937536/configuration/src/main/java/io/camunda/configuration/Secrets.java#L109)
Fix: document it in `_secrets.md` and in the generated property reference, with the AWS and GCP quota
use case spelled out.

#### Correction to the cluster variables page (from camunda/camunda-docs#9792)

`docs/components/admin/cluster-variables.md`

**12. An unconfigured store is described as failing silently. It raises an incident.**
The page states that without a configured store "references are never resolved, and any job that
depends on them stays unactivated with no error pointing at the cause". In fact every reference fails
permanently and the job receives a `SECRET_RESOLUTION_ERROR` incident naming the reference, which
also contradicts our own incidents page. Same root cause as item 7.
Fix: replace with the incident behavior, and link the troubleshooting page.

#### Missing security statement on the authorizations page (from camunda/camunda-docs#9383)

`docs/components/concepts/access-control/authorizations.md`

**13. The page does not say that broker-side resolution bypasses the `SECRET` resource entirely.**
An admin reading this page reasonably concludes that withholding `SECRET` withholds access to secret
values. It does not: `SECRET:READ` and `SECRET:REVEAL` gate only `POST /v2/secrets/list` and
`POST /v2/secrets/resolve`. Any process model may reference any secret, and any worker polling that
job type receives the plaintext with no `SECRET` grant anywhere. The troubleshooting page states this
correctly, but that is not the page an admin reads when designing access control.
Evidence: [SecretServices.java#L178](https://github.com/camunda/camunda/blob/f391a6211805ba9c8b684c6d9b82719ea0937536/service/src/main/java/io/camunda/service/SecretServices.java#L178)
Fix: add an explicit note to the `SECRET` entry, and define `READ` (list reference names) versus
`REVEAL` (read values) while we are there.

### Motivation

The documentation for this feature is published and is what customers will follow for the 8.10 alpha.
Six of these thirteen points are statements that contradict the implemented behavior, and two of them
are in the release notes, which is the most-read page of the set. The concrete risks:

- A user follows the release notes, puts a secret reference in an output mapping, and silently
  persists a secret-shaped literal into process variables with no error anywhere.
- An operator sizes a cache around a 20 second TTL that cannot be configured, since anything below
  one minute is rejected at startup.
- An operator hitting the most common first failure, no store configured, is sent by our own runbook
  to create a secret in a store that does not exist.
- An admin assumes the `SECRET` authorization resource controls who can obtain secret values, when
  broker-side resolution does not consult it at all. That is a security-model misunderstanding on a
  security page.
- A worker author has no warning that echoing job variables back through job completion writes the
  plaintext into state and exports, undoing the central promise of the feature.

The remaining points are gaps in the flow that leave operators without the property, the tenancy
model, or the failure semantics they need to run this.

### Acceptance criteria

- [ ] Concept page: replace `WAITING_FOR_SECRET_RESOLUTION` with observable signals
- [ ] Concept page: rewrite multi-store wording for a single `default` store per physical tenant
- [ ] Concept page: clarify precisely who is affected when no store is configured, and that the
      outcome is an incident rather than a delay
- [ ] Concept page and property reference: document `camunda.processing.engine.secrets.wake-delay`,
      and correct the latency tuning guidance
- [ ] Concept page: add a warning that a worker writing the value back persists the plaintext
- [ ] Concept page: document the per-physical-tenant store and cache model
- [ ] Runbook: drop row 1 of the diagnosis table and add the `NOT_FOUND - No secret store configured`
      row
- [ ] Release notes: correct the cache TTL default to `20m`
- [ ] Release notes: remove output mappings from the supported locations
- [ ] Release notes: reframe the file store as a supported production option on Kubernetes
- [ ] Configuration reference: document `camunda.secrets.max-concurrency`
- [ ] Cluster variables page: replace the silent-failure claim with the incident behavior
- [ ] Authorizations page: state that broker-side resolution bypasses `SECRET`, and define `READ`
      versus `REVEAL`

### Links

- Parent: https://github.com/camunda/camunda/issues/56591
- Reviewed merged PRs: camunda/camunda-docs#9703, camunda/camunda-docs#9717,
  camunda/camunda-docs#9751, camunda/camunda-docs#9770, camunda/camunda-docs#9792,
  camunda/camunda-docs#9383
- Related open documentation issues: #60326 (information architecture and terminology), #60336
  (sidebar, alpha flagging, release and upgrade notes), #60965 (secret store availability in SaaS)
- Open documentation PRs to check against before implementing: camunda/camunda-docs#9866 (terminology),
  camunda/camunda-docs#9830 (input mappings and connector coexistence)

### Breakdown

Findings were verified individually against the implementation, so they can be split by target page
and worked in parallel. Suggested grouping, ordered by risk:

1. Release notes corrections (items 8, 9, 10). Smallest change, highest exposure.
2. Concept page corrections and additions (items 1 to 6).
3. Runbook and cluster variables corrections (items 7 and 12). Same root cause, fix together.
4. Authorizations page security note (item 13).
5. Configuration reference property (item 11).

Further findings from the same review were deliberately left out of this issue, including the absent
Helm `orchestration.secretStore` documentation, the missing SaaS story, the sidebar orphaning of both
new pages, and the risk that the hand-edited expression endpoint prose is overwritten by the next REST
spec sync. Those can be filed separately if we decide to act on them.

berkaycanbc:
## Description

Fixes camunda/camunda#62709 — a set of factual errors and documentation gaps in the merged
Centralized Secret Resolution docs, found during a DRI review of every merged PR under
camunda/camunda#56591 against the actual implementation.

Six commits, one per affected page (plus two small style follow-ups):

- **Release notes** (`8100-release-notes.md`): the cache TTL default was stated as 20 seconds; it's
  20 minutes (anything below one minute is rejected at startup). Output mappings were listed as a
  supported location for `camunda.secrets.NAME`; only input mappings and expressions actually detect
  references. The file-based store was framed as development/testing only, though it's the mechanism
  backing SaaS and a first-class Helm option.
- **Secret resolution concept page** (`secret-resolution-and-job-activation.md` + the
  `_processing.md` property partial): replaced a reference to an engine-internal, unobservable job
  state with the signals that are actually observable; stated the one-store-per-physical-tenant rule
  explicitly (the page read as if multiple stores per tenant were supported); corrected what happens
  with no secret store configured (a permanent incident, not a silent delay); documented
  `camunda.processing.engine.secrets.wake-delay` (previously undocumented, and the property that
  actually governs steady-state activation latency, not `interval`) and corrected the tuning table
  accordingly; added a warning that a worker writing the resolved value back (job completion,
  output mapping, connector result, error message) persists the plaintext into process variables and
  every exported record; and added a short section on per-physical-tenant scoping of the store, cache,
  and scheduler retry state. Also fixed a pre-existing broken anchor link to the incidents page.
- **Secret resolution runbook** (`secret-resolution-incidents.md`): the diagnosis table's first row
  matched a log line that doesn't fire when no store is configured, sending operators to "create the
  secret" in a store that doesn't exist. Replaced it with the actual log signature.
- **Configuration reference** (`_secrets.md`): documented `camunda.secrets.max-concurrency`, which was
  missing entirely despite being the relevant knob for staying inside an AWS/GCP API rate quota.
- **Cluster variables page** (`cluster-variables.md`): corrected the same no-store-configured claim as
  above (permanent incident, not silent non-activation).
- **Authorizations page** (`authorizations.md`): added a note that `SECRET:READ`/`SECRET:REVEAL` gate
  only the `/v2/secrets` API endpoints, not broker-side resolution — any process model can reference
  any configured secret regardless of `SECRET` grants — and defined what `READ` and `REVEAL` actually
  control.

The two `style(docs)` commits remove em dashes from my own added prose (per
`.github/instructions/content.instructions.md`) and re-run Prettier on the four files whose tables I
resized (the originals were Prettier-clean; verified with `git diff -w` that only whitespace changed).

**Validation:** ran `npx prettier --check` on all seven changed files (clean). Did not run the full
`npm run build` — this environment has no `node_modules` and a from-scratch install would exceed the
"few minutes" budget `AGENTS.md` sets for doc-only changes. Instead manually verified every new/changed
link target: confirmed each target file exists, and for anchor links confirmed the destination heading
text matches Docusaurus's default slug algorithm.

## PR Checklist


christinaausley (dismissed):
Looks good to me!

giorgionaps:
@berkaycanbc I think you might want to merge changes from https://github.com/camunda/camunda-docs/pull/9883 into this branch as I feel like they might overlap a bit just from the first glance

berkaycanbc:
Checked this by fetching #9883's branch and dry-running `git merge origin/bcan/62709-secret-docs-fixes` on top of it: it merges clean, no conflicts, on all four overlapping files (`secret-resolution-and-job-activation.md`, `secret-resolution-incidents.md`, `authorizations.md`, `_secrets.md`). Every specific sentence/row I touch is different from every one #9883 touches, so merge order doesn't matter mechanically.

There was one real duplication though: `authorizations.md` would have ended up with two sections both explaining `READ` vs `REVEAL` on `SECRET`, worded differently, in two places. Trimmed mine down to just the two things #9883's section doesn't cover. That broker-side resolution never checks `SECRET` at all regardless of whether authorization is enabled (broader than the auth-disabled-mode note in #9883), and the exact-match-only resource key gotcha (a `camunda.secrets.*` prefix grant matches nothing). Left the READ/REVEAL definition and the role-default behavior to your section.

Also picked up your `WAITING_FOR_SECRET_RESOLUTION` fix on the runbook page (pushed directly to this branch) via rebase. Good catch, I'd only fixed that reference on the concept page and missed it on the incidents page.

docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/_secrets.md
```diff
@@ -134,3 +134,24 @@ One cache is created per configured store.
...
+| :-------------------------------- | :------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ | :--------------------------------- |
+| `camunda.secrets.max-concurrency` | Maximum number of sequential backend calls in flight at once for such a store, shared by every such store this physical tenant configures. Lower it to stay within a provider's rate limit (for example, an AWS Secrets Manager or GCP Secret Manager API quota); raising it above what a single request can use has no further effect. Must be at least `1`. | `max(8, 2 × available processors)` |
```
ajanoni-camunda: 🔧 **Wording + slight overpromise.** Two things here:

1. "Maximum number of **sequential** backend calls **in flight at once**" is self-contradictory. The property is `max-concurrency` and bounds *concurrent* calls, so this should read "concurrent backend calls in flight".

2. "Lower it to **stay within** a provider's rate limit" slightly overpromises. The property's own code doc notes the bound "only approximates a provider quota measured in calls per second rather than guaranteeing it", and that a request small enough to resolve in a single call takes **no** permit (so actual concurrency can exceed this value by one call per concurrent caller). Suggest softening to "help stay within / approximate a provider's rate limit".

Same text appears in the env-var row (line 154).

docs/components/concepts/secret-resolution-and-job-activation.md
```diff
@@ -43,15 +47,15 @@ Two kinds of failure are treated differently:
 
 The broker tracks retry state per store rather than per secret and holds it in memory only. The retry state resets when the broker restarts or the partition changes leader. During backoff, the scheduler skips the store, so its references do not consume batch capacity that a healthy store can use.
```
ajanoni-camunda: 🙃 Minor (point 2 residual): the issue explicitly flagged "tracks retry state **per store**" as multi-store phrasing to remove. It reads fine now that one-store-per-tenant is established (it contrasts per-store vs per-secret granularity), but for full consistency with the singular rewrite consider "tracks retry state for its store rather than per secret".

docs/components/concepts/secret-resolution-and-job-activation.md
```diff
@@ -84,11 +88,15 @@ A job worker does not need to handle secret resolution. While a job waits for a
 
 A worker receives the job with the placeholders replaced by the resolved values. If your worker logs its input variables, the resolved secret values appear in plaintext.
 
+:::warning
+The guarantees below describe what Camunda stores, not what a worker does with the value it received. A worker can still write the resolved value back into process variables, for example through a completed job's variables, an output mapping, a connector result, or an error message. Doing so writes it to process variables and exposes it in runtime state and every exported record as plaintext, the same as any other variable value. Write resolved secret values into a worker's own request to the credential's consumer, not back into process variables.
+:::
+
 You don't need to make client-side changes. Existing workers, clients, and job worker libraries continue to work with a cluster that resolves secrets.
```
ajanoni-camunda, line 95: ❓ **Implemented but undocumented behavior, worth adding?** This section tells worker authors everything "continues to work", but per ADR-0007 (`zeebe/docs/adr/0007-810-job-waiting-for-secret-resolution-state.md`, Accepted 8.10) and the engine code, worker lifecycle commands on a job that is *parked* for secret resolution are **rejected with `INVALID_STATE`** (`CompleteJob`, `FailJob`, `ThrowError`; `UpdateJob` timeout is also rejected).

The user-visible race the ADR calls out: if a job's secret **expires from the cache** (after `cache.ttl`, default 20m) between activations, a worker completing the job **late** now gets `INVALID_STATE` and "the work of that activation is lost", the job is re-parked, re-resolved, and handed out again. Neither this page nor the runbook mentions `INVALID_STATE` anywhere.

Suggest a short note here (or in the runbook) so worker authors know a late completion of a secret-referencing job can be rejected.

docs/components/concepts/access-control/authorizations.md
```diff
@@ -107,6 +107,12 @@ The following table lists all resources that support authorization in the Orches
...
 
+### Secret authorizations don't cover broker-side resolution
+
+`SECRET:READ` and `SECRET:REVEAL` gate only the `/v2/secrets` API endpoints (`POST /v2/secrets/list` and `POST /v2/secrets/resolve`, respectively). They don't gate the broker resolving `camunda.secrets.<name>` references for job activation. Any process model may reference any configured secret, and any worker that receives a job with a resolved reference sees the value in plaintext, with no `SECRET` grant involved anywhere in that path. See [secret resolution and job activation](../secret-resolution-and-job-activation.md).
```
ajanoni-camunda: **Added claim beyond the issue — please confirm against the implementation.** The issue's point 13 only asked for the "broker resolution bypasses `SECRET`" note (which is correct). This sentence adds detail the issue did not state: that the resource key is the full `camunda.secrets.MY_KEY` reference and that "only an exact match or `*` is supported, so `camunda.secrets.*` matches nothing". Plausible, but it's a specific authorization-matching claim not covered by the issue's cited evidence — worth a quick DRI/code check before merge so we don't ship a subtly wrong matcher rule.

ajanoni-camunda (dismissed):
✅ Approved with some minor comments

docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/_secrets.md
berkaycanbc: Good catch, thanks. Reworded to "concurrent backend calls in flight" and softened to "help stay within" per `Secrets.java`'s own doc: a request that resolves in a single call takes no permit and isn't counted, so the bound approximates rather than guarantees the provider quota. Fixed in f47ba8040.

docs/components/concepts/secret-resolution-and-job-activation.md
berkaycanbc: Agreed, that was leftover multi-store phrasing. Changed to "tracks retry state for its store rather than per secret". Fixed in f47ba8040.

docs/components/concepts/secret-resolution-and-job-activation.md
berkaycanbc: Verified against the ADR (zeebe/docs/adr/0007-810-job-waiting-for-secret-resolution-state.md, Accepted 8.10) and `JobCommandPreconditionValidator`: CompleteJob/FailJob/ThrowError all pass `List.of(ACTIVATABLE, ACTIVATED)` as valid states, so a parked job gets `INVALID_STATE` naming the state (`"Expected to %s job with key '%d', but it is in state '%s'"`). Added a paragraph right after the "continues to work" line covering the rejection and the timed-out-then-re-parked race the ADR calls out. Fixed in f47ba8040.

ajanoni-camunda (approved):
✅ LGTM

PR #9889, opened 2026-09-10
Filter hub apis

alexronquillo:
## Description

SaaS API docs for Camunda Hub should ignore SM-only endpoints. Same the other way around. We identify an endpoint's availability via `x-availability`. This change introduces a new pre-gen filter step:

1. For every yaml file, only include paths that match the expected SaaS or SM environment.
2. If a path has no remaining endpoints, remove it.
3. Cleanup any references to paths/endpoints that were filtered out.

## PR Checklist


christinaausley (dismissed):
Nice this is good! Two minor things:

1. `Object.keys(methods).length > 0` counts every key on a Path Item Object instead of just HTTP methods. If a path has `summary`/`description`/`parameters` alongside operations and all its operations get filtered out, those leftover keys keep the count > 0 so an operation-less path could survive instead of being removed. Do any Hub spec path mixes shared metadata like that with fully-filterable operations?
2. Is `x-availability` set per-operation or per-path? `filterPaths()` only checks each key's own value so it wouldn't catch a path-level extension meant to cover all its operations...right?

also minor: no fallback for an unexpected `x-availability` value and the two filter files are ~95% duplicated... could be one module parameterized by environment?

alexronquillo:
Thanks @christinaausley 

> Do any Hub spec path mixes shared metadata like that with fully-filterable operations?

Not right now, but that doesn't mean it won't happen in the future. It wouldn't matter for the docs because they get rendered at the endpoint level, not at the path level. However, it might cause some weirdness in Postman. In any case, I've updated the script to drop any paths that don't have an operation or a reference to an operation, even if they have other metadata like you mentioned.

> Is `x-availability` set per-operation or per-path?

My understanding is we only want this at the operation level. This may also only be a temporary solution, as a better solution is just to have SaaS and SM spec files so no filtering is needed. All in all, I don't think we need to overengineer this. Good point, though.

> no fallback for an unexpected x-availability value

If someone gives another value, it just won't show up in the docs, and we'll know why. We handle the three known cases:

1. SaaS-only
2. SM-only
3. Both

I think that's good enough for now. If you or another reviewer disagrees, we can cover the edge cases.

> could be one module parameterized by environment?

Great point, abstracted 👍 I added a test suite, as well.

Can you give it another pass? 🙏

christinaausley (approved):

PR #9918, opened 2026-09-14
docs(hub): document the business value dashboard and add business value access to the Hub roles table

OmranAbazid:
## Description

Documents the Business Value Dashboard introduced in Camunda 8.10: a new organization-level **Business Value** page in Camunda Hub for tracking process outcomes and setting targets.

Closes https://github.com/camunda/camunda-hub/issues/27251
Epic: https://github.com/camunda/product-hub/issues/3543

**New page** — `components/hub/organization/analyze-operations/business-value-dashboard.md`:

- How to open the dashboard and select an environment.
- Availability and permissions, linking to the central Hub roles matrix.
- How metrics are calculated, including the completed-instances-only rule.
- Automation rate and cycle time definitions, including root-process-only scope for automation rate.
- The portfolio view and process view, with per-metric interpretation guidance.
- Setting and updating cycle time and automation rate targets.
- Filters and their effect on results.
- Moving from a missed target to an investigation in Operate and Optimize.
- Multi-tenancy behavior.

**Supporting changes:**

- `manage-users/index.md`: added business value access to the roles and permissions table, merged into the existing Optimize column since the two are granted by the same roles.
- Sidebar entry and navigation cards under **Organization > Analyze operations**.
- 8.10 release notes entry under **Camunda Hub**.

### Notes for reviewers

Two items need confirmation from the DRI:

1. **Release notes placement.** The entry sits under `8.10.0-alpha5` because that was the most recent section. If the dashboard first ships in a different tag or at GA, it should move.
2. **UI labels.** The page has no screenshots, and labels such as **Set target**, **Update target**, and **Off-target details** are taken from the issue description rather than the shipped UI.

The permission model follows the final decision recorded in camunda/camunda-hub#27251, including that the standalone `Optimize` role does not grant access.

## PR Checklist


janxc15 (changes requested):
Nice page overall — structure and the metric definitions read well. Four things on
terminology and UI labels, all in `docs/components/hub/organization/analyze-operations/business-value-dashboard.md`.

---

**1. `Activity` is a section name, not a filter (lines 111, 64, 77)**

```
| Activity    | Limits the portfolio view to processes with recent activity, where available, so retired processes are hidden. |
```

The dashboard ships four filters: Environment, Process, Date range, and Version
(the last one only renders once a process is selected). There's no activity
filter — see `BusinessValueFilters.tsx` in `camunda-hub`.

Additionally, the portfolio and process tables call the section
**Volume** (lines 64 and 77), but the UI heading is **Activity**. Worth renaming
those rows to match. 

---

**2. Avoid "KPI" — use "metric" and "target" instead**

Lines 63, 76, 81, 87, 95 introduce "KPI" as a third term alongside "metric" and
"target", without defining how the three relate. Users won't know what it refers to:
"KPI" appears exactly once in the shipped UI (a helper line in the Set targets
modal). Everything else users see says **target**.

Suggestion: **metric** for the thing being measured, **target** for the threshold
someone sets on it. Concretely:

- line 63 — "How many configured KPI targets are currently met" → "How many configured targets are currently met"
- line 76 — "Per-KPI value and status" → "Per-metric value and status"
- line 81 — "for the KPIs this process supports" → "for the metrics this process supports"
- line 87 — table header `| KPI | Direction |` → `| Metric | Direction |`
- line 95 — "Click **Set target** for the KPI" → "...for the metric"

---

**3. `Update target` isn't a real label (lines 81, 95)**

There's no Update target button. The UI has a single **Set target** button; the
modal it opens is titled **Set targets** and saves with **Save targets** —
same control whether or not a target already exists.

Two options: change both lines to **Set target** to match what shipped, or, if
the distinct label was the intended design, file it as a bug against the dashboard
and keep the docs as-is until it lands. Happy either way — just flagging that as
written the page describes a button users won't find.

---

**4. `Off-target details` → `Off-target processes` (lines 68, 119)**

The section heading in the UI is **Off-target processes**, with the subtitle
"Ranked by how many targets are off first, then by how far off."

---

**5. Self-Managed access isn't covered**

The *Availability and permissions* section is entirely SaaS; an SM reader lands here and finds nothing that applies
to them. We need to at least add a sentence saying that on Self-Managed
access is determined by the cluster rather than by a Hub role.

OmranAbazid:
Thanks — all five addressed in 9cb86ef, verified against `camunda-hub`.

| # | Change | Verified against |
| :--- | :--- | :--- |
| 1 | Removed the `Activity` filter row; `Version` now notes it appears once a process is selected | `BusinessValueFilters.tsx` |
| 1b | Volume → Activity | `L0_SECTIONS` in `l0-portfolio/sections.ts` |
| 2 | KPI → metric / target at all five lines | — |
| 3 | Single **Set target** → **Set targets** dialog → **Save targets** | `SetTargetModal.tsx` 235, 471 |
| 4 | **Off-target processes**, plus corrected ranking wording | `OffTarget.tsx` 111 |

On 1b I also propagated Activity to the description, lead sentence, nav cards, and release notes — leaving "volume" in the intro while the tables said "Activity" would've been worse than either choice.

**Point 5** was wrong in a different way than flagged. `use-business-value-access.sm.ts` says it gates on "the Hub org-level `admin:bi` grant, not the per-cluster Optimize access". Both variants return `hasBiManagementPermissions`. So instead of an SM-specific sentence, the section now states one rule for both editions.

### Follow-ups, resolved in b3b0a42

- **Standalone `Optimize` role** — confirmed it doesn't grant Hub access, so the note stays as written.
- **Coverage and attainment** — merged into a single **Target coverage and attainment** row, matching `TargetOverview.tsx`.
- **Release notes placement** — the dashboard didn't ship in alpha5, so the entry moved out of that section into a new `## 8.10.0` minor release section.

janxc15 (dismissed):

giorgionaps (approved):
lgtm!

PR #9739, opened 2026-08-28
docs(self-managed): document the Multi-Region RDBMS architecture and its AWS EKS implementation

leiicamundi:
## Description

Closes #9547

Documents the **Multi-Region RDBMS** architecture: one Orchestration Cluster stretched across three or more regions, backed by a relational secondary storage whose replication is the database's responsibility.

The distinguishing property is that a region loss does not stop the engine. With one replica of every partition per zone, losing a zone leaves a majority, so Zeebe elects new leaders and keeps processing instead of waiting for an operator. And because there is one database rather than one per region, failback has no snapshot-and-restore step.

The concept is deployment-platform agnostic. Only the reference implementation is AWS-specific, and the pages are structured so another platform can be added without rewriting the concept.

### New pages

Three pages, one per user need:

| Page | Answers |
| --- | --- |
| `concepts/multi-region/multi-region-rdbms.md` | Which multi-region strategy do I pick, and why does this one keep processing? |
| `.../amazon-eks/multi-region-rdbms.md` | How do I stand this up on AWS? |
| `operational-tasks/multi-region-rdbms-ops.md` | A region just went down. What do I run? |

The operational runbook is deliberately short compared to the [dual-region one](https://github.com/camunda/camunda-docs/blob/main/docs/self-managed/deployment/helm/operational-tasks/dual-region-ops.md): most of that procedure exists to restore a quorum that this topology never loses.

Reference material is not duplicated. The pages link to the existing [zone-aware clusters](https://github.com/camunda/camunda-docs/blob/main/docs/self-managed/components/orchestration-cluster/zeebe/configuration/zone-aware-clusters.md) and [RDBMS multi-region support](https://github.com/camunda/camunda-docs/blob/main/docs/self-managed/concepts/databases/relational-db/configuration.md#multi-region-support) sections rather than restating them.

### Updated pages, so the architecture is reachable

- `resilience-tiers.md`: adds Multi-Region RDBMS as a third strategy, with a decision row and a third comparison column. Existing rows are preserved and a `Regions`, `Secondary storage`, and `Optimize` row were added, since those are now the differentiators.
- `zone-aware-clusters.md`: links from **Related resources**, since this architecture is the main consumer of zone awareness.
- `databases/relational-db/configuration.md`: links from **Multi-region support**, which currently states the constraint without pointing at an architecture built on it.
- `reference-architecture/kubernetes.md`: lists the new implementation under Amazon EKS.
- `sidebars.js`: three entries.

### Diagrams

[...]

PR #3449, opened 2024-03-11
docs(choosing-the-dmn-hit-policy): Fix formatting

tobiasschaefer:
## Description

Fixes this formatting issue:

image

I fixed the trivial issue but there is more and I leave it up to you to decide what is correct:
* in one place it says "by means of an execution listener attached to the end of your business rule task." However, execution listeners are currently not support by C8, only by C7.
* the whole page assumes that the hit policy is visualized by a single letter, e.g. "U" for "Unique". But actually, it never only shows the single letter in the UI. If I'm not mistaken this was the old behaviour.

## When should this change go live?

- [ ] This change is not yet live and should not be merged until {ADD_DATE} (apply `hold` label or convert to draft PR)?
- [x] There is no urgency with this change.
- [ ] This change or page is part of a marketing blog, conference talk, or something else on a schedule.
- [ ] This functionality is already available but undocumented.
- [x] This is a bug fix or security concern.

## PR Checklist

- [ ] I have added changes to the relevant `/versioned_docs` directory, or they are not for an **already released version**.
- [ ] I have added changes to the main `/docs` directory (aka `/next/`), or they are not for **future versions**.
- [ ] My changes require an [Engineering review](https://github.com/camunda/camunda-docs/blob/main/howtos/documentation-guidelines.md#review-process), and I've assigned an engineering manager or tech lead as a reviewer, or my changes do not require an Engineering review.
- [x] My changes require a [technical writer review](https://github.com/camunda/camunda-docs/blob/main/howtos/documentation-guidelines.md#review-process), and I've assigned @christinaausley as a reviewer, or my changes do not require a technical writer review.

PR #3440, opened 2024-03-07
EXPERIMENT: check-version workflow should be skipped from the start

pepopowitz:
DO NOT MERGE!

## Description

Experimental PR for #3415. 

And I will do this by adding the check-versions:skip label when creating the PR.

PR #9924, opened 2026-09-15
docs(self-managed): document the ECS dual-region replication lag settings [ready]

leiicamundi:
## Description

Docs counterpart to camunda/camunda-deployment-references#3438, which pins these settings in the ECS dual-region reference architecture.

The ECS dual-region page describes Aurora Global as secondary storage and says the tier is active-standby, but says nothing about what happens to exported data when the writer is promoted. Aurora replicates asynchronously, so a promotion can leave the new writer missing whatever had not reached it, and what closes that gap is Camunda holding back Zeebe log compaction until the database confirms, not anything Aurora does.

This adds a **Secondary storage replication lag** subsection under **Architecture decisions**, alongside the existing networking-mode and VPC-source decisions, covering the four settings the reference architecture pins and why.

Two points are worth having in writing, because both are natural misreadings:

- **`pause-on-max-lag-exceeded` is not a data-loss control.** Acknowledgement is gated on confirmed replication either way. What it decides is whether the exporter keeps writing to a database that is already behind, or stops and says so.
- **`max-lag` is not compared against a lag figure Aurora reports.** Under `LOG_SEQ` the engine measures the age of the oldest exporter position still waiting for its log sequence number, so the value is the longest replication interruption to ride out, not an acknowledgement delay. The reference architecture pins `PT1H` rather than the `PT15M` engine default, because a cross-region writer promotion under sustained load runs past 15 minutes.

The section links to [multi-region support](/self-managed/concepts/databases/relational-db/configuration.md#multi-region-support) for the vendor support list and the delay-backoff fallback rather than restating them.

Notes:

- EFS is elastic, so the guidance is to monitor storage growth and throughput rather than to size the volume. The EKS sibling differs here, because its disk is a fixed PVC.
- No sidebar change, since this is a subsection of an existing page.
- Kept as a draft until camunda/camunda-deployment-references#3438 merges, because it documents the values that PR pins.
- Related: #9739 documents the same mechanism for the EKS multi-region architecture.

## PR Checklist


leiicamundi:
Fixed in `732fb3ee8`: `async-replication.max-lag` = `PT15M` now has its own row and the count says four rather than three. The prose referenced `max-lag` without ever stating its value, so the configuration was not reproducible from the page.

Separately, and not from this review: the section originally said pausing makes the degradation "bounded". That was wrong and is corrected in `e3f592daf`. `RdbmsExporter.export()` throws while the exporter is paused, so its position stops advancing and log compaction stays blocked exactly as it does while replication lags — the Zeebe log grows either way. What pausing buys is back-pressure and a loud failure, not disk headroom. Same correction in camunda/camunda-deployment-references#3438.

docs/self-managed/deployment/containers/cloud-providers/amazon/aws-ecs-dual-region.md
```diff
@@ -205,6 +205,22 @@ BYO-VPC is the preferred path for customers integrating with an existing AWS lan
...
+
+The reference architecture enables it and pins the three settings that decide what it delivers:
+
+| Setting                                       | Value     | Why                                                                                                                                               |
+| --------------------------------------------- | --------- | ------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `async-replication.enabled`                   | `true`    | Off by default. Without it the exporter acknowledges records the standby has not received, and a failover loses them.                             |
+| `async-replication.type`                      | `LOG_SEQ` | Reads Aurora's own replication position. It is the engine default, but only some vendors support it, so it is pinned where the vendor is known.   |
+| `async-replication.pause-on-max-lag-exceeded` | `true`    | Decides the failure mode once `max-lag` is exceeded. It is not a data-loss control: acknowledgement is gated on confirmed replication either way. |
```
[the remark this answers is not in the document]
leiicamundi: Valid finding, fixed. The `async-replication.max-lag` row is in the table and the surrounding prose says "the four settings", so the count matches.

One update since: the pinned value is now `PT1H`, not `PT15M`. `PT15M` turned out to be the engine default (`ReplicationConfiguration.DEFAULT_MAX_LAG`) rather than a value sized for cross-region Aurora, and a reviewer on the reference-architecture PR measured failovers running past it under sustained load. Both the reference architecture and this page now pin `PT1H`.

The page also explains what the value is actually compared against, because the property name is misleading: under `LOG_SEQ` the engine measures the age of the oldest exporter position still waiting for its log sequence number to be confirmed, not a lag figure Aurora reports.

Fixed in 732fb3ee8 (row added) and c7a4478b4 (value raised, semantics documented).

leiicamundi:
Second review triaged.

**Scope mismatch** — fixed. The description said three settings while the table documents four; it now says four. `max-lag` gained its own row in `732fb3ee8` after the first review.

**`PAUSE_WORST_CASE` timing** — already addressed in `c7a4478b4`, which landed after the review ran. The paragraph now distinguishes the two cases explicitly: while nothing is queued a secondary below `min-sync-replicas` is reported as worst-case lag and pauses at the next poll whatever the budget, and once positions are queued the queue-head age governs instead, so that case waits out `max-lag` like any other. That matches `TimeMonitoringReplicationSignalStrategy.computePauseLag()`, which returns `queueHeadAge.orElse(PAUSE_WORST_CASE)` when the quorum is short.

Note the same commit raised the documented value from `PT15M` to `PT1H`, matching camunda/camunda-deployment-references#3438. The EKS multi-region reference is still on `PT15M`; I have flagged that drift on camunda/camunda-deployment-references#3442 rather than changing it before #3438 merges.

docs/self-managed/deployment/containers/cloud-providers/amazon/aws-ecs-dual-region.md
```diff
@@ -205,6 +205,27 @@ BYO-VPC is the preferred path for customers integrating with an existing AWS lan
...
+
+`pause-on-max-lag-exceeded` is worth understanding before you change it. It is not a data-loss control, and it is not a disk control either. Records are only ever acknowledged once Aurora confirms them, so no data is lost either way, and the Zeebe log grows either way: the exporter position cannot advance past unconfirmed records, so compaction stays blocked for as long as replication is behind, paused or not.
+
+What it decides is whether the exporter keeps writing to a database that is already lagging, or stops and says so. Paused, the exporter raises an error that reaches the exporter metrics and the broker log rather than degrading quietly, and it stops adding load to the database that needs to catch up. Zeebe keeps processing throughout, and the APIs and web applications that read secondary storage serve stale data until Aurora recovers. Size the EFS volume and alert on replication lag regardless of this setting.
```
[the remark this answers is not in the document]
leiicamundi: Correct, and the wording was misleading. Fixed in 90ca2f124.

EFS is elastic, so a prolonged replication outage never hits a capacity wall the way the EKS sibling's fixed PVC does. It grows stored bytes and burns throughput for as long as it lasts, which surfaces as cost rather than a full disk, and no amount of up-front sizing prevents that. The page now says to monitor EFS storage growth and throughput, and alert on replication lag, instead of telling operators to size the volume.

The same sentence was in the reference architecture's `locals.tf`, fixed there in camunda/camunda-deployment-references@c3c88cb05. The EKS multi-region values file keeps "size the disk", which is accurate there because the volume really is a fixed PVC.

leiicamundi:
Picking up the two suppressed findings from the latest review as well, since both were right.

**Quorum loss is not always immediate.** The page claimed a missing secondary "trips the pause immediately past any finite budget". That only holds while the replication queue is empty: `computePauseLag()` returns `PAUSE_WORST_CASE` under `queueHeadAge.isEmpty() && statuses.size() < minSyncReplicas`, and otherwise falls through to the queue-head age. With positions already queued, quorum loss waits out `max-lag` like any other interruption. The two cases are now separated in 90ca2f124, and the same correction went to both reference-architecture PRs (camunda/camunda-deployment-references#3438, #3440), which carried the same sentence.

**Setting count.** The page itself said "the four settings" and listed four rows; it was this PR description that still said three, written before `max-lag` was added to the table. Description updated.

leiicamundi:
Latest suppressed nit taken too: the `max-lag` row read as though the value delayed every acknowledgement, which contradicted the paragraph under it. Confirmed positions are acknowledged as soon as Aurora reports them, so the row and the paragraph now both call it a threshold on the age of the oldest unconfirmed position. Fixed in 65063bfc5, and mirrored to both reference-architecture PRs, which carried the same phrasing.

leiicamundi:
Both suppressed nits from the last pass taken, in 009fbe7d3.

- The table listed bare suffixes with no prefix anywhere in the section, so none of the four strings was copyable. The intro now names the `camunda.data.secondary-storage.rdbms.` prefix and says the table shortens it.
- "A secondary that drops below `min-sync-replicas`" inverted the comparison. `min-sync-replicas` is a replica count, and `LsnReplicationSignalStrategy` compares `statuses.size()` against it, so what falls below the minimum is the number of in-sync replicas.

docs/self-managed/deployment/containers/cloud-providers/amazon/aws-ecs-dual-region.md
```diff
@@ -205,6 +205,31 @@ BYO-VPC is the preferred path for customers integrating with an existing AWS lan
...
+| --------------------------------------------- | --------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `async-replication.enabled`                   | `true`    | Off by default. Without it the exporter acknowledges records the standby has not received, and a failover loses them.                                                      |
+| `async-replication.type`                      | `LOG_SEQ` | Reads Aurora's own replication position. It is the engine default, but only some vendors support it, so it is pinned where the vendor is known.                            |
+| `async-replication.max-lag`                   | `PT1H`    | Threshold on the age of the oldest unconfirmed exporter position, which is what the next setting reacts to. Not an acknowledgement delay. Raised from the `PT15M` default. |
+| `async-replication.pause-on-max-lag-exceeded` | `true`    | Decides the failure mode once `max-lag` is exceeded. It is not a data-loss control: acknowledgement is gated on confirmed replication either way.                          |
```
entangled90: 🔧 I think we should keep the pause-on-max-lag-exceeded off.
I [commented](https://github.com/camunda/camunda-deployment-references/pull/3438#issuecomment-5694145754)  in the code PR as well

docs/self-managed/deployment/containers/cloud-providers/amazon/aws-ecs-dual-region.md
```diff
@@ -205,6 +205,31 @@ BYO-VPC is the preferred path for customers integrating with an existing AWS lan
...
+
+`max-lag` is not compared against a lag figure Aurora reports, and it doesn't delay acknowledgement: confirmed positions are acknowledged as soon as Aurora reports them. Under `LOG_SEQ`, the engine measures how long the oldest exporter position has been waiting for its log sequence number to be confirmed, so the value is really the longest replication interruption you want the deployment to ride out. A cross-region writer promotion under sustained load can run past the `PT15M` engine default, which would pause the exporter during the exact event this architecture treats as routine.
```
entangled90: `is not compared against a lag figure Aurora reports` this is true but a bit misleading. It's explained a bit better later on.

docs/self-managed/deployment/containers/cloud-providers/amazon/aws-ecs-dual-region.md
```diff
@@ -205,6 +205,31 @@ BYO-VPC is the preferred path for customers integrating with an existing AWS lan
 
 The full validation contract — including the plan-time checks that fail with a descriptive error when a constraint is missing — lives in [`terraform/vpc/README.md`](https://github.com/camunda/camunda-deployment-references/blob/main/aws/containers/ecs-dual-region-fargate/terraform/vpc/README.md) in the reference repository.
 
+### Secondary storage replication lag
+
+Aurora Global Database replicates asynchronously, so a writer promotion can leave the new writer missing whatever had not reached it yet. What closes that gap is Camunda, not Aurora: with asynchronous replication monitoring enabled, the RDBMS exporter acknowledges a record to the broker only once the database reports it replicated, which holds back Zeebe log compaction so the missing records are replayed from the log.
+
+The reference architecture enables it and pins the four settings that decide what it delivers. All four sit under the `camunda.data.secondary-storage.rdbms.` prefix, shortened in the table below:
```
entangled90: We have explanations for these properties in the properties reference section as they map 1 to 1 with them. Should we point back there instead of repeating it here? It's harder to maintain long term if we have on copy for each reference architecture

leiicamundi:
All three suppressed nits from the last pass taken, in 250cb55ba. Each was an overstatement.

- **Compaction.** "Blocked for as long as replication is behind" was too strong. The exporter position cannot advance past unconfirmed records, so compaction is held back while records stay unconfirmed; once the queue drains, `drainConfirmed()` has acknowledged through the last confirmed entry and compaction resumes up to that position.
- **`DELAY`.** Database-confirmed acknowledgement is a `LOG_SEQ` property, not a property of async replication monitoring as a whole. The opening sentence now scopes the claim to the strategy this page pins and says `DELAY` waits a fixed interval instead.
- **Stale reads.** Attributing them to the lagging secondary was wrong: Camunda connects through Aurora's global writer endpoint, so reads never come from the secondary. What falls behind is the writer itself, because a paused exporter stops feeding it, and it recovers when exporting resumes rather than when replication catches up.

docs/self-managed/deployment/containers/cloud-providers/amazon/aws-ecs-dual-region.md
leiicamundi: Agreed, and done in 9e81ab1e1. It is back to the engine default of `false` here, in the ECS reference architecture (camunda/camunda-deployment-references#3438) and in the EKS one (camunda/camunda-deployment-references#3449), which had inherited the same `true`.

Your reasoning holds up against the code. Pausing protects nothing that acknowledgement does not already protect: records are reported safe to the broker only after confirmed replication whether or not pausing is on, and the Zeebe log is held by the unacknowledged position either way. What it adds is back-pressure and a loud error, at the cost of secondary storage stopping on its own. That is a deliberate choice, not a default to inherit from an architecture someone copies.

One consequence worth flagging: with pausing off, `max-lag` does nothing at all. `DefaultReplicationController` only ever compares it inside the pause check, so the whole `PT15M` to `PT1H` discussion is inert until someone enables pausing. I kept the `PT1H` value and documented it as a pre-sized budget, so enabling pausing stays a one-line change rather than a change plus a re-derivation. Happy to drop the property instead if you would rather the architecture not pin an inert value.

docs/self-managed/deployment/containers/cloud-providers/amazon/aws-ecs-dual-region.md
leiicamundi: Fair, and the sentence is gone in 9e81ab1e1.

It was there to pre-empt a misreading of the property name, but it defined the value by what it is not and then explained it properly two paragraphs later. The section now leads with the mechanism instead: exporting and acknowledging are separate steps, and `max-lag` bounds how long the oldest unacknowledged position may sit. The property semantics themselves are left to the multi-region support reference.

docs/self-managed/deployment/containers/cloud-providers/amazon/aws-ecs-dual-region.md
leiicamundi: You are right, and I have restructured it in 9e81ab1e1.

The table no longer carries a "Why" column explaining what each property does. It now has "Why this value here" and records only the architecture-specific choice: why the monitoring is not optional on a stack that treats a writer failover as routine, why `PT1H` is sized for a cross-region promotion, and why pausing is left at its default. The paragraph above it points at [multi-region support](/self-managed/concepts/databases/relational-db/configuration.md#multi-region-support) for what the properties mean, including vendor support and the `DELAY` alternative, so the semantics live in one place.

The vendor support list and the delay-backoff description are gone from this page for the same reason.

leiicamundi:
All three suppressed findings from the last pass taken, in ba4359950, with one correction back.

**Flow control: you are right, and it changes the framing.** "Zeebe keeps processing throughout" was too strong. Worth being precise about what causes it, though: the acknowledged position is held back by the replication stall whether or not pausing is on, so the export backlog grows either way, and a large enough backlog triggers flow control regardless. Pausing neither causes that nor prevents it. The page now says that first, before the trade-off, so the halt is attributed to the stall rather than to the flag. Propagated to both reference architectures, which carried the same sentence (camunda/camunda-deployment-references#3438, #3449).

**The exporter does raise an exception, though.** `RdbmsExporter.export()` opens with:

```java
if (!replicationController.isReplicationInSync()) {
  throw new ExporterException(
      String.format("[RDBMS Exporter P%d] Exporting paused: DB-reported lag exceeded maxLag. Retry later.", partitionId));
}
```

You are right that the pause path itself only sets the state, logs a warning and records metrics. The exception comes from each later `export()` call. The page now describes both rather than collapsing them.

**Retention.** The sentence separated cause from purpose badly. A slow replica is exactly what holds the acknowledgement back and makes the log grow, so it is the cause; it is just not what the records are kept for, because a replica catches up from the writer and never from Zeebe. The retained records matter when the writer itself is lost. Both paths are now spelled out instead of one being excluded.

**LSN failure scope.** Correct, and narrowed. An unsupported vendor, a non-global Aurora instance, or a user without the required privileges fails while the exporter starts. A replication status read that fails later hits the `catch` in `checkReplication()` and is retried at the next poll.

leiicamundi:
All three suppressed findings taken, in 59984777f.

**Privileges are not checked at startup on the Aurora path, and you are right that the page implied they were.** `ReplicationLsnProviderFactory` only calls `validateRequiredPrivileges` for non-Aurora PostgreSQL and for MSSQL; both Aurora branches go straight to `createAuroraGlobalProvider()`. The startup claim is now scoped to an unsupported vendor or a non-global Aurora instance, and the page says the privileges are exercised by the periodic status reads instead, which are caught and retried.

**Pause trigger.** Corrected to name the threshold: pausing stops writes once `max-lag` is exceeded, or immediately if the required quorum is unavailable while nothing is queued. Ordinary lag below the budget changes nothing.

**What `max-lag` measures.** One clause restored before the "inert" sentence, since calling something a budget without saying what it bounds is not usable. Kept deliberately short: it says the value bounds the age of the oldest position still waiting for its log sequence number and is not an Aurora-reported figure, and leaves the full property semantics to the multi-region support reference, per @entangled90's point about not duplicating it per architecture.

leiicamundi:
Both suppressed findings taken, in f35784fbe. The first is a good catch.

**Flow control is opt-in here.** The page presented throttling as an automatic consequence of a growing backlog. [Configure flow control](/self-managed/operational-guides/configure-flow-control/configure-flow-control.md) says write rate limits are enabled by default in SaaS and disabled in Self-Managed, so on this deployment the backlog does not slow the engine by itself. Reworded to say the limits are off unless you turn them on.

**Quorum.** Right, acknowledgement waits for `min-sync-replicas` confirmations rather than every replica, and this architecture leaves that at one. The paragraph now names the quorum, so a replica outside it falling behind is not presented as holding acknowledgement back.

leiicamundi:
Both suppressed findings taken, in c96005d55.

**Quorum pause timing.** Correct, it is not synchronous. `DefaultReplicationController` evaluates quorum only inside its scheduled check, so the pause lands at the next replication poll rather than the moment the quorum drops. Reworded.

**Pausing does add pressure of its own.** Fair, "neither causes nor prevents it" was too flat. The configurable write limits key on the backlog rather than on this flag, which is the accurate version and is what the paragraph now says. The following paragraph adds what pausing does contribute: the exporter stops consuming records, and an exporter that falls behind holds up the other exporters on its partition. The acknowledgement guarantee is unchanged either way.

leiicamundi:
All three suppressed findings taken, in 50b1f333b. Each was an overstatement, and the first two are worth spelling out.

**Compaction does not follow this exporter alone.** [Exporters](/self-managed/concepts/exporters.md) says compaction happens "up to the lowest acknowledged position across all exporters and the stream processor". The page implied this exporter's acknowledgement released disk by itself. Holding its position back is enough to keep segments on disk, but releasing them also needs every other exporter and the stream processor to have moved past, so a second lagging exporter would still pin the lower bound. Reworded to say that.

**"No data is lost either way" was an absolute claim standing in for a comparative one.** The point is only that pausing changes nothing about what a promotion can lose, since acknowledgement waits for confirmed replication with the flag on or off. It is not a promise of zero loss: an unplanned promotion can still leave unacknowledged records absent from the promoted Aurora until they are replayed from the retained log. Scoped accordingly.

**Static limit and throttling are separate switches.** Confirmed in the defaults: `flowControl.write.enabled: false` and `flowControl.write.throttling.enabled: false` are independent, and only `throttling` reacts to the export backlog. The page said "enable flow control", which would let someone turn on the static limit and expect backlog-driven behaviour. It now names `throttling` and says the static limit alone does nothing here.

docs/self-managed/deployment/containers/cloud-providers/amazon/aws-ecs-dual-region.md
```diff
@@ -205,6 +205,35 @@ BYO-VPC is the preferred path for customers integrating with an existing AWS lan
...
+
+What pausing does not change: a replication stall holds the acknowledged position back whether or not you pause, so the export backlog grows either way. Self-Managed ships with both the static write limit and backlog throttling disabled, so that backlog does not slow the engine on its own. They are separate switches in [flow control](/self-managed/operational-guides/configure-flow-control/configure-flow-control.md): only `throttling` reacts to the export backlog, so turning on the static limit alone does nothing here. Either way they key on the backlog rather than on this flag, so pausing neither turns them on nor exempts you from them.
```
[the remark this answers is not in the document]
leiicamundi: Right on both counts, fixed in 72fc12365.

Calling the static limit a no-op was wrong: it caps the rate of all new records, so it does bound how fast the backlog can grow. What it does not do is react to how large the backlog already is. Only `throttling` adapts to it.

The closing sentence was also wrong to say both switches key on the backlog. It now says neither keys on this flag, which is the claim that actually mattered there.

leiicamundi:
Both suppressed findings from the last pass addressed in ebcd2e0a2, one by fixing it and one by stepping around a disagreement.

**Runtime failures: you are right, and the blanket claim was wrong.** `DefaultReplicationController.onFlush()` does not retry:

```java
} catch (final Exception e) {
  paused.set(true);
  log.error("[RDBMS Exporter P{}] Failed to capture replication state after flushing exporter position {}. "
      + "Exporting will remain paused until replication checks recover.", partitionId, exporterPosition, e);
}
```

So a failed marker capture pauses exporting, while a failed status read in `checkReplication()` is logged and retried at the next poll. The page now separates the two instead of promising a retry for everything after startup.

**Compaction: I am not taking this one as written, because it reverses the previous round.** The wording you flagged came from adopting the earlier finding on this same line, which cited [`concepts/exporters.md`](/self-managed/concepts/exporters.md). That page says:

> Log compaction then occurs up to the lowest acknowledged position across all exporters and the stream processor

which is what the sentence paraphrased. Rather than flip back and forth on a Zeebe internal, I have cut the mechanism from this page entirely. It now says only what an ECS operator needs: an unacknowledged position keeps segments on disk, and releasing them depends on the slowest consumer on the partition, with a link to the exporters page for the full rule. If the exporters page itself is inaccurate, that is worth a separate fix there rather than divergent wording here.

christinaausley (dismissed):
This looks really good and has a bunch of review depth! Fixed some small wording issues which were survivors of heavy revision rather than new problems.

docs/self-managed/deployment/containers/cloud-providers/amazon/aws-ecs-dual-region.md
```diff
@@ -205,6 +205,37 @@ BYO-VPC is the preferred path for customers integrating with an existing AWS lan
...
+
+A replica counted towards that quorum falling behind is what holds the acknowledgement back, so it is what makes the log grow. It is not what the log is kept for. A replica catches up from the writer, never from Zeebe. The retained records matter when the writer itself is lost: the promoted one resumes from its own position, and Zeebe replays the gap.
+
+The reference architecture pins four properties under `camunda.data.secondary-storage.rdbms.`, shortened in the table below. [Multi-region support](/self-managed/concepts/databases/relational-db/configuration.md#multi-region-support) documents what each one does, including which vendors support `LOG_SEQ` and how the `DELAY` alternative behaves. The table records only which values this architecture picks and why.
+
+| Setting                                       | Value     | Why this value here                                                                                                                           |
+| --------------------------------------------- | --------- | --------------------------------------------------------------------------------------------------------------------------------------------- |
```
entangled90, line 219: 🔧 I think we should keep the pause-on-max-lag-exceeded off.
I [commented](https://github.com/camunda/camunda-deployment-references/pull/3438#issuecomment-5694145754)  in the code PR as well

docs/self-managed/deployment/containers/cloud-providers/amazon/aws-ecs-dual-region.md
```diff
@@ -205,6 +205,37 @@ BYO-VPC is the preferred path for customers integrating with an existing AWS lan
...
+| `async-replication.max-lag`                   | `PT1H`    | Sized for a cross-region promotion under load, which runs past the `PT15M` default.                                                           |
+| `async-replication.pause-on-max-lag-exceeded` | `false`   | The engine default, kept deliberately.                                                                                                        |
+
+Under `LOG_SEQ`, `max-lag` bounds the age of the oldest exporter position still waiting for its log sequence number to be confirmed, not a lag figure Aurora reports. It does nothing while `pause-on-max-lag-exceeded` is `false`, because the engine compares it only inside the pause check. It is pinned anyway so the budget is already sized if you turn pausing on, which is then a one-line change.
+
+Whether to pause is left to you. Turning it on stops the exporter writing to Aurora once that budget is exceeded, or at the next replication poll if the required quorum is unavailable while nothing is queued. Ordinary lag below the budget changes nothing.
```
entangled90, line 227: > Whether to pause is left to you.
Not sure but this feels more like an LLM speaking rather documentation 🤷

entangled90: Not sure if we need to got his much in details about the replication poll?

I feel like we should not go too much into details into what those settings mean as they are already explained in the property reference, we should just link there.

We should add details that are relevant for this particular artifacture (explaining why this setting is used (e.g. `LOG_SEQ`)

docs/self-managed/deployment/containers/cloud-providers/amazon/aws-ecs-dual-region.md
leiicamundi: Confirmed, and it already is: the table pins `pause-on-max-lag-exceeded` to `false`, the engine default, since 9e81ab1e1. This thread is a re-post of the earlier one on the same line, which was addressed there and then went outdated when the paragraph moved.

Your reasoning from camunda/camunda-deployment-references#3438 is what the page now records: pausing protects nothing that acknowledgement does not already protect, because a record is reported safe to the broker only after confirmed replication either way. It is left as an opt-in for operators who have lag alerting and want the stall to be loud.

docs/self-managed/deployment/containers/cloud-providers/amazon/aws-ecs-dual-region.md
leiicamundi: Both taken, in 8ec66801e. You are right on both counts and they had the same cause: the section had grown a running commentary on the engine instead of a record of this architecture's choices.

**"Whether to pause is left to you."** Gone. It was addressing the reader about a decision rather than documenting anything, and the sentence after it already says what turning pausing on costs.

**Too much on the settings themselves.** Five paragraphs are now one. Removed: the `max-lag` mechanics under `LOG_SEQ`, the next-replication-poll timing, the quorum-pause condition, and the whole flow-control aside on the static write limit versus `throttling`. All of it is engine behavior that belongs in [multi-region support](/self-managed/concepts/databases/relational-db/configuration.md#multi-region-support), which the paragraph above the table already links to. What is left is the operational trade for this stack: turn pausing on only once you have lag alerting, because writes to Aurora stop and secondary storage stays stale until replication recovers.

**Why `LOG_SEQ` here.** That row said "pinned because the vendor is known and fixed here", which is not a reason. It now says it: LSN monitoring is the preferred strategy and Aurora Global Database with PostgreSQL supports it, where the `DELAY` fallback would add a static wait to every acknowledgement.

The exporting-versus-acknowledging paragraphs above the table are kept on purpose. They are not property semantics; they are why the Zeebe log grows on this architecture, which is what makes the EFS growth and the lag alerting below them matter. Say the word if you want those cut back too.

leiicamundi:
Copilot's latest pass raises no inline findings. Its summary asks for one thing, and it is the thing that was just deliberately removed, so recording why rather than acting on it:

> The `LOG_SEQ` section does not explain what `max-lag` measures, which may lead readers to interpret it as an Aurora-reported lag threshold.

That explanation was on the page until 8ec66801e, and @entangled90 asked for it to go: the properties map one to one with [multi-region support](/self-managed/concepts/databases/relational-db/configuration.md#multi-region-support), and a copy of their semantics in every reference architecture is what rots. The paragraph above the table links there and says the table records only which values this architecture picks and why. The misreading Copilot worries about is fixed in the reference page if it is worth fixing, not in a fourth copy here.

leiicamundi:
@entangled90 I believed having addressed all the comments. @christinaausley once @entangled90 approved, feel free to merge, ty!

christinaausley (approved):
Looks good! I verified the replication behavior against the existing RDBMS documentation and Camunda implementation and updated the section to clarify the points you raised:

- Camunda retains the source records needed to recover a replication gap- it does not repair Aurora replication.
- The required replica must confirm the flush marker before the exporter acknowledges the position.
- Under `LOG_SEQ`, `max-lag` is compared with the age of the oldest exporter position still waiting for confirmation, not with a lag value reported by Aurora.
- The EFS guidance now avoids an absolute capacity claim and focuses on storage growth, throughput, and cost.

leiicamundi:
Thanks @christinaausley

entangled90 (approved):

PR #9925, opened 2026-09-15
docs(hub): fix username-claim/user-id-claim conflation in auth docs

p-wunderlich:
## Summary

Camunda Hub has two similarly-named but distinct legacy 8.9 OAuth2/JWT config keys:

- `camunda.modeler.oauth2.token.user-id-claim` — the 8.9 **principal-identity** claim. This is translated onto CSL's `camunda.security.authentication.oidc.username-claim` at startup (default `sub`).
- `camunda.modeler.oauth2.token.username-claim` — Hub's own **display-name** claim (default `name`). This is deliberately **never** translated onto CSL, since doing so would key authorization on a display name.

Two docs.camunda.io pages conflated these:

- The 8.9→8.10 migration table in `890-to-8100.md` mapped the *username-claim* key to the CSL setting, when it should map the *user-id-claim* key. Added a clarifying note that `username-claim` is unrelated to this migration.
- `identity.md` documented the CSL `username-claim` setting's default as `name` instead of its actual default, `sub`, in both the YAML example and the env-var table.

Traced and verified directly against `LegacyAuthConfigEnvironmentPostProcessor` in `camunda/camunda-hub#28558` (merged), and cross-checked the `sub` default against this repo's own `_security.md` and `connect-external-identity-provider.md`.

## Test plan

- [x] `npx prettier --write` on both changed files — no formatting changes needed
- [x] Reviewed diff to confirm changes are limited to the intended cells/lines


docs/self-managed/components/hub/configuration/identity.md
```diff
@@ -29,7 +29,7 @@ camunda:
       oidc:
         issuer-uri: https://keycloak.example.com/auth/realms/camunda-platform
         client-id: web-modeler
-        username-claim: name # optional, default: name
+        username-claim: preferred_username # optional, default: sub
```
jfriedenstab, line 32: 💭 I think `preferred_username` could be confusing here, because it's not meant to be the unique identifier of a user. What about using `oid` as the example (from Azure AD)?

docs/self-managed/components/hub/configuration/identity.md
```diff
@@ -40,7 +40,7 @@ camunda:
...
 | `CAMUNDA_SECURITY_AUTHENTICATION_OIDC_CLIENTID`      | Client ID of the Camunda Hub application configured in your identity provider.                                                                                                                                                 | `web-modeler`                                               | -             |
-| `CAMUNDA_SECURITY_AUTHENTICATION_OIDC_USERNAMECLAIM` | [optional]<br/>Token claim used to assign usernames.                                                                                                                                                                           | `preferred_username`                                        | `name`        |
+| `CAMUNDA_SECURITY_AUTHENTICATION_OIDC_USERNAMECLAIM` | [optional]<br/>Token claim used to assign usernames.                                                                                                                                                                           | `preferred_username`                                        | `sub`         |
```
jfriedenstab, line 43: Same here. Probably we should also update the description. The [OC property reference](https://docs.camunda.io/docs/next/self-managed/components/orchestration-cluster/core-settings/configuration/properties/#camundasecurityauthenticationoidc) describes it as:

> The JWT claim that identifies a user. Extracted from a token, this claim value becomes the user's username. This setting is evaluated on any token-based access, regardless of the underlying OIDC/OAuth flow.

I would just copy the first sentence which holds true for Hub, too. The rest does not fit exactly I think.

jfriedenstab (changes requested):
Thanks for taking care of this 👍🏻.

jfriedenstab (approved):

Issue #9927, opened 2026-09-15
Camunda Hub final sweep


alexronquillo:
Walk through the Camunda Hub-related documentation to catch and fix any remaining issues.

Tracked issues:

- [ ] [Document Git sync upload limit](https://camunda.slack.com/archives/C07USKREKDL/p1789463876484549?thread_ts=1789459697.461659&cid=C07USKREKDL)
- [ ] [Fix Helm migration guide table formatting for better readability](https://camunda.slack.com/archives/C0ADXE9F8N7/p1789149862002059)
- [ ] [Single-file deployment now possible in Test mode](https://camunda.slack.com/archives/C07USKREKDL/p1789365611086449?thread_ts=1788529590.179799&cid=C07USKREKDL)

PR #9933, opened 2026-09-15
Fix table format

alexronquillo:
## Description

Improve readability of large tables.

This PR adds a new "ExpandableTable" component that allows users to see a fullscreen view of the table. It's also intended to help users recognize that the large table has more content offscreen (sometimes that's not obvious).

In this PR, I added it around a very small table, too, mainly for the purpose of demonstration/testing. I can remove it from the small table before merging.

Related to: https://github.com/camunda/camunda-docs/issues/9927

## PR Checklist


christinaausley:
Nice UX idea! the "more content offscreen" signal is for sure a gap and the drop-shadow/expand pattern is a reasonable way to surface it

A couple things before merge:

1. `#### Camunda Hub` section wraps its table in `<ExpandableTable title="Deprecated orchestration keys">` ... I think copy-paste from the orchestration section above it. Should reference Camunda Hub, not orchestration, right?
2. `<button className="close-button"><Icon icon="bx:x" /></button>` is icon-only with no `aria-label`, so screen readers just announce "button." We could add `aria-label="Close"` but this is a nitpick.
3. No Escape-key handler to close the modal - right now the only way out is that unlabeled close button or clicking the backdrop.
4. No body-scroll lock while the modal is open - worth confirming the page doesn't still scroll behind the fixed overlay?
5. Possible CSS no-op: `.expandable-table-container table { padding: 15px; }` padding on a `<table>` element is unreliable under default `border-collapse`; worth double-checking it actually renders, or move it to a wrapping div?
6. `@iconify/react` is a new runtime package, not just icon data (`@iconify-json/logos` was already there). Per `repo.instructions.md` npm additions need a docs-infra discussion first... has that happened? Also curious if two inline SVGs would avoid the dependency entirely.

Also noted you already flagged removing it from the small demo table, and the Connectors table doesn't appear wrapped in this diff - intentional?

alexronquillo:
Thanks for the feedback @christinaausley 

> #### Camunda Hub section wraps its table in <ExpandableTable title="Deprecated orchestration keys"> ... I think copy-paste from the orchestration section above it. Should reference Camunda Hub, not orchestration, right?

Yes, this was a copy-paste issue, but as I mentioned in the description, it was only for testing/showcasing purposes. It was intended to be removed before merging anyway. I removed in in the latest commit ✅ 

> `<button className="close-button"><Icon icon="bx:x" /></button>` is icon-only with no aria-label, so screen readers just announce "button." We could add aria-label="Close" but this is a nitpick.

Good catch. Added ✅ 

> No Escape-key handler to close the modal - right now the only way out is that unlabeled close button or clicking the backdrop.

Good catch. Added ✅ 

> No body-scroll lock while the modal is open - worth confirming the page doesn't still scroll behind the fixed overlay?

Good catch. Added ✅ 

> Possible CSS no-op: .expandable-table-container table { padding: 15px; } padding on a <table> element is unreliable under default border-collapse; worth double-checking it actually renders, or move it to a wrapping div?

This does render. I think the padding is visually helpful, but feel free to let me know if you think otherwise 👍 

> @iconify/react is a new runtime package, not just icon data (@iconify-json/logos was already there). Per repo.instructions.md npm additions need a docs-infra discussion first... has that happened? Also curious if two inline SVGs would avoid the dependency entirely.

I think that directive in repo.instructions.md is really to prevent AI from introducing new dependencies unchecked rather than us from introducing the dependencies, right? In any case, using the inline svg works, too, so I removed the dependency. ✅

mesellings (approved):
Love this, great idea and implementation to a traditional tech writing issue - let's hope we don't need to worry about printing the site, but until then let's get this merged and we can iterate as need be.

Btw, agree this should be an opt-in for tables, rather than a default style for all tables - probably only needed for a low number of tables (usually property/class related ones).

Issue #9795, opened 2026-09-02
Property reference (core-settings/configuration/properties) is missing ~120 camunda.* properties, 2 entire sections, and has 5 wrong property names


deepthidevaki:
## Summary

The unified-configuration property reference at
https://docs.camunda.io/docs/next/self-managed/components/orchestration-cluster/core-settings/configuration/properties/
(source: `docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/*.md`
in this repo) is missing roughly 120 real, currently-bindable `camunda.*` properties, is missing
two entire sections outright, and has 5 property names that are simply wrong (won't bind to the
real property).

This was found by regenerating `dist/src/main/config/defaults.yaml` from current
`camunda/camunda` source (`configuration/src/main/java/io/camunda/configuration/`) and diffing it
against every partial on this page, then verifying each candidate by hand against the Java source
and the raw doc content to rule out false positives (deprecated legacy aliases, `Map`/`List`
fields that collapse in generated metadata, and Spring relaxed-binding-equivalent spelling
variants). Every item below was confirmed as a real, bindable property absent from every partial
on this page, in every plausible spelling.

**Explicitly out of scope for this issue** (already tracked elsewhere, do not duplicate):
- `camunda.physical-tenants.*` — tracked in #9729.
- Non-kebab-case spelling variants that still bind correctly (e.g. `advertisedHost` vs.
  `advertised-host`, `rocksdb` vs. `rocks-db`) — tracked in #9688.
- `camunda.data.primary-storage.rocks-db.*` missing from the separate broker-config reference page
  — tracked in #9687 (that's a different page from this one; `rocks-db` **is** present on this
  page already).
- `camunda.processing.evaluate-boundary-event-correlation-key-in-activity-scope` — tracked in
  #9380 (note: that issue currently says "no unified `camunda.*` name exists yet" — it does now,
  this is it, worth updating that issue's config-table task).

### `camunda.document.*`

Not on this reference page at all. (It does have example-driven coverage on the
[document handling configuration](/self-managed/concepts/document-handling/configuration/helm.md)
page, but that page isn't a systematic property table and doesn't cover every field below —
either way, every other `camunda.*` section gets a full table on *this* reference page, and this
one doesn't.)

| Property | Default | Notes |
|---|---|---|
| `camunda.document.default-store-id` | `null` | |
| `camunda.document.thread-pool-size` | `null` | |
[...]

PR #9935, opened 2026-09-15
docs: work in progress restructuring of the backup & restore

panagiotisgts:
documentation for 8.10

## PR Checklist


christinaausley:
@panagiotisgts Let me know when you are ready for a final review, or if you need help answering any questions/putting the docs together now?

docs/self-managed/operational-guides/backup-restore/elasticsearch/restore-snapshot.md
```diff
@@ -0,0 +1,572 @@
...
+sidebar_label: "Restore Elasticsearch/OpenSearch snapshot"
+keywords:
+  ["backup", "backups", "restore", "elasticsearch", "opensearch", "snapshot"]
+description: "Learn how to restore Elasticsearch or OpenSearch snapshots from a Camunda 8 Self-Managed backup."
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+### Prerequisites
```
deepthidevaki, line 13: Here, mention that this is part of the overall restore process and link to the parent page.

docs/self-managed/operational-guides/backup-restore/elasticsearch/restore-snapshot.md
```diff
@@ -0,0 +1,572 @@
...
+└── 1749130104
+3/
+├── 1748937221
+└── 1749130104
+```
+
+   </summary>
+</details>
+
+### 3. Stop all components apart from Elasticsearch/OpenSearch
```
deepthidevaki, line 383: This is not required when using the new restore API.

deepthidevaki (changes requested):
🚀  Thanks for the restructure. Overall it looks good.

:x: The restore process for ES/OS is a bit confusing. The difference between the legacy approach (stop all components) vs restore during the broker-recovery mode is unclear.
🔧 Unrelated to this PR - The  pages still documents breaking changes introduced in 8.8. This could be confusing to the readers.


More comments from Claude:

Technical/endpoint accuracy

1. **Wrong `/mode` response shape** — `docs/self-managed/operational-guides/backup-restore/rdbms/restore.md` (~L2020, "Switch the cluster into recovery mode"): example shows `plannedChanges` as a flat array of operations (`{changeId, plannedChanges: [{operation, mode}, ...]}`). Per `cluster.yaml`'s `ClusterModeChangeResponse`, `plannedChanges` must be an array of `{physicalTenantId, operations: [...]}`. The PR's own `/restore` examples nest this correctly — only the `/mode` example is flattened.

2. **Stale step anchor** — `docs/self-managed/operational-guides/backup-restore/elasticsearch/restore.md:1546`: links to `#3-trigger-the-restore`, but in this file that step is `### 4. Trigger the restore` (ES has an extra step vs. RDBMS). Looks like text copy-pasted from `rdbms/restore.md` without adjusting the anchor.

3. **Duplicated `/v2` in curl example** — `docs/self-managed/operational-guides/backup-restore/rdbms/restore.md:2040`: `curl "${ORCHESTRATION_CLUSTER_API}/v2/topology"` where `ORCHESTRATION_CLUSTER_API` is already exported as `.../v2`, producing `/v2/v2/topology`. The ES equivalent correctly omits the extra `/v2`.

All other endpoints checked against Zeebe's actual specs (`v2/backups.yaml`, `v2/cluster.yaml`, `v2/cluster-admin.yaml`) — `/backups/runtime`, `/backups/history`, `PATCH /v2/mode`, `/v2/restore`, `/v2/topology`, cluster-admin equivalents, `physicalTenantId`, `overrides`, `dryRun`, permissions — all matched correctly.

Broken links

4. `docs/components/admin/cluster-admin.md:39` and `docs/self-managed/concepts/physical-tenants/index.md:76` both link to `backup-and-restore.md#back-up-a-cluster-with-multiple-physical-tenants`, but this PR renames that heading to `### Multiple Physical Tenants`, breaking the anchor to `#multiple-physical-tenants`.

5. No orphaned references to the deleted `in-process-restore.md` remain — links were correctly repointed, and `sidebars.js` is updated for the new `es-restore-snapshot` page.

Style

6. Garbled/incomplete sentence in `docs/self-managed/operational-guides/backup-restore/rdbms/restore.md` (~L2034): "Requests sent earlier are rejected with `409`. verifying that all brokers are in recovery mode using either the Cluster API or the Management API." — missing lead-in clause, lowercase sentence start.

Structure & readability (user perspective)

 Landing page (`backup-and-restore.md`)
7. The "REST API"/"Management API" explainer (~L74–107) sits between the routing cards and the Physical Tenants section, unrelated to "which page do I need" — dilutes the page's job as a router. *Suggestion:* move to a reference doc.
8. No guidance anywhere on this page for Legacy vs. Restore API — the PR's own stated core decision axis is invisible until the reader is already inside `restore.md`. *Suggestion:* add a short "which restore approach do I use?" section up front.
9. Cards route through an extra in-page click (top-level cards reveal a second card set) instead of linking directly to the destination page.

 each `restore.md` (elasticsearch and rdbms)
10. `elasticsearch/restore.md` ~L370: mid-procedure in the Legacy tab it tells the reader "this page's Restore API tab contains the recovery-mode procedure" and then keeps going with Legacy steps anyway — confusing placement; should be a callout at the top of the tab instead.
11. Tabs nested 3 levels deep (approach → backup listing → Kubernetes/Manual) risk readers losing track of which layer they're in.
12. Nothing tells the reader up front how to determine whether their cluster is Legacy or Restore API (implied by version, never stated as an explicit rule).

 `restore-snapshot.md` (new page)
13. No back-reference/breadcrumb to the parent `restore.md` procedure — a reader arriving via search or sidebar can't tell it's a sub-step rather than a standalone page.
14. Its "stop all components" step is written for full-downtime Legacy restores but reused verbatim in the Restore API tab, where components may legitimately stay running, with no approach-specific callout.
15. Sidebar lists it as a sibling to backup/restore, visually implying peer status rather than a nested step.

Cross-cutting
16. Both `restore.md` files set `hide_table_of_contents: true` despite being 556 and 769 lines with nested tabs — removes the best navigation aid on exactly the pages that need it most.
17. Heading hierarchy diverges between ES and RDBMS Legacy sections (ES nests steps as H3 under an H2; RDBMS puts steps at H2 alongside Overview/Prerequisites), undermining the "familiar with one, navigate the other" goal.
18. Destructive-action warnings (data loss, partial-failure states, user-ID mismatches) are present and consistently styled — a strength, not an issue.

Under incident pressure
19. Reaching an actual command requires landing page → storage card → correct tab → jump to `restore-snapshot.md` and back, with no single-page quick reference for either approach.
20. The disabled ToC on both long restore pages removes the fastest way to jump straight to "trigger the restore" mid-incident.

docs/self-managed/operational-guides/backup-restore/assets/icon-considerations.png
leiicamundi: can't we use existing icons sets?

docs/self-managed/operational-guides/backup-restore/webapps-backup.md
```diff
@@ -9,6 +9,10 @@ keywords: ["backup", "backups"]
 import Tabs from "@theme/Tabs";
 import TabItem from "@theme/TabItem";
 
+:::tip
+From Camunda 8.10, you can also use the Orchestration Cluster REST API to create, query, and manage history backups for Operate and Tasklist. See the [Orchestration Cluster REST API reference](/apis-tools/orchestration-cluster-api-rest/specifications/take-history-backup.api.mdx).
```
leiicamundi: duplication, please extract it in a reusable block

docs/self-managed/operational-guides/backup-restore/backup-and-restore.md
```diff
@@ -79,18 +49,26 @@ Using an RDBMS as secondary storage unlocks three new capabilities not available
...
 
-:::note
-We recommend using the Unix timestamp as the backup ID.
+:::info
+The RDBMS backup path covers Zeebe, Operate, Tasklist, and Admin. Management Identity and Optimize are not covered by this path. Optimize always stores its data in Elasticsearch or OpenSearch, regardless of the Orchestration Cluster's secondary storage. If you use Optimize alongside an RDBMS-backed Orchestration Cluster, back up and restore Optimize independently using the [standalone Optimize backup procedure](./optimize-backup-and-restore.md). You do not need to switch the Orchestration Cluster to the Elasticsearch / OpenSearch backup path.
 :::
 
-:::note
-When using the RDBMS path, backup IDs are auto-generated by the cluster and do not need to be managed manually.
+:::warning
```
leiicamundi: to be verified (coherence)

docs/self-managed/operational-guides/backup-restore/backup-and-restore.md
```diff
@@ -9,46 +9,16 @@ description: "Learn how to back up and restore your Camunda 8 Self-Managed compo
...
-:::
+<ZeebeGrid zeebe={storageCards} />
 
-:::warning
-Because Optimize and an RDBMS-backed Orchestration Cluster are backed up independently on separate schedules, they may be restored to **different points in time**. This means Optimize analytics data — dashboards, reports, and historical process data — may not be consistent with the restored state of the Orchestration Cluster.
+:::tip Disaster recovery context
+If you are using backups as the foundation of a cross-region recovery strategy, see [Cold Recovery](../../concepts/multi-region/cold-recovery.md) for the architecture, RTO/RPO targets, and recovery flow. Cold Recovery builds on the procedures in this guide.
```
leiicamundi: should be in the related pages / associated topic, not a tip

docs/self-managed/operational-guides/backup-restore/react-components/_card-data.js
```diff
@@ -1,55 +1,67 @@
...
   {
-    link: "../elasticsearch/es-backup#elasticsearch--opensearch-prerequisites",
-    title: "Prerequisites",
-    image: IconGear,
+    link: "#relational-databases-rdbms",
+    title: "RDBMS",
+    image: IconRdbms,
     description:
-      "Set up a snapshot repository in Elasticsearch or OpenSearch and configure component backup storage. Covers all components including Optimize.",
+      "Relational database management system (RDBMS) as secondary storage.",
```
leiicamundi: to be verified: why we switch from ES to RDBMS here

docs/self-managed/operational-guides/backup-restore/rdbms/restore.md
```diff
@@ -2,6 +2,7 @@
 id: rdbms-restore
 title: "Restore a backup (RDBMS)"
 sidebar_label: "Restore a backup"
+hide_table_of_contents: true
```
leiicamundi: why?

leiicamundi (commented):
## Review summary

The restructure is a real improvement: routing by secondary storage first, then by backup vs restore, answers the two questions operators actually arrive with. The Restore API content is detailed and the destructive-action warnings are consistently placed.

Before it can merge, nine items need fixing. Two of them break the build, and one would mislead an operator during an incident. I checked every API claim against the OpenAPI specs committed in this repo (`api/camunda/v2/*.yaml`) and every cross-reference against the rest of the docs tree, so each point below cites what it was verified against.

### Blocking

1. **The two restore pages do not compile.** `elasticsearch/restore.md:13` imports `esRestoreCards` and `rdbms/restore.md:27` imports `rdbmsRestoreCards`, but `_card-data.js` exports only `storageCards`, `esCards` and `rdbmsCards` (lines 11, 34, 52). The `ZeebeGrid` import on both pages is also unused. Either add the card data and render it, or drop both imports.

2. **The Restore API flow tells operators to tear down the cluster it just put in recovery mode.** `elasticsearch/restore.md:157` sends the reader to `restore-snapshot.md`, which instructs them to start the full Camunda stack (lines 33 to 39) and then stop all components (line 383). The Restore API requires the brokers to stay up in recovery mode until `POST /restore` completes. Following the link as written destroys the recovery state. @deepthidevaki flagged the same thing on line 383.

3. **The `/mode` response example is the wrong shape.** `rdbms/restore.md:89` shows `plannedChanges` as a flat list of operations. `api/camunda/v2/cluster.yaml:361-382` defines it as a list of `ClusterModeChangePlannedChange`, which requires `physicalTenantId` and `operations`. The `/restore` examples in this same PR get it right, so only this one example is off.

4. **Renaming the Physical Tenants heading breaks six links in five files.** Details inline on `backup-and-restore.md:166`. Worth knowing: `docusaurus.config.js` sets `onBrokenLinks: "throw"` but never sets `onBrokenAnchors`, which defaults to `warn` in Docusaurus 3. CI will not catch this.

5. **The RDBMS card promises coverage that does not exist.** `_card-data.js:58` says "all components", but `rdbms/backup.md:32` states that the RDBMS path does not cover Optimize or Management Identity. The text it replaced said so correctly.

6. `rdbms/restore.md:106` calls `/v2/v2/topology` because `ORCHESTRATION_CLUSTER_API` already ends in `/v2` (line 67). Line 324 in the same file gets it right.

7. `elasticsearch/restore.md:314` links to `#3-trigger-the-restore`, but that step is `### 4. Trigger the restore` on this page. It is step 3 only on the RDBMS page.

8. `rdbms/restore.md:442` sends RDBMS operators to the Elasticsearch procedure, which has different secondary-storage steps.

9. The deleted `in-process-restore` page has no redirect in `static/.htaccess`. The PR checklist calls for one.

Items 1, 6, 7, 8 and 9 were also raised by the Copilot review, and I confirmed each one against the files.

### Non-blocking

The three new icons add roughly 580 KB because they are saved as 16-bit grayscale instead of 8-bit RGBA like the existing set. `hide_table_of_contents: true` on both restore pages makes them the two longest pages in `docs/` without a table of contents, and the only non-generated ones apart from a 70-line page. Several headings use title case or contain colons, which the style guide rules out, and two pages start at H3. Details inline.

### Checked and clean

No orphaned references to the deleted page remain anywhere. The three `storageCards` anchors all resolve. Apart from the `/mode` example, every endpoint, query parameter and permission I checked matches the specs: `GET /backups/runtime` with `prefix`, `GET /backups/history` with `prefix` and `verbose`, `dryRun` on `/restore`, and the `BACKUP:RESTORE` permission. The backup ID guidance was consolidated into `backup-and-restore.md:68` rather than lost. I also checked `camunda-deployment-references`, which has no links into this doc tree, so nothing breaks on the reference architecture side.

Two dangling anchors I found are pre-existing on `main` and not caused by this PR: `#backup-process` from `elasticsearch-without-cluster-privileges.md:471` and `#why-you-should-use-backup-and-restore` from `elasticsearch/backup.md:181`. Worth a separate fix.

docs/self-managed/operational-guides/backup-restore/react-components/_card-data.js
```diff
@@ -1,55 +1,67 @@
 // this makes use of the Zeebe react grid component
 
 import IconGear from "../assets/icon-prereqs.png";
```
leiicamundi: `IconGear` is no longer used anywhere in this file now that both "Prerequisites" cards are gone. Remove the import, or bring back a prerequisites card.

Worth a second look either way: dropping those two cards removes the only navigation path from the landing page to the prerequisites, and the restore pages still depend on them (`elasticsearch/restore.md:82`, `rdbms/restore.md:56` both link to `./backup.md#prerequisites`). If that removal was deliberate, ignore this.

docs/self-managed/operational-guides/backup-restore/react-components/_card-data.js
```diff
@@ -1,55 +1,67 @@
...
+import IconConsiderations from "../assets/icon-considerations.png";
```
leiicamundi: Can these reuse the existing icon set instead of adding new files?

If new icons are needed, they should at least be re-encoded. All six icons are 515x515, but the three added here are saved as 16-bit gray+alpha while the existing ones are 8-bit RGBA:

| file | size | encoding |
| --- | --- | --- |
| `icon-prereqs.png` (existing) | 5.1 KB | 8-bit RGBA |
| `icon-restore.png` (existing) | 7.4 KB | 8-bit RGBA |
| `icon-backup.png` (existing) | 9.2 KB | 8-bit RGBA |
| `icon-considerations.png` (new) | 176 KB | 16-bit gray+alpha |
| `icon-elasticsearch.png` (new) | 189 KB | 16-bit gray+alpha |
| `icon-rdbms.png` (new) | 228 KB | 16-bit gray+alpha |

That is about 580 KB added for three icons at the same display size, roughly 20 to 30 times the existing ones. Running them through `pngquant` or exporting at 8-bit should bring them in line.

docs/self-managed/operational-guides/backup-restore/react-components/_card-data.js
```diff
@@ -1,55 +1,67 @@
...
+    link: "#relational-databases-rdbms",
+    title: "RDBMS",
+    image: IconRdbms,
     description:
-      "Set up a snapshot repository in Elasticsearch or OpenSearch and configure component backup storage. Covers all components including Optimize.",
+      "Relational database management system (RDBMS) as secondary storage.",
   },
+  {
+    link: "#considerations",
+    title: "Additional Information",
```
leiicamundi: Title case. The style guide asks for sentence case in titles and headers (`howtos/technical-writing-styleguide.md:118`). Use "Additional information".

The label is also vague next to "Elasticsearch / OpenSearch" and "RDBMS", which name a storage choice. Since the card points at `#considerations`, "Considerations" would match both the heading and the style guide's ask for descriptive headings.

docs/self-managed/operational-guides/backup-restore/react-components/_card-data.js
```diff
@@ -1,55 +1,67 @@
...
-    description:
-      "Configure Zeebe continuous backup storage and your RDBMS backup tooling. Zeebe and RDBMS can be backed up independently Covers Zeebe, Operate, and Tasklist only.",
-  },
   {
     link: "../rdbms/rdbms-backup",
     title: "Create a backup",
     image: IconCloud,
     description:
-      "Enable decoupled continuous backups for Zeebe, Operate, and Tasklist. Zeebe automatically takes regular snapshots; the RDBMS is backed up separately using native database tools. Optimize is not supported.",
+      "Create a backup of all components while using an RDBMS as secondary storage.",
```
leiicamundi: Blocking: this says "all components", but the page it links to says the opposite. `rdbms/backup.md:32` reads "covers **Zeebe**, **Operate**, **Tasklist**, and **Admin**. It does **not** cover **Management Identity** or **Optimize**." The `Considerations` table on the landing page agrees with `backup.md`.

The text this replaced was accurate on that point. Suggest something like "Create a backup of the Orchestration Cluster components while using an RDBMS as secondary storage. Optimize is backed up separately."

Same wording on line 65 for the restore card, and the Elasticsearch cards on lines 40 and 47 now say "all components" too. There it happens to be true, which makes the phrase mean two different things across the two card sets.

docs/self-managed/operational-guides/backup-restore/react-components/_card-data.js
```diff
@@ -1,55 +1,67 @@
...
   },
   {
     link: "../rdbms/rdbms-restore",
-    title: "Restore a backup",
+    title: "Restore a backup ",
```
leiicamundi: Trailing space in the title: `"Restore a backup "`. It renders into the card label. The matching Elasticsearch card on line 44 does not have it.

docs/self-managed/operational-guides/backup-restore/backup-and-restore.md
```diff
@@ -9,46 +9,16 @@ description: "Learn how to back up and restore your Camunda 8 Self-Managed compo
...
 
-:::info
-The RDBMS backup path covers Zeebe, Operate, Tasklist, and Admin. Management Identity and Optimize are not covered by this path. Optimize always stores its data in Elasticsearch or OpenSearch, regardless of the Orchestration Cluster's secondary storage. If you use Optimize alongside an RDBMS-backed Orchestration Cluster, back up and restore Optimize independently using the [standalone Optimize backup procedure](./optimize-backup-and-restore.md). You do not need to switch the Orchestration Cluster to the Elasticsearch / OpenSearch backup path.
-:::
+<ZeebeGrid zeebe={storageCards} />
 
-:::warning
-Because Optimize and an RDBMS-backed Orchestration Cluster are backed up independently on separate schedules, they may be restored to **different points in time**. This means Optimize analytics data — dashboards, reports, and historical process data — may not be consistent with the restored state of the Orchestration Cluster.
+:::tip Disaster recovery context
```
leiicamundi: This belongs in a related-topics section rather than a tip.

The page is now a router: pick your secondary storage, then pick backup or restore. A tip about Cold Recovery sitting between the storage cards and the storage sections interrupts that choice with a topic the reader has not reached yet. It is a pointer to adjacent reading, not advice on the step in front of them.

Moving it to a short related-links block at the end of the page keeps the routing path clean and still surfaces it for anyone who needs it.

docs/self-managed/operational-guides/backup-restore/backup-and-restore.md
```diff
@@ -79,18 +49,26 @@ Using an RDBMS as secondary storage unlocks three new capabilities not available
...
-When using the RDBMS path, backup IDs are auto-generated by the cluster and do not need to be managed manually.
+:::warning
+Because Optimize and an RDBMS-backed Orchestration Cluster are backed up independently on separate schedules, they may be restored to **different points in time**. This means Optimize analytics data — dashboards, reports, and historical process data — may not be consistent with the restored state of the Orchestration Cluster.
 :::
 
+### Backup IDs
+
+When using Elasticsearch or OpenSearch as secondary storage, each component backup is identified by a user-supplied integer backup ID. The backup ID must be greater than the ID of any previous backup. Using Unix timestamps as backup IDs is a common practice to ensure uniqueness and chronological order. When using RDBMS as secondary storage, backup IDs are auto-generated by the cluster and do not need to be managed manually.
+
 The steps outlined on this page are generally applicable for any kind of deployment but might differ slightly depending on your setup.
```
leiicamundi: This sentence no longer matches the page. It says "the steps outlined on this page", but after the restructure this page has no steps. It routes to the pages that do.

It also sits inside `### Backup IDs`, where it is unrelated to the backup ID rules above it.

Either delete it or move it to the pages that carry the procedures.

On the coherence question I had marked here: the RDBMS backup ID behaviour was not lost, it was folded into line 68, and that reads correctly now. The only leftover is this sentence.

docs/self-managed/operational-guides/backup-restore/backup-and-restore.md
```diff
@@ -185,45 +163,7 @@ You can see further examples from Camunda consultants in the [Backup and Restore
...
-ORCHESTRATION_CLUSTER_MANAGEMENT_API=http://localhost:9600
-
-curl $ORCHESTRATION_CLUSTER_MANAGEMENT_API/actuator/health
-```
-
-</summary>
-</details>
-
-## Back up a cluster with multiple Physical Tenants
+### Multiple Physical Tenants
```
leiicamundi: Blocking: renaming this heading changes its anchor from `#back-up-a-cluster-with-multiple-physical-tenants` to `#multiple-physical-tenants`, and six links across five files still use the old one:

- `docs/components/admin/cluster-admin.md:39`
- `docs/self-managed/concepts/physical-tenants/index.md:76`
- `docs/self-managed/concepts/physical-tenants/storage-isolation.md:634`
- `docs/self-managed/concepts/physical-tenants/storage-isolation.md:652`
- `docs/self-managed/concepts/physical-tenants/authorization-model.md:25`
- `docs/self-managed/operational-guides/backup-restore/elasticsearch/backup.md:124`

The first two are edited by this PR to drop the `in-process-restore` link, so the stale anchor is on a line the PR already touches.

The cheapest fix is to keep the anchor stable: `### Multiple Physical Tenants {#back-up-a-cluster-with-multiple-physical-tenants}`. That also preserves any external links to the published page. Otherwise all six callers need updating.

One thing that makes this easy to miss: `docusaurus.config.js:32` sets `onBrokenLinks: "throw"` but never sets `onBrokenAnchors`, which defaults to `warn` in Docusaurus 3. The build will pass with these links broken.

The heading also dropped from `##` to `###`, so it is now nested under `## Considerations`. Backing up a multi-tenant cluster is a procedure rather than a consideration, so `##` may be the better level.

docs/self-managed/operational-guides/backup-restore/webapps-backup.md
```diff
@@ -9,6 +9,10 @@ keywords: ["backup", "backups"]
 import Tabs from "@theme/Tabs";
 import TabItem from "@theme/TabItem";
 
+:::tip
```
leiicamundi: This tip is duplicated. `zeebe-backup-and-restore.md:9` has the same block with only the backup type and the link target changed:

- here: "manage history backups for Operate and Tasklist" pointing at `take-history-backup.api.mdx`
- there: "manage runtime backups" pointing at `take-runtime-backup.api.mdx`

Two copies is where this kind of note starts drifting, and the 8.8 breaking-changes warning right below it is already duplicated across both pages in the same way. Extracting a shared MDX partial that takes the backup type and link as props would cover all four instances.

One accuracy point while it is being touched: `api/camunda/v2/backups.yaml:307` notes that history backups are "Only available on clusters whose secondary storage is Elasticsearch or OpenSearch". That constraint is worth stating here, since this page does not otherwise say it.

docs/self-managed/operational-guides/backup-restore/rdbms/restore.md
```diff
@@ -2,6 +2,7 @@
 id: rdbms-restore
 title: "Restore a backup (RDBMS)"
 sidebar_label: "Restore a backup"
```
leiicamundi: Why disable the table of contents here?

This makes the two restore pages the longest pages in `docs/` without one. Across the whole `docs/` tree, only three non-generated pages set `hide_table_of_contents: true`, and two of them are added by this PR:

- `rdbms/restore.md`, 769 lines (this PR)
- `elasticsearch/restore.md`, 556 lines (this PR)
- `configuration-mapping.md`, 70 lines

The other 340 pages that set it are generated API reference pages, which have their own navigation.

I suspect the reason is that the numbered steps live inside `<TabItem>` blocks, so the table of contents would interleave the Restore API steps with the legacy ones. If that is it, the table of contents is the symptom and the tab nesting is the cause. During an incident this page is exactly where someone needs to jump straight to "trigger the restore" without scrolling through 700 lines.

Splitting the legacy procedure into its own page, the way `restore-snapshot.md` was split out, would let both pages keep a usable table of contents.

docs/self-managed/operational-guides/backup-restore/rdbms/restore.md
```diff
@@ -22,20 +23,396 @@ description: "Learn how to restore a Camunda 8 Self-Managed backup using a relat
...
+A Restore API recovery runs in four phases, driven by two API requests:
+
+1. **Entering recovery mode**: every broker deactivates its partitions and switches to a restricted partition manager. While the cluster is in recovery mode it processes no work, and only read-only operations and restore remain available.
+2. **Restoring secondary storage**: while the cluster is in recovery mode, restore the RDBMS to the intended point that the primary storage backup aligns to.
+3. **Restoring the partitions**: the cluster plans a single change that, for every broker and partition, first drops the local partition data and then restores that partition from the selected backups. The steps of that plan run one at a time across the cluster.
+4. **Returning to processing**: once every partition is restored, the same change switches all brokers back to `PROCESSING` and the partitions become active again.
+
+Both requests are non-blocking. Each is acknowledged as soon as the cluster accepts the change and returns the `changeId` of the cluster configuration change that carries it out.
+
+### Restore API prerequisites
```
leiicamundi: This is the first heading on the page and it is an H3, so the page jumps from the frontmatter title straight to H3. The style guide says "Do not skip heading levels. Within a page, sibling headings at the same depth should describe parallel concepts" (`howtos/technical-writing-styleguide.md:118`).

It also diverges from the Elasticsearch page, which starts at `## About restoring a backup` and keeps the legacy steps nested as H3 under `## Restore Application (Legacy)`. This page puts `## Overview` (415), `## Prerequisites` (422) and `## Step 1` (440) at H2 instead.

Since the stated goal is that someone familiar with one page can navigate the other, the two should use the same skeleton.

docs/self-managed/operational-guides/backup-restore/rdbms/restore.md
```diff
@@ -22,20 +23,396 @@ description: "Learn how to restore a Camunda 8 Self-Managed backup using a relat
...
+
+The response returns the ID of the cluster change and the operations it will apply. The plan contains one `ModeChangeOperation` and one `AwaitModeChangeOperation` per broker:
+
+<details>
+<summary>Example response</summary>
+
+```json
+{
+  "changeId": "7",
+  "plannedChanges": [
```
leiicamundi: Blocking: wrong response shape. `plannedChanges` here is a flat array of operations, but the spec groups operations by physical tenant.

`api/camunda/v2/cluster.yaml:361-382`:

```yaml
plannedChanges:
  description: The operations that will be applied to complete the change, grouped by the physical tenant they belong to.
  type: array
  items:
    $ref: "#/components/schemas/ClusterModeChangePlannedChange"
ClusterModeChangePlannedChange:
  type: object
  required:
    - physicalTenantId
    - operations
```

So it should be:

```json
{
  "changeId": "7",
  "plannedChanges": [
    {
      "physicalTenantId": "default",
      "operations": [
        { "operation": "ModeChangeOperation", "mode": "RECOVERING" },
        { "operation": "AwaitModeChangeOperation", "mode": "RECOVERING" }
      ]
    }
  ]
}
```

The `/restore` examples in this PR already nest it correctly (line 221 here, and `elasticsearch/restore.md:179`), so this is the only one out of step.

Separately, the Elasticsearch page shows no `/mode` response at all at its step 1. Since both pages document the same endpoint, they should either both show it or both link to the API reference.

docs/self-managed/operational-guides/backup-restore/rdbms/restore.md
```diff
@@ -387,3 +761,9 @@ See [back up and restore Optimize independently](../optimize-backup-and-restore.
 If you previously backed up Camunda Hub data, restore it using the same RDBMS restore tools.
 
 See [backup and restore Camunda Hub data](../modeler-backup-and-restore.md) for more details.
+
+## How RDBMS restore works
```
leiicamundi: "How RDBMS restore works" used to open this page and now closes it, after 760 lines of procedure.

It explains the exporter position and how Zeebe aligns primary and secondary storage, which is what tells an operator whether they are about to lose data. Someone following the procedure top to bottom reaches it only after the restore has run.

Moving it back near the top, or reducing it to a short paragraph up front that links down to the full explanation, would restore the ordering that was there before.

docs/self-managed/operational-guides/backup-restore/elasticsearch/restore.md
```diff
@@ -59,564 +56,304 @@ Based on this, we can look in the [matrix versioning of 8.8](https://helm.camund
...
-camunda_optimize_1748937221_8.8.0_part_1_of_2
-camunda_optimize_1748937221_8.8.0_part_2_of_2
-camunda_webapps_1748937221_8.8.0_part_1_of_5
-camunda_webapps_1748937221_8.8.0_part_2_of_5
-camunda_webapps_1748937221_8.8.0_part_3_of_5
-camunda_webapps_1748937221_8.8.0_part_4_of_5
-camunda_webapps_1748937221_8.8.0_part_5_of_5
-camunda_zeebe_records_backup_1748937221
-```
+### Step 1: Restore Elasticsearch/OpenSearch Snapshots
```
leiicamundi: Heading style. `howtos/technical-writing-styleguide.md:118` asks for sentence case and says "In Markdown, do not include a colon in your headers." This heading breaks both, and it is not alone:

- `elasticsearch/restore.md:352` `### Step 1: Restore Elasticsearch/OpenSearch Snapshots`
- `elasticsearch/restore.md:356` `### Step 2: Restore Zeebe Cluster`
- `elasticsearch/restore.md:507` `### Step 3: Start all Camunda 8 components`
- `rdbms/restore.md:440` `## Step 1: Restore Zeebe from its primary storage backup`
- `rdbms/restore.md:742` `## Step 2: Start all Camunda 8 components`

The Restore API sections on both pages already use the compliant form (`### 1. Switch the cluster into recovery mode`). Matching that would fix the colon, the capitalization and the inconsistency between the two halves of each page in one pass.

docs/self-managed/operational-guides/backup-restore/elasticsearch/restore.md
```diff
@@ -630,7 +367,7 @@ The following specific prerequisites are required when restoring the Zeebe Clust
 
 ### Restore Zeebe Cluster
```
leiicamundi: Duplicate anchor. This heading auto-generates the id `restore-zeebe-cluster`, which line 356 already claims explicitly via `{#restore-zeebe-cluster}`. Two elements end up with the same id, and which one an anchor link reaches is left to the browser.

Renaming this one, for example to `### Run the restore application`, resolves it and also describes the step better than repeating the parent heading.

The same page has `## Prerequisites` at line 23 and `### Prerequisites` at line 358. Docusaurus will suffix the second one to `#prerequisites-1`, so nothing breaks today, but a link written against either is easy to get wrong later.

docs/self-managed/operational-guides/backup-restore/elasticsearch/restore-snapshot.md
leiicamundi: Agreeing with @deepthidevaki here, with the structural part spelled out.

The page opens on `### Prerequisites` with no H2 above it, so it starts mid-hierarchy and reads as a fragment lifted out of another page. The style guide asks not to skip heading levels (`howtos/technical-writing-styleguide.md:118`).

It also has no link back to the procedure it belongs to. Arriving from search or the sidebar, there is nothing saying this is step 3 of the Elasticsearch restore.

A short intro above this heading would fix both: one sentence naming it as part of the restore, a link to `./restore.md`, and an H2 to open the body.

The sidebar placement reinforces the same problem: `sidebars.js:2232` lists it as a sibling of `es-backup` and `es-restore`, which presents a sub-step as a peer procedure.

docs/self-managed/operational-guides/backup-restore/elasticsearch/restore-snapshot.md
leiicamundi: Confirming @deepthidevaki's point with what it breaks upstream, since this is the most serious issue in the PR.

`elasticsearch/restore.md:157` routes the Restore API flow here: "With the cluster in recovery mode, restore the Elasticsearch/OpenSearch snapshots". But this page then tells the reader to start the full Camunda stack (lines 33 to 39) and to stop all components (this step).

Recovery mode requires the brokers to stay up until `POST /restore` completes. An operator following the link as written tears down the cluster they just put into recovery mode, then has no way to issue the restore call.

The two flows need different snapshot procedures:

- Legacy: stop everything, restore snapshots, start the restore application.
- Restore API: brokers stay up in recovery mode, restore snapshots only, no start or stop steps.

Splitting the start/stop steps into a legacy-only section and keeping the snapshot restore itself shared would let both flows link here safely. Whatever the shape, the Restore API tab needs to stop linking to a procedure that shuts the cluster down.

leiicamundi:
## Cross-repo verification

Recording what I checked outside this repo so it does not get redone. All of this came back clean, except one pre-existing inconsistency that is **not** caused by this PR.

### Helm chart, camunda/camunda-platform-helm

Checked the values used by the legacy restore snippets against `charts/camunda-platform-8.10/values.yaml` (chart `15.0.0-alpha1`, appVersion `8.10.x`):

| Value used in the docs | Chart | Result |
| --- | --- | --- |
| `orchestration.enabled` | `values.yaml:2481` | correct |
| `orchestration.env` | `values.yaml:2649` | correct |
| `orchestration.command` | `values.yaml:2661` | correct |
| `connectors.enabled` | `values.yaml:2229` | correct |
| `optimize.enabled` | `values.yaml:3131` | correct |
| `elasticsearch.enabled` | `values.yaml:3486` | correct |

The 8.10 chart has no backup or restore specific value, so driving the legacy restore through `orchestration.command` and `orchestration.env` is the right approach. The top-level key is `orchestration`, not the pre-8.8 `zeebe`, and both pages use it correctly.

`zeebe.restore.ignoreFilesInTarget` at `rdbms/restore.md:740` is an application property rather than a chart value, so the naming is fine there.

### Reference architecture, camunda/camunda-deployment-references

No links into the backup and restore doc tree, so nothing breaks on that side. The `infraex-misc` bot comment above lists `aws-ecs.md`, `aws-ecs-dual-region.md` and `operator-based-infrastructure.md`, but none of those are in this PR's diff.

### API specs

Every endpoint, query parameter and permission checked against `api/camunda/v2/backups.yaml`, `cluster.yaml` and `cluster-admin.yaml`. Only the `/mode` response example is wrong, covered inline on `rdbms/restore.md:89`. Specifically confirmed as correct: `GET /backups/runtime` with `prefix` (`backups.yaml:61,81`), `GET /backups/history` with `prefix` and `verbose` (`367,389,397`), `dryRun` on `/restore`, and `BACKUP:RESTORE`.

### Pre-existing, out of scope for this PR

Two things I found while checking the above. Neither is in this PR's diff, so please do not treat them as review items. Flagging them only because the restructure puts the affected pages next to each other, which is what made them visible.

The backup store environment variable differs between the two restore pages. `rdbms/restore.md` uses `CAMUNDA_DATA_PRIMARYSTORAGE_BACKUP_STORE`, which matches the broker reference at `docs/self-managed/components/orchestration-cluster/zeebe/configuration/broker.md:347`. `elasticsearch/restore.md:419` and `:522` use `CAMUNDA_DATA_BACKUP_STORE`. Both sets of lines are untouched by this PR. The repository-name property has three spellings across the docs as well, and `_secondary-storage.md:91` is already marked "Needs verification" against #9795, so this likely belongs in that issue.

Two anchors are already dangling on `main`: `#backup-process` from `elasticsearch-without-cluster-privileges.md:471`, and `#why-you-should-use-backup-and-restore` from `elasticsearch/backup.md:181`. Neither heading exists in `backup-and-restore.md` before or after this PR.

### Versioning

The bot flagged that changes landed only in `docs/`. That looks right to me: the Restore API is documented as 8.10 and later, and `in-process-restore.md` never existed under `versioned_docs/`. Worth a second opinion from a tech writer, since the deleted page was published under `/docs/next/`, which is also why the `.htaccess` redirect still matters.

christinaausley:
Thanks for taking this on! I am happy to help among all of the feedback, just let me know. My thoughts:

I like the "what's your secondary storage, and are you backing up or restoring?" entry point and the link cleanup is nice!

I built the branch locally and it looks good -- a few things to consider:

## Worth fixing before merge

**1. No redirect for the deleted page.** `in-process-restore.md` is removed but `static/.htaccess` isn't touched. Our convention is to add a redirect at the top of that file whenever a page is removed. Mitigating factor: this page only ever existed in `docs/`, never in a released version, so only `/docs/next/…/in-process-restore/` will 404, but it's a cheap addition.

**2. The three new icons are very heavy.** ... versus their existing siblings in the same folder. Could you resize and compress them to match?

**3. `restore-snapshot.md` renders its title twice.** The frontmatter `title: "Restore Elasticsearch/OpenSearch snapshot"` is followed immediately by `## Restore Elasticsearch/OpenSearch snapshots` so the page opens with an H1 and a near-identical H2 (also singular vs. plural). I'd drop the H2 and promote its subsections one level.

**4. Malformed `<details>` blocks 10 occurrences in the new page.** The pattern is:

```
<summary>Example Output</summary>
<summary>
```

The second, unclosed `<summary>` ends up wrapping the body content in the rendered HTML which isn't valid inside `<details>`. It still expands and collapses so it's cosmetic rather than broken, but since the new file copies it 10 times it'd be good to drop the stray tag. (The same pattern pre-exists in `backup.md` separate cleanup not yours to fix here.)

**5. Trailing space in a card title.** `_card-data.js:61` has `title: "Restore a backup "`.

**6. The RDBMS restore card contradicts the backup card next to it.** The backup card says "Optimize and Management Identity are not included," but the restore card says "Perform a restore of all components while using an RDBMS as secondary storage." The previous copy explicitly called out that Optimize isn't supported... worth keeping that caveat?

## Structure and consistency (non-blocking)

**7.** `elasticsearch/restore.md` uses two numbering conventions: `### 1. Switch the cluster into recovery mode` in the Restore API tab, and `### Step 1: …` in the legacy tab. The legacy tab also goes H3 to H2 (`## Restore Zeebe Cluster`) to H3 (`### Step 3: …`), so the step sequence straddles heading levels and there's no Step 2 at that level.

**8.** `## Overview` mid-page in `rdbms/restore.md` (legacy tab) restarts the outline at H2, and the style guide steers away from bare "Overview" headings.

**9.** Anchor collision in `elasticsearch/restore.md`: both `## Prerequisites` and `### Prerequisites` exist, so the second builds as `#prerequisites-1`. Nothing links to it today but it's fragile if someone adds a cross-reference later.

**10.** Same concept, different tab labels: `label="Legacy approach"` (ES) vs. `label="Restore Application (Legacy)"` (RDBMS).

**11.** Same for section naming: "Restore a cluster with multiple Physical Tenants" vs. "Multiple physical tenant restore" vs. "Multiple Physical Tenants". Worth settling on one form and casing.

**12.** Sentence case per the style guide: "Restore Zeebe Cluster", "Restore Application (Legacy)", "Available Backups of Zeebe Partitions", "Available Backups on Elasticsearch/OpenSearch", `<summary>Example Output</summary>`, and the "Additional Information" card title.

**13.** `### 1. Restore [Templates](https://www.elastic.co/…)` puts a link inside a heading, which makes for an awkward anchor.

**14.** `assets/icon-prereqs.png` is now unreferenced in the backup-restore card data.

**15.** Prose on the landing page:

- "The procedure is influenced depending on the selection of the cluster's secondary storage." to something like "The procedure depends on which secondary storage your cluster uses."
- "can continue service other physical tenants" to "continue serving".
- `### Backup IDs` now ends with "The steps outlined on this page are generally applicable…", a general statement stranded under a specific heading.
- The card reading "Additional Information" links to a section titled "Considerations"... aligning the two would help.

**16.** The landing page drops the `:::info` describing the 8.8 Orchestration Cluster composition, along with the note that Management Identity isn't covered. That caveat now survives only inside the RDBMS-specific box... was dropping it for the ES path intentional?

**17.** `### Choose a restore approach` currently sits under `## Considerations` between "Storage backends and Physical Tenants" and "ContextPath". Given it's the central decision this restructure is organized around, it might deserve a more prominent position?

## One bigger-picture structural note

Nothing below needs to block this PR... it's a larger conversation and the restructure is already a big step forward. But I think the current shape will get harder to maintain, so it's worth deciding now rather than after the next provider or approach lands.

### Two full procedures inside `<Tabs>` on one page

Both restore guides put the Restore API path and the legacy Restore Application path in tabs on a single page. That's what's driving most of the individual issues I flagged above...

It also has costs that aren't obvious in the diff:

- `hide_table_of_contents: true` had to be set on both pages, because a TOC interleaving two procedures would be incoherent. That leaves a 602-line page (ES) and an 825-line page (RDBMS) with no on-page navigation at all... rough for an operator working through a restore under pressure.
- Browser find-in-page doesn't reach the unselected tab, so someone searching for a legacy flag or error string on the Restore API tab silently gets nothing.
- Deep links are unreliable. Support can't confidently send "here's the step you're stuck on," because the anchor resolves but the correct tab isn't necessarily selected.

I'd suggest splitting each approach into its own page, with the storage-level `restore.md` acting as a short chooser that covers shared prerequisites and points to the two paths:

```
elasticsearch/
  restore.md                 # chooser + shared context
  restore-api.md             # 8.10+ path
  restore-application.md     # legacy path
  restore-snapshot.md        # shared sub-procedure (already extracted here ✔)
```

Each page then gets a working TOC, one coherent heading outline, no anchor collisions, and reliable deep links and when the legacy path is eventually retired it's one file deletion plus a redirect, rather than unpicking a tab from the middle of a large page.

Worth noting this is the same move you already made with `restore-snapshot.md`, so it'd be applying an existing decision in this PR consistently rather than a new direction.

### Substantial duplication between the ES and RDBMS restore pages

Identically-named sections now exist in both `elasticsearch/restore.md` and `rdbms/restore.md`. I diffed the bodies:

| Section | Similarity |
| :--- | ---: |
| Track the Restore API operation | 100% |
| Handle a failed Restore API operation | 99% |
| Data directory is not empty | 98% |
| Retry a Restore API operation externally | 96% |
| Restore API prerequisites | 94% |
| Multiple physical tenant restore | 93% |

That's roughly 10 KB of byte-identical or near-identical prose maintained in two places, and it will drift... a fix to the retry guidance in one file won't reach the other.

The rest genuinely differ (Prerequisites 15%, Switch into recovery mode 32%, Trigger the restore 64%), so this isn't an argument for merging the pages. It's an argument for extracting just the storage-agnostic Restore API sections into partials. The repo already has that convention with 69 `_`-prefixed partials, including a `partials/` directory pattern in the self-managed tree (`import Cluster from './partials/\_cluster.md'`), so it'd fit right in.

### The restore-approach decision is under-signposted

The landing page gives the storage decision real estate: cards, plus a full comparison table. The Restore API vs. Restore Application decision... arguably the more consequential fork for anyone on 8.10... gets three sentences under `## Considerations`, sandwiched between "Storage backends and Physical Tenants" and "ContextPath".

A short comparison table mirroring the storage one would carry a lot of weight here: downtime required, whether brokers restart, Physical Tenant support, minimum version, and when you'd pick each. That single table would also let the individual restore pages stop re-explaining the choice.

panagiotisgts:
@christinaausley  Thanks for reminding me the partials, had totally forgotten their existence. :+1:  I've now split the restore approaches into distinct pages with routing cards present on the header for `restore` on each database type to follow the _guided_ approach the other pages had.

PR #9938, opened 2026-09-15
docs(self-managed): fix literal backticks in External PostgreSQL sample values

reem-sab:
## Description

On the **Use external PostgreSQL** page, the "Connection details" prerequisites
block is a ` ```yaml ` fenced code block whose values are wrapped in Markdown
backticks:

    host: `db.example.com`
    port: `5432`
    username: `postgres`
    password: `examplePassword`

Inside a fenced code block, backticks are literal, so they render verbatim instead
of as inline code. This drops the backticks so the sample is clean, valid YAML and
consistent with the same page's later **Example usage** block, which already uses
these values without backticks (`host: "db.example.com"`, `port: 5432`, and so on).

Applied to the current default version (8.9) and Next (`/docs`).

## PR Checklist


reem-sab:
Verified against the same page. The sample values were wrapped in backticks inside a code block, where backticks show up literally instead of as formatting. Removing them makes the sample match the "Example usage" block further down the page, which already lists the same values without backticks. Parses as valid YAML and passes the repo's Prettier check.

PR #9942, opened 2026-09-15
Rewrite backup and restore procedures to use the REST API, with actuator as the backward-compatible alternative

### LINKED ISSUE #9728: Rewrite backup and restore procedures to use the REST API, with actuator as the backward-compatible alternative
## Context

Follow-up from #9662, agreed with @deepthidevaki in review.

The backup and restore guides currently document the actuator endpoints as the primary surface. Camunda 8.10 adds equivalent Orchestration Cluster REST API endpoints for backup, exporting control, and restore, at both tenant and cluster scope. These are the endpoints new users should be using.

#9662 added Physical Tenant scoping notes to the existing pages but deliberately did **not** switch the primary surface, because the Elasticsearch backup page alone is ~800 lines and applies to every 8.10 user, not just multi-tenant deployments. That change is behavior-affecting for all readers and needs its own PR.

## Decision from review

> We should rewrite the existing backup process with the new REST endpoints, and provide the actuator endpoints as an alternative and backward compatible option. A new customer should use the REST endpoints. An existing user can easily find the documentation with the actuator endpoints, and should be encouraged to migrate to the REST endpoint. The deprecation of the actuator endpoint is not decided, but it is likely that we move all similar endpoints to REST and deprecate actuator endpoints eventually.

## Deliverables

- Rewrite the primary backup procedure in `operational-guides/backup-restore/elasticsearch/backup.md` around the REST endpoints, keeping actuator as a clearly labeled backward-compatible alternative (tabs or a subsection, not a parallel page).
- Apply the same treatment to `backup-and-restore.md` and `in-process-restore.md`, so the REST flow is the documented default end to end.
- Add migration guidance for existing users on the actuator path, mapping each actuator call to its REST equivalent.
- Document exporting control (`/v2/exporting/pause`, `/v2/exporting/resume`) alongside backup, since a history backup requires `BACKUP:CREATE` and `EXPORTER:PAUSE` together.
- Cover both scopes throughout: tenant-scoped `/physical-tenants/{physicalTenantId}/v2/...` and cluster-wide `/cluster/v2/...`, including the optional `physicalTenantId` query parameter.
- Replace the literal endpoint paths used in #9662 with links to generated API reference pages once those exist (see dependency below).
- Do **not** state a deprecation timeline for the actuator endpoints. Deprecation is not decided.

## Dependency

Sequenced behind the generated cluster API reference page (#8886, item 2). That work is now unblocked — @deepthidevaki confirmed on #8886 that `cluster-api.yaml` is complete and covers the physical-tenant parameters, and that `https://github.com/camunda/camunda/tree/main/dist/src/main/resources/api/cluster` is the stable location for a sync workflow.

This issue can start before that lands, but endpoint references should link to generated pages rather than literal paths, so the final pass depends on it.

## Coordinate

Two `<!-- TODO(physical-tenants-day-2) -->` markers were left in `backup-and-restore.md` and `in-process-restore.md` for per-tenant backup-store configuration properties and backend-specific restore fallback steps. Reviewer: @houssain-barouni.

## Related

- Epic: #8886
- PR: #9662

christinaausley:
## Description

Resolves #9728. Rewrites the backup/restore documentation to make the Orchestration Cluster REST API the primary documented path, with the management (actuator) API as a clearly labeled backward-compatible alternative:

- `elasticsearch/backup.md`: every step with a REST equivalent (soft-pause exporting, webapps/history backup, Zeebe/runtime backup, resume) now shows REST as the default tab and actuator as the alternative tab. Optimize and native ES/OpenSearch snapshot calls are unchanged (no REST equivalent exists).
- `backup-and-restore.md`: new "REST API" section as the primary reference, with a migration table mapping each actuator call to its REST equivalent.
- `in-process-restore.md`: was already REST-first; only its Physical Tenant path documentation needed correcting.

Two corrections from the original issue text, verified against the actual generated API spec pages:
- The issue described a `/physical-tenants/{physicalTenantId}/v2/...` path prefix for backup/exporting/restore/mode endpoints. That pattern doesn't exist for these endpoints ... verified directly against `change-cluster-mode`, `change-cluster-mode-as-cluster-admin`, `get-cluster-topology`, and the backup/exporting specs. The real pattern: an unprefixed path resolves to your own tenant (via credentials), and `/cluster/v2/...` takes an optional `physicalTenantId` query parameter for cluster-admin calls. This also fixes a real bug already merged via #9662.
- The issue assumed a history backup requires `BACKUP:CREATE` and `EXPORTER:PAUSE` together. `take-history-backup`'s spec only requires `BACKUP:CREATE`. pausing exporting first is documented as a recommended practice (avoids log compaction), not an enforced pairing.

**Out of scope, flagged separately:** the same incorrect path pattern also appears in `physical-tenants/index.md`, `api-routing.md`, `authorization-model.md`, `multi-tenancy/physical-tenants.md`, and `cluster-admin.md`. Those describe different (older, core v2 API) endpoints where I couldn't confirm the same fix applies... evidence from `get-status`'s spec suggests path-prefixed tenant addressing may be real for that endpoint family. Left untouched pending a dedicated look, rather than risk propagating a different error into core pages.

Anytime.

## PR Checklist


docs/self-managed/operational-guides/backup-restore/elasticsearch/backup.md
```diff
@@ -215,42 +237,83 @@ During a hot backup, the Zeebe cluster remains fully operational:
...
 
-The `/actuator/backupRuntime` API then creates a consistent backup of each partition while processing continues. The “wait for backup to complete” steps in this guide only poll backup status and do not introduce any additional pause in processing beyond the initial soft export pause.
+The runtime backup step below then creates a consistent backup of each partition while processing continues. The “wait for backup to complete” steps in this guide only poll backup status and do not introduce any additional pause in processing beyond the initial soft export pause.
 
 ### 2. Start the web applications backup (Operate / Tasklist)
 
-This step uses the [web applications management backup API](/self-managed/operational-guides/backup-restore/webapps-backup.md).
+This is a history backup: it covers the exported data Operate and Tasklist read from secondary storage, and is only available when secondary storage is Elasticsearch or OpenSearch.
```
panagiotisgts: 🔧  Would rephrase this as, or something similar:

"This step initiates taking a coordinated snapshot on Elasticsearch/OpenSearch, where Operate and Tasklist historic data are present."

docs/self-managed/operational-guides/backup-restore/backup-and-restore.md
```diff
@@ -93,10 +93,47 @@ When using the RDBMS path, backup IDs are auto-generated by the cluster and do n
 
 The steps outlined on this page are generally applicable for any kind of deployment but might differ slightly depending on your setup.
 
+### REST API
+
+With Camunda 8.10, the Orchestration Cluster (Zeebe and the web applications) exposes backup, exporting control, and restore as [Orchestration Cluster REST API](/apis-tools/orchestration-cluster-api-rest/orchestration-cluster-api-rest-overview.md) endpoints under `/v2/...`. Use the REST API for new deployments: it's authenticated and authorized like the rest of the Orchestration Cluster API, works the same way whether you run one Physical Tenant or several, and is the surface Camunda expects to keep building on.
+
+| Operation               | Endpoint                                                                                                                        |
```
panagiotisgts: 💭  Not sure if we should list all API available calls here, we already do refer to the respective one in the respective Backup/Restore guide steps. We can include them in the beginning of the respective guide as a cooking recipe style guide.

panagiotisgts (commented):
Not really a fan of _"This is a x "_ or _"This step uses x"_ prefix in the steps, but maybe that's personal preference.

docs/self-managed/operational-guides/backup-restore/elasticsearch/backup.md
christinaausley: Good call, updated...dropped the "This is a X:" framing here and for the parallel line in step 8 (runtime backup), for consistency:

"This step initiates a coordinated snapshot on Elasticsearch or OpenSearch, where Operate and Tasklist historic data are present. It is available only when secondary storage is Elasticsearch or OpenSearch."

docs/self-managed/operational-guides/backup-restore/backup-and-restore.md
christinaausley: Fair point 👍 that table was duplicating what each step already links to. Removed it and kept the intro paragraph, the permissions summary, and the pointer to the backup guide, so the REST API section still explains the concepts without re-listing every endpoint. Let me know if you'd still like a lightweight "at a glance" list added to the top of the backup guide itself-- happy to add that separately if it's useful for someone jumping straight past the step-by-step?

christinaausley:
@panagiotisgts 

> Not really a fan of "This is a x " or "This step uses x" prefix in the steps, but maybe that's personal preference.

I dropped the two "This is a X:" framing sentences I introduced (steps 2 and 8, above). For "This step uses X" specifically: that phrasing predates this PR, it's the existing convention throughout this doc (Optimize's step 3, the original actuator steps, etc.), and I extended it into the new REST tabs to match the existing voice rather than introduce a second style.

Given you've flagged it as possibly personal preference, I'd rather not do a blanket rewrite of all 13 instances as part of this PR... that's a bigger voice change than "add REST endpoints" and would touch pre-existing content this PR isn't otherwise changing. Happy to pick it up as its own pass across the backup/restore doc family if the team wants to move away from that pattern generally. Marking this one non-blocking for now unless you feel strongly otherwise?

deepthidevaki:
Do you need my review here? @panagiotisgts Could you review it?

docs/self-managed/operational-guides/backup-restore/elasticsearch/backup.md
```diff
@@ -172,39 +176,57 @@ As noted in the [Management API](../backup-and-restore.md#management-api) sectio
 
 ### 1. Soft pause exporting in Zeebe
 
-This step uses the [management API](/self-managed/components/orchestration-cluster/zeebe/operations/management-api.md?exporting=softPause#exporting-api).
+This will continue exporting records, but not delete those records (log compaction) from Zeebe. This makes the backup a hot backup, as covered in the [why you should use backup and restore](../backup-and-restore.md#why-you-should-use-backup-and-restore). Pausing exporting is recommended before a backup to avoid log compaction removing data the backup still needs; it is not enforced by either API as a precondition of the backup call itself.
```
panagiotisgts: > pausing exporting is recommended before a backup to avoid log compaction removing data the backup still needs; it is not enforced by either API as a precondition of the backup call itself.

It's actually required not recommended to have state consistency. The enforcement part is correct tho 👍

docs/self-managed/operational-guides/backup-restore/elasticsearch/backup.md
```diff
@@ -104,6 +104,8 @@ As of Camunda 8.8, the `/actuator` endpoints for backups have been moved to `/ac
 
 To create a backup, complete the following [backup process](#back-up-process).
 
+Steps that changed with Camunda 8.10 show both APIs: the Orchestration Cluster REST API is the default tab and the recommended path for new deployments, and the management (actuator) API is available as a backward-compatible alternative for existing automation. See [REST API](../backup-and-restore.md#rest-api) for the concepts shared across steps, including how each management API call maps to its REST equivalent.
```
panagiotisgts: ❌  Better just add a note that by default the REST API should be used

docs/self-managed/operational-guides/backup-restore/elasticsearch/backup.md
```diff
@@ -172,39 +176,57 @@ As noted in the [Management API](../backup-and-restore.md#management-api) sectio
...
+      If the request fails, verify that all brokers are running and retry.
+      :::
+
+         <details>
+            <summary>Example output</summary>
+            <summary>
+
+            :::note
+            Yes, 204 is the expected result and indicates a successful soft pause.
+            :::
```
panagiotisgts: That's redundant I think, we already mention this above.

docs/self-managed/operational-guides/backup-restore/elasticsearch/backup.md
```diff
@@ -215,42 +237,83 @@ During a hot backup, the Zeebe cluster remains fully operational:
...
+         </details>
+
+      </TabItem>
+
+   </Tabs>
 
 ### 3. Start the Optimize backup
 
-This step uses the [Optimize management backup API](/self-managed/operational-guides/backup-restore/optimize-backup.md).
+Optimize is not covered by the Orchestration Cluster REST API. This step always uses the [Optimize management backup API](/self-managed/operational-guides/backup-restore/optimize-backup.md), regardless of which API you use for the other steps.
```
panagiotisgts:
```suggestion
Optimize is not covered by the Orchestration Cluster backup REST API. The [Optimize management backup API](/self-managed/operational-guides/backup-restore/optimize-backup.md) must be used, regardless of which API you use for the other steps.
```

docs/self-managed/operational-guides/backup-restore/elasticsearch/backup.md
```diff
@@ -274,75 +337,97 @@ curl -XPOST "$OPTIMIZE_MANAGEMENT_API/actuator/backups" \
...
 ### 4. Wait for the web applications backup to complete
 
-This step uses the [web applications management backup API](/self-managed/operational-guides/backup-restore/webapps-backup.md).
+   <Tabs groupId="backup-api">
+      <TabItem value="rest" label="REST API" default>
 
-```bash
-curl -s "$ORCHESTRATION_CLUSTER_MANAGEMENT_API/actuator/backupHistory/$BACKUP_ID"
-```
+      This step uses [check a history backup](/apis-tools/orchestration-cluster-api-rest/specifications/get-history-backup.api.mdx), which requires the `BACKUP:READ` permission.
```
panagiotisgts:
```suggestion
     [Query the historic backup](/apis-tools/orchestration-cluster-api-rest/specifications/get-history-backup.api.mdx). This requires the `BACKUP:READ` permission.
```

docs/self-managed/operational-guides/backup-restore/elasticsearch/backup.md
```diff
@@ -644,95 +729,170 @@ This remains relevant if you run Optimize, which still relies on the former expo
 
 ### 8. Create the Zeebe broker backup
 
-This step uses the [Zeebe management backup API](/self-managed/operational-guides/backup-restore/zeebe-backup-and-restore.md).
+This step creates a runtime backup, covering Zeebe's own primary storage independently of secondary storage.
```
panagiotisgts:
```suggestion
This step creates a runtime backup, covering Zeebe's own primary storage alone.
```

docs/self-managed/operational-guides/backup-restore/elasticsearch/backup.md
```diff
@@ -644,95 +729,170 @@ This remains relevant if you run Optimize, which still relies on the former expo
...
+
+   </Tabs>
 
 ### 9. Wait for the Zeebe backup to complete before proceeding
 
-This step uses the [Zeebe management backup API](/self-managed/operational-guides/backup-restore/zeebe-backup-and-restore.md).
+   <Tabs groupId="backup-api">
+      <TabItem value="rest" label="REST API" default>
+
+      This step uses [check a runtime backup](/apis-tools/orchestration-cluster-api-rest/specifications/get-runtime-backup.api.mdx), which requires the `BACKUP:READ` permission.
```
panagiotisgts:
```suggestion
   [Query the runtime backup](/apis-tools/orchestration-cluster-api-rest/specifications/get-runtime-backup.api.mdx). This requires the `BACKUP:READ` permission.
```

I think in general this way of phrasing this seems clearer. Not gonna suggest it for all references.

docs/self-managed/operational-guides/backup-restore/elasticsearch/backup.md
```diff
@@ -644,95 +729,170 @@ This remains relevant if you run Optimize, which still relies on the former expo
...
+         <details>
+            <summary>Example output</summary>
+            <summary>
 
-         :::note
-         Yes, 204 is the expected result and indicates a successful resume.
-         :::
+            :::note
+            Yes, 204 is the expected result and indicates a successful resume.
+            :::
```
panagiotisgts: Same duplication

docs/self-managed/operational-guides/backup-restore/backup-and-restore.md
```diff
@@ -93,10 +93,33 @@ When using the RDBMS path, backup IDs are auto-generated by the cluster and do n
 
 The steps outlined on this page are generally applicable for any kind of deployment but might differ slightly depending on your setup.
 
+### REST API
+
+With Camunda 8.10, the Orchestration Cluster (Zeebe and the web applications) exposes backup, exporting control, and restore as [Orchestration Cluster REST API](/apis-tools/orchestration-cluster-api-rest/orchestration-cluster-api-rest-overview.md) endpoints under `/v2/...`. Use the REST API for new deployments: it's authenticated and authorized like the rest of the Orchestration Cluster API, works the same way whether you run one Physical Tenant or several, and is the surface Camunda expects to keep building on. Each step of the [back up process](./elasticsearch/backup.md#back-up-process) links to the specific REST endpoint it uses.
```
panagiotisgts:
```suggestion
With Camunda 8.10, the Orchestration Cluster (Zeebe and the web applications) exposes backup, exporting control, and restore as [Orchestration Cluster REST API](/apis-tools/orchestration-cluster-api-rest/orchestration-cluster-api-rest-overview.md) endpoints under `/v2/...`, and is preferable over the Management API. Authentication and authorization is required, similar to the rest of the Orchestration Cluster API. Behaves the same way whether you run one Physical Tenant or several, and is the surface Camunda expects to keep building on. Each step of the [back up process](./elasticsearch/backup.md#back-up-process) links to the specific REST endpoint it uses.
```

docs/self-managed/operational-guides/backup-restore/in-process-restore.md
```diff
@@ -142,7 +142,7 @@ Requests that combine `backupIds` with `from` or `to`, that specify a time range
 
 ### 4. Track the restore
 
-While a restore is in flight, the restore status reports progress per broker and per partition:
+While a restore is in flight, [the restore status](/apis-tools/orchestration-cluster-api-rest/specifications/get-restore-status.api.mdx) reports progress per broker and per partition:
```
panagiotisgts:
```suggestion
While a restore is in flight, progress of the procedure can be queried from the [recovery API](/apis-tools/orchestration-cluster-api-rest/specifications/get-restore-status.api.mdx) :
```

panagiotisgts (dismissed):
Thanks 👍 Some further suggestions. Now based on [this thread](https://camunda.slack.com/archives/C0B0GFXTSQH/p1789474317517829), there might be some structural changes in the way the guides are layed out but the gist of it should remain the same. I'm taking a look at it and happy for your feedback as well.

christinaausley:
Resolved all @panagiotisgts thank you! Ready for final review.

giorgionaps (approved):
Approving 👌🏼

PR #9944, opened 2026-09-16
docs(hub): document the plain-text warning and secret reference rules for credential fields

crobbins215:
## Description

Documents the plain-text secret warning that [camunda/camunda-hub#28737](https://github.com/camunda/camunda-hub/pull/28737) added to credential fields, closing [camunda/camunda-hub#28718](https://github.com/camunda/camunda-hub/issues/28718). Before that change, a sensitive credential field accepted a raw password or API key with no feedback at all, and a cluster with no secrets produced a silent, empty suggestion list. The same PR also anchored secret reference matching to a token boundary, which changes which values count as a reference.

Both changes are user-visible and neither was documented, so this PR covers them on the two pages under `hub/organization/credentials/`.

### [Manage credentials](https://github.com/camunda/camunda-docs/blob/main/docs/components/hub/organization/credentials/index.md)

- New **Reference a secret from a credential field** section, expanded from the two sentences that were there before. It states that the reference is the whole value of the field, which characters a secret key can contain, and where a reference is recognized. A table shows how `camunda.secrets.AWS_SECRET_KEY`, `foo.camunda.secrets.AWS_SECRET_KEY`, `camunda.secrets.camunda.secrets.AWS_SECRET_KEY`, and a plain-text value are each read.
- New **Store sensitive values as secrets, not plain text** section. It leads with the rule the warning exists to enforce, then quotes both messages verbatim and covers the **Open Clusters** link that SaaS adds to each one but Self-Managed does not. Saving a plain-text value is still allowed, so the section says the value stays exposed until it is replaced with a reference, rather than presenting the warning as something to dismiss.
- The **Configure** step of the create wizard now points at that section and notes that selecting a sensitive field offers the secrets on the selected clusters.
- New entry under **Known limitations** for the whole-value check: a composite value such as `Bearer camunda.secrets.TOKEN` is flagged even though its reference resolves.

### [Configure credentials in the modeling interface](https://github.com/camunda/camunda-docs/blob/main/docs/components/hub/organization/credentials/modeling-interface.md)

The properties panel dialogs share the credential field component with the wizard, so they show the same warning. Two additions:

- The plain-text warning on a sensitive field, with the same guidance to replace the value with a reference, and a link to the reference syntax on the main credentials page rather than a second copy of it.
- What the token-boundary change means here: a value that embeds `camunda.secrets.` mid-word holds no reference, so it is reported as plain text rather than as a missing secret, and the plain-text warning replaces the missing secret warning on that field.

Credentials documentation exists only in `/docs`, so there is nothing to backport.

@camunda/tech-writers for a writing review.

## PR Checklist


crobbins215:
Hey @alexronquillo me again, I thought its worth making these changes based on https://github.com/camunda/camunda-hub/issues/28718. Able to review, tweak and merge if your happy?

docs/components/hub/organization/credentials/index.md
```diff
@@ -42,16 +42,47 @@ Credentials and [connector secrets](/components/hub/organization/manage-clusters
...
+
+| Field value                                      | How it is read                                                                                                      |
+| ------------------------------------------------ | ------------------------------------------------------------------------------------------------------------------- |
+| `camunda.secrets.AWS_SECRET_KEY`                 | A reference to the secret `AWS_SECRET_KEY`.                                                                         |
+| `foo.camunda.secrets.AWS_SECRET_KEY`             | Not a reference, because `camunda.secrets.` is in the middle of a word. The value is stored and sent as plain text. |
+| `camunda.secrets.camunda.secrets.AWS_SECRET_KEY` | A reference to a secret with the key `camunda`, because the key ends at the next period.                            |
+| `hunter2`                                        | Plain text. Camunda Hub warns you, and stores the value as you entered it.                                          |
```
alexronquillo: As an example, this feels unexpected. Should we choose something that's more obvious/realistic?

docs/components/hub/organization/credentials/index.md
```diff
@@ -42,16 +42,47 @@ Credentials and [connector secrets](/components/hub/organization/manage-clusters
...
+
+To reference a secret from a credential field, enter `camunda.secrets.` followed by the secret key as the whole value of the field. For example, reference a secret with the key `AWS_SECRET_KEY` as follows:
 
 ```
 camunda.secrets.AWS_SECRET_KEY
 ```
 
+A secret key can contain letters, digits, underscores, and hyphens. The key ends at the first character outside that set, so a key cannot contain a period.
+
+A reference is recognized at the start of a value, or directly after a character that is not a letter, digit, underscore, or period. Anywhere else, `camunda.secrets.` is part of a longer word:
```
alexronquillo, line 55: We show the following examples:

- Reference recognized at the start of a value
- Plain text recognized by being part of a longer word (includes a prefix)
- Reference recognized at the start of a value (again, but with different key implications)
- Plain text recognized by a lack of the `camunda.secrets.` prefix

This lead-in sentence suggests another way to recognize references, namely directly after a character that is not a letter, digit, underscore, or period.

Because of this, I think we should include a fifth example, something like:

```
(camunda.secrets.AWS_SECRET_KEY)
```

The leading parenthesis is not a letter, digit, underscore, or period, therefore, `camunda.secrets.AWS_SECRET_KEY` is recognized as a reference. The closing parenthesis is also not included in the key because it's not allowed.

docs/components/hub/organization/credentials/index.md
```diff
@@ -42,16 +42,47 @@ Credentials and [connector secrets](/components/hub/organization/manage-clusters
...
 Credential fields use `camunda.secrets.MY_API_KEY`, without braces. This is not the same as the `{{secrets.MY_API_KEY}}` syntax you use in a [connector field](/components/connectors/use-connectors/index.md#using-secrets). Use `camunda.secrets.` inside a credential, and `{{secrets.}}` in connector fields that support secrets.
 :::
 
+### Store sensitive values as secrets, not plain text
+
+Store every sensitive value, such as a password or API key, as a connector secret, and reference it from the credential field. A credential field accepts any text you type, so a value entered directly is stored as you typed it, outside the secrets vault.
+
+To guide you to a secret, Camunda Hub highlights a sensitive field and warns you when its value is not a secret reference:
+
+> You're storing this as plain text. Create a cluster secret and reference it as camunda.secrets.NAME to avoid exposing sensitive information.
```
alexronquillo: Is "cluster secret" the correct term here?

docs/components/hub/organization/credentials/index.md
```diff
@@ -42,16 +42,47 @@ Credentials and [connector secrets](/components/hub/organization/manage-clusters
...
+
+Saving is still allowed, so the value stays exposed until you replace it with a reference. The warning clears as soon as the field references a secret.
+
+If the clusters you selected hold no secrets yet, the field says so instead of showing an empty suggestion list:
+
+> No cluster secrets found. Create one, then reference it here.
+
+In Camunda 8 SaaS, both messages carry an **Open Clusters** link that opens in a new tab, so your part-finished credential survives the detour. Self-Managed shows no link, because connector secrets come from the connector runtime configuration.
```
alexronquillo: It feels like we're mixing "cluster secret", "credential", and "connector secret" in this section

docs/components/hub/organization/credentials/index.md
```diff
@@ -89,7 +120,9 @@ To create a credential, select **Create credential**, and complete the three ste
 
 2. **Configure**: name the credential, choose which clusters it applies to, and fill in the fields for the credential type you selected.
 
-   Camunda suggests an ID for the credential based on the name you enter. You can change the ID while you are creating the credential, but not afterwards. For a sensitive field, enter a reference to an existing secret, such as `camunda.secrets.AWS_SECRET_KEY`, rather than the value itself.
+   Camunda suggests an ID for the credential based on the name you enter. You can change the ID while you are creating the credential, but not afterwards. For a sensitive field, enter a reference to an existing secret, such as `camunda.secrets.AWS_SECRET_KEY`, rather than the value itself. Select the field to pick from the secrets that exist on the clusters you selected.
+
+   Camunda Hub highlights a sensitive field and warns you when its value is not a secret reference. **Continue** stays enabled, so replace the value with a reference before you save. See [Store sensitive values as secrets, not plain text](#store-sensitive-values-as-secrets-not-plain-text).
```
alexronquillo:
```suggestion
   Camunda Hub highlights a sensitive field and warns you when its value is not a secret reference. **Continue** stays enabled, so replace the value with a reference before you save. See [store sensitive values as secrets, not plain text](#store-sensitive-values-as-secrets-not-plain-text).
```

docs/components/hub/organization/credentials/modeling-interface.md
```diff
@@ -42,8 +42,12 @@ To create a credential from the properties panel:
 4. Fill in the fields for this credential type. For a sensitive field, enter a reference to a secret that already exists on the cluster, using `camunda.secrets.` followed by the secret key, such as `camunda.secrets.AWS_SECRET_KEY`.
 5. Save the credential. Camunda Hub creates it on the cluster and selects it on the connector task.
 
+Camunda Hub highlights a sensitive field and warns you when its value is not a secret reference. Saving is still allowed, so replace the value with a reference to keep the sensitive value in the secrets vault. For the reference syntax, see [Reference a secret from a credential field](./index.md#reference-a-secret-from-a-credential-field).
```
alexronquillo:
```suggestion
Camunda Hub highlights a sensitive field and warns you when its value is not a secret reference. Saving is still allowed, so replace the value with a reference to keep the sensitive value in the secrets vault. For the reference syntax, see [reference a secret from a credential field](./index.md#reference-a-secret-from-a-credential-field).
```

alexronquillo (changes requested):
Thanks! I left some comments. Once those are addressed, I'll do one more quick line edit pass 👍

docs/components/hub/organization/credentials/index.md
crobbins215: Agreed — swapped for `wJalrXUtnFEMI/K7MDENG/bPxRfiCYEXAMPLEKEY` in 796ccbe32. The row only works if the reader recognises the value as something they'd paste into a secret key field, and `hunter2` doesn't carry that.

docs/components/hub/organization/credentials/index.md
crobbins215: Good catch, thanks — the lead-in stated a rule the table never demonstrated. Added `(camunda.secrets.AWS_SECRET_KEY)` in 796ccbe32, with your reasoning as the explanation: the reference is recognised after the opening parenthesis, and the key ends at the closing one, because neither character can be part of a key.

Checked it against the matcher it documents, `(?:^|[^A-Za-z0-9_.])(camunda\.secrets\.[A-Za-z0-9_-]+)` in `missing-secret.ts`.

docs/components/hub/organization/credentials/index.md
crobbins215: Fair — the field's wording shouldn't decide the page's. Both product strings are gone in 89006db0d; the page now describes the two nudges in its own terms and says secret throughout. Nothing on the page says cluster secret any more.

docs/components/hub/organization/credentials/index.md
crobbins215: Fixed in 796ccbe32, and 89006db0d finished it. The prose now uses one term, secret. A credential is the configuration that uses secrets, and connectors can use secrets too, so qualifying them as connector or cluster secrets in this section blurred all three.

docs/components/hub/organization/credentials/index.md
crobbins215: Applied as written in 796ccbe32.

docs/components/hub/organization/credentials/modeling-interface.md
crobbins215: Applied as written in 796ccbe32.

crobbins215:
Intentional — the hub credentials pages exist only in `/docs`, so there's no `versioned_docs/version-8.9` copy to mirror into.

crobbins215:
@alexronquillo — all six are addressed and the threads are resolved, so it's ready for your line edit pass whenever you have a moment.

**Fixed**

- `hunter2` read as nothing a reader would actually paste into a secret key field; the plain-text row now uses `wJalrXUtnFEMI/K7MDENG/bPxRfiCYEXAMPLEKEY`. (796ccbe32)
- The lead-in stated that a reference is recognised after a non-name character, and no example showed it. Added `(camunda.secrets.AWS_SECRET_KEY)` with your reasoning, checked against the matcher it documents. (796ccbe32)
- Terminology: the prose now says secret throughout, and the two product strings are no longer quoted, so cluster secret has left the page entirely. A credential is the configuration that uses secrets, and connectors can use secrets too, so qualifying them in that section blurred all three. (796ccbe32, 89006db0d)
- Both link-text suggestions taken as written. (796ccbe32)

**Left alone**

`connector secret` outside the section you flagged — the comparison table near the top, and the link text on `modeling-interface.md`. The page they point at is titled **Manage connector secrets**, so renaming inside the credentials docs alone would just move the mismatch. Happy to take the whole subtree onto one term as a follow-up if you'd rather it were consistent.

alexronquillo (approved):

PR #9953, opened 2026-09-16
docs(operate): clarify incident retry does not re-verify underlying resolution

### LINKED ISSUE #63192: Operate: incident Retry (single and bulk) reports success without re-checking underlying resolution
### 

### 

### 

### 

### 

### Description

Clicking **Retry Incident** in Operate — from either the per-incident button or the Processes-list bulk **Retry** — clears the incident and reports success without any check that the underlying problem was actually resolved. The incident only returns once something polls for the job again; with no worker running it never returns, so a user who has fixed nothing sees a healthy instance.

The bulk path makes this worse: a batch operation reporting "12 of 12 completed" reads as a stronger success guarantee than the single-incident button, and it is equally unfounded.

Split out from parent issue #62727 (problem 2 of 2, covering both parts 2 and 3).

### Steps to reproduce

Tested on C8Run 8.10.0-alpha5 with a file secret store configured, using process A from #62727 (two parallel service tasks whose input mappings reference a secret that does not exist).

**Single-incident retry**

1. Start one instance of process A. Do **not** start a worker.
2. Send one job activation request for process A's job type (it returns zero jobs). Reload the instance — both incidents are now present.
3. Click **Retry Incident** on both. Do **not** create the secrets.
4. Reload again.

**Bulk retry**

5. Start several more instances of process A and send one activation request so they all raise incidents.
6. On the Processes page, filter to Incidents, select all rows, and use the toolbar **Retry**.
7. Open the resulting batch operation from Operations and check the instances afterwards.

### Current behavior

**Single-incident retry.** After step 3, both incident rows are `RESOLVED` and `hasIncident` is `false`, while the secrets are still absent from the store. The UI shows a healthy instance.

**Bulk retry.** With 6 instances (12 incidents) selected, the batch operation "Resolve Incident" reported **Batch state: Completed** and all **12 items Completed**, one row per incident with its process instance key and incident key. Immediately afterwards `hasIncident` was `false` on all 6 instances, and the secrets were still missing on disk. One subsequent activation request brought all 6 incidents back.

The retry path is [`useResolveIncident.ts`](https://github.com/camunda/camunda/blob/a62ec903fc80fe8bfa7f53381d78fbf340515e0c/operate/client/src/modules/mutations/incidents/useResolveIncident.ts#L40-L49). It calls `updateJob(jobKey, {retries: 1})`, then `resolveIncident(incidentKey)`, then polls `GET /v2/incidents/{key}` until the state is no longer `ACTIVE`. That condition is satisfied as soon as the incident record is resolved, which says nothing about whether the secret can now be resolved.

### Expected behavior

After a retry (single or bulk), the UI should reflect whether the underlying problem is still present rather than reporting success as soon as the incident record is marked resolved. Re-raising the incident promptly, or not clearing it until resolution has actually been re-attempted, would both avoid the false signal.

Whatever is done for the single-incident button should apply to the batch-operation result as well, since "N of N completed" reads as a much firmer guarantee.

### Workaround

Keep a worker connected for the job type. A retry then re-raises the incident promptly if the secret is still missing.

### Version

- Component Version: 8.10.0-alpha5

### Rootcause

`useResolveIncident.ts` treats "incident record no longer `ACTIVE`" as success. Nothing observes whether the follow-up job activation and secret resolution actually succeeded. The batch bulk-retry uses the same success criterion at the operation level and inherits the same false signal.

Separately, resolution is requested from the job activation path, so with no activation attempt there is no resolution attempt and no incident. If that lazy-resolution behaviour is intended it needs documenting, since absence of incidents does not imply absence of secret problems when no worker is connected.

### Solution ideas

- After a retry, wait for the follow-up resolution attempt before reporting success, or surface that the retry is pending rather than complete.
- Apply the same criterion to bulk retry / batch operation completion so "N of N completed" only appears when resolution was actually re-attempted.
- If the lazy-resolution behaviour is intended, document it.

### Links

- Parent: #62727
- Epic: #56555
- Related: #62711 (incident message names the wrong cause), #61355 (Operate UX for secret resolution)

ajanoni-camunda:
## Description

Clarifies that resolving or retrying an incident in Operate (single **Retry**, per-instance retry, and bulk/batch **Retry**) marks the incident record as resolved and triggers a retry, but does not re-verify that the underlying cause was fixed. For job incidents, the cause is only re-checked when a worker next activates the job. If no worker is connected for the job's type, the incident does not reappear even when the problem persists, so the instance can look healthy without being fixed.

This addresses the documentation gap called out in camunda/camunda#63192 ("If the lazy-resolution behaviour is intended it needs documenting, since absence of incidents does not imply absence of secret problems when no worker is connected").

Changes (Next / 8.10, `docs/` only — 8.10 is not yet a `versioned_docs` release):

- `docs/components/concepts/incidents.md`: note that a job incident is re-checked only on the next activation; with no worker connected a new incident is not raised even when the problem persists.
- `docs/components/operate/userguide/resolve-incidents-update-variables.md`: note that **Retry** does not re-verify the fix; keep a worker connected.
- `docs/components/operate/userguide/selections-operations.md`: note that a batch retry's "N of N completed" confirms the retries ran, not that the causes are gone.
- `docs/components/concepts/secret-resolution-incidents.md`: state that absence of incidents does not imply absence of secret problems when no worker is connected, plus the keep-a-worker-connected workaround.

Closes: camunda/camunda#63192

## PR Checklist


docs/components/concepts/secret-resolution-incidents.md
```diff
@@ -51,6 +51,8 @@ Resolve the incident only after fixing the underlying cause. Resolving the incid
 
 You don't need to redeploy or make client-side changes. Once the reference resolves successfully, the process instance continues from where it stopped.
 
+Resolution is only re-attempted when the job is next activated. If no worker is connected for the job's type, the broker does not activate the job, so it does not request resolution again and the incident does not reappear even when the secret is still missing. Absence of incidents does not imply absence of secret problems when no worker is connected. Keep a worker connected so the incident re-raises promptly if the cause is not yet fixed.
```
HeleneW-dot: :warning: Does this ` If no worker is connected for the job's type, the broker does not activate the job..` point also apply to the other retry scenarios? ie `### Retry secret injection` and `### Retry after reducing the size`. If so, lets add this part there too

HeleneW-dot (dismissed):
Nothing blocking, just a question re. whether this notice should also be applied to the other retry section 👍

giorgionaps (dismissed):
Committed wording edits directly to the branch, lgtm!

docs/components/concepts/secret-resolution-incidents.md
ajanoni-camunda: 👍 Done

giorgionaps (approved):
lgtm

PR #9957, opened 2026-09-17
docs(self-managed): document OIDC startup resilience and the redirect URI constraint

megglos:
## Description

Camunda 8.10 changes two things about OIDC in the Orchestration Cluster. This PR documents each one where its reader meets it. All prose follows ASD-STE100.

**The cluster starts when an identity provider is unreachable.** It contacts a provider at the first request that needs it, and not at startup. Only the requests that need that provider fail, and they recover without a restart. A per-minute warning replaces the failed startup as the signal to watch.

- [Debugging the authentication flow](docs/self-managed/components/orchestration-cluster/admin/debugging-authentication.md) — the behavior, the warning, and what to check. This is where an operator lands when the cluster is up but authentication fails.
- [Connect to an identity provider](docs/self-managed/components/orchestration-cluster/admin/connect-external-identity-provider.md) — a note at the restart step that a successful start no longer proves the IdP is reachable.
- 8.10 release notes, and a `Change` announcement, because log-based alerting has to take over from the failed startup.

**An incomplete OIDC redirect URI fails the cluster at startup.** The rules themselves are not new: a value without a `/sso-callback` path could not complete a browser login on 8.9 either, because the IdP sent the browser to a path the cluster does not serve. 8.10 reports such a value at startup instead of at the next login. That matters for one case — a cluster that only serves API clients never uses its redirect URI, so an incorrect value can stay unknown on 8.9 and stop the start on 8.10.

- A `Breaking change` announcement with the action to take before upgrading.
- [8.9 to 8.10 upgrade guide](docs/self-managed/upgrade/components/890-to-8100.md) — a new Orchestration Cluster section with the rules and the accepted and rejected values.
- The redirect URI section of the IdP setup guide, and the `redirect-uri` property rows, as the permanent description.

Implemented in the Camunda security library (camunda/camunda-security-library#656, camunda/camunda-security-library#672, released in 1.0.2), adopted by camunda/camunda#62935, and tracked by camunda/camunda#61967.

## PR Checklist


megglos:
note, this is WIP and I'm not sure about the current place in the docs, need to revise and potentially find a better fit

docs/reference/announcements-release-notes/8100/8100-announcements.md
```diff
@@ -749,6 +749,42 @@ Camunda Hub and Optimize accept their existing authentication settings in 8.10 a
...
 
+<div className="release-announcement-row">
+<div className="release-announcement-badge">
+<span className="badge badge--breaking-change">Breaking change</span>
+</div>
+<div className="release-announcement-content">
+
+#### OIDC redirect URI must include a callback path
+
+The Orchestration Cluster validates the OIDC redirect URI at startup. It does not start if the value has no `/sso-callback` path. This rule is not new: such a value also could not complete a browser login on 8.9. 8.10 reports the value at startup, and not at the next login. A cluster that only serves API clients never uses its redirect URI, so an incorrect value can stay unknown until the upgrade. The default value `{baseUrl}/sso-callback` is correct, and so are Helm deployments that set `orchestration.security.authentication.oidc.redirectUrl`.
```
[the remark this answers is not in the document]
megglos: Adopted, with a correction to the correction. The 1.0.2 validator indeed accepts any servable callback path, so the text no longer attributes `/sso-callback` to the validation. But `/sso-callback` is still the only path an Orchestration Cluster serves the callback on: the webapp chain's `securityMatcher` is built from `SecurityPathPort#webappPaths()` alone (`ScopedWebappSecurityChainBuilder`, the `securityMatcher(matchers)` call), and the OC's implementation hardcodes `/sso-callback/**`. A custom path therefore passes startup and then never receives the callback.

b1c70e3cee states the validation rules, and names the served path separately.

docs/self-managed/components/orchestration-cluster/admin/connect-external-identity-provider.md
```diff
@@ -127,6 +127,15 @@ You may need to customize the redirect URI in advanced scenarios, such as:
 
 Regardless of customization, the redirect URI must always point to the `/sso-callback` endpoint of your Orchestration Cluster deployment.
 
+The Orchestration Cluster validates the redirect URI at startup. It does not start if the value is not a complete callback URL. Set a value that:
+
+- starts with `{baseUrl}`, or has an `http` or `https` scheme and a host,
+- has a port between 1 and 65535, if it has a port,
+- ends with `/sso-callback`,
+- has no fragment (`#`).
```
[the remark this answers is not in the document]
megglos: Adopted the same way as the announcement: the bullet now reads "has a callback path", and a following sentence says the cluster does not serve a path other than `/sso-callback`, so the login does not complete. `/login/oauth2/code/{registrationId}` passes validation but is not routed to the webapp chain in this product.

docs/self-managed/upgrade/components/890-to-8100.md
```diff
@@ -359,6 +359,34 @@ Next, move the files from the "General" process application to the new one:
...
+### OIDC redirect URI must include a callback path
+
+Camunda 8.10 validates the OIDC redirect URI at startup. The Orchestration Cluster does not start if the value is not a complete callback URL. This applies to `camunda.security.authentication.oidc.redirect-uri`, and to `camunda.security.authentication.providers.oidc.<provider-id>.redirect-uri`.
+
+A correct value:
+
+- starts with `{baseUrl}`, or has an `http` or `https` scheme and a host,
+- has a port between 1 and 65535, if it has a port,
+- ends with `/sso-callback`,
+- has no fragment (`#`).
```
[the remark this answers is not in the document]
megglos: Adopted. The rules now describe the validator, and the accepted/rejected table gains a row for `https://camunda.example.com/callback` — "Accepted, but the login does not complete" — which is the distinction that was missing. Templated hosts are covered by "starts with `{baseUrl}`, or has an `http` or `https` scheme and a host"; I kept that phrasing rather than listing `{baseScheme}://{baseHost}`, because an upgrader checking one value does not need the full placeholder set.

docs/self-managed/components/orchestration-cluster/admin/debugging-authentication.md
```diff
@@ -77,6 +78,21 @@ LOGGING_LEVEL_IO_CAMUNDA_SECURITY=DEBUG
...
+While a provider is unreachable:
+
+- Those requests fail with a server error, and not with an authentication error. This includes a browser login and an API request with a token from that provider.
+- All other requests succeed.
+- Each new request tries again. The cluster serves the failed traffic again when the provider answers. You do not need to restart the cluster.
+- The cluster logs one warning each minute for each step that fails: a client registration, a token decoder, or a UserInfo mapping. The warning gives the provider, its issuer, and the scope.
+
+An unreachable provider no longer stops the cluster from starting. This warning is your only signal that part of the authentication traffic fails. Monitor it in your log pipeline.
+
+If you see the warning, make sure that the discovery endpoint `<issuer-uri>/.well-known/openid-configuration` is reachable from the cluster. See [Test the IdP directly](#test-the-idp-directly).
```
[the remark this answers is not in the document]
megglos: Adopted. The step now points at the endpoint the warning names, and says that this is the discovery endpoint only for a provider configured with an issuer URI, and the configured `jwk-set-uri` or `user-info-uri` otherwise.

docs/reference/announcements-release-notes/8100/8100-announcements.md
```diff
@@ -749,6 +749,42 @@ Camunda Hub and Optimize accept their existing authentication settings in 8.10 a
 </div>
 </div>
 
+<div className="release-announcement-row">
+<div className="release-announcement-badge">
+<span className="badge badge--breaking-change">Breaking change</span>
+</div>
+<div className="release-announcement-content">
+
+#### OIDC redirect URI must include a callback path
```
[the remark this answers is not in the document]
megglos: Declining, on the split the cited convention itself draws: the announcements file takes "breaking changes, deprecations ... anything that requires action before upgrading", and the release-notes file takes "every notable feature, enhancement, and fix". A tightened validation that can stop a cluster from starting is the first, not the second, and it is covered twice already — a `Breaking change` announcement and a section in the 8.9 to 8.10 upgrade guide. A release-notes entry would repeat that under a heading upgraders do not read for required actions.

The startup-resilience change is in the release notes because it is a behavioral improvement that shipped, and it needs no action.

docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/_security.md
```diff
@@ -105,7 +105,7 @@ script-src-attr 'none'.
...
-| `camunda.security.authentication.oidc.redirect-uri`                                          | The URI for redirects from the OIDC provider to the Orchestration Cluster after user login.                                                                                                                                                                                                                                                                                     | `http://localhost:8080/sso-callback` |
+| `camunda.security.authentication.oidc.redirect-uri`                                          | The URI for redirects from the OIDC provider to the Orchestration Cluster after user login. The value is validated at startup and must end with `/sso-callback`. See [redirect URI](/self-managed/components/orchestration-cluster/admin/connect-external-identity-provider.md#redirect-uri).                                                                                   | `http://localhost:8080/sso-callback` |
```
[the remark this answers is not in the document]
megglos: Same as the thread on this file's line 135 — adopted. Both rows now read "must include a callback path", followed by the path the cluster serves.

docs/self-managed/components/orchestration-cluster/admin/debugging-authentication.md
```diff
@@ -77,6 +78,21 @@ LOGGING_LEVEL_IO_CAMUNDA_SECURITY=DEBUG
...
+While a provider is unreachable:
+
+- Those requests fail with a server error, and not with an authentication error. This includes a browser login and an API request with a token from that provider.
+- All other requests succeed.
+- Each new request tries again. The cluster serves the failed traffic again when the provider answers. You do not need to restart the cluster.
+- The cluster logs one warning each minute for each step that fails: a client registration, a token decoder, or a UserInfo mapping. The warning gives the provider, its issuer, and the scope.
+
+An unreachable provider no longer stops the cluster from starting. This warning is your only signal that part of the authentication traffic fails. Monitor it in your log pipeline.
+
+The warning names the endpoint that did not answer. Make sure that this endpoint is reachable from the cluster. This is the discovery endpoint `<issuer-uri>/.well-known/openid-configuration` for a provider that is configured with an issuer URI. It is the configured `jwk-set-uri` or `user-info-uri` for a provider that sets these endpoints. See [Test the IdP directly](#test-the-idp-directly).
```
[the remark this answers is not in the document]
megglos: Adopted — you are right on both counts. `DeferredOidcResolution.report` warns with `"Failed to resolve {}. …"` and the subject is built by `describeProvider` (`'camunda' (issuer https://idp/realms/camunda)`) plus the scope; the endpoint only shows up in the attached exception. The section now gives the alertable signature — `WARN` from `io.camunda.security.spring.oidc.DeferredOidcResolution`, message starting `Failed to resolve` — states what the subject does and does not contain, and sends the reader to the exception for the endpoint before testing it.

docs/self-managed/components/orchestration-cluster/core-settings/configuration/partials/_security.md
```diff
@@ -105,7 +105,7 @@ script-src-attr 'none'.
...
-| `camunda.security.authentication.oidc.redirect-uri`                                          | The URI for redirects from the OIDC provider to the Orchestration Cluster after user login.                                                                                                                                                                                                                                                                                     | `http://localhost:8080/sso-callback` | Yes                             |
+| `camunda.security.authentication.oidc.redirect-uri`                                          | The URI for redirects from the OIDC provider to the Orchestration Cluster after user login. The value is validated at startup and must include a callback path. The cluster serves the callback at `/sso-callback`. See [redirect URI](/self-managed/components/orchestration-cluster/admin/connect-external-identity-provider.md#redirect-uri).                                | `http://localhost:8080/sso-callback` | Yes                             |
```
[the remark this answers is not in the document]
megglos: Adopted. Both environment-variable rows (`CAMUNDA_SECURITY_AUTHENTICATION_OIDC_REDIRECTURI`, `CAMUNDA_SECURITY_AUTHENTICATION_PROVIDERS_OIDC_<provider-id>_REDIRECTURI`) now carry the same startup-validation sentence and link as the dotted-property rows. The deprecated Helm `orchestration.security.authentication.oidc.redirectUrl` row is deliberately left as-is: the chart appends `/sso-callback` to that value itself, so a "must include a callback path" note would be wrong there.

megglos:
@copilot please review again

docs/self-managed/components/orchestration-cluster/admin/debugging-authentication.md
```diff
@@ -77,6 +78,21 @@ LOGGING_LEVEL_IO_CAMUNDA_SECURITY=DEBUG
...
+- All other requests succeed.
+- Each new request tries again. The cluster serves the failed traffic again when the provider answers. You do not need to restart the cluster.
+- The cluster logs one warning each minute for each step that fails: a client registration, a token decoder, or a UserInfo mapping.
+
+An unreachable provider no longer stops the cluster from starting. This warning is your only signal that part of the authentication traffic fails. Monitor your log pipeline for `WARN` entries of the logger `io.camunda.security.spring.oidc.DeferredOidcResolution`. Each entry starts with `Failed to resolve`.
+
+The warning gives the provider, its issuer, and the scope. It does not give the endpoint that did not answer. Read the exception that the warning attaches to find that endpoint. It is the discovery endpoint `<issuer-uri>/.well-known/openid-configuration` for a provider that is configured with an issuer URI. It is the configured `jwk-set-uri` or `user-info-uri` for a provider that sets these endpoints. Then make sure that the cluster can reach that endpoint. See [Test the IdP directly](#test-the-idp-directly).
```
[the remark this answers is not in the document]
megglos: Adopted — verified in CSL 1.0.2 and you are right on both paths. `DeferredJwtDecoder.decode` calls `delegate.get().decode(token)`, so a configured `jwk-set-uri` is fetched outside the resolution wrapper and produces no `Failed to resolve` warning. `CachingOidcClaimsProvider.fetchEntry` catches a UserInfo fetch failure, logs `UserInfo fetch failed for issuer '…' at '…'` at `ERROR`, and returns the negative entry.

The warning guidance is now limited to discovery, and the two explicit endpoints have their own bullet: `jwk-set-uri` fails token validation with no `Failed to resolve` entry (raise `org.springframework.security` to `DEBUG`), and `user-info-uri` gives the `CachingOidcClaimsProvider` `ERROR` with issuer and endpoint. The UserInfo bullet also states the consequence your finding implies — the request succeeds with the token claims only, so the user can lose groups, roles, or tenants that UserInfo adds.

docs/self-managed/upgrade/components/890-to-8100.md
```diff
@@ -359,6 +359,37 @@ Next, move the files from the "General" process application to the new one:
 4. Select your new process application.
 5. Click **Move**.
 
+## Orchestration Cluster
+
+### OIDC redirect URI must include a callback path
+
+Camunda 8.10 validates the OIDC redirect URI at startup. The Orchestration Cluster does not start if the value is not a complete callback URL. This applies to `camunda.security.authentication.oidc.redirect-uri`, and to `camunda.security.authentication.providers.oidc.<provider-id>.redirect-uri`.
+
+A correct value:
```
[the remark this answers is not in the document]
megglos: Adopted. The list is now introduced as "The startup check accepts a value that:", and the paragraph below it opens with "A value that passes this check is not always a working configuration." before it names `/sso-callback` as the only served path. I applied the same wording to the equivalent list in `connect-external-identity-provider.md#redirect-uri`.

bmudric (approved):

docs/reference/announcements-release-notes/8100/8100-announcements.md
```diff
@@ -749,6 +749,42 @@ Camunda Hub and Optimize accept their existing authentication settings in 8.10 a
...
+
+The Orchestration Cluster validates the OIDC redirect URI at startup. It does not start if the value has no callback path. The cluster serves the callback at `/sso-callback`. This rule is not new: such a value also could not complete a browser login on 8.9. 8.10 reports the value at startup, and not at the next login. A cluster that only serves API clients never uses its redirect URI, so an incorrect value can stay unknown until the upgrade. The default value `{baseUrl}/sso-callback` is correct, and so are Helm deployments that set `orchestration.security.authentication.oidc.redirectUrl`.
```
StevePascoe: Maybe rephrase:
"The Orchestration Cluster validates the OIDC redirect URI at startup and does not start if the value has no callback path. The cluster serves the callback at '/sso-callback'. In 8.9 this check was performed at login rather than startup.  Clusters serving only API clients will never use the redirect URI, so a misconfiguration may not be detected until you upgrade to 8.10. The default '{baseURL}/sso-callback' is the correct value, as are the Helm deployments that set `orchestration.security.authentication.oidc.redirectUrl`.

docs/reference/announcements-release-notes/8100/8100-announcements.md
```diff
@@ -749,6 +749,42 @@ Camunda Hub and Optimize accept their existing authentication settings in 8.10 a
...
+
+<div className="release-announcement-row">
+<div className="release-announcement-badge">
+<span className="badge badge--change">Change</span>
+</div>
+<div className="release-announcement-content">
+
+#### Orchestration Cluster starts when an identity provider is unreachable
+
+The Orchestration Cluster contacts an OIDC provider at the first request that needs it, and not at startup. A provider that is down no longer stops the cluster from starting. The cluster serves all requests that it can authenticate without that provider, and fails only the requests that need it. Each new request tries again, thus the traffic recovers when the provider answers, and you do not need a restart. A failed request writes a warning that gives the provider, its issuer, and the scope, at most once each minute for each of these.
```
StevePascoe: I think we need to give an example here (e.g. we would honor a valid access token, but fail when trying to refresh it)

StevePascoe: do we need to be explicit on how this would fail? I would expect on login we'd get a not available style failure where on an expired token I'd expect unauthorized errors.

docs/reference/announcements-release-notes/8100/8100-announcements.md
```diff
@@ -749,6 +749,42 @@ Camunda Hub and Optimize accept their existing authentication settings in 8.10 a
...
+
+**Action:** If you used a failed startup to detect an unreachable identity provider, alert on that warning instead. The warning follows the traffic, so add a synthetic request if you must detect an outage that no user request reaches.
```
StevePascoe: that warning? we should be explicit "alert on the <blah> warning"

docs/reference/announcements-release-notes/8100/8100-release-notes.md
```diff
@@ -55,6 +55,20 @@ Use the new **Business Value** page in Camunda Hub to track process outcomes usi
...
 
+### Orchestration Cluster
+
+#### Startup no longer depends on a reachable identity provider
+
+<div class="release"><span class="badge badge--long" title="This feature affects Self-Managed">Self-Managed</span><span class="badge badge--medium" title="This feature affects Camunda 8 Run">Camunda 8 Run</span><span class="badge badge--medium" title="This feature affects Orchestration Cluster">Orchestration Cluster</span></div>
+
+The Orchestration Cluster contacts an OIDC provider at the first request that needs it, and not at startup. A provider that is still starting, or that is down, no longer stops the cluster from starting.
+
+- Only the requests that need the unreachable provider fail. All other requests succeed.
```
StevePascoe: I would add some examples here (e.g. login requests fail, token refreshes fail)

docs/reference/announcements-release-notes/8100/8100-release-notes.md
```diff
@@ -55,6 +55,20 @@ Use the new **Business Value** page in Camunda Hub to track process outcomes usi
...
+- Each new request tries again. The traffic recovers when the provider answers, and you do not need a restart.
```
StevePascoe, line 67: Does the customer need to worry about the provider being flooded with requests from us when it recovers? I'm assuming we try and fail with no re-try or queuing happening, yes?

docs/self-managed/components/orchestration-cluster/admin/connect-external-identity-provider.md
```diff
@@ -127,6 +127,15 @@ You may need to customize the redirect URI in advanced scenarios, such as:
 
 Regardless of customization, the redirect URI must always point to the `/sso-callback` endpoint of your Orchestration Cluster deployment.
 
+The Orchestration Cluster validates the redirect URI at startup. It does not start if the value is not a complete callback URL. The check accepts a value that:
+
+- starts with `{baseUrl}`, or has an `http` or `https` scheme and a host,
```
StevePascoe: I'd put https first... (i.e. lead with the recommended)

christinaausley (dismissed):
Thanks! I made a few small docs edits if you wanted to take a look.

Two things to confirm:
1. Is it accurate to use browser login requests and token-validation requests as the examples of traffic that can fail while the provider is unreachable?
2. Should we explicitly document that Camunda does not queue or retry failed requests in the background and only the next incoming request attempts resolution again? I left the current wording at “each new request tries again” to avoid over-specifying without confirmation.

docs/reference/announcements-release-notes/8100/8100-announcements.md
megglos: Moot now, and thank you for the rewrite — I took the "In 8.10, the cluster reports …" fix into the row it still applies to.

The direction changed after your review: camunda/camunda-security-library#682, released in 1.0.4, turned every one of these configuration checks into a warning. The cluster starts, and it logs what it found. So there is no breaking change and no upgrade action, the row is now a `Change`, and the section in the 8.9 to 8.10 upgrade guide is gone.

docs/self-managed/components/orchestration-cluster/admin/connect-external-identity-provider.md
megglos: Adopted — the list now reads "an `https` or `http` scheme".

docs/reference/announcements-release-notes/8100/8100-announcements.md
megglos: Adopted — "alert on the `DeferredOidcResolution` warning instead".

docs/reference/announcements-release-notes/8100/8100-announcements.md
megglos: Adopted, and your example is the one the text was missing. The row now says that a session the cluster authenticated before the outage keeps its access token until the token expires, and that the refresh which follows fails and ends the session.

megglos: Adopted, and you are right that there are two outcomes — with one correction on which is which.

- A request that needs a provider the cluster never resolved gets a **server error**, by design. `DeferredOidcResolution` rethrows the original exception unchanged, and `DeferredJwtDecoder` wraps a failed build in `JwtDecoderInitializationException`, which is not a `JwtException`, so Spring Security does not read it as a refused credential. A browser login and a bearer-token API request both land here.
- An **expired session** is where your unauthorized error comes from, and it is a different code path: `OAuth2RefreshTokenFilter` fails the refresh, runs the logout handler, and throws `OAuth2AuthenticationException`. No `Failed to resolve` warning is written for it.

Both are now named in the announcement, the release notes, and the debugging guide.

docs/reference/announcements-release-notes/8100/8100-release-notes.md
megglos: Adopted — the bullet names browser login requests and token-validation requests, and a second bullet covers the expired-session case.

docs/reference/announcements-release-notes/8100/8100-release-notes.md
megglos: Good question, and the answer is no — I checked it in the library rather than assume it.

There is no queue and no attempt in the background. One request makes one attempt, so the provider sees the rate of the requests that need it, and nothing more. `DeferredOidcResolution.memoizeOnSuccess` deliberately holds no lock across the resolution — the comment in it says a burst would otherwise wait one discovery timeout after another and run the request threads out — so the concurrent requests each try once and the first result wins. That result is then kept, so the attempts stop as soon as the provider answers. The requests already in flight at that moment are the only overlap.

I added this to the debugging guide, and a short form of it to the release notes.

megglos:
@christinaausley — two answers, and one apology.

**The apology first:** your `technical review` commit (c77fd62f83) was lost. I rebased the branch onto `main` to clear a conflict and force-pushed without carrying it, which dropped it. It is restored in a60e1c0639, minus the two hunks that touched the redirect-URI upgrade section, because that section is gone — see below.

**1. Are browser login and token-validation requests the right examples?** Yes, and there is a third case the text was missing. A request that needs a provider the cluster never resolved gets a server error, which is by design: the resolution rethrows the original exception, and the deferred decoder wraps a failed build in an exception Spring Security does not read as a refused credential. A session that was authenticated before the outage is different — it keeps its access token until the token expires, and then the refresh fails and the session ends with an authentication error. Both are now in the text.

**2. Should we say there is no queue and no background retry?** Yes — I verified it and added it. One request makes one attempt, the cluster keeps the first result it gets, and nothing runs in the background. "Each new request tries again" was correct, only incomplete.

**One more thing that changed since your review:** camunda/camunda-security-library#682, released in 1.0.4, turned the OIDC configuration checks into warnings. The cluster now starts and logs what it found, so the breaking-change announcement is a `Change`, the 8.9 to 8.10 upgrade section is removed, and a new "Review the startup warnings" section covers the two loggers. Worth a second look at that part.

megglos:
@StevePascoe @bmudric @christinaausley ready for another 👀

christinaausley (approved):
Looks good to me!

PR #9962, opened 2026-09-17
docs(best-practices): note secondary storage index sharding is not partition-aware

ChrisKujawa:
## Description

Adds a note to the sizing guide's secondary storage considerations: Elasticsearch/OpenSearch write throughput scales with primary shard count, not node count, and all Zeebe partitions currently export into the same indices. Scaling partitions up does not reshard secondary storage automatically. This is easy to miss when the surrounding horizontal-scaling guidance otherwise reads as "add more Elasticsearch/OpenSearch nodes."

Applied to the docs/ (next) and versioned_docs/version-8.8 and version-8.9 copies of sizing-self-managed.md. Version 8.7 has no equivalent page.

Related to [camunda/camunda#44833](https://github.com/camunda/camunda/issues/44833).

## PR Checklist


docs/components/best-practices/architecture/sizing-self-managed.md
```diff
@@ -315,9 +315,13 @@ Increase CPU and memory per broker. Note that there are **diminishing returns**
 
 ## Secondary storage considerations
 
-The resource tables above assume Elasticsearch as the secondary storage backend. If you are using a different backend:
+The resource tables above assume Elasticsearch as the secondary storage backend.
 
-- **OpenSearch:** Similar resource profile to Elasticsearch. The tables above generally apply.
+All Zeebe partitions currently export into the same Elasticsearch indices. Write throughput for an index scales with primary shard count, not with the number of Elasticsearch nodes, and scaling Zeebe partitions doesn't reshard indices automatically. Reassess shard count for your indices when you scale partitions. See the [Elasticsearch shards documentation](/self-managed/concepts/secondary-storage/managing-secondary-storage.md#shards).
```
ChrisKujawa: ℹ️ Probably make sense to move some details into the sizing pages instead of having this in the concepts page https://github.com/camunda/camunda-docs/issues/10003

RomanJRW: 💭 True that it doesn't reshard, but the indices are short lived anyway and new ones will be created with more shards if configured.

I think it's good to document this, I am just wondering if we can avoid mentioning the reshard, as I would say _new_ indices are what need changing rather than the existing ones. Wdyt?

RomanJRW (dismissed):

afgambin (approved):
@ChrisKujawa I did a pass and lgtm now

docs/components/best-practices/architecture/sizing-self-managed.md
ChrisKujawa: I think it depends on the use case, e.g., for long-lived running instances, it might not be so short-lived actually. Also this is different between old Zeebe ES/OS Exporters and Camnuda exporter. 

I guess yes we could expand a little more as a follow-up (like with #10003 ), but would keep this PR as it is for now, and see it as improvement. Ok?

PR #9968, opened 2026-09-18
fix(connectors): correct response mapping, env var typos, and inbound lifecycle docs

christinaausley:
## Description

Resolves recurring end-user and support feedback across three connectors:

- **AWS Bedrock** - adds an "Extract data from the response" section documenting how to parse `InvokeModel` vs. `Converse` responses (verified against connector source and AWS Bedrock API docs), fixes a broken response-mapping example for the Jamba-instruct model (indexed into the wrong field), and clarifies system prompt configuration. Closes the underlying idea behind Support ticket SUPPORT-28688.
- **Kafka connector** - documents that editing **Consumer Group ID** or **Offsets** on an already-deployed element can cause unexpected behavior (verified against connector source), and clarifies that upgrading from a version without the **Consume unmatched events** checkbox doesn't silently change existing elements. Generalizes both into `inbound-lifecycle.md` per the explicit ask in camunda-docs#4789.
- **Connector Runtime env vars** - fixes incorrect `CAMUNDA_CLIENT_*` env var spellings (extra underscore before `ID`/`ADDRESS`/`SECRET`) in `connector-sdk.md` and `host-custom-connector.md`, verified against the canonical properties reference, and cross-links to it to avoid future drift. Resolves SUPPORT-26704.

All technical claims were verified against real connector source (camunda/connectors) or AWS documentation rather than taken at face value from the support tickets, several of which contained garbled or stale values.

Only `docs/` (next) is updated; no `versioned_docs/version-8.9/` changes.

## PR Checklist

- [x] The commit history of this PR is cleaned up, using `{type}(scope): {description}` commit message(s)
- [x] My changes are for an upcoming minor release and are in the `/docs` directory.
- [x] I added a DRI, team, or delegate as a reviewer for technical accuracy and grammar/style


christinaausley:
This is ready for review 👍

PR #9973, opened 2026-09-18
self-managed(docs): migration guide to fix cloud credentials misuse

### LINKED ISSUE #6778: Document-store AWS credentials and region double as ambient cloud SDK configuration
## Description

`global.documentStore.type.aws.*` is the chart's only source of `AWS_ACCESS_KEY_ID`,
`AWS_SECRET_ACCESS_KEY` and `AWS_REGION` for three components that use them for purposes unrelated to
document storage, and a fourth that could:

| Component | What it uses them for | Activated by |
|---|---|---|
| connectors | `DefaultCredentialsProvider` for AWS connector tasks (Lambda, SQS, SNS, DynamoDB, Bedrock, Textract) | the per-task element-template "Default Credentials Chain (Hybrid/Self-Managed only)" option |
| optimize | credentials **and** `AWS_REGION` to sign AWS OpenSearch requests | `global.opensearch.aws.enabled` / `optimize.database.opensearch.aws.enabled` |
| web-modeler-restapi | `AWS_REGION` for its own S3 client (`AwsCloudStorageProperties` has no region property) | `global.documentStore.activeStoreId=aws` |
| identity | credentials and region for documented Aurora/RDS IAM database auth | nothing the chart renders — only a hand-written `identity.env` `SPRING_DATASOURCE_URL` override |

## Expected vs Actual Behavior

Expected: document-store credentials configure the document store. Actual: they are also the de facto
credential source for AWS connector tasks, Optimize's OpenSearch connection, and restapi's S3 client.

Consequences:

- Document-store S3 credentials become ambient AWS credentials for BPMN-authored connector tasks, so a
  process author selecting the default credentials chain silently borrows the platform's document-store
  IAM identity. Least-privilege violation.
- Changing or disabling the document store silently degrades Optimize's OpenSearch authentication —
  `useAwsCredentials()` catches the resolution failure, logs at `INFO`, and falls back to basic auth.
- The coupling obstructs IRSA adoption (`SUPPORT-29235` → `SUPPORT-30767` → product-hub#3388).
- On EKS with IRSA, the chart's `AWS_REGION` suppresses the webhook's cluster-region injection, so removing
  it silently switches these components to the cluster's region rather than leaving them region-less.
- `existingSecret: ""` still fails to render across the remaining consumers.

## Acceptance Criteria

- [ ] Each of the four components has its own supported way to supply AWS credentials and region, independent of `global.documentStore.*`.
- [ ] Migration guidance published for each, split by credential mode: IRSA-on-EKS users inherit credentials and the cluster region from the webhook, so they need explicit region configuration only when the bucket or OpenSearch domain sits in another region; static-key users lose credentials *and* region outright and need a replacement surface per component.
- [ ] ADR recorded (next free ID `0095`), amending ADR-0052 and ADR-0089.
- [ ] `breaking-change` label applied; InfraEx and #oc-reliability-testing have acknowledged.

## Additional Context

Verification method matters here — three traps produced false signals during investigation:
`camunda/identity` and `camunda/connectors-bundle` ship a single uber-jar so filename scans see nothing;
AWS SDK v2 jars are module-named (`s3-*.jar`, `regions-*.jar`) so an `aws-sdk` pattern scores zero; and
pnpm registry cache entries under `root/.cache/pnpm/**` look like installed dependencies.

Settled: the EKS IRSA webhook does inject the region. Verified read-only against `camunda-ci-eks` with a
throwaway annotated ServiceAccount and a server-side dry-run pod (admission runs, no pod created):

```
AWS_STS_REGIONAL_ENDPOINTS=regional
AWS_DEFAULT_REGION=eu-central-1
AWS_REGION=eu-central-1
AWS_ROLE_ARN=arn:aws:iam::…:role/…
AWS_WEB_IDENTITY_TOKEN_FILE=/var/run/secrets/eks.amazonaws.com/serviceaccount/token
```

The value is the cluster's own region. Injection is flag-gated upstream —
`amazon-eks-pod-identity-webhook` `pkg/handler/handler.go:230` emits the pair only when
`--aws-default-region` is set, empty by default (`main.go:82`) — and managed EKS sets it.

The webhook also never overwrites a region already present in the pod spec (`handler.go:201-203`).
Repeating the probe with `AWS_REGION=us-west-2` preset returned `AWS_REGION=us-west-2` and no
`AWS_DEFAULT_REGION`, so the chart's value wins wherever both apply.

Two gaps remain: EKS Pod Identity, the newer association-based mechanism, was not tested — creating an
association needs AWS-side permissions — and non-EKS clusters get nothing from either mechanism.

## References

`SUPPORT-29235`, `SUPPORT-30767`, `INC-29235`, product-hub#3388, #3741, #6724, #6735, #5026, #2836.

aabouzaid:

## PR Checklist


hisImminence (dismissed):
Nice addition — covers AWS/GCP/Azure and IAM-federation paths clearly, and it's placed correctly as a sibling `###` under "Update your values file to 8.10".

A few things worth cleaning up before/after merge:

1. **Em dashes**: `content.instructions.md` §2 says to avoid em dashes (—) in favor of shorter sentences. This adds 8 of them (e.g. `**Connectors** — used by the AWS SDK...`, `...silently fall back to Basic authentication — verify the OpenSearch connection...`). Not caught by Vale, so needs a manual split into two sentences or a comma/colon.
2. **Heading depth**: `##### Static access key / secret key` and `##### IRSA / EKS Pod Identity` are H5 — the rest of this doc tops out at H4. Docusaurus's default TOC only renders to H3, so these won't show in the sidebar TOC. Suggest flattening to H4 with a bold lead-in instead.
3. **Naming consistency**: AWS examples use `webmodeler-aws-credentials` / `role/webmodeler-role` for the Camunda Hub component, but the GCP block for the same component uses `camunda-hub-gcp-credentials`. Since the doc elsewhere treats `webModeler.*` as the legacy name being migrated away from, worth aligning on `camunda-hub-*` in the AWS examples too.
4. **Minor grammar**: "Workload Identity (GKE) is the annotation-based equivalent of IRSA, on `<component>.serviceAccount.annotations`:" is missing a verb — maybe "...equivalent of IRSA, configured on `<component>.serviceAccount.annotations`:".
5. **PR title typo**: "migration guie" → "migration guide".

None of these block correctness of the guidance, so approving — but please fix at least #1 and #2 before merge.

aabouzaid:
All comments have been addressed :heavy_check_mark: 
The last step is the review by @camunda/tech-writers.

christinaausley (approved):
Made a few adjustments if you want to confirm:

- Replaced the stale `webModeler.restapi` reference with `camundaHub.restapi`.
- Renamed the Web Modeler REST API heading to Camunda Hub REST API.
- Corrected the AWS workload identity guidance: IRSA and EKS Pod Identity provide credentials, but not `AWS_REGION`.
- Made the GCP Secret examples consistent by using `gcp-credentials` and clarifying that users should replace it with their actual Secret name.

One question before merge: can you confirm from the 8.9 and 8.10 chart templates or rendered manifests that AWS document-store credentials were propagated to Connectors, Optimize, and Web Modeler REST API in 8.9, while Azure document-store credentials were never propagated to those components? This determines whether the migration scope and Azure section are correct.

Feel free to revert any changes you disagree with 👍

aabouzaid:
The approval has been dismissed due to the rebase to fix the merge conflict.
No change since last commit :heavy_check_mark:

hisImminence:
## crev automated review

Ran `crev` (correctness, sensitive-data, adr-conformance, devils-advocate, verifier, escalation-assessor specialists) against this PR. 4 findings (all P2):

1. **`docs/self-managed/upgrade/helm/890-to-8100.md:476`** — Camunda Hub `envFrom` shorthand (`camundaHub.restapi:`) is not a valid values path; the chart reads `camundaHub.restapi.envFrom`. As written, this key is a no-op and users following it will end up with no injected credentials for Hub. Suggested fix:
   ```yaml
   camundaHub:
     restapi:
       envFrom:
         - secretRef:
             name: camunda-hub-aws-credentials
   ```

2. **`docs/self-managed/upgrade/helm/890-to-8100.md:403`** — New "Migrate document-store cloud credentials" section only exists in `docs/` (Next). The maintained `versioned_docs/version-8.9/self-managed/upgrade/helm/890-to-8100.md` doesn't have it, so 8.9 users reading the current released docs before upgrading won't see this guidance. Should be duplicated into the 8.9 versioned page if it applies there.

3. **`docs/self-managed/upgrade/helm/890-to-8100.md:565`** — GKE Workload Identity example only shows the `iam.gke.io/gcp-service-account` annotation. That alone isn't sufficient — pods also need Workload Identity Federation enabled on the cluster and an IAM binding granting the Kubernetes SA impersonation rights on the Google SA. Worth linking/stating those prerequisites so users don't apply only the annotation and end up without credentials.

4. **`docs/self-managed/upgrade/helm/890-to-8100.md:473`** — `envFrom` shortcut only mentions loading `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY`, but the preceding Optimize/Camunda Hub examples state those components also need `AWS_REGION`. Users taking this shortcut will lose the region requirement silently.

Devils-advocate ran 11 hypotheses; 3 promoted (above, #2-4), rest dropped as ungroundable or low-severity after verification (Azure connector claim, AWS example region value, GCP secret-name reuse, Hub service-account path, Identity region asymmetry — all checked out).

---
🤖 Generated with [crev](https://github.com/camunda) — flagging for human review per repo escalation policy (no override policy found, defaults to always-require-review).

hisImminence (dismissed):
Approving with the crev findings above noted for follow-up (particularly #1, the broken `envFrom` values path — worth a quick fix before merge if not already addressed).

aabouzaid:
@hisImminence

```
1. docs/self-managed/upgrade/helm/890-to-8100.md:476
```

This is just because the `camundaHub` key is a [deep copy](https://github.com/camunda/camunda-platform-helm/blob/a2e788aefb87d53edcd42c130eafa5df8c4df725/charts/camunda-platform-8.10/templates/camunda-hub/_helpers.tpl#L11) of the `webModeler,` so not all keys are shown.

```
2. docs/self-managed/upgrade/helm/890-to-8100.md:403
```

Expected as it's only from 8.9 to 8.10 upgrade.

```
3. docs/self-managed/upgrade/helm/890-to-8100.md:565
```

Not so accurate, as only the Kubernetes ServiceAccount requires annotations, according to the [official docs](https://docs.cloud.google.com/kubernetes-engine/docs/how-to/workload-identity#kubernetes-sa-to-iam).

Also, the migration only covers the Kubernetes-related part, not the entire GCP IAM setup (which is handled outside the Helm chart).

```
4. docs/self-managed/upgrade/helm/890-to-8100.md:473
```

Enhanced.

hisImminence (approved):

PR #9803, opened 2026-09-03
docs(self-managed): document phased Camunda Hub database migration

eamonnmoloney:
## Description

Document the required Camunda Hub database migration sequence when upgrading Self-Managed from 8.9 to 8.10.

The guide now covers the `quiesce`, `migrate`, and `normal` phases; the required backup and validation gates; workload checks between phases; fresh 8.10 installations; and the point after which restoring 8.9 requires restoring the pre-migration database backup.

Related to camunda/camunda-platform-helm#6787 and camunda/camunda-platform-helm#6788.

## When should this change go live?

- [ ] This is a bug fix, security concern, or something that needs **urgent release support**. (add `bug` or `support` label)
- [ ] This is already available but undocumented and should be released within a week. (add `available & undocumented` label)
- [x] This is on a **specific schedule** and the assignee will coordinate a release with the Documentation team. (create draft PR and/or add `hold` label)
- [ ] This is part of a scheduled **alpha or minor**. (add alpha or minor label)
- [ ] There is **no urgency** with this change (add `low prio` label)

## PR Checklist

- [x] The commit history of this PR is cleaned up, using [`{type}(scope): {description}` commit message(s)](https://github.com/camunda/camunda-docs/blob/main/CONTRIBUTING.MD#commit-message-header-formatting)

- [x] My changes are for **an upcoming minor release** and are in the `/docs` directory.
- [ ] My changes are for an **already released minor** and are in a `/versioned_docs` directory.

- [ ] I added a DRI, team, or delegate as a reviewer for technical accuracy and grammar/style:
  - [ ] [Engineering team review](https://github.com/camunda/camunda-docs/blob/main/howtos/documentation-guidelines.md#review-process)
  - [ ] [Technical writer review](https://github.com/camunda/camunda-docs/blob/main/howtos/documentation-guidelines.md#review-process) via `@camunda/tech-writers` unless working with an embedded writer.

PR #9976, opened 2026-09-18
docs(self-managed): [TOPO 1/2] document the 8.10 deployment topology and its install and upgrade paths

eamonnmoloney:
## Description

> | # | PR | Phase | Purpose |
> | --- | --- | --- | --- |
> | **1/2** | **this PR** | **Topology** | **Documents the 8.10 release-shape topology: Hub, Orchestration Cluster, and Optimize releases, their install/upgrade paths, and Physical Tenants across releases** ← you are here |
> | 2/2 | [docs(self-managed): document Optimize identity isolation for Physical Tenants](https://github.com/camunda/camunda-docs/pull/10020) | Optimize isolation | Documents how one Management Identity isolates multiple Optimize releases (the audience/role model, the `roleName` requirement) and a logical-tenant-ID collision limitation |

1/2 lays the release-topology groundwork; 2/2 layers directly on top of it (its base branch is this PR's `docs/8.10-deployment-topology`) and cross-links the Optimize identity-isolation details into the install/topology pages this PR introduces.

Camunda 8.10 changes the *shape* of a Self-Managed deployment, not just its contents: `global.topology.mode` introduces four release roles, so one management release running Camunda Hub and Management Identity can serve many independently deployed Orchestration Clusters, each hosting Physical Tenants with their own Optimize release.

The docs hadn't caught up. Before this PR:

- `global.topology.mode` appeared in exactly one page, `configure/multi-namespace.md`, filed under *configure* as though it were an optional variation rather than the production baseline.
- The `optimize` release role was **undocumented anywhere**, despite being the only supported way to run Optimize per Physical Tenant.
- Physical Tenant concept docs were thorough, but nothing explained how tenants change the Helm **release shape** — per-tenant Optimize, the four index-prefix families, or that declaring any tenant silently stops the `default` tenant inheriting root exporters.
- `install/index.md` framed the choice by secondary storage, steering a greenfield reader to an all-in-one release.

### What this adds

**Vision** — what an 8.10 deployment is meant to look like:

- `reference-architecture/deployment-topology.md`: the four release roles with their real requirements and chart-version applicability, the `1 Hub → N clusters → M tenants → M Optimize` cardinality, the mixed-version topology, why the split exists, and its limits.
- `deployment/helm/configure/configuration-responsibilities.md`: Helm configures how and where a component runs and connects; the application configures what it does; `extraConfiguration` is the hook between them.

**Mechanics** — how to build it:

- `deployment/helm/install/topology/`: the multi-namespace guide split by release role — `index`, `hub-release`, `orchestration-release`, `optimize-release`, and `physical-tenants`.
- `upgrade/helm/combined-to-split-topology.md`: converting an existing combined release, including what broker storage prevents from moving.

`configure/multi-namespace.md` is removed with a redirect; its content lives in `install/topology/`. Its three inbound links are retargeted to precise anchors.

### Verification

Every values example was extracted from the rendered pages and templated against chart `15.0.0-alpha5` with the Helm v4 CLI, rather than written by hand:

| Check | Result |
| --- | --- |
| `hub` mode | Hub + Management Identity, **0** StatefulSets |
| `orchestration` mode | 1 StatefulSet + Connectors, no Identity/Hub/Optimize |
| `optimize` mode | Optimize templates only |
| 8.7 `orchestration` | Zeebe, Zeebe Gateway, Operate, Tasklist, Connectors, Optimize |
| 8.10 Hub + `architecture: legacy` | `type: zeebeGateway`, split readiness URLs, no Orchestration Admin |
| Tenants without a pinned issuer | fails with the documented error |

This caught two defects in the drafts: the Optimize example didn't render (missing OIDC issuer), and `publicIssuerUrl` does **not** satisfy the Physical Tenant issuer requirement although it does satisfy Optimize's. Both are fixed and the distinction is now documented.

### Grounded in

- [ADR 0095 — release roles for multi-cluster deployment topology](https://github.com/camunda/camunda-platform-helm/blob/main/docs/adr/0095-adopt-release-roles-for-multi-cluster-deployment-topology.md)
- ADR 0098 — backport the orchestration role to supported charts (`camunda/camunda-platform-helm#7012`), and the `#7008`–`#7012` stack that adds `orchestration` mode to the 8.7, 8.8, and 8.9 charts
- [ADR 0091 — `extraConfiguration` as the application configuration mechanism](https://github.com/camunda/camunda-platform-helm/blob/main/docs/adr/0091-adopt-component-extraconfiguration-as-the-standard-application-configuration-mechanism.md)
- `camunda/product-hub#3562`, whose docs criteria the configuration-responsibilities page satisfies

### Reviewer notes

- **Needs sign-off.** `combined-to-split-topology.md` describes a procedure engineering has not yet ratified — ADR 0095 explicitly leaves it undefined. Please review that page against what the Distribution team will actually support.
- **One known broken anchor.** The announcements entry links to `890-to-8100#migrate-camunda-hub`, which lands with #9803. The build passes; the warning clears when that merges.
- **Not split into a stack.** `onBrokenLinks: "throw"` plus bidirectional links between the vision and mechanics pages means no subset builds on its own. The four commits are the reviewable units.
- Follow-up, not in this PR: a split-topology reference lane in `camunda-deployment-references`, blocked on `camunda/camunda-deployment-references#3399`.

## PR Checklist

- [x] I added a redirect for a renamed or deleted page to the .htaccess file.


bkenez (changes requested):
**Request Changes**

The release-role structure makes sense, but the installation examples and several behavior claims need correcting before publication.

1. **Enable the exporter feeding standalone Optimize.** The [orchestration example](https://github.com/camunda/camunda-docs/blob/9d221f5c1dc98a0731feca82215930a770873f0a/docs/self-managed/deployment/helm/install/topology/orchestration-release.md#L128) renders no legacy exporter: local Optimize is disabled and `orchestration.exporters.zeebe.enabled` defaults to `false`. Installing Optimize separately cannot enable it in the orchestration release. Please configure the writer explicitly, including its destination, credentials, and matching index prefix.

2. **Add routing to the Optimize example.** The [provided values](https://github.com/camunda/camunda-docs/blob/9d221f5c1dc98a0731feca82215930a770873f0a/docs/self-managed/deployment/helm/install/topology/optimize-release.md#L46) render a Service and Deployment, but no Ingress or HTTPRoute. Sharing the orchestration namespace does not make its Ingress route to this separate release. Please include routing matching the advertised redirect URL, or make externally managed routing an explicit prerequisite.

3. **Gate mixed-version guidance on the chart implementation.** The [8.7 Hub example](https://github.com/camunda/camunda-docs/blob/9d221f5c1dc98a0731feca82215930a770873f0a/docs/self-managed/deployment/helm/install/topology/hub-release.md#L167) fails schema validation against current chart `15.0.0-alpha5`: `architecture`, `operateServiceName`, and `tasklistServiceName` are rejected. The referenced chart PRs #7008–#7012 remain open. Please hold this guidance until those changes land and specify minimum chart versions.

4. **Distinguish tenant client registration from Hub inventory.** The [tenant status claim](https://github.com/camunda/camunda-docs/blob/9d221f5c1dc98a0731feca82215930a770873f0a/docs/self-managed/deployment/helm/install/topology/physical-tenants.md#L81) does not match the render. `physicalTenants[].components.optimize` registers the client in Identity, but generates no corresponding tenant Optimize entry in Hub. Please remove the claim that it appears as unreachable, or document how that inventory entry is actually configured.

5. **Correct the default-tenant exporter warning.** The [warning](https://github.com/camunda/camunda-docs/blob/9d221f5c1dc98a0731feca82215930a770873f0a/docs/self-managed/deployment/helm/install/topology/physical-tenants.md#L41) says declaring any tenant silently stops default-tenant exporting. The runtime synthesizes an undeclared `default` from root configuration and retains its root broker configuration. Explicit default configuration and exporter assignments behave differently, and missing required assignments can fail startup. Please document those cases separately.

6. **Do not exclude RDBMS-backed clusters from Optimize.** The [install overview](https://github.com/camunda/camunda-docs/blob/9d221f5c1dc98a0731feca82215930a770873f0a/docs/self-managed/deployment/helm/install/index.md#L35) conflates the cluster’s secondary storage with Optimize’s input. RDBMS plus an Elasticsearch exporter renders successfully, and broker configuration supports both. Optimize cannot read RDBMS directly, but it can consume the additional Elasticsearch/OpenSearch record stream.

The combined-to-split procedure also needs the requested engineering sign-off and migration/rollback validation. It should explicitly preserve Optimize’s **application index prefix**, not only its reader prefix, to retain existing reports and configuration.

Validation: focused Helm 4.3.0 renders against Helm main `baf7640`, plus runtime source inspection. No live deployment or migration rehearsal.

docs/self-managed/deployment/helm/install/index.md
```diff
@@ -5,33 +5,46 @@ sidebar_label: Install
...
+| --------------------------- | ------------------------------------------------------------------------------------------------------------------------------ | ------------------------------------------------------ |
+| Elasticsearch or OpenSearch | [Using external Elasticsearch](/self-managed/deployment/helm/configure/database/elasticsearch/using-external-elasticsearch.md) | High throughput, and any deployment that uses Optimize |
+| Relational database         | [Install with RDBMS](./helm-with-rdbms.md)                                                                                     | Organizations that standardize on relational databases |
+| Embedded H2                 | [Quick install](./quick-install.md)                                                                                            | Evaluation only. Single broker, not for production     |
+
+:::warning
+Optimize requires Elasticsearch or OpenSearch and can't read from a relational database. An Orchestration Cluster on RDBMS secondary storage can't be analyzed with Optimize.
```
hisImminence, line 35: From my understanding I would say: it can be analyzed but only if Optimize is deployed with an Elasticsearch and the Orchestration cluster also exports into the Elasticsearch.

Suggestion: Something like: "An Orchestration Cluster on RDBMS secondary storage can be analyzed only with a separate Elasticsearch or OpenSearch and Optimize deployed."?

docs/self-managed/deployment/helm/install/topology/optimize-release.md
```diff
@@ -0,0 +1,130 @@
...
+
+`optimize.database.elasticsearch.prefix` is a reader prefix. It must exactly equal the writer prefix of the exporter for this tenant in the Orchestration Cluster.
+
+A mismatch doesn't fail. Optimize starts successfully against the wrong or an empty record set, and displays no process data. Similar-looking prefixes are not sufficient.
+
+`CAMUNDA_OPTIMIZE_ELASTICSEARCH_SETTINGS_INDEX_PREFIX` is different: it names where Optimize writes its own indices, and must be unique per Optimize release and distinct from every writer prefix.
+
+For the full prefix model across all releases, see [isolate every index prefix family](./physical-tenants.md#isolate-every-index-prefix-family).
+
+## Give each release its own identity
```
hisImminence, line 102: ❓ @eamonnmoloney, would that allow us to not deploy one identity with each Optimize while being able to limit the access per user/ group? (I thought the limitation was that we would need to install one mgmt idenenty to each Optimize if we want to guarantee this level of separation).

docs/self-managed/deployment/helm/install/index.md
```diff
@@ -5,33 +5,46 @@ sidebar_label: Install
...
 
-:::tip Decision criteria
-- Use **quick install** if you want to evaluate quickly without external dependencies.
-- Use **install with RDBMS** if your organization standardizes on relational databases.
-- Use **production install** if you need security, scalability, and operational features.
-- If you don’t have required infrastructure, see [deploy required dependencies](/self-managed/deployment/helm/configure/operator-based-infrastructure.md).
+| Use case                                        | Topology                                                   | Installation guide                                                              |
+| ----------------------------------------------- | ---------------------------------------------------------- | ------------------------------------------------------------------------------- |
+| Testing, evaluation, local development          | One `combined` release                                     | [Quick install](./quick-install.md)                                             |
```
hisImminence, line 16: Do we miss the Optimize-release entry and link in this list?

docs/self-managed/reference-architecture/deployment-topology.md
```diff
@@ -0,0 +1,112 @@
...
+| Management Identity presets and Camunda Hub cluster inventory  | The chart, from `global.topology.clusters` |
+
+Camunda 8.10 bundles no Elasticsearch, PostgreSQL, or Keycloak subcharts. Provision these before you install. See [deploy required dependencies](/self-managed/deployment/helm/configure/operator-based-infrastructure.md).
+
+## Choose your topology
+
+| Your situation                                            | Use                                                                                                                                                                             |
+| --------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Evaluating Camunda, or developing locally                 | A `combined` release. See [quick developer install](/self-managed/deployment/helm/install/quick-install.md)                                                                     |
```
hisImminence, line 106: Same question as before - does this list miss the Optimize only release entry &  link?

docs/self-managed/reference-architecture/kubernetes.md
```diff
@@ -166,6 +169,18 @@ For configuration details, see:
 
 The Orchestration Cluster can be configured to authenticate with OIDC by connecting to the Management Identity service deployed in this namespace.
 
+#### Optimize releases
+
+Each Physical Tenant in an Orchestration Cluster is served by one Optimize release, deployed with `global.topology.mode: optimize`. That release deploys Optimize and nothing else.
+
+One Optimize instance reads exported records from a single Elasticsearch or OpenSearch index prefix, so it can serve exactly one tenant. This applies to the default Physical Tenant too: a cluster with no additional tenants still needs one Optimize release if you want analytics.
+
+Each Optimize release requires its own OIDC client, audience, redirect URL, and context path, and it connects to the same secondary storage the Orchestration Cluster exports to. Its reader prefix must exactly match that tenant's exporter writer prefix. Optimize requires Elasticsearch or OpenSearch and can't use an RDBMS.
```
hisImminence, line 178: Please clarify the question/ limitation potentially with @Ben-Sheppard: Optimize requires its own management identity if access of users/roles/groups should be limited - if it connects to one mgmt. identity all of these have access to all Optimize resources?

hisImminence (changes requested):
Thanks @eamonnmoloney for this massive improvement and clarity for the Helm Chart deployments and topology! Besides some follow up quesiotns/ comments - really nice work! 

Wondering if we should track one follow up after this is merged: Rewrite the production install page to a check list, linking to further details...

alexronquillo:
@eamonnmoloney My plan is to review this by EOW (25. Sept)

docs/self-managed/reference-architecture/kubernetes.md
eamonnmoloney: This is address in https://github.com/camunda/camunda-docs/pull/10020/changes#diff-3a5dec42b44fc1150c1bc6e171a6c112cfe34f31ed88fd7c599e09dad27809a5

docs/self-managed/deployment/helm/configure/configuration-responsibilities.md
```diff
@@ -0,0 +1,100 @@
...
+description: "Helm configures how and where a Camunda component runs and connects. The application configures what it does. extraConfiguration is the hook between them."
+---
+
+Helm configures how and where a Camunda component runs and connects. The application configures what it does. `<component>.extraConfiguration` is the hook between the two.
+
+Neither layer is more authoritative than the other. They answer different questions, and knowing which question you're asking tells you where a setting belongs.
+
+Starting with Camunda 8.10, chart values that existed only to proxy a single application property are deprecated in favor of `extraConfiguration`. The chart keeps the Kubernetes surface it's responsible for and stops mirroring the application's own configuration.
+
+## Which layer owns a setting
```
alexronquillo: I like the approach of this section, but I suggest we follow one through-line from the header to the result in the table. Using this sentence from the introduction, "knowing which question you're asking tells you where a setting belongs", I suggest something like this:

## Where settings belong

_What_ you're configuring determines _where_ the setting belongs:

| Configuring | Examples | Belongs in |
| :--- | :--- | :--- |
| Application behavior    | Feature flags, toggles, log levels, security and authorization behavior, Spring Boot properties, and anything else that changes application logic | `<component>.extraConfiguration` |
| Kubernetes infrastructure    | Resource requests and limits, affinity and scheduling, service accounts, volumes, replica counts, deployment strategy                             | `values.yaml`                    |
| Connectivity and credentials | External endpoints, database URLs, secondary storage hosts, TLS certificates, secret references, Ingress and Gateway wiring                       | `values.yaml`                    |
| Cross-component coordination | Release role (`global.topology.mode`), the Hub cluster inventory, shared authentication identifiers                                               | `values.yaml`                    |

What do you think?

docs/self-managed/deployment/helm/configure/configuration-responsibilities.md
```diff
@@ -0,0 +1,100 @@
...
+| Application configuration    | Feature flags, toggles, log levels, security and authorization behavior, Spring Boot properties, and anything else that changes application logic | `<component>.extraConfiguration` |
+| Kubernetes infrastructure    | Resource requests and limits, affinity and scheduling, service accounts, volumes, replica counts, deployment strategy                             | `values.yaml`                    |
+| Connectivity and credentials | External endpoints, database URLs, secondary storage hosts, TLS certificates, secret references, Ingress and Gateway wiring                       | `values.yaml`                    |
+| Cross-component coordination | Release role (`global.topology.mode`), the Hub cluster inventory, shared authentication identifiers                                               | `values.yaml`                    |
+
+Application property names are the same whichever deployment method you use, so what you learn about `camunda.security.*` or `camunda.physical-tenants.*` transfers from Helm to Docker, to a JAR, or to ECS. Chart values don't transfer, which is why they're limited to the deployment layer.
+
+## Deliver application configuration
```
alexronquillo: In my opinion, "deliver" is an overloaded term with deployment connotations. "application configuration" is also a mouthful. I suggest we use something like this, if it's still equally accurate:

```suggestion
## Provide application settings
```

If you agree, we'll need to make similar updates throughout this PR.

docs/self-managed/concepts/physical-tenants/index.md
```diff
@@ -117,3 +117,11 @@ Isolation is enforced by validating the resolved `provider, bucket/container, pa
 For configuration examples covering shared buckets with per-tenant paths, dedicated buckets per tenant, and GCP prefix isolation, see [document store storage](./storage-isolation.md#document-store-storage).
 
 For the storage backends used by tenant-scoped data, see [secondary storage](../secondary-storage/index.md) and [document handling configuration](../document-handling/configuration/index.md).
+
+## Deploying Physical Tenants with Helm
+
+The Helm chart delivers tenant configuration rather than modeling it: there's no `orchestration.physicalTenants` values key, and tenants are declared as `camunda.physical-tenants.*` application configuration through `orchestration.extraConfiguration`.
```
alexronquillo: We should link to `configuration-responsibilities.md` and clean up the language, if we go with the suggestions in that file.

docs/self-managed/reference-architecture/deployment-topology.md
```diff
@@ -0,0 +1,112 @@
...
+| --------------- | -------------- | ---------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `hub`           | 8.10 only      | Camunda Hub and Management Identity      | `identity.enabled: true`, OIDC authentication, and it's the only release that declares `global.topology.clusters`                                                                                                                                                                                                 |
+| `orchestration` | 8.7 and later  | One Orchestration Cluster and Connectors | `global.identity.auth.enabled: true`, `identity.enabled: false`, a reachable `global.identity.service.url`, and the workload enabled. Requirements differ by chart version, see [per-version requirements](/self-managed/deployment/helm/install/topology/orchestration-release.md#requirements-by-chart-version) |
```
leiicamundi: `8.7 and later` is not true of any published chart, and the `Deploys` cell is only correct for 8.10.

On availability: the latest published 8.7, 8.8 and 8.9 charts (12.13.8, 13.13.2, 14.10.1) contain no `global.topology` key. The backports merged on 2026-09-22 (#7009, #7010, #7011) without a version bump, so those same version numbers were already published from artifacts without the feature. I rendered published 12.13.8 with and without `global.topology.mode: orchestration` and the output is byte-identical after normalizing the generated secrets, so the value is accepted and ignored.

On what it deploys: ADR 0098 clause 5 says an 8.7 orchestration release "retains Zeebe, Zeebe Gateway, Operate, Tasklist, Connectors, and Optimize", which your own verification table in the description confirms. So on 8.7 the orchestration release still runs Optimize in-release, which cuts against the one-Optimize-release-per-tenant model this page builds.

Suggested: replace `8.7 and later` with the first chart version per line that actually ships the role, and split the `Deploys` cell so 8.7 is described separately.

docs/self-managed/reference-architecture/deployment-topology.md
```diff
@@ -0,0 +1,112 @@
...
+| `optimize`      | 8.10 only      | Optimize only                            | `optimize.enabled: true`, `global.noSecondaryStorage: false`, an enabled Elasticsearch or OpenSearch backend with a non-empty host, an OIDC issuer, a reachable Management Identity URL, and `optimize.contextPath` when the chart renders this release's routing                                                 |
```
leiicamundi: The `optimize` role does not exist in the published 8.10 chart.

```
$ helm template camunda camunda/camunda-platform --version 15.0.0-alpha5 -f optimize-values.yaml
Error: values don't meet the specifications of the schema(s) in the following chart(s):
camunda-platform:
- at '/global/topology/mode': value must be one of 'combined', 'hub', 'orchestration'
```

The role landed on `main` in #6884 on 2026-09-10. The `camunda-platform-8.10-15.0.0-alpha5` tag is from 2026-09-04, so it predates it. Since `Chart.yaml` on `main` still reads `15.0.0-alpha5`, templating a working copy of `main` succeeds and templating the published chart fails, which is most likely how this passed your verification.

This needs a minimum chart version, and realistically it needs the next alpha published before the page can ship.

docs/self-managed/reference-architecture/deployment-topology.md
```diff
@@ -0,0 +1,112 @@
...
+| `combined`      | 8.7 and later  | Every enabled component in one release   | None beyond normal component configuration. This is the default                                                                                                                                                                                                                                                   |
+
+Camunda Hub and its cluster inventory exist only in the 8.10 chart, so `hub` and `optimize` are 8.10-only roles. The 8.7, 8.8, and 8.9 charts support `combined` and `orchestration` only.
```
leiicamundi: The second sentence is right about intent and wrong about the present tense. Today the published 8.7, 8.8 and 8.9 charts support neither `combined` nor `orchestration` as a topology value, because `global.topology` is not in them at all. `combined` is the effective behavior only because it is the behavior those charts always had.

Worth stating explicitly that `orchestration` on those lines requires a chart version that has not shipped yet, otherwise a reader on 12.13.8 sets the value, sees no error, and ends up with a combined release.

docs/self-managed/deployment/helm/install/topology/optimize-release.md
```diff
@@ -0,0 +1,130 @@
...
+Optimize requires Elasticsearch or OpenSearch and can't use a relational database, even when the Orchestration Cluster uses one for its own secondary storage.
+
+## Create `optimize-values.yaml`
+
+This example is the Optimize release for the Physical Tenant `tenanta` in the cluster `production-a`. Use the OpenSearch equivalents where applicable.
+
+```yaml
+global:
+  topology:
+    mode: optimize
```
leiicamundi: This whole page is blocked on a chart release. `mode: optimize` is rejected by the published `15.0.0-alpha5`:

```
- at '/global/topology/mode': value must be one of 'combined', 'hub', 'orchestration'
```

Nothing else on the page can be exercised until the role ships, including the prefix-matching guidance, which is the part most likely to be misapplied.

Two smaller things that will matter once it does render, and which I could not verify because the render fails:

- @bkenez's point 2 about routing is still unanswered. The example sets `optimize.contextPath` but the release produces no Ingress or HTTPRoute of its own, so the advertised `redirectUrl` has nothing serving it unless routing is managed externally. That belongs in the prerequisites.
- Line 35 lists `global.identity.auth.publicIssuerUrl` as one way to satisfy the OIDC issuer requirement, while `orchestration-release.md:147` and `physical-tenants.md:18` both say `publicIssuerUrl` does not satisfy the issuer requirement. The description explains that the distinction is real and deliberate, but a reader hitting these two pages in either order will read it as a contradiction. Consider saying on this page why Optimize is the exception.

docs/self-managed/deployment/helm/install/topology/physical-tenants.md
```diff
@@ -0,0 +1,141 @@
...
+          optimize: # Optimize for the default tenant
+            enabled: true
+            clientId: optimize-production-a-default
+            audience: optimize-production-a-default-api
+            roleName: Optimize production-a default
+            redirectUrl: https://production-a.example.com/optimize-default
+            secret:
+              existingSecret: optimize-production-a-default-oidc
+              existingSecretKey: client-secret
+        physicalTenants:
```
leiicamundi: `physicalTenants` on a cluster record is rejected by the published chart, and the behavior it documents is not what the chart produces on `main` either.

Published chart:

```
$ helm template camunda camunda/camunda-platform --version 15.0.0-alpha5 -f hub-values-with-physicaltenants.yaml
Error: values don't meet the specifications of the schema(s) in the following chart(s):
camunda-platform:
- at '/global/topology/clusters/0': additional properties 'physicalTenants' not allowed
```

The published `topologyCluster` definition accepts only `authorizations`, `components`, `contextPaths`, `host`, `id`, `name`, `namespace`, `releaseName` and `version`, with `additionalProperties: false`. The key exists on `main`, added by #6884 and #7095, both after the alpha5 tag. So this section needs the next chart alpha published before it can ship.

Rendering it against `main` also confirms @bkenez's point 4, which is a separate problem that the release will not fix. The tenant reaches the Identity presets:

```
- name: "Optimize Production A tenanta"
  id: "optimize-production-a-tenanta"
- name: "Optimize Production A tenanta API"
  audience: "optimize-production-a-tenanta-api"
- name: "Optimize production-a tenanta"
  description: "Grants full access to Optimize Production A tenanta"
```

But the Hub cluster inventory in `camunda-web-modeler-restapi-configuration` lists exactly one Optimize component for that cluster, the cluster-level one:

```
- name: Optimize
  type: optimize
  urls:
    webapp: "https://orchestration.example.com/optimize"
    readiness: "http://camunda-optimize.orchestration.svc.cluster.local:80/optimize/api/readyz"
```

`tenanta` appears nowhere in the inventory. I checked the rendered clusters block directly rather than grepping the whole manifest, because the Identity hits are easy to mistake for inventory hits.

That makes line 81 wrong: a tenant mapped in Hub does not show an unreachable Optimize, it shows nothing at all, which is a different thing for an operator to diagnose. Either drop that sentence or document how a tenant Optimize is meant to reach the Hub inventory.

docs/self-managed/deployment/helm/install/topology/hub-release.md
```diff
@@ -0,0 +1,241 @@
...
+  topology:
+    mode: hub
+    clusters:
+      - id: legacy-a
+        name: Legacy A
+        namespace: orchestration-legacy
+        releaseName: camunda
+        host: legacy-a.example.com
+        version: "8.7.38"
+        architecture: legacy
```
leiicamundi: Confirming @bkenez's point 3 with a current render, because the referenced chart PRs have since merged and the point could look resolved. It is not.

```
$ helm template camunda camunda/camunda-platform --version 15.0.0-alpha5 -f hub-values-chart87.yaml
Error: values don't meet the specifications of the schema(s) in the following chart(s):
camunda-platform:
- at '/global/topology/clusters/0/components/orchestration': additional properties 'operateServiceName', 'tasklistServiceName' not allowed
- at '/global/topology/clusters/0': additional properties 'architecture' not allowed
```

#7008 to #7012 all merged on 2026-09-22, so the fields exist on `main`. They are not in the published `15.0.0-alpha5` artifact, which was tagged 2026-09-04. The published `topologyOrchestrationComponent` accepts `serviceName`, `gatewayServiceName`, `restUrl` and `readinessUrl` from this example, but not `operateServiceName` or `tasklistServiceName`.

So the example is correct for the chart that is coming and unusable with the chart a reader can install right now.

docs/self-managed/deployment/helm/install/topology/orchestration-release.md
```diff
@@ -0,0 +1,202 @@
...
+description: "Install an execution plane: a Helm release with global.topology.mode set to orchestration, running one Orchestration Cluster and Connectors."
+---
+
+An orchestration release is an execution plane. It runs one Orchestration Cluster and Connectors, and connects to the Management Identity service in the Hub release.
+
+Install it after the [Hub release](./hub-release.md) is healthy. You can install any number of orchestration releases against one Hub.
+
+## What an orchestration release deploys
+
+`global.topology.mode: orchestration` deploys the Orchestration Cluster and Connectors, and suppresses the local Management Identity workload.
```
leiicamundi: This under-describes what the role does. I set `camundaHub.enabled: true` alongside `mode: orchestration` and the render still produced only the Connectors Deployment and the Orchestration Cluster StatefulSet, so the role suppresses Camunda Hub and Optimize as well, not just Management Identity.

Worth saying so directly. It is the same reassurance `optimize-release.md:24` gives ("gates off every component except Optimize, so you don't disable each one by hand"), and it removes any doubt about whether a converted values file needs manual cleanup.

docs/self-managed/deployment/helm/install/topology/orchestration-release.md
```diff
@@ -0,0 +1,202 @@
...
+
+orchestration:
+  enabled: true
+  contextPath: /orchestration
+  security:
+    authentication:
+      oidc:
+        clientId: orchestration
+        audience: orchestration-api
+        redirectUrl: https://orchestration.example.com/orchestration
```
leiicamundi: This example triggers the one deprecation warning the PR's own new page tells readers to avoid. Rendering it against `15.0.0-alpha5` emits:

```
[camunda][warning] DEPRECATION: The Helm values file key "orchestration.security.authentication.oidc.redirectUrl" is deprecated and will be removed in chart v16 (Camunda 8.11). Configure this via "orchestration.extraConfiguration" instead.
```

It is the only warning this example produces, and the Hub example produces none. Meanwhile `configuration-responsibilities.md:12` states that "chart values that existed only to proxy a single application property are deprecated in favor of `extraConfiguration`".

Since this is the flagship orchestration example for 8.10, it should use the form it is steering people toward, or say in a line why it does not yet.

docs/self-managed/deployment/helm/install/topology/index.md
```diff
@@ -0,0 +1,131 @@
...
+- Network policies that permit Domain Name System (DNS) traffic and the required cross-namespace service traffic.
+
+Camunda 8.10 bundles no Elasticsearch, PostgreSQL, or Keycloak subcharts, so these must exist before you install. See [deploy required dependencies](/self-managed/deployment/helm/configure/operator-based-infrastructure.md).
+
+The examples use `camunda` as the release name in every namespace, `hub` as the Hub namespace, and `orchestration` as the orchestration namespace. If you change a release name or namespace, update every Kubernetes service name that references it.
+
+Select a supported chart version from the [Helm chart version matrix](https://helm.camunda.io/camunda-platform/version-matrix/), then set it before installation:
+
+```sh
+export HELM_CHART_VERSION=<15.x-chart-version>
```
leiicamundi: One variable will not carry this guide. Line 29 tells the reader each Orchestration Cluster release can deploy from the 8.7, 8.8, 8.9 or 8.10 chart, but `orchestration-release.md:158` and `optimize-release.md:115` both reuse `$HELM_CHART_VERSION`. Someone following the mixed-version path installs every release from the 15.x chart, which is the one case the topology explicitly does not want.

A per-role variable would fix it, for example `HUB_CHART_VERSION` and `ORCHESTRATION_CHART_VERSION`, with the Hub one pinned to 15.x since `hub-release.md:16` already says the Hub release is always the 8.10 chart.

docs/self-managed/upgrade/helm/combined-to-split-topology.md
```diff
@@ -0,0 +1,143 @@
...
+
+### Step 2: install the Hub release in a new namespace
+
+Create `hub-values.yaml` with `global.topology.mode: hub`, and a `global.topology.clusters` record whose component client IDs, audiences, redirect URLs, and secrets exactly match what the existing combined release already uses. Reusing the existing identifiers is what lets Hub adopt the running cluster instead of registering a second one.
+
+Follow [install the Hub release](/self-managed/deployment/helm/install/topology/hub-release.md), and install into a new namespace. Don't reuse the combined release's namespace.
+
+Project the workload client secrets into the Hub namespace as well. Kubernetes Secrets are namespace-scoped.
+
+At this point Hub and Management Identity are running twice: once in the combined release, once in the new Hub release. Both read the same external databases, so confirm your Hub database supports the concurrent connections before you continue.
```
leiicamundi: Framing this as a connection-count question understates it. Two Management Identity instances against one database is not only a pool-sizing concern: Identity runs schema initialization and additive client, resource-server, permission and role provisioning at startup, and step 2 has both instances doing that against the same database with the same topology identifiers.

This is exactly the part of the procedure the description already flags as needing engineering sign-off, so I would not guess at the answer. But the sentence currently reads as though checking `max_connections` covers it, and a reader will act on that.

Either state the concurrency constraint explicitly once Distribution confirms it, or say plainly that running two Identity instances against one database is unvalidated and gate the step behind a rehearsal.

docs/self-managed/upgrade/helm/combined-to-split-topology.md
```diff
@@ -0,0 +1,143 @@
...
+Stop here and roll back if the cluster doesn't appear. Nothing has changed in the execution plane yet.
+
+### Step 4: convert the combined release to an orchestration release
+
+Update the existing release's values:
+
+- Set `global.topology.mode: orchestration`.
+- Set `identity.enabled: false`.
+- Set `global.identity.service.url` to the Management Identity service in the Hub namespace.
+- Set `camundaHub.enabled: false`.
```
leiicamundi: This step is redundant. `mode: orchestration` already suppresses Camunda Hub. I rendered the orchestration example with `camundaHub.enabled: true` explicitly set and the output still contained only Connectors and the Orchestration Cluster StatefulSet.

Harmless, but it invites the reader to think the role needs manual help, which is the opposite of the point the topology pages make. Dropping the line, or noting that the role handles it, is cleaner.

docs/reference/announcements-release-notes/8100/8100-announcements.md
```diff
@@ -638,6 +638,40 @@ By default, this mount used an `emptyDir`, so no PVC cleanup is required. Howeve
...
+</div>
+<div className="release-announcement-content">
+
+#### Deployment topology release roles
+
+Camunda 8.10 adds `global.topology.mode` to the Helm chart, so a release declares its role in the wider deployment: `combined`, `hub`, `orchestration`, or `optimize`. One `hub` release running Camunda Hub and Management Identity can serve many independently deployed `orchestration` releases, and an `optimize` release deploys Optimize alone, so each Physical Tenant gets its own Optimize instance.
+
+`combined` remains the default and preserves existing single-release behavior, so no existing deployment changes on upgrade. For a new production deployment, the split topology is the baseline.
+
+The `orchestration` role is also available in the 8.7, 8.8, and 8.9 charts, so one 8.10 Hub release can manage Orchestration Cluster releases on any supported chart version. `hub` and `optimize` are 8.10-only roles, because Camunda Hub and its cluster inventory don't exist in the earlier charts.
```
leiicamundi: This sentence is not true of any published chart, and a release note is the worst place for that.

The latest published 8.7, 8.8 and 8.9 charts are 12.13.8, 13.13.2 and 14.10.1, and none of them contains a `global.topology` key. The backports merged on 2026-09-22 (#7009, #7010, #7011) without a version bump, so those version numbers were already on helm.camunda.io built from artifacts without the feature. Rendering published 12.13.8 with `global.topology.mode: orchestration` produces output byte-identical to omitting it.

A reader acting on this announcement today sets the value, gets no error, and quietly ends up with a combined release. Please gate this paragraph on the patch releases that carry the role, and name the minimum version for each line.

docs/reference/announcements-release-notes/8100/8100-announcements.md
```diff
@@ -638,6 +638,40 @@ By default, this mount used an `emptyDir`, so no PVC cleanup is required. Howeve
...
+<div className="release-announcement-badge">
+<span className="badge badge--change">Change</span>
+</div>
+<div className="release-announcement-content">
+
+#### Camunda Hub database migration phases
+
+The 8.9 to 8.10 Camunda Hub database migration is controlled by `camundaHub.upgrade.phase`. Use `quiesce` to stop all Hub workloads so you can take a verified database backup, `migrate` to run the startup schema migration on a single pod without serving traffic, and `normal` to restore serving capacity. Fresh installs stay on `normal`.
+
+**Action:** Run the phases in order as part of your 8.9 to 8.10 upgrade, and plan a maintenance window: Hub serves no traffic in `quiesce` or `migrate`. The migration isn't backward compatible, so take a verified database backup first. See [migrate Camunda Hub](/self-managed/upgrade/helm/890-to-8100.md#migrate-camunda-hub).
```
leiicamundi: The known broken anchor will not clear on its own. The description says the warning goes away when #9803 merges. #9803 merged on 2026-09-23, and the anchor is still broken.

Headings on `origin/main` in `890-to-8100.md` now slug to `migrate-web-modeler-and-console-to-camunda-hub`, `back-up-the-hub-database` and `restore-hub-traffic`. There is no `migrate-camunda-hub`.

So rebasing will not fix this. Either point the link at `#migrate-web-modeler-and-console-to-camunda-hub`, or ask for an explicit `migrate-camunda-hub` heading in that page.

docs/reference/announcements-release-notes/8100/whats-new-in-810.md
```diff
@@ -513,6 +513,22 @@ Camunda for Slack joins Camunda for Microsoft Teams as a second chat platform se
...
 
+### A Hub plane and one or more execution planes
+
+The 8.10 Helm chart adds `global.topology.mode`, so each release declares its role in the deployment: `combined`, `hub`, `orchestration`, or `optimize`. One `hub` release running Camunda Hub and Management Identity can serve many independently deployed `orchestration` releases, each with its own lifecycle, scaling, and upgrade schedule.
+
+The new `optimize` role deploys Optimize alone. Because one Optimize instance reads a single index prefix, this is what lets each [Physical Tenant](/self-managed/concepts/multi-tenancy/physical-tenants.md) have its own Optimize instance.
+
+`combined` remains the default, so existing deployments are unchanged by the upgrade. For a new production deployment, the split topology is the baseline.
+
+The `orchestration` role is available in the 8.7, 8.8, and 8.9 charts too, so one 8.10 Hub can manage Orchestration Clusters across supported chart versions, and each cluster upgrades on its own schedule.
```
leiicamundi: Same problem as the matching paragraph in `8100-announcements.md:653`. The `orchestration` role is not in any published 8.7, 8.8 or 8.9 chart. Published 12.13.8, 13.13.2 and 14.10.1 have no `global.topology` key, and setting the mode on them is silently ignored.

Whatever wording you settle on in the announcement should be mirrored here so the two release notes do not drift.

leiicamundi (changes requested):
## Review: blocked on chart availability

Both existing reviews are still open, and the branch has not moved since they were written (head is still `9d221f5c`), so nothing from @bkenez or @hisImminence has been addressed yet. Rather than repeat their points, I re-ran the chart verification from scratch. The result changes the picture: most of what this PR documents cannot be installed from any chart that has been published.

### The main problem

The description says every values example was templated against chart `15.0.0-alpha5`. That version string is ambiguous right now. `Chart.yaml` on `camunda-platform-helm@main` still reads `15.0.0-alpha5` while carrying commits made after the `camunda-platform-8.10-15.0.0-alpha5` tag (2026-09-04). Templating a working copy of `main` passes. Templating the published artifact does not.

Against `camunda/camunda-platform:15.0.0-alpha5` pulled from helm.camunda.io:

```
$ helm template camunda camunda/camunda-platform --version 15.0.0-alpha5 -f optimize-values.yaml
Error: values don't meet the specifications of the schema(s) in the following chart(s):
camunda-platform:
- at '/global/topology/mode': value must be one of 'combined', 'hub', 'orchestration'
```

```
$ helm template ... -f hub-values-with-physicaltenants.yaml
- at '/global/topology/clusters/0': additional properties 'physicalTenants' not allowed
```

```
$ helm template ... -f hub-values-chart87.yaml
- at '/global/topology/clusters/0/components/orchestration': additional properties 'operateServiceName', 'tasklistServiceName' not allowed
- at '/global/topology/clusters/0': additional properties 'architecture' not allowed
```

The published `topologyCluster` definition accepts only `authorizations`, `components`, `contextPaths`, `host`, `id`, `name`, `namespace`, `releaseName` and `version`, with `additionalProperties: false`. The `optimize` role reached `main` in #6884 on 2026-09-10, six days after the alpha5 tag.

The older chart lines fail more quietly. The latest published 8.7, 8.8 and 8.9 charts (12.13.8, 13.13.2, 14.10.1) contain no `global.topology` key at all. The backports merged on 2026-09-22 in #7009, #7010 and #7011 without a version bump, so those version numbers were already on helm.camunda.io pointing at artifacts that predate the feature. Setting `global.topology.mode: orchestration` on published 12.13.8 renders output byte-identical to omitting it, once the randomly generated secrets are normalized. The reader gets a combined release and no error.

Of the four documented roles, only `hub` and `orchestration` on the 8.10 chart render today. `optimize-release.md`, the `physicalTenants` mapping, "Describe a chart 8.7 cluster" and the entire mixed-version section describe a chart that nobody can install yet.

This is a release-sequencing question, not a writing one. The pages are accurate about where the chart is heading. What they need is a minimum chart version per role, and a merge gated on those charts being published.

### Two earlier findings that still hold

@bkenez's points 1 and 3 survive against the published chart, so please do not treat them as stale now that #7008 to #7012 have merged:

- Point 3 stands. The 8.7 Hub record still fails schema validation, output above. The chart PRs merged, but the fix is unreleased.
- Point 1 stands. `orchestration.exporters.zeebe.enabled` defaults to `false` (`values.yaml:2664-2666`), and the orchestration example renders only `autoconfigure-camunda-exporter: true` with no legacy Elasticsearch exporter. `physical-tenants.md:91` and `optimize-release.md:94` both require the Optimize reader prefix to equal the legacy exporter writer prefix. A reader who follows both pages end to end lands on exactly the silent empty-dashboard failure the pages warn about.

### What I checked and found correct

- The `[camunda][error]` contract is exact. Removing `global.identity.service.url` from the orchestration example gives `[camunda][error] global.topology.mode=orchestration requires global.identity.service.url to reach Management Identity.`
- `orchestration.profiles`, `orchestration.secretStore.physicalTenants` (including the tenant-id character set and the shared-provider rule) and `global.noSecondaryStorage` all match the chart.
- The `grep 'DEPRECATION'` workflow in `configuration-responsibilities.md` works. `NOTES.txt:176` renders the warnings, so `helm install` and `helm upgrade` do print them.
- ADR 0095 lines 164 to 165 say the combined-to-split move "is not defined by this ADR", so that reviewer note is accurate.
- Every internal link and anchor across the 11 changed pages resolves, with one exception flagged inline.
- Nothing in `camunda-deployment-references` links to the removed `configure/multi-namespace.md`, so the InfraEx watcher has no breakage to report from our side.

### Structure

I went looking for duplication and layering problems and found very little. Across roughly 1,200 lines of new prose there are two exactly repeated sentences. The reference and how-to split is clean: `deployment-topology.md` explains why, the `install/topology/` pages explain how, and neither restates the other. Page sizes are sensible, the largest being `hub-release.md` at 241 lines. The `.htaccess` redirect and the sidebar entries are correct. I have no structural changes to ask for.

### Verdict

The structure is the strongest part of this PR and I want it merged. What blocks it is that it publishes install instructions for chart capabilities absent from every published chart, including two release notes telling readers the `orchestration` role is available on 8.7, 8.8 and 8.9 today. Details inline.

Verification: Helm 4.2.4 against `camunda/camunda-platform` 15.0.0-alpha5, 14.10.1, 13.13.2 and 12.13.8 pulled from helm.camunda.io, plus schema and template inspection of `camunda-platform-helm@main`. No live cluster, no upgrade rehearsal.

leiicamundi (commented):
Consolidated into the review above. Please read that one instead; this entry is an artifact of the GitHub API, which does not allow deleting a submitted review.

docs/self-managed/deployment/helm/install/topology/hub-release.md
```diff
@@ -0,0 +1,241 @@
...
+
+Install it first. For the prerequisites, Secrets, and network policies this page assumes, see [install the deployment topology](./index.md).
+
+## What a Hub release deploys
+
+`global.topology.mode: hub` deploys Camunda Hub and Management Identity, and suppresses the chart's default Orchestration Cluster, Optimize, and Connectors workloads. You don't configure disabled components in this release.
+
+The `hub` role requires the 8.10 chart. Camunda Hub and `global.topology.clusters` don't exist in the 8.7, 8.8, and 8.9 charts, so the Hub release is always the 8.10 chart even when it manages older clusters.
+
+Camunda Hub contains Web Modeler and Console. Console isn't a separate deployment in 8.10.
```
alexronquillo: I recommend we keep references to old applications, like Web Modeler and Console, out of this guide. Instead, we should refer users to the migration guide, if they're interested. That's because the migration guide targets existing users who are familiar with these concepts, whereas this documentation targets both existing and new users, many of which might not understand what we're referring to with this comment.

docs/self-managed/deployment/helm/install/topology/hub-release.md
```diff
@@ -0,0 +1,241 @@
...
+description: "Install the management plane: a Helm release with global.topology.mode set to hub, running Camunda Hub and Management Identity."
+---
+
+The Hub release is the management plane. It runs Camunda Hub and Management Identity, and it owns the inventory of every Orchestration Cluster in the deployment.
+
+Install it first. For the prerequisites, Secrets, and network policies this page assumes, see [install the deployment topology](./index.md).
+
+## What a Hub release deploys
+
+`global.topology.mode: hub` deploys Camunda Hub and Management Identity, and suppresses the chart's default Orchestration Cluster, Optimize, and Connectors workloads. You don't configure disabled components in this release.
```
alexronquillo: I agree with the framework of setting expectations with introductory sections like this. However, referring to technical details like `global.topology.mode` before providing the context for those details could confuse readers.

I recommend we stick to a higher-level introduction, and only refer to code details below, after the **Create hub-values.yaml** section. This aligns with our general docs strategy for guides: 

1. Instruct the reader what to do
2. Explain that instruction to the reader

docs/self-managed/deployment/helm/install/topology/hub-release.md
```diff
@@ -0,0 +1,241 @@
...
+| `namespace`, `releaseName`        | Used to generate in-cluster service endpoints                                                  |
+| `host`                            | Public hostname of the Orchestration Cluster                                                   |
+| `version`                         | The Camunda version deployed by that release                                                   |
+| `architecture`                    | `unified` (default) or `legacy`. Set `legacy` for a chart 8.7 cluster                          |
+| `contextPaths`                    | Sub-paths each component is served on                                                          |
+| `components.<component>`          | Enabled state, `clientId`, `audience`, `redirectUrl`, and `secret` for each workload component |
+| `components.<component>.roleName` | A per-cluster role name, instead of the shared canonical role                                  |
+| `physicalTenants`                 | One entry per Physical Tenant that runs its own Optimize release                               |
+
+## Create `hub-values.yaml`
```
alexronquillo: Should we introduce this section with a sentence explicitly clarifying this is an example?

docs/self-managed/deployment/helm/install/topology/hub-release.md
```diff
@@ -0,0 +1,241 @@
...
+
+## Provider setup outside the chart
+
+Management Identity can administer clients only in an external Keycloak instance, and only when you provide Keycloak administrator credentials.
+
+For Microsoft Entra ID or a generic OIDC provider, first complete the provider setup, including Management Identity's confidential client, initial administrator claims, Hub clients, mapping rules, and each workload client. Then add the matching client IDs and audiences to the topology records. Management Identity initializes only its permission and role model from those records. See [external OIDC provider](/self-managed/deployment/helm/configure/authentication-and-authorization/external-oidc-provider.md).
+
+## Role assignment across clusters
+
+By default, every cluster contributes permissions to the canonical `Orchestration` and `Optimize` roles. Assigning either role grants access to every declared cluster of that component type. Set `components.<component>.roleName` to a unique value in each Hub topology entry when users must be authorized per cluster.
```
alexronquillo:
```suggestion
By default, every cluster of a given component type shares the canonical `Orchestration` or `Optimize` role. Assigning a user one of these roles grants access to every cluster of that type. To authorize a user to a specific cluster only, set components.<component>.roleName to a unique value in that cluster's Hub topology entry.
```

docs/self-managed/deployment/helm/install/topology/index.md
```diff
@@ -0,0 +1,131 @@
...
+sidebar_label: "Install the deployment topology"
+title: "Install the Camunda 8.10 deployment topology"
+description: "Install Camunda 8.10 Self-Managed as separate Hub, Orchestration Cluster, and Optimize Helm releases."
+---
+
+import HelmV4Required from '../../_partials/_helm-v4-required.md'
+
+Install Camunda 8.10 Self-Managed as separate Helm releases: one Hub release, one release per Orchestration Cluster, and one Optimize release per Physical Tenant.
+
+This is the baseline topology for a new 8.10 production deployment. Each release declares its role through `global.topology.mode`, so the Hub plane and each execution plane have independent lifecycles. For the reasoning, the release-role reference, and the limits of this model, see [Camunda 8.10 deployment topology](/self-managed/reference-architecture/deployment-topology.md).
```
alexronquillo: > the **Hub plane** and each execution plane

I don't think we're consistent with our terms. For example:

- The Hub release is the **management plane** (hub-release.md)
- ...still bundle **Hub-plane** dependencies (orchestration-release.md)

I recommend we use "management plane" consistently throughout these docs.

docs/self-managed/deployment/helm/install/topology/hub-release.md
```diff
@@ -0,0 +1,241 @@
...
+| `architecture`                    | `unified` (default) or `legacy`. Set `legacy` for a chart 8.7 cluster                          |
+| `contextPaths`                    | Sub-paths each component is served on                                                          |
+| `components.<component>`          | Enabled state, `clientId`, `audience`, `redirectUrl`, and `secret` for each workload component |
+| `components.<component>.roleName` | A per-cluster role name, instead of the shared canonical role                                  |
+| `physicalTenants`                 | One entry per Physical Tenant that runs its own Optimize release                               |
+
+## Create `hub-values.yaml`
+
+```yaml
+global:
```
alexronquillo: Should we also include an exhaustive reference table with explanations like we do for other configurations?

docs/self-managed/deployment/helm/install/topology/orchestration-release.md
```diff
@@ -0,0 +1,202 @@
...
+:::
+
+For Keycloak, Management Identity creates every declared client. For another OIDC provider, provision the clients before applying the Helm releases.
+
+Generated internal service URLs in the Hub inventory use Kubernetes service DNS, so this pattern supports multiple namespaces in the same Kubernetes cluster. For workloads in another Kubernetes cluster, provide equivalent cross-cluster DNS and routing, or configure explicit `grpcUrl`, `restUrl`, `readinessUrl`, `operateUrl`, `tasklistUrl`, `adminUrl`, and component web application URL overrides. Set each orchestration release's `global.identity.service.url` to an address from which it can reach Management Identity.
+
+## Declare Physical Tenants
+
+Physical Tenants are application configuration, not chart values. There's no `orchestration.physicalTenants` values key. Declare tenants as `camunda.physical-tenants.*` through `orchestration.extraConfiguration`:
```
alexronquillo: Link to configuration-responsibilities?

docs/self-managed/concepts/physical-tenants/index.md
alexronquillo: Link to configuration-responsibilities?

docs/self-managed/upgrade/helm/combined-to-split-topology.md
```diff
@@ -0,0 +1,143 @@
+---
+id: combined-to-split-topology
+sidebar_label: Move to the split topology
+title: Move from a combined release to the split topology
+description: Plan and execute the move from a single combined Camunda 8.10 Helm release to separate Hub, Orchestration Cluster, and Optimize releases.
+---
+
+Move an existing single-release Camunda 8.10 deployment to the split topology: one Hub release, one release per Orchestration Cluster, and one Optimize release per Physical Tenant.
+
+This is a topology change, not a version upgrade. It doesn't change any component version, and it isn't required. A `combined` release remains supported and remains the chart default.
```
alexronquillo:
```suggestion
This is a topology change, not a version upgrade. It doesn't change any component version, and it isn't required. A `combined` release remains both supported and the chart default.
```

docs/self-managed/upgrade/helm/combined-to-split-topology.md
```diff
@@ -0,0 +1,143 @@
...
+
+## When to make this move
+
+Make it when you need something the combined release can't give you:
+
+- Several Orchestration Clusters sharing one Camunda Hub and one Management Identity.
+- Independent upgrade, scaling, or removal of a cluster without touching the management plane.
+- Physical Tenants with a separate Optimize instance per tenant.
+
+If none of those apply, staying on a combined release is a reasonable long-term choice.
```
alexronquillo: Is it just "reasonable", or is it "recommended"?

docs/self-managed/upgrade/helm/combined-to-split-topology.md
```diff
@@ -0,0 +1,143 @@
...
+| Optimize              | Yes            | Its state is in Elasticsearch or OpenSearch, reached by index prefix                                                      |
+| Orchestration Cluster | **No**         | Broker PVCs are StatefulSet volume claim templates holding partition logs and snapshots, which are the live process state |
+
+Secondary storage doesn't substitute for broker storage. Installing a fresh orchestration release creates a new, empty Orchestration Cluster: the Operate and Tasklist data in Elasticsearch describes process instances whose authoritative state lives in the broker volumes you left behind.
+
+:::warning
+Switching an existing release to `global.topology.mode: hub` suppresses its Orchestration Cluster StatefulSet. The PVCs remain, but the cluster stops. Never flip a release running your only Orchestration Cluster to `hub` mode.
+:::
+
+## Strategy 1: keep the cluster in place (recommended)
```
alexronquillo:
```suggestion
## Strategy 1: Keep the cluster in place (recommended)
```

Typically, we keep the first word capitalized in cases like this and the following steps:

**Step 1: Inventory what the combined release owns**

docs/self-managed/upgrade/helm/combined-to-split-topology.md
```diff
@@ -0,0 +1,143 @@
...
+
+Keep the existing release and namespace as the orchestration release, and stand up a new Hub release alongside it. Broker storage and cluster identity never move, so there's no process-state cutover.
+
+### Step 1: inventory what the combined release owns
+
+Record, from your current values file and cluster:
```
alexronquillo: Style: start with _where_ and end with _what_:

```suggestion
From your current values file and cluster, record:
```

docs/self-managed/upgrade/helm/combined-to-split-topology.md
```diff
@@ -0,0 +1,143 @@
...
+
+See [install an Optimize release](/self-managed/deployment/helm/install/topology/optimize-release.md).
+
+### Step 6: clean up
+
+- Confirm no workload still resolves the old in-release Management Identity or Hub service names.
+- Inventory the OIDC clients, resource servers, permissions, and roles. Identity initialization is additive, so the combined release's objects still exist. Remove only what no release uses.
+- Retire the old Hub hostname and its TLS certificate, or redirect it.
+
+## Strategy 2: drain and re-create
```
alexronquillo:
```suggestion
## Strategy 2: Drain and recreate
```

alexronquillo (changes requested):
Thanks @eamonnmoloney I left some comments from a stylistic/structural perspective.

Issue #9983, opened 2026-09-18
(Tenants) Fix cross-page technical contradictions and restated facts that have drifted


christinaausley:
Relates to https://github.com/camunda/camunda-docs/issues/8886. Fix cross-page technical contradictions and restated facts that have drifted:

**Branch:** tenant-issue-15-guide
**Priority:** Mixed — two items below are marked Critical because they cause silent misconfiguration or startup failure, not just a wrong sentence.
**Root cause, per the review:** the same fact is written out in full on 3–7 different pages instead of stated once and linked. Every contradiction below is a restatement that drifted from the original. Fixing this properly means correcting the wrong copies now, *and* reducing duplication so it can't drift again (see the canonicalization task at the end).

**Do not touch:** the review separately verified the following as already accurate — leave their content as-is: `java-client/physical-tenants.md`, `cluster-scaling.md`, `890-to-8100.md`, `cluster-admin.md`'s unauthenticated-status note, and everything listed under "§1.21 Verification results" and "§5 Verified correct" in the source review (tenant ID regex, `camunda.physical-tenants.<id>` prefix, per-tenant `security.initialization` requirement, secret store/cache validation, RDBMS uppercase normalization, Oracle `database-vendor-id`, ES/OS overlapping-prefix validation, document-store nesting rules, the `503`/`Retry-After: 5` behavior, retry backoff, the readiness/schema-init metrics, `BACKUP`/`EXPORTER` resource types and role grants, `cluster-admin` OIDC/Basic config, `Camunda-Physical-Tenant` header, client builder methods/env vars, `camunda.clients.*` properties, and `POST /actuator/cluster/purge?physicalTenant=`).

### CRITICAL 1. `partitions-count` is not a real property — it's `partition-count`

Every per-tenant cluster-sizing example uses the wrong spelling. The real property (confirmed in `configuration/.../Cluster.java:89` and `dist/src/main/config/defaults.yaml:210`, env var `CAMUNDA_CLUSTER_PARTITIONCOUNT`) is `partition-count`. `partitions-count` appears nowhere in the codebase — Spring relaxed binding will not map it, so it's **silently ignored** and the tenant starts with one partition, with **no startup error**. This is the only defect in the whole review that produces a silently broken configuration rather than a startup failure.

[...]

PR #9986, opened 2026-09-18
(Tenants) Fix cross-page technical contradictions and restated facts that have drifted

christinaausley:
## Description

This PR fixes documentation drift in the Physical Tenants guidance for Self-Managed deployments and aligns the examples with the verified runtime behavior. Closes https://github.com/camunda/camunda-docs/issues/9983.

It corrects:
- incorrect per-tenant examples that use `partitions-count` instead of `partition-count`
- incomplete copy-pasteable configuration examples that omit required `security.initialization` blocks for explicitly configured non-default tenants when authorization is enabled
- inconsistent guidance about default-tenant routing on the REST `/v2/...` API versus the actuator surface
- contradictory wording around tenant deletion versus purge
- invalid example tenant IDs such as `tenant-a` in runnable commands and examples

These updates reduce the risk of silent misconfiguration and startup failure for users following the docs, while keeping the guidance consistent with the actual behavior of the cluster.

The changes also clarify the important distinction between:
- routing unscoped REST requests to the default tenant
- actuator operations such as purge, which do not follow the same default-tenant rule
- disabling a tenant in configuration versus purging data or removing a tenant’s configuration

This change is documentation-only and does not alter verified runtime behavior or unrelated tenant docs.


docs/components/admin/cluster-admin.md
```diff
@@ -34,7 +34,7 @@ Cluster admin protects the operations served under the `/cluster/v2/...` path pr
 | Recovery            | `POST /cluster/v2/restore`, `PATCH /cluster/v2/mode`                                                 |
 | Partition placement | `POST /cluster/v2/rebalance`                                                                         |
 
-Each cluster-wide endpoint also accepts an optional `physicalTenantId` query parameter, which narrows the same cluster-admin operation to a single Physical Tenant without switching to the tenant-scoped API. Omitting the parameter targets every Physical Tenant.
+The Recovery endpoints (`POST /cluster/v2/restore` and `PATCH /cluster/v2/mode`) also accept an optional `physicalTenantId` query parameter, which narrows the same cluster-admin operation to a single Physical Tenant without switching to the tenant-scoped API. Omitting the parameter targets every Physical Tenant. The other cluster-wide endpoints don't take this parameter; Status and topology already report per-tenant detail in the response body.
```
deepthidevaki: Other endpoints such as backups and exporting also takes physicalTenantId as parameter. https://github.com/camunda/camunda/blob/main/zeebe/gateway-protocol/src/main/proto/v2/cluster-admin.yaml 

I think, there is no need to call out every endpoint here. When we add new endpoints this doc gets outdated. Instead it would be better to just link to the api spec page  - because it would be auto-updated from the open-api spec file.

docs/self-managed/operational-guides/backup-restore/backup-and-restore.md
```diff
@@ -220,7 +220,7 @@ Both scopes serve the same operations:
 | History backup    | `/physical-tenants/{physicalTenantId}/v2/backups/history`       | `/cluster/v2/backups/history`       |
 | Exporting control | `/physical-tenants/{physicalTenantId}/v2/exporting`             | `/cluster/v2/exporting`             |
 
-Each cluster-wide endpoint also accepts an optional `physicalTenantId` query parameter, which narrows the same cluster-admin call to one tenant. Omit it to target every Physical Tenant.
+These backup and exporting endpoints don't accept a `physicalTenantId` query parameter to narrow the call to one tenant; only the Recovery endpoints (`POST /cluster/v2/restore` and `PATCH /cluster/v2/mode`) do. A cluster-wide backup or exporting call always fans out to every configured tenant.
```
deepthidevaki: :x: This is not correct. See https://github.com/camunda/camunda/blob/main/zeebe/gateway-protocol/src/main/proto/v2/cluster-admin.yaml

deepthidevaki (changes requested):
Thanks. There are some minor corrections.

christinaausley:
Thanks @deepthidevaki! I removed the endpoint-specific claims that only Recovery endpoints accept `physicalTenantId` and that backup/exporting cluster-wide calls cannot be narrowed to one tenant. Now the docs avoid trying to enumerate parameter support in prose and point to the generated ocluster REST API reference so the page stays aligned as the spec changes.

I also adjusted the backup/restore wording so cluster-wide requests without a tenant parameter fan out to all tenants without implying every cluster-wide call always does.

PR #9999, opened 2026-09-21
Revert "chore(deps): lock file maintenance (#8552)"

alexronquillo:
This reverts commit e95b1c934bba8246dc33fecc85c17421ab344525.


kellervater (approved):
weird that only the package-lock file got adapted in the first place.
tbh: never heard of a pure lockfile maintenance. 
anyways. the revert is legit for this specific PR.

PR #10015, opened 2026-09-22
docs(variables): document var propagation for duplicate names

### LINKED ISSUE #62940: Document that a local variable blocks propagation of any later write with the same name
### 

### Description

When a variable already exists locally in a scope — for example, one created by an
input mapping, or by a start execution listener completing its job with
variables — any later write of that same name from within that scope updates the
local copy instead of propagating to the process instance. The local copy, along
with that update, is discarded once the scope is exited.

This follows directly from two rules that are each documented on their own:
variable propagation stops at the first scope that already holds a variable of that
name, and local variables are removed when their scope exits unless explicitly
propagated with an output mapping. The "Local variables" section already gestures at
the second half of this ("Remember: Local variables are removed when a scope is
exited unless you explicitly propagate them with output mappings"), but doesn't
spell out the specific consequence: a same-named write from later in that scope
silently updates the doomed local copy rather than reaching the process instance,
with no error or signal that it happened.

Support context [here](https://camunda.slack.com/archives/C0C1N9A3GQH/p1789561734024309).

### Motivation

A completion command that includes this kind of write returns success. No incident
is raised, nothing rejects the request. The value is simply gone from the rest of
the process the moment the scope that held it closes. This is easy to trigger by
accident — naming a variable the same as an existing local one anywhere in that
scope — and expensive to diagnose without documentation calling out the
consequence explicitly.

### Suggested phrasing 
draft:

```markdown
:::warning A local variable blocks later writes of the same name
If a later operation writes a variable with the same name from inside the scope, variable propagation finds the local variable first. For example, this can happen when another job completes or an input mapping writes the variable.

The later write updates the local variable instead of the process instance. When the scope exits, Camunda discards the local variable and its updated value. The write succeeds without indicating that the new value remained local to the scope.

For example, an input mapping creates a local variable `x`. The element's job later completes with a new value for `x` without an output mapping. The new value updates the local `x`, not the process instance. The next element in the process still sees the previous value of `x`.

To make a variable visible outside the scope where it was created, propagate it explicitly with an [output mapping](#output-mappings).
:::
```

### Acceptance criteria

- [x] Add a short callout in the "Local variables" section of variables.md
      (docs.camunda.io), placed directly after the existing "When to use local
      variables" tip block
- [x] Callout explains: a later write of the same name from within that scope
      updates the local variable, not the process instance, and that update is
      discarded along with the scope when it exits
- [x] Point readers to output mappings as the way to propagate a value out of the
      scope it was created in
- [x] No separate edit needed elsewhere: this section is already linked from the
      "start listeners" part of execution-listeners.md and from the "variable
      mappings" section of user-tasks.md, so one edit here is discoverable from both

### Links

Related to https://jira.camunda.com/browse/SUPPORT-34571

HeleneW-dot:
related to [#62940](https://github.com/camunda/camunda/issues/62940)

## Description
Add a more explicit notice to make variable propagation behaviour more clear in the case of local same named variables.

## PR Checklist


HeleneW-dot:
@giorgionaps I also added the notice to the past versions but in a different section as the structure changed, lmk if it doesnt fit there

giorgionaps (approved):
Thanks for making the changes! I updated the wording a bit, lgtm

PR #10016, opened 2026-09-22
docs(optimize): document removal of the static API access token as an 8.10 breaking change

megglos:
## Description

Documents the removal of Optimize's static API access token as an 8.10 breaking change. See [camunda/camunda#63349](https://github.com/camunda/camunda/issues/63349).

Under the Camunda Security Library, which is the default in 8.10, Optimize accepts only OIDC bearer tokens on its API. Requests that carry the static token from `api.accessToken` (environment variable `OPTIMIZE_API_ACCESS_TOKEN`) get a `401`, on both the Optimize API and the external variable ingestion endpoint. The product decision is to keep the removal and announce it, so these docs stop presenting the static token as a supported option.

Research behind the scope:

- **8.9 behavior:** the single 8.9 Self-Managed stack chose its public API decoder from `api.jwtSetUri` alone, and fell back to the static token decoder when that key was unset. The static token therefore worked in 8.9 with OIDC login configured, and it was the only accepted API credential when `api.jwtSetUri` was unset.
- **Affected deployments:** SaaS never set the token, and `camunda-operator` always sets the JWKS URI. Both Helm charts (8.9 and 8.10) render `api.jwtSetUri` whenever Optimize auth is enabled, so the static token was already ignored there. The affected group is manual, Docker, and Docker Compose installs, plus Helm installs with Optimize auth disabled.

Changes:

- `8100-announcements.md`: new breaking-change entry, `#optimize-static-api-access-token-is-no-longer-supported`.
- `system-configuration.md`: rewrote the **External API** section, and marked the `api.*` and `security.*` rows that 8.10 deprecates or no longer uses, so the status is visible in the table rather than only in a note.
- `890-to-8100.md`: moved the token out of the flat "keys with no replacement" list into its own section with the symptom, the migration action, and a note to check the YAML spelling as well as the environment variable.
- `optimize-api-authentication.md`: the Self-Managed note no longer offers the shared token. Also fixed a broken `#public-api` anchor and pointed the audience step at `camunda.security.authentication.oidc.audiences`.
- Four Optimize API example pages: replaced the `accessToken: mySecret` examples with bearer tokens.
- `authentication-to-optimize.md`: new **Authenticate API requests** section, which the announcement links to.

New prose follows ASD-STE100.

Validation: `npm run build` passes, and none of the new links appear in the broken-anchor warnings.

## PR Checklist

---


docs/self-managed/components/optimize/configuration/system-configuration.md
```diff
@@ -112,20 +110,15 @@ These values relate to Optimize data import.
...
-configured an error will be thrown and all requests to the External API will get rejected. If both are configured then
-the `jwtSetUri` will take precedence and the `accessToken` will be ignored.
+These values control how Optimize validates tokens on its External REST API.
 
-:::note
-In Camunda 8.10, set `camunda.security.authentication.oidc.jwk-set-uri` and `camunda.security.authentication.oidc.audiences` instead of `api.jwtSetUri` and `api.audience`. The `api.accessToken` setting is no longer used, and you can remove it. See [legacy configuration keys](/self-managed/upgrade/components/890-to-8100.md#legacy-security-configuration-keys-are-deprecated) for the full mapping.
-:::
+In Camunda 8.10, the External REST API accepts only OIDC bearer tokens. Configure `camunda.security.authentication.oidc.jwk-set-uri` and `camunda.security.authentication.oidc.audiences`. If you configure no key set, Optimize rejects all requests to the External REST API. For the upgrade steps, see [Optimize static API access token is no longer supported](/reference/announcements-release-notes/8100/8100-announcements.md#optimize-static-api-access-token-is-no-longer-supported).
```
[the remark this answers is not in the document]
megglos: Agreed, fixed in 80c7467. The section now names `issuer-uri` and `audiences` as the settings to configure, states that Optimize gets the JWK set URI from the issuer's discovery document, and presents `jwk-set-uri` as the override for a different URI. This matches the shared security property reference, which describes `jwk-set-uri` as overriding the discovered value.

docs/self-managed/components/optimize/configuration/system-configuration.md
```diff
@@ -112,20 +110,15 @@ These values relate to Optimize data import.
...
-| api.jwtSetUri   | SPRING_SECURITY_OAUTH2_RESOURCESERVER_JWT_JWK_SET_URI | null          | Complete URI to get public keys for JWT validation, e.g. `https://weblogin.cloud.company.com/.well-known/jwks.json`                            |
-| api.audience    | CAMUNDA_OPTIMIZE_API_AUDIENCE                         | optimize      | Optimize tries to match this with the `aud` field contained in the JWT token. Only used when `jwtSetUri` is set.                               |
+| YAML path       | Environment variable                                  | Default value | Description                                                                                                                            |
+| --------------- | ----------------------------------------------------- | ------------- | -------------------------------------------------------------------------------------------------------------------------------------- |
+| api.accessToken | OPTIMIZE_API_ACCESS_TOKEN                             | null          | Not supported in 8.10. Through 8.9, a static shared token that a client sends to the secured REST API in the authorization header.     |
```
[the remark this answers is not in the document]
megglos: Good catch, fixed in 80c7467. The `api` block in the sample `service-config.yaml` no longer presents the static token as one of two supported options. It now points readers at the `camunda.security.authentication.oidc` settings and marks `accessToken` as 8.9 only. The keys themselves stay in the sample, because the file mirrors the configuration Optimize ships.

megglos:
@copilot review

I addressed both findings from your first review in 80c7467c68. Please re-review the full diff, including the wording of the new sections in `8100-announcements.md`, `890-to-8100.md`, and `authentication-to-optimize.md`.

docs/self-managed/components/optimize/configuration/system-configuration.md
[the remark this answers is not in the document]
megglos: Addressed in caf14b202f. The sample now marks `accessToken` as unsupported in 8.10 and fallback-only (`optimize.security.csl.enabled=false`), and notes that 8.11 removes the fallback. I keep the key in the sample because 8.10 still ships it for that fallback, and the configuration table lists it with the same marker.

megglos:
Cycle 2 addressed in caf14b202f:

- **Fallback migration path** (`8100-announcements.md` and `890-to-8100.md`): the `Action` paragraph now ends after the removal step, and a separate `Alternative` paragraph states that the fallback keeps `api.accessToken` in use, that 8.11 removes the fallback, and that clients must move before the 8.11 upgrade.
- **Sample `service-config.yaml`**: `accessToken` is marked unsupported in 8.10 and fallback-only. The key stays in the sample because 8.10 still ships it for that fallback.

megglos:
@copilot please review the full diff of this PR once more as a reviewer, not as an implementer.

Scope:
- Technical accuracy of the statements about `api.accessToken`, `api.jwtSetUri`, `api.audience`, `optimize.security.csl.enabled`, and the `camunda.security.authentication.oidc.*` replacements.
- Consistency between `8100-announcements.md`, `890-to-8100.md`, `system-configuration.md`, `service-config.yaml`, `authentication-to-optimize.md`, and the Optimize API pages.
- Link targets and heading anchors.
- Style: brief ASD-STE100 sentences, and no description of internal behavior.

Report findings as comments. Do not push commits unless a finding is a clear factual error, and explain any commit you push. If you have no findings, say so explicitly.

megglos:
Cycle 3 addressed:

- **Wording vs the 8.9 fallback**: `optimize-api-authentication.md`, `authentication-to-optimize.md`, and the `api.accessToken` row in `system-configuration.md` now state that only the temporary 8.9 fallback still accepts the static token. The concepts page links to its own fallback section.
- **Wrong base URL variable**: the SaaS tip in `optimize-api-authentication.md` named `CAMUNDA_TASKLIST_BASE_URL`. It now names `CAMUNDA_OPTIMIZE_BASE_URL`, which matches the request example below it and the client-credentials table on the same page. Good catch, this one predates the PR.

@copilot please review the current diff again as a reviewer. Report findings, or state explicitly that you have none.

docs/self-managed/components/optimize/configuration/system-configuration.md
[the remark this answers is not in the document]
megglos: Reverted in d77e80365c: this PR no longer touches the reference `service-config.yaml`. The whole file still advertises legacy properties (`security.auth.ccsm.*`, `security.auth.token.secret`, `security.responseHeaders.*`), so we adopt it away from them in one separate change instead of patching the `api` block here. The 8.10 status of `api.accessToken`, `api.jwtSetUri`, and `api.audience` stays documented in the External API table on this page, in the upgrade guide, and in the announcement.

megglos:
Scope update: this PR no longer changes `docs/self-managed/components/optimize/configuration/service-config.yaml` (reverted in d77e80365c).

The reference configuration still lists other legacy properties, so adopting it away from legacy keys happens in a follow-up change rather than in the `api` block alone. The breaking change itself remains documented in the announcement, the 8.9-to-8.10 upgrade guide, the Optimize system configuration table, the Optimize authentication concept page, and the Optimize API pages.

megglos:
Follow-up for the reference configuration: #10021

docs/apis-tools/optimize-api/report/get-report-ids.md
```diff
@@ -38,23 +38,23 @@ The response contains a list of IDs of the reports existing in the collection wi
...
 
-Assuming you want to retrieve all report IDs in the collection with the ID `1234` and have configured the accessToken `mySecret`, this is what it would look like:
+To get all report IDs in the collection with the ID `1234`, send the following request:
 
 GET `/api/public/report?collectionId=1234`
 
 #### Request header
 
-`Authorization: Bearer mySecret`
+`Authorization: Bearer <TOKEN>`
```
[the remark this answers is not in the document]
megglos: Valid, fixed in 8c6a7dbc93. All remaining Optimize API pages now use the `Authorization: Bearer <TOKEN>` placeholder and a `Token incorrect or missing in HTTP header.` 401 description: `import-entities.md`, `delete-process-definition-data.md`, `variable-labeling.md`, `external-variable-ingestion.md`, `configuration/enable-sharing.md`, `configuration/disable-sharing.md`, `dashboard/delete-dashboard.md`, `report/delete-report.md`, and `report/get-data-export.md` (both request headers). The "Let's assume ... this is what it would look like" lead-ins on those pages are now short instructions.

`tutorial.md` keeps its `accessToken` variable, because that value comes from the OIDC token endpoint, not from `api.accessToken`.

docs/reference/announcements-release-notes/8100/8100-announcements.md
```diff
@@ -787,6 +787,25 @@ Starting with Camunda 8.10, Optimize authenticates through the [Camunda Security
...
+</div>
+<div className="release-announcement-content">
+
+#### Optimize static API access token is no longer supported
+
+In Camunda 8.10, Self-Managed Optimize accepts only OIDC bearer tokens on its API. The static token from `api.accessToken` (environment variable `OPTIMIZE_API_ACCESS_TOKEN`) is not accepted. Requests with this token get a `401` response. This applies to the [Optimize API](/apis-tools/optimize-api/overview.md) and to the [external variable ingestion](/apis-tools/optimize-api/external-variable-ingestion.md) endpoint.
+
+**Action:** Change these API clients to OIDC bearer tokens. Then remove `api.accessToken` from your configuration.
+
+**Alternative:** To keep the 8.9 behavior for the duration of 8.10, set `optimize.security.csl.enabled=false` and keep `api.accessToken`. Camunda 8.11 removes this fallback. Change the API clients before you upgrade to 8.11.
```
StevePascoe: What happens during an upgrade from 8.9 to 8.10? 

I would rephrase this "Camunda 8.11 removes this fallback. Change the API clients before you upgrade to 8.11."  to "Camunda 8.11 will remove this fallback, please update your API clients before upgrading to 8.11"

docs/self-managed/concepts/authentication/authentication-to-optimize.md
```diff
@@ -35,6 +35,12 @@ Note the following:
 If you deploy with the Camunda Helm chart, you don't need to set these directly. The chart continues to read the same `global.identity.auth.optimize.*` values you already use, and renders them into the properties above for you.
 :::
 
+## Authenticate API requests
+
+The Optimize API accepts OIDC bearer tokens. A client gets a token from your IdP with the client credentials flow, then sends the token in the `Authorization` header. Add the audience that your IdP issues for the client to `camunda.security.authentication.oidc.audiences`.
+
+Starting with Camunda 8.10, the static token `api.accessToken` is not supported. Only the temporary [8.9 fallback](#fall-back-to-the-89-security-stack) still accepts it. For the token steps, see [Optimize API authentication](/apis-tools/optimize-api/optimize-api-authentication.md).
```
StevePascoe: is this the right phrasing? It's still supported but is now on the path to deprecation in 8.11.  

Maybe we rephrase this "Starting with Camunda 8.10, the static token `api.accessToken` is not supported." to "Starting with Camunda 8.10 support with the static token 'api.accessToken' is being phased out, with full deprecation targeting 8.11. To continue to use the static token in 8.10 while you transition to OIDC bearer tokens, please refer to [Optimize API authentication](/apis-tools/optimize-api/optimize-api-authentication.md)"

docs/self-managed/upgrade/components/890-to-8100.md
```diff
@@ -497,10 +497,17 @@ The following keys are no longer used in 8.10. Optimize still starts if you leav
 - `CAMUNDA_OPTIMIZE_SECURITY_AUTH_TOKEN_SECRET`
 - `CAMUNDA_OPTIMIZE_SECURITY_AUTH_COOKIE_MAX_SIZE`
 - `CAMUNDA_OPTIMIZE_SECURITY_AUTH_COOKIE_SAME_SITE_ENABLED`
-- `OPTIMIZE_API_ACCESS_TOKEN`
 - `security.responseHeaders.X-XSS-Protection`
 
-If you use `OPTIMIZE_API_ACCESS_TOKEN` to call the Optimize API, switch those callers to OIDC bearer tokens.
+### Static API access token is not supported
```
StevePascoe: Similar comment to the above. We're phasing support out.

StevePascoe (changes requested):

docs/reference/announcements-release-notes/8100/8100-announcements.md
megglos: Good question, and it is the core of the change: with 8.10 defaults, the rejection starts the moment you upgrade. The Camunda Security Library is on by default in 8.10, and it validates only OIDC bearer tokens, so an 8.9 client that sends the static token gets a `401` on its next call. There is no grace period unless you set `optimize.security.csl.enabled=false`. I made this explicit in the upgrade guide ("Directly after the upgrade, Optimize rejects each request …") and moved the client change before the upgrade in the **Action** line.

On the rephrasing: I kept present tense and dropped `please`, because the style guide bans `please` in instructional steps (`.github/instructions/content.instructions.md` §1). The sentence now reads: "This fallback is available for the duration of 8.10. Camunda 8.11 removes the fallback and the static token, so complete the move before you upgrade to 8.11." Happy to switch to "will remove" if you prefer the future tense.

megglos: Good question, and it is the core of the change: the rejection starts on the first call after the upgrade. The Camunda Security Library is the 8.10 default, and it validates only OIDC bearer tokens, so an 8.9 client that sends the static token gets a `401`. There is no grace period unless you opt into `optimize.security.csl.enabled=false`. The upgrade guide now says this directly ("Directly after the upgrade, Optimize rejects each request …"), and the **Action** line moves the client change before the upgrade.

I took your "will remove" phrasing. I left out `please`, because the style guide bans it in instructional steps (`.github/instructions/content.instructions.md` §1). The sentence now reads: "Camunda 8.11 will remove this fallback and the legacy configuration keys. Migrate these API clients during 8.10."

I also added a **Check if you are affected** paragraph to the upgrade guide: the Helm chart and SaaS configure OIDC and never set `api.accessToken`, so only operators who set the property or `OPTIMIZE_API_ACCESS_TOKEN` themselves are affected.

docs/self-managed/concepts/authentication/authentication-to-optimize.md
megglos: I checked the behavior before writing this, and the token is not supported in 8.10 with default settings: the Camunda Security Library is on by default and validates only OIDC bearer tokens, so requests with the static token get a `401`. That regression is what this PR documents ([camunda/camunda#63349](https://github.com/camunda/camunda/issues/63349)). "Still supported, deprecation targeting 8.11" would tell operators they can upgrade and carry on, and their API clients would break.

Your point about the transition path is right though, so I reframed it in ab6e9b21d6 with the opt-in made explicit:

"Camunda 8.10 no longer accepts the static token `api.accessToken` on this API. The token works only if you opt into the [8.9 security fallback](#fall-back-to-the-89-security-stack), which Camunda 8.11 will remove. Migrate these API clients to OIDC bearer tokens during 8.10."

The same framing is now in the announcement, the upgrade guide, the Optimize API authentication page, and the `api.accessToken` row of the system configuration table.

docs/self-managed/upgrade/components/890-to-8100.md
megglos: Same answer as on the concepts page: with 8.10 defaults the static token is rejected with a `401`, so the wording keeps "no longer accepted", and the opt-in fallback carries the transition. Changes in this section:

- Heading is now **Static API access token is no longer accepted**.
- The first paragraph states that Optimize rejects these requests directly after the upgrade.
- New **Check if you are affected** paragraph: the Helm chart and SaaS configure OIDC and do not set `api.accessToken`.
- **Alternative** now reads: the token works only with the 8.9 security fallback, Camunda 8.11 will remove the fallback and the legacy keys, migrate the clients during 8.10.

megglos:
@StevePascoe thanks, addressed in 1eb6f6651a, ab6e9b21d6, and 1e02e8eea7, with answers in each thread.

Summary of the wording now used across the announcement, the 8.9-to-8.10 upgrade guide, the Optimize authentication concept page, the Optimize API authentication page, and the system configuration table:

- Camunda 8.10 no longer accepts the static token. With 8.10 defaults, Optimize rejects those requests with a `401` directly after the upgrade.
- The token works only if you opt into the 8.9 security fallback (`optimize.security.csl.enabled=false`). Camunda 8.11 will remove that fallback and the legacy configuration keys, so migrate the clients during 8.10.
- New **Check if you are affected** paragraph in the upgrade guide: the Camunda Helm chart and SaaS configure OIDC and do not set `api.accessToken`, so only operators who set the property or `OPTIMIZE_API_ACCESS_TOKEN` themselves are affected.

Ready for another look.

docs/apis-tools/optimize-api/optimize-api-authentication.md
```diff
@@ -81,9 +81,7 @@ All Optimize API requests except [the health readiness](./health-readiness.md) e
    ```
 6. Capture the value of the `access_token` property and store it as your token.
 
-:::note
-The Optimize API can also be configured in a Self-Managed environment to authenticate using a single shared access token. See [External API Configuration](/self-managed/components/optimize/configuration/system-configuration.md#external-api) for the configuration required to access the public API using a specific token.
-:::
+Through Camunda 8.9, Self-Managed also accepted a single shared access token on this API. Camunda 8.10 no longer accepts it. The token works only if you opt into the 8.9 security fallback, which Camunda 8.11 will remove. See [Optimize static API access token is no longer supported](/reference/announcements-release-notes/8100/8100-announcements.md#optimize-static-api-access-token-is-no-longer-supported).
```
StevePascoe, line 84: let's change this to the "8.9 component specific configuration fallback"
security fallback sounds scary. (This better aligns to how Ben's talking about it in his doc changes)

docs/reference/announcements-release-notes/8100/8100-announcements.md
```diff
@@ -787,6 +787,25 @@ Starting with Camunda 8.10, Optimize authenticates through the [Camunda Security
...
+</div>
+<div className="release-announcement-content">
+
+#### Optimize static API access token is no longer supported
+
+In Camunda 8.10, Self-Managed Optimize accepts only OIDC bearer tokens on its API. The static token from `api.accessToken` (environment variable `OPTIMIZE_API_ACCESS_TOKEN`) is not accepted. Requests with this token get a `401` response. This applies to the [Optimize API](/apis-tools/optimize-api/overview.md) and to the [external variable ingestion](/apis-tools/optimize-api/external-variable-ingestion.md) endpoint. The Camunda Helm chart and SaaS configure OIDC for this API, so this change affects you only if you set the token yourself.
+
+**Action:** Change these API clients to OIDC bearer tokens before you upgrade to 8.10. Then remove `api.accessToken` from your configuration.
+
+**Alternative:** The static token works only if you opt into the 8.9 security fallback with `optimize.security.csl.enabled=false`. Camunda 8.11 will remove this fallback and the legacy configuration keys. Migrate these API clients during 8.10.
```
StevePascoe, line 802: similar comment here - the 'security fallback' sounds scary

docs/self-managed/concepts/authentication/authentication-to-optimize.md
```diff
@@ -35,6 +35,12 @@ Note the following:
 If you deploy with the Camunda Helm chart, you don't need to set these directly. The chart continues to read the same `global.identity.auth.optimize.*` values you already use, and renders them into the properties above for you.
 :::
 
+## Authenticate API requests
+
+The Optimize API accepts OIDC bearer tokens. A client gets a token from your IdP with the client credentials flow, then sends the token in the `Authorization` header. Add the audience that your IdP issues for the client to `camunda.security.authentication.oidc.audiences`.
+
+Camunda 8.10 no longer accepts the static token `api.accessToken` on this API. The token works only if you opt into the [8.9 security fallback](#fall-back-to-the-89-security-stack), which Camunda 8.11 will remove. Migrate these API clients to OIDC bearer tokens during 8.10. For the token steps, see [Optimize API authentication](/apis-tools/optimize-api/optimize-api-authentication.md).
```
StevePascoe, line 42: same comment - security fallback is scary. (we should also change the path to get rid of 'security-stack')

StevePascoe (changes requested):

christinaausley (dismissed):
This looks good! I made one follow-up docs correction: the external variable ingestion page still documented the legacy `access_token` query parameter as an authentication option so I removed that and made the `Authorization` header required there as well.

One thing to confirm: under the 8.10 CSL/default security stack, `/api/ingestion/variable` no longer accepts the `access_token` query parameter in addition to rejecting `api.accessToken`/`OPTIMIZE_API_ACCESS_TOKEN`, correct? The docs now assume OIDC bearer tokens via the `Authorization` header only matching the rest of the PR 👍

docs/apis-tools/optimize-api/optimize-api-authentication.md
megglos: Renamed to "the 8.9 component-specific configuration fallback" throughout, and the section is now "Fall back to the 8.9 component-specific configuration" (anchor and inbound links updated). Applied consistently to the other "legacy security stack" mentions in the PR as well. `93a15b66c5`

docs/reference/announcements-release-notes/8100/8100-announcements.md
megglos: Renamed to "the 8.9 component-specific configuration fallback" throughout, and the section is now "Fall back to the 8.9 component-specific configuration" (anchor and inbound links updated). Applied consistently to the other "legacy security stack" mentions in the PR as well. `93a15b66c5`

docs/self-managed/concepts/authentication/authentication-to-optimize.md
megglos: Renamed to "the 8.9 component-specific configuration fallback" throughout, and the section is now "Fall back to the 8.9 component-specific configuration" (anchor and inbound links updated). Applied consistently to the other "legacy security stack" mentions in the PR as well. `93a15b66c5`

megglos:
Confirmed, and it goes further than 8.10: Optimize dropped `access_token` query parameter support in 2022 (commit `231780b`), so it is not accepted in 8.7, 8.8, or 8.9 either, and `optimize.security.csl.enabled=false` does not bring it back — the code has no query-parameter resolver at all. Your edit is right, and I applied the same correction to the 8.7, 8.8, and 8.9 versioned pages in `4fc8e8a`.

megglos:
@StevePascoe @christinaausley read for another 👀

Ben-Sheppard (commented):
Solid breaking-change documentation overall. A few wording issues: the "Alternative" notes frame the static-token removal as a regression fallback rather than a deliberate migration, reintroduce "legacy" inconsistently with this PR's own "component-specific configuration" phrasing, and we'd communicate the removal timeline differently than naming 8.11 outright.

docs/reference/announcements-release-notes/8100/whats-new-in-810.md
```diff
@@ -313,7 +313,7 @@ Camunda Hub and Optimize both continue to accept their existing authentication s
 
 With the move to the [Camunda Security Library](/reference/glossary.md#camunda-security-library-csl) (CSL), Optimize adopts the same authentication and session handling as the Orchestration Cluster components.
 
-The legacy Optimize login and API security keys are deprecated in favor of `camunda.security.*` and removed in 8.11, along with the legacy security stack and its `optimize.security.csl.enabled=false` fallback. `CAMUNDA_OPTIMIZE_IDENTITY_BASE_URL` is not deprecated and stays in use for user lookups. See [legacy configuration keys](/self-managed/upgrade/components/890-to-8100.md#legacy-security-configuration-keys-are-deprecated) for the full key mapping.
+The legacy Optimize login and API security keys are deprecated in favor of `camunda.security.*` and removed in 8.11, along with the legacy component-specific configuration and its `optimize.security.csl.enabled=false` fallback. `CAMUNDA_OPTIMIZE_IDENTITY_BASE_URL` is not deprecated and stays in use for user lookups. See [legacy configuration keys](/self-managed/upgrade/components/890-to-8100.md#legacy-security-configuration-keys-are-deprecated) for the full key mapping.
```
Ben-Sheppard: 🔧 NABD — This line states "removed in 8.11" as settled fact, but that's carried over from existing text this PR only lightly touched (swapping "security stack" for "component-specific configuration"). Flagging since it's the same timeline-communication issue as the newer additions below, but the underlying wording here is pre-existing and out of scope for this PR — likely one to fold into the separate PR reworking this terminology.

docs/reference/announcements-release-notes/8100/8100-announcements.md
```diff
@@ -840,6 +840,25 @@ Starting with Camunda 8.10, Optimize authenticates through the [Camunda Security
...
+</div>
+<div className="release-announcement-content">
+
+#### Optimize static API access token is no longer supported
+
+In Camunda 8.10, Self-Managed Optimize accepts only OIDC bearer tokens on its API. The static token from `api.accessToken` (environment variable `OPTIMIZE_API_ACCESS_TOKEN`) is not accepted. Requests with this token get a `401` response. This applies to the [Optimize API](/apis-tools/optimize-api/overview.md) and to the [external variable ingestion](/apis-tools/optimize-api/external-variable-ingestion.md) endpoint. The Camunda Helm chart and SaaS configure OIDC for this API, so this change affects you only if you set the token yourself.
+
+**Action:** Change these API clients to OIDC bearer tokens before you upgrade to 8.10. Then remove `api.accessToken` from your configuration.
+
+**Alternative:** The static token works only if you opt into the 8.9 component-specific configuration fallback with `optimize.security.csl.enabled=false`. Camunda 8.11 will remove this fallback and the legacy configuration keys. Migrate these API clients during 8.10.
```
Ben-Sheppard: ⚠️ Please address — Three issues with this note:
1. **Regression framing**: this links readers to the "Fall back to the 8.9 component-specific configuration" section, which is written for "if CSL causes a regression, revert temporarily." The static token removal isn't a regression — it's a deliberate change some users currently rely on. This note should say so directly (e.g. "you may currently rely on the static token; we recommend migrating to OIDC bearer tokens; if you need more time, the fallback below temporarily restores 8.9 behavior") rather than implying the fallback exists because something broke.
2. **"Legacy"**: "the legacy configuration keys" is inconsistent with "component-specific configuration" used earlier in the same sentence for the same thing.
3. **Timeline**: "Camunda 8.11 will remove this fallback..." commits to a specific version. We'd communicate this differently — e.g. "planned for removal in a future release" — rather than naming 8.11 outright.

docs/self-managed/upgrade/components/890-to-8100.md
```diff
@@ -499,10 +499,19 @@ The following keys are no longer used in 8.10. Optimize still starts if you leav
...
+
+**Check if you are affected:** The Camunda Helm chart and Camunda 8 SaaS do not set `api.accessToken`. They configure OIDC for the Optimize API. You are affected only if you set the property or the `OPTIMIZE_API_ACCESS_TOKEN` environment variable yourself, for example as an Optimize property override or an extra environment variable in your Helm values, or in a manual or Docker Compose installation.
+
+Optimize logs an obsolete-property warning for the environment variable only. Check your YAML configuration for the `api.accessToken` key too.
+
+**Action:** Change these API clients to [OIDC bearer tokens](/apis-tools/optimize-api/optimize-api-authentication.md) before you upgrade. Then remove the setting.
+
+**Alternative:** The static token works only if you opt into the 8.9 component-specific configuration fallback with `optimize.security.csl.enabled=false`. Camunda 8.11 will remove this fallback and the legacy configuration keys. Migrate these API clients during 8.10.
```
Ben-Sheppard: ⚠️ Please address — Identical text to the Alternative note in `8100-announcements.md`, same three issues: (1) frames the token removal as a regression fallback rather than a deliberate migration, (2) reintroduces "legacy" inconsistently with "component-specific configuration" in the same sentence, (3) commits to naming 8.11 for the removal rather than phrasing it as the current aim.

docs/self-managed/concepts/authentication/authentication-to-optimize.md
```diff
@@ -35,6 +35,12 @@ Note the following:
 If you deploy with the Camunda Helm chart, you don't need to set these directly. The chart continues to read the same `global.identity.auth.optimize.*` values you already use, and renders them into the properties above for you.
 :::
 
+## Authenticate API requests
+
+The Optimize API accepts OIDC bearer tokens. A client gets a token from your IdP with the client credentials flow, then sends the token in the `Authorization` header. Add the audience that your IdP issues for the client to `camunda.security.authentication.oidc.audiences`.
+
+Camunda 8.10 no longer accepts the static token `api.accessToken` on this API. The token works only if you opt into the [8.9 component-specific configuration fallback](#fall-back-to-the-89-component-specific-configuration), which Camunda 8.11 will remove. Migrate these API clients to OIDC bearer tokens during 8.10. For the token steps, see [Optimize API authentication](/apis-tools/optimize-api/optimize-api-authentication.md).
```
Ben-Sheppard: ⚠️ Please address — "...which Camunda 8.11 will remove" — same as the other two new instances, we'd communicate this timeline differently rather than naming the version.

docs/reference/announcements-release-notes/8100/8100-announcements.md
megglos: Applied in `5fddb54e44`. The note now reads: "**Alternative:** If you need more time, set `optimize.security.csl.enabled=false`. This opts into the 8.9 component-specific configuration fallback, and the static token works again. Camunda plans to remove this fallback and the component-specific configuration keys in a future release. Migrate these API clients as soon as you can." I made the same change in the two other new sentences that named 8.11 (`optimize-api-authentication.md` and the `api.accessToken` row in `system-configuration.md`).

docs/self-managed/upgrade/components/890-to-8100.md
megglos: Applied in `5fddb54e44`. The note now reads: "**Alternative:** If you need more time, set `optimize.security.csl.enabled=false`. This opts into the 8.9 component-specific configuration fallback, and the static token works again. Camunda plans to remove this fallback and the component-specific configuration keys in a future release. Migrate these API clients as soon as you can." I made the same change in the two other new sentences that named 8.11 (`optimize-api-authentication.md` and the `api.accessToken` row in `system-configuration.md`).

docs/self-managed/concepts/authentication/authentication-to-optimize.md
megglos: Applied in `5fddb54e44`. The note now reads: "**Alternative:** If you need more time, set `optimize.security.csl.enabled=false`. This opts into the 8.9 component-specific configuration fallback, and the static token works again. Camunda plans to remove this fallback and the component-specific configuration keys in a future release. Migrate these API clients as soon as you can." I made the same change in the two other new sentences that named 8.11 (`optimize-api-authentication.md` and the `api.accessToken` row in `system-configuration.md`).

docs/reference/announcements-release-notes/8100/whats-new-in-810.md
megglos: Agreed, this text is pre-existing. I left it for the follow-up PR on this terminology.

christinaausley:
Tech review done 👍 Some small wording adjustments.

A few things --

1. The page contradicts itself on the removal timeline. This PR adds five "Camunda plans to remove … in a future release" statements. Seven existing ones say "removed in 8.11", two of them on `8100-announcements.md` itself, lines 870 and 887, within two entries of your new line 855.

I saw your reply to @Ben-Sheppard deferring the pre-existing wording to the terminology follow-up, which is fair for text you only brushed. But the deferral runs backwards here: this PR introduces the contradiction rather than inheriting it. Can the new text say 8.11 for now, with the follow-up softening all of them together? Or has 8.11 actually slipped... in which case that follow-up is the more urgent PR?

2. The versioned-docs changes need confirming, and aren't in the description. This also edits `external-variable-ingestion.md` in 8.7, 8.8, and 8.9- removing the documented `access_token` query parameter and renaming the header from `Authentication` to `Authorization`. That's the only part changing already-released behavior docs so if it's wrong it's wrong for users on shipped versions.

Can you confirm the query parameter genuinely never worked on that endpoint in 8.7-8.9 rather than being removed because it's gone in 8.10?

3. The announcement may under-warn Helm users. Your research says the affected group includes Helm installs with Optimize auth disabled, since the chart only renders `api.jwtSetUri` when auth is enabled. The upgrade guide covers that case- the announcement states that "The Camunda Helm chart and SaaS configure OIDC for this API." A Helm user with auth disabled could stop reading there. Can it carry the same qualifier?

Two non-blocking follow-ups: the rewritten External API section drops the old "if neither is configured an error will be thrown" behavior without an 8.10 equivalent, and `security.auth.token.lifeMin` is left unmarked in the table directly above `security.auth.token.secret` ("Not used in 8.10") - worth confirming it's still reflected?

PR #10025, opened 2026-09-22
docs(concepts): document draining deletion for process definitions

### LINKED ISSUE #59567: Document draining deletion behavior for process definitions
## Context

Docs updates for **draining deletion** — camunda/camunda#55930.

Deleting a process definition with active instances no longer blocks the cluster-wide command distribution queue. Instead of waiting for physical removal, the deletion marks the definition as `DRAINING` on each partition (acknowledged immediately, freeing the queue). New instances of that version are blocked; running instances drain to completion; each partition physically removes the definition asynchronously once its last active instance finishes or is cancelled, transitioning to the existing `DELETING`/deleted lifecycle.

This changes documented behavior in several places.

### AC1 — Explain draining in the resource-deletion concept page
Update `docs/components/concepts/resource-deletion.md`:
- Deleting a process definition with active instances is no longer rejected; the definition enters `DRAINING`.
- New instances are blocked (`NOT_FOUND`); start events are deactivated at delete time even though the physical record still lingers while draining.
- Running instances drain to completion; physical removal and transition to deleted happen asynchronously per-partition once the last active instance finishes or is cancelled.
- Fix the **Limitations** section — "You cannot delete a process definition that has one or more running process instances" is now false and must be rewritten.

### AC2 — Document the `DRAINING` lifecycle state
Where process definition lifecycle states are documented, add `DRAINING`: its meaning, entry/exit conditions, and the difference from the existing `DELETING` state.

### AC3 — Document the new Prometheus metric
Add `zeebe_process_definitions_draining_count` (and any sibling metrics) to `docs/self-managed/operational-guides/monitoring/metrics.md`. Confirm the exact metric name, type, and labels with engineering before publishing.

### AC4 — Breaking behavioral change entry
Add to the 8.10 announcements/release notes and the `docs/self-managed/upgrade/components/890-to-8100.md` upgrade guide: deleting a process definition with active instances now succeeds (draining) instead of being rejected. Callers relying on the old rejection behavior are affected.

### AC5 — API surfaces the state
- Process definition GET/search API: new state field exposing `DRAINING` (generated spec — coordinate with the API repo).
- `docs/apis-tools/orchestration-cluster-api-rest/specifications/delete-resource.api.mdx` is auto-generated from OpenAPI — behavioral description changes land upstream, not by hand-editing the mdx.
- Update the gRPC `DeleteResource` RPC description in `docs/apis-tools/zeebe-api/gateway-service.md`.

### AC6 — `deleteHistory` timing change
Update the **Historic data** section of `docs/components/concepts/resource-deletion.md`: the `deleteHistory` batch operation no longer runs immediately on submit — it is deferred until the definition is physically deleted on all partitions (draining instances are still exporting events). Clarify timing and eventual-consistency implications.

### AC7 — Operate UI draining state
Update `docs/components/operate/userguide/delete-resources.md`: if Operate surfaces a `DRAINING` / "waiting for instances" state, document it and how a user sees which instances are keeping the definition alive. Gated on the Operate UI actually shipping this.

## Open questions (resolve before writing)
- Exact Prometheus metric name(s), type, and labels.
- Whether `DRAINING` is exposed in the v2 API and the exact field name.
- Whether the Operate/Tasklist UI ships the draining signal in the same release.

Backports to older versioned docs will be handled in separate issues.

danielkelemen:
## Description

Document draining deletion fix at multiple places:

- Operate: document draining state when deleting process definitions
- Metrics: add zeebe_process_definitions_draining_count
- Zeebe-api: update DeleteResource RPC for draining deletion
- Concepts: document draining deletion for process definitions
- Document in 8.9, 8.10 documentations

## PR Checklist

closes https://github.com/camunda/camunda/issues/59567


docs/reference/announcements-release-notes/8100/8100-announcements.md
```diff
@@ -291,6 +291,26 @@ The [Get decision instance](/apis-tools/orchestration-cluster-api-rest/specifica
...
 
+<div className="release-announcement-row">
+<div className="release-announcement-badge">
+<span className="badge badge--change">Change</span>
+</div>
+<div className="release-announcement-content">
+
+#### Deleting a process definition with running instances defers history deletion
+
+The delete resource endpoint behaves the same, with one difference for a process definition that still has running instances. Instead of being deleted synchronously, the definition [drains](/components/concepts/resource-deletion.md#draining): new instances are blocked immediately, running instances continue to completion, and the definition is removed automatically afterwards.
```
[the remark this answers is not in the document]
danielkelemen: It was added for 8.8 & 8.9 in a patch release so we can still announce it for 8.10 main, I think.

versioned_docs/version-8.8/components/concepts/resource-deletion.md
```diff
@@ -61,10 +58,24 @@ A [call activity](/components/modeler/bpmn/call-activities/call-activities.md) r
...
+| :--------- | :-------------------------------------------------------------------------------------------------------------------------------------- |
+| `ACTIVE`   | Deployed and able to create new instances.                                                                                              |
+| `DRAINING` | Deleted for new use (new instances are blocked) while running instances drain. The record is retained until the last instance finishes. |
+| `DELETING` | The last instance has drained and the definition is being physically removed on the partition. This is a brief internal transition.     |
+| `DELETED`  | Fully removed.                                                                                                                          |
+
+The Orchestration Cluster API process definition `state` field exposes `ACTIVE`, `DRAINING`, and `DELETED`. You can also track draining definitions with the `zeebe_process_definitions_draining_count` [metric](/self-managed/operational-guides/monitoring/metrics.md) and the draining indicator in [Operate](../operate/userguide/delete-resources.md).
```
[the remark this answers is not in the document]
danielkelemen: DRAINING is part of the latest API spec, it is just probably out of sync for docs.

versioned_docs/version-8.9/components/concepts/resource-deletion.md
```diff
@@ -43,9 +43,28 @@ Deleting `Version 2` before `Version 3` produces the same behavior, except `Vers
...
+| `ACTIVE`   | Deployed and able to create new instances.                                                                                              |
+| `DRAINING` | Deleted for new use (new instances are blocked) while running instances drain. The record is retained until the last instance finishes. |
+| `DELETING` | The last instance has drained and the definition is being physically removed on the partition. This is a brief internal transition.     |
+| `DELETED`  | Fully removed.                                                                                                                          |
+
+The Orchestration Cluster API [process definition `state` field](/apis-tools/orchestration-cluster-api-rest/specifications/get-process-definition.api.mdx) exposes `ACTIVE`, `DRAINING`, and `DELETED`. You can also track draining definitions with the `zeebe_process_definitions_draining_count` [metric](/self-managed/operational-guides/monitoring/metrics.md) and the draining indicator in [Operate](../operate/userguide/delete-resources.md#delete-process-definition).
```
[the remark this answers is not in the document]
danielkelemen: DRAINING is part of the latest API spec, it is just probably out of sync for docs.

versioned_docs/version-8.9/components/concepts/resource-deletion.md
```diff
@@ -65,6 +84,8 @@ If you only want to delete process instance data, see [process instance deletion
 
 Historic data deletion runs asynchronously. Depending on the amount of data, it may take time for the data to be removed and for it to disappear from Operate and Tasklist.
 
+If the definition is [draining](#draining), history deletion is deferred: the `deleteHistory` operation does not run when you submit the delete, but only after the definition has been physically deleted on all partitions. Draining instances keep exporting events until they finish, so their historic data cannot be removed before then.
```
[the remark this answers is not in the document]
danielkelemen: Again, spec out of sync for docs.

christinaausley (dismissed):
I reviewed the draining deletion docs and made a few small corrections if you want to take a look 😄 

Two questions before merge:

1. The 8.8 and 8.9 generated Orchestration Cluster API specs for `get-process-definition` still show `state` as only `ACTIVE`/`DELETED`, while the concept pages say the API exposes `DRAINING`. Should the generated specs be updated for the affected patch docs or should the versioned concept wording avoid linking that claim until the generated specs are updated?
2. Can you confirm the exact metric contract for `zeebe_process_definitions_draining_count` ... name, gauge type, labels, and the partition-leader caveat with engineering before merge?

versioned_docs/version-8.9/self-managed/upgrade/prepare-for-upgrade.md
```diff
@@ -42,6 +42,7 @@ This section lists actions you must complete or review before upgrading to Camun
...
+| <span className="label-highlight red" style={{whiteSpace: "nowrap"}}>Action required</span>  | Orchestration Cluster: draining deletions   | If you deleted process definitions with running instances (draining) on 8.8, you must upgrade to **8.9.19 or later**. Earlier 8.9 patches cannot replay the `DRAINING` and `FULLY_DELETED` events and fail to recover the draining state on startup. See [resource deletion](../../components/concepts/resource-deletion.md#draining).                                                                                                                                                                                                                                                                                                     |
```
tanjinyi, line 45: 🤔 do we want to highlight the version of 8.8 that would first carry draining process definitions as well?

docs/components/concepts/resource-deletion.md
```diff
@@ -43,9 +43,24 @@ Deleting `Version 2` before `Version 3` produces the same behavior, except `Vers
...
+- **Running instances continue to completion.** The deletion does not cancel them.
+- **Physical removal is asynchronous and per-partition.** Once a partition's last active instance of the definition finishes or is canceled, that partition removes the definition and transitions it to the deleted state.
+
+A process definition moves through the following lifecycle states:
+
+| State      | Meaning                                                                                                                                 |
+| :--------- | :-------------------------------------------------------------------------------------------------------------------------------------- |
+| `ACTIVE`   | Deployed and able to create new instances.                                                                                              |
+| `DRAINING` | Deleted for new use (new instances are blocked) while running instances drain. The record is retained until the last instance finishes. |
```
tanjinyi: 🤔 would `marked as deleted` make more sense over `deleted for new use` here?

versioned_docs/version-8.8/components/operate/userguide/delete-resources.md
```diff
@@ -15,8 +15,10 @@ To delete a process definition from the **Processes** page, take the following s
 ![operate-view-process-filters](./img/delete-resources/process-filters.png)
 
 :::note
-Make sure the selected process definition version has no running instances, otherwise it is not possible to delete a process definition.
+In Operate, you can only delete a process definition version that has no running instances; the **Delete** button is disabled otherwise.
```
tanjinyi: 💭 Here, the paragraph is attempting to say that Operate will not allow process definitions with running process instances to be deleted via the UI right? I assume this is an 8.8 and 8.9 limitation

tanjinyi (dismissed):
Preapproving with minor comments.

versioned_docs/version-8.9/self-managed/upgrade/prepare-for-upgrade.md
danielkelemen: Good idea, I'll update this.

docs/components/concepts/resource-deletion.md
danielkelemen: Agree, this does sound weird now to me as well. I'll rephrase this.

versioned_docs/version-8.8/components/operate/userguide/delete-resources.md
danielkelemen: Yes, we left the UI as is so it still expects the running instances to be gone, even with draining.

It's even for 8.10. UI remains unchanged, API allows deletion with running instances.

danielkelemen:
> 1. The 8.8 and 8.9 generated Orchestration Cluster API specs for `get-process-definition` still show `state` as only `ACTIVE`/`DELETED`, while the concept pages say the API exposes `DRAINING`. Should the generated specs be updated for the affected patch docs or should the versioned concept wording avoid linking that claim until the generated specs are updated?

@christinaausley: do 8.8 & 8.9 get regenerated periodically or only on demand? Should I do it? I'm not familiar with this part.

8.10 shows DRAINING: https://docs.camunda.io/docs/next/apis-tools/orchestration-cluster-api-rest/specifications/get-process-definition/#:~:text=%5BACTIVE%2C-,DRAINING,-%2C%20DELETED%5D

versioned_docs/version-8.8/components/operate/userguide/delete-resources.md
tanjinyi: i think the 8.10 UI has a different wording for this? we should make sure that the paragraph is unchanged for 8.10 too, right?

christinaausley:
> do 8.8 & 8.9 get regenerated periodically or only on demand? Should I do it?

Periodically, and automatically... you don't need to run anything by hand. [`.github/workflows/sync-rest-api-docs.yaml`](.github/workflows/sync-rest-api-docs.yaml) runs on a cron (`0 15 * * 4`, Thursdays 15:00 UTC) and also accepts `workflow_dispatch`. After syncing `next`, it loops over `versions.json` (`["8.9", "8.8", "8.7"]`) and, for each version that has both an `api/camunda/version-$version` directory here and a matching `stable/$version` branch in `camunda/camunda`, copies the spec across and runs `npm run api:generate -- camunda "$version"`. If anything changed it opens a PR on the `update-rest-doc` branch.

But regenerating won't fix this, which I think is the actual answer to my question. The gap isn't stale generated output it's the source specs. I checked all three:

| Spec source | `state` enum |
| :--- | :--- |
| `api/camunda/v2/process-definitions.yaml` (next) | `ACTIVE`, `DRAINING`, `DELETED` |
| `api/camunda/version-8.9/process-definitions.yaml` | `ACTIVE`, `DELETED` |
| `api/camunda/version-8.8/camunda-openapi.yaml` (single-file) | no `DRAINING` |

Since the sync overwrites those files wholesale from `stable/8.9` and `stable/8.8`, editing them in this repo would just be reverted next Thursday. So the change belongs upstream in `camunda/camunda`:

- `stable/8.9` to `zeebe/gateway-protocol/src/main/proto/v2/process-definitions.yaml`
- `stable/8.8` to `zeebe/gateway-protocol/src/main/proto/rest-api.yaml` (8.8 and below still use the single-file spec)

Once that goes, the weekly sync picks it up and opens the docs PR on its own.

Given that, two ways forward...I'm fine with either, your call:

1. Land the upstream spec change first then merge this as-is. Cleanest, but gated on a `camunda/camunda` backport.
2. Merge now and soften the 8.8/8.9 concept wording so it doesn't assert the API exposes `DRAINING`, with a follow-up issue to restore it after the specs sync. The 8.10 wording can stay as it is, since `next` is already correct.

One nuance worth keeping in mind for either option: the versioned specs track the **tip** of `stable/8.x`, not individual patches. So even after the upstream fix, the 8.9 reference will show `DRAINING` across all of 8.9, even though it only exists from 8.9.19+. The `x-properties-added-in-version` extension (already used in `version-8.9/process-definitions.yaml`) only renders minor-level "Added in X.Y" badges, so it can't express a patch floor. That's an argument for keeping the patch requirement explicit in the prose the way you already have it in `prepare-for-upgrade.md`.

Separately my second question is still open - the metric contract for `zeebe_process_definitions_draining_count`. The PR documents it in `metrics.md` as a gauge, reported on the partition leader, with labels `physicalTenant` and `partition`. I'd just like an engineering 👍 that those four things are exactly right before we publish them.

danielkelemen:
`stable/8.9` has draining already: https://github.com/camunda/camunda/blob/stable/8.9/zeebe/gateway-protocol/src/main/proto/v2/process-definitions.yaml#L454-L463

`stable/8.8` as well: https://github.com/camunda/camunda/blob/stable/8.8/zeebe/gateway-protocol/src/main/proto/rest-api.yaml#L6601-L6610

danielkelemen:
> the metric contract for zeebe_process_definitions_draining_count. The PR documents it in metrics.md as a gauge, reported on the partition leader, with labels physicalTenant and partition. I'd just like an engineering 👍 that those four things are exactly right before we publish them.

That is all correct. If needed @tanjinyi can also confirm it.

